PR #1801, opened 2014-09-18
xunit: rename testsuite attribute skip to skipped

llipavsky:
Currently, xunit output looks like:

```
<testsuite ... skip="0">
```

However, all examples of JUnit results (and also some other xUnits that look similar) either ignore this attribute completely, or use `skipped` instead of skip. So the xunit looks like this:

```
<testsuite ... skipped="0">
```

This small issue has one bad inpact: Bamboo JUnit parser will not consider the file to be valid when `skip` is used instead of `skipped`. So I wrote this patch to change it to `skipped`. I did not find any xunit specification that would state that the attribute should be `skip` while `skipped` is used quite often in existing JUnit files.

However, this change might have some impact to existing users of Robot Framework xunit output, who might already have their custom parsers relying on `skip`. But are there any?

### DIFF
--- a/atest/robot/output/xunit.robot
+++ b/atest/robot/output/xunit.robot
@@ -86,8 +86,8 @@ Get XUnit Nodes
     [Return]  ${nodes}
 
 Stats Should Be
-    [Arguments]    ${elem}    ${tests}    ${failures}    ${skip}
+    [Arguments]    ${elem}    ${tests}    ${failures}    ${skipped}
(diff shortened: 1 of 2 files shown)


pekkaklarck:
xUnit consumers have got more strings and nowadays `skip` isn't anymore generally supported. We now got #2978 about changing it and it's basically a duplicate of this issue. Probably better anyway to keep that open and this one closed. The proposed change in this issue is nevertheless going to be implemented.

Issue #1506, opened 2014-06-30
Python 3 support


spooning:
> _Originally submitted to [Google Code](http://code.google.com/p/robotframework/issues/detail?id=1506) by sabbir.shafi on 18 Aug 2013_

Later versions of Python 2.7 does not support robot framework...

spikergit1:
Word on the street (https://github.com/robotframework/robotframework/issues/1912#issuecomment-73610630) is that RF 2.9 will drop j/python 2.5 support.  Does that mean that the python3 support can be / will be integrated in RF 2.9?

pekkaklarck:
I doubt we have time for that. The main goal of dropping 2.5 support is making it easier to add 3.x support later. If required changes are not too big, we can consider that already in RF 2.9, though.

userzimmermann:
### Status update :)

The latest release of my [Python 3 compatibility port](https://github.com/userzimmermann/robotframework/tree/python3) is __2.8.7__:

https://github.com/userzimmermann/robotframework/releases/tag/python3-2.8.7
https://pypi.python.org/pypi/robotframework-python3/2.8.7

Complete diff to original __2.7.8__:

https://github.com/userzimmermann/robotframework/compare/2.8.7...python3-2.8.7

Further I merged with recent commit https://github.com/robotframework/robotframework/commit/3a7571d23030166244c8dd1785346b41680dadbf which already contains many Python 2.5 incompatible code updates as discussed in #1928 and lots of code restructuring.
I got a lot of merge conflicts :D but they are solved.
There are only a few new __utests__ and __atests__ which need some more compatibility adaptions.

You can always get the latest complete code diff at:

https://github.com/userzimmermann/robotframework/compare/master...python3

I also started to create issues for __atests__ FAILing on Python 3:

https://github.com/userzimmermann/robotframework/issues

So anyone can participate in solving them. They generally only need workarounds in the __atests__ themselves.

And the best thing is... actually not the best but very close ;)
__It is still running on Python 2.6!!!__ (At least with Python 2.6.8)
I remember there were some issues with Python 2.6 in the beginning of my fork. I can't remember exactly. But I declared my fork incompatible with Python 2.6 because of them and defined 2.7 as the only compatible Python 2 version. I was just curious and ran the tests of the current code with Python 2.6 and it worked. Maybe the earlier issues were only related to older Python 2.6.x versions or somehow resolved by using __six__.
I would appreciate if someone could run the tests with some older Python 2.6 version.

[...]

PR #1811, opened 2014-10-02
Python 3 compatibility

userzimmermann:
For everyone interested in my Python 2.7/3.3+ compatibility fork. To have a central place for comparing and discussing the code. See #1506 for the history of this fork.


pekkaklarck:
Ok, we have now dropped Python 2.5 support (#1928).

We'e also already changed all occurrences of `except Exception, err` to use `except Exception as err` instead. It would be great if @userzimmermann or someone else could take the current code and see how to make it Python 3 compatible. Instead of having one big PR with all changes, it would be easier for us to accept smaller PRs fixing one or few incompatibilities. All changes getting us closer to official Python 3 compatibility would be great.

I close this PR now because there have been way too many changes in the code base for us to be able to merge it.

userzimmermann:
I had not much time for this fork in the last few months. But I want to continue. I will try to start getting in sync with the latest commits in the next days...

pekkaklarck:
Awesome!

PR #1836, opened 2014-10-23
New exact_match option for Get Lines Matching Regexp, which uses re.search instead of re.match

HelioGuilherme66:
The original keyword 'Get Lines Matching Regexp' is using '^%s$' and re.match. This caused me some troubles in finding lines with some content inside. With this new option, regular expressions are parsed with re.search and not using the '^' and '$' positional markers. Acceptance tests are included.


HelioGuilherme66:
Please consider this as the final version of the PR.
Thanks.

src/robot/libraries/String.py
```diff
@@ -237,29 +237,38 @@ def get_lines_matching_pattern(self, string, pattern, case_insensitive=False):
...
         Python regular expression syntax in general and how to use it
-        in Robot Framework test data in particular. A line matches
-        only if it matches the ``pattern`` fully. Notice that to make
-        the match case-insensitive, you need to embed case-insensitive
-        flag into the pattern.
-
-        Lines are returned as one string catenated back together with
+        in Robot Framework test data in particular. By default ``exact_match``
+        is True causing line matches only if it matches the ``pattern`` fully.
+        When passing ``exact_match`` as False, this keyword uses the matching
```
pekkaklarck: `exact_match` doesn't need to be exactly `True` or `False` after we recently introduced `is_truthy` that you also use here. Take a look at other `get_lines_xxx` keywords for examples how to explain that.

src/robot/libraries/String.py
```diff
@@ -237,29 +237,38 @@ def get_lines_matching_pattern(self, string, pattern, case_insensitive=False):
...
         Examples:
-        | ${lines} = | Get Lines Matching Regexp | ${result} | Reg\\\\w{3} example |
-        | ${ret} = | Get Lines Matching Regexp | ${ret} | (?i)FAIL: .* |
-
-        See `Get Lines Matching Pattern` and `Get Lines Containing String` if
-        you do not need full regular expression powers (and complexity).
+        | ${lines} = | Get Lines Matching Regexp | ${result} | Reg\\\\w{3} example |   |
+        | ${ret} = | Get Lines Matching Regexp | ${ret} | (?i)FAIL: .* |   |
+        | ${result} = | Get Lines Matching Regexp | ${input} | ^ABC | False |
+        | ${result} = | Get Lines Matching Regexp | ${input} | ^ABC.* | True |
```
pekkaklarck: Good that you've added examples. I'd like them to be enhanced slightly:
- Last two examples should be in different order with the one using default behavior first.
- Using `exact_match=False` would be much more explicit than just `False`. There's no need to have `True` at all because it is the default.
- Both should have same pattern to make it explicit that the value of `exact_match` is interesting. Using `^` in the pattern requiring exact match is confusing.

src/robot/libraries/String.py
```diff
@@ -237,29 +237,38 @@ def get_lines_matching_pattern(self, string, pattern, case_insensitive=False):
...
+
+        See `Get Lines Matching Pattern` and `Get Lines Containing
+        String` if you do not need full regular expression powers (and
+        complexity).
```
pekkaklarck: A note about `exact_match` being new in 2.9 needs to added. You could explain it where `exact_match` is explained in general, but it might be easier just to add a note like this to the end of the docstring:

> `exact_match` argument is new in Robot Framework 2.9. In earlier versions exact match was always required.

atest/testdata/standard_libraries/string/get_matching_lines.robot
```diff
@@ -49,6 +49,9 @@ Get Lines Matching Regexp When Input Is Empty
 Get Lines Matching Regexp When Pattern Is Empty
     Test Get Lines Matching Regexp    ${INPUT}    ${EMPTY}    ${EMPTY}
 
+Get Lines Matching Regexp Matching Two Words
+    Test Get Lines Matching Regexp    ${INPUT}    more|row    ${EMPTY}
+
```
pekkaklarck: What's the purpose of this new test?

atest/testdata/standard_libraries/string/get_matching_lines.robot
```diff
@@ -64,6 +67,27 @@ Get Lines Matching Regexp With Case-Insensitive
     Test Get Lines Matching Regexp    ${INPUT}    (?i).*LINE    Third line
     Test Get Lines Matching Regexp    ${INPUT}    .*LINE.*    ${EMPTY}
 
+Get Lines Containing Regexp Matching Two Words
+    Test Get Lines Containing Regexp    ${INPUT}    more|row    One more row here
+
```
pekkaklarck: Oh, this is clearly a pair of the above new test. I don't think either of them are really needed. They seem to basically test that Python handles `|` correctly in regexps.

atest/testdata/standard_libraries/string/get_matching_lines.robot
```diff
@@ -76,6 +100,11 @@ Test Get Lines Matching Pattern
...
-    [Arguments]    ${input}    ${pattern}    ${expected}
-    ${actual} =    Get Lines Matching Regexp    ${input}    ${pattern}
+    [Arguments]    ${input}    ${pattern}    ${expected}    ${exact_match}=True
+    ${actual} =    Get Lines Matching Regexp    ${input}    ${pattern}    ${exact_match}
+    Should Be Equal    ${actual}    ${expected}
+
+Test Get Lines Containing Regexp
+    [Arguments]    ${input}    ${pattern}    ${expected}    ${exact_match}=False
+    ${actual} =    Get Lines Matching Regexp    ${input}    ${pattern}    ${exact_match}
     Should Be Equal    ${actual}    ${expected}
```
pekkaklarck: `${exact_match}` argument to these keywords seems to unused. I'd actually prefer the first one not to pass `exact_match` at all to `Get Lines Matching Regexp` to verify that that is the default behavior. The latter should preferably use the named argument syntax (e.g. `exact_match=false`) to test the argument name.

pekkaklarck:
1. The PR looks really good.
2. I added some line notes about docs and tests, but they ought to be easy to fix.
3. Please add yourself to AUTHORS.txt too.
4. This PR contains 21 commits, most of which are merges. Could you either take a look at squashing the commits or just do a new PR with same content?

atest/testdata/standard_libraries/string/get_matching_lines.robot
HelioGuilherme66: Please note the ${EMPTY} result when exact_match=true. It is one of the problematic examples. I added more examples to documentation to illustrate this.

HelioGuilherme66:
All done, squash, doc and AUTHORS.txt.

pekkaklarck:
I noticed few more problems:
- There are two commits still. The first one added the later removed `Get Lines Containing Regexp`.
- There are trailing empty spaces in code. We try to avoid them.
- Example has quotes around strings.

I'll fix all these and possible other problems myself.

pekkaklarck:
I squashed commits and committed them as 4c178d399c39aa35d0bfefbf96a9aba26a52ac76. Essentially this PR is merged but apparently GitHub didn't notice that.

Other fixes I mentioned above were done in 02fe31f6fd0fc4fc73dbcfeaf706121804eb97b9.

PR #1852, opened 2014-11-06
Implement float support to FOR IN RANGE loops (fixes #1850)

### LINKED ISSUE #1850: Support using floats in FOR IN RANGE loop
Test steps: 
:FOR	${temperature}	IN RANGE	-40	86	11.5
          log ${temperature}

Expected results:
-40; -28.5; -17; -5.5 ...
Actual results:
-40; -29; -18; -7 ...

nicolae-chedea:
FOR IN RANGE loop works now even with float arguments.


pekkaklarck:
This definitely looks good enough to be merged. I see few places for additional clean-up, but I can do that afterwards.

It would be great if you could still squash commits and resolve merge conflicts. My guess still is that conflicts are caused by AUTHORS.txt. I doubt any other file has been changed since you started this work.

pekkaklarck:
I merged this pull request after squashing commits in 169e095b24cea207a75a1e463d320094668d0c06. Further discussion can continue in issue #1850.

Issue #1273, opened 2014-06-29
Tidy should split ELSE, ELSE IF, and AND to own rows


spooning:
> *Originally submitted to [Google Code](http://code.google.com/p/robotframework/issues/detail?id=1273) by @jussimalinen on 6 Nov 2012*

Based on issue in ride: <a href="https://code.google.com/p/robotframework-ride/issues/detail?id=1067" rel="nofollow">https://code.google.com/p/robotframework-ride/issues/detail?id=1067</a>

In the Robot Framework 2.74, the Run Keyword If has optional ELSE and ELSE IF arguments. Tidy places these arguments in a single line:
Run Keyword If	'${status}'=='PASS'	Do something	ELSE IF	'${status}'=='FAIL'	Do other thing	ELSE	Fail

It would provide better reading, if tidy could place the these arguments in separate lines:
Run Keyword If	'${status}'=='PASS'	Do something
...	ELSE IF	'${status}'=='FAIL'	Do other thing
...	ELSE	Fail

pekkaklarck:
Pull request #1889 was great otherwise, but it didn't split long lines before ELSE/ELSE IF. For example, 

```robotframework
*** Test Cases ***
Without ELSE
  Run Keyword If  expr  KW 1  1  2  3  4  5

With ELSE
  Run Keyword If  expr  KW 1  1  2  3  4  5  ELSE  KW 2  1  2  3  4  5
```

is currently split like

```robotframework
*** Test Cases ***
Without ELSE
    Run Keyword If    expr    KW 1    1    2    3    4
    ...    5

With ELSE
    Run Keyword If    expr    KW 1    1    2    3    4    5
    ...    ELSE    KW 2    1    2    3    4
    ...    5
```

It's inconsistent that KW 1 is split differently depending is there ELSE (or ELSE IF) after it, and also that KW 1 and KW 2 are split differently in the latter example.

I will fix this inconsistency myself, but I first want to clean up RowSplitter in general. It took me quite long time to understand how the original splitting worked in different corner cases and how new ELSE (IF) splitting fit into the bigger picture.

pekkaklarck:
Commit 49b62d489f9a9d6daf462b10e9dad8b2fbeb1a59 fixes inconsistent splitting of long ELSE (IF) lines. Accidentally started commit message row with `#1273` and git considered that row to be a comment. Anyway, this issue is ready for final review and testing.

pekkaklarck:
[...]

PR #1889, opened 2015-01-02
Tidy ELSE and ELSE IF onto separate lines

guykisel:
This is for #1273. This approach is only good for ELSE/ELSE IF - it's not the generic approach of preserving ellipsis.


robotci:
Tests passed.



http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/53/Tests passed.

jussimalinen:
Looks good, thanks!

PR #1890, opened 2015-01-02
Contribution instructions

guykisel:
[skip ci]


robotci:
Tests passed.



http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/54/Tests passed.

robotci:
Tests passed.



http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/55/Tests passed.

laurentbristiel:
shouldn't we include a sentence about documentation? Just mentioning that when the usage of the product or a lib is modified, user documentation should be updated accordingly, and pointing people to a file with more explanation(doc/userguide/README?).

guykisel:
That's a good idea. I'll add something about that.

jussimalinen:
Looks good! We had a quick chat about this with @pekkaklarck and as we have written all other docs in rest, it would be better to continue with one wiki format and use the same here. Github supports also rest for contribution instructions.

As the same doc will be shown when creating a new issue, there probably should be a chapter about creating issues.

I wonder why the [skip ci] didnt work... It says in the jenkins plugin that [skip ci] should be written to "pull request body". I suppose this is not the same as the first comment at pull request?

guykisel:
Oh, I went with markdown because https://github.com/blog/1184-contributing-guidelines only mentions markdown. I didn't know you could use rst! I'll convert it.

guykisel:
Also, I forgot about [skip ci] initially and edited it in later.

guykisel:
I already had a section about creating issues, but it wasn't explicitly saying so in the title. I've updated it to be more obvious.

pekkaklarck:
Merged this PR in 388e8f2109b827daa2955903deecb85e3d67a589. I guess the PR wasn't closed because I squashed commits and thus the committed revisions weren't exactly same as here. Anyway, thanks @guykisel for the initial instructions!

PR #1892, opened 2015-01-09
Fixed typos in user guide

vtuomain:
Typos corrected in local copy and verified that html user guide could be created accoording to instructions.


robotci:
Admin, can CI be enabled?

pekkaklarck:
Typo fixes look good but "coercion" has been used on purpose. Is there a semantic difference that would make "conversion" better suited in these cases?

pekkaklarck:
This PR was replaced with #1895 that's already merged. Thanks @vtuomain!

Issue #1897, opened 2015-01-19
Clean-up reference to RF 2.6 and older from User Guide and other documentation


laurentbristiel:
In the section [remote library interface](http://robotframework.googlecode.com/hg/doc/userguide/RobotFrameworkUserGuide.html?r=2.7.6#remote-library-interface) of the "User Guide", I bumped into sentences like "Robot Framework 2.1 contains generic" whereas the current version is 2.8.7. 

I thought I could just fix this into "starting from Robot Framework 2.1 etc.", but I wonder if we should not clean up all the references to Robot Framework version < 2.5. This version is quite old now and I guess for 99% of the readers, those references to version 2.1 are irrelevant.

I did not find any guideline in the source code of how/when we start to remove reference to older version. Maybe I missed it.

I could do a PR for the cleaning.

pekkaklarck:
Sound very good to me. I think we could remove references to 2.5 and perhaps also to 2.6 at the same time. Latest 2.6 release was on Oct 2011, so it's over three years old too.

Do you have interest to go through also standard library docs? That could be a separate PR.

jussimalinen:
Yeah, If someone is still interested in versions < 2.7 (some kind of test automation historians :D) they can get the old user guides.

laurentbristiel:
Just proposed a PR for User Guide updates.
For standard librares, I am OK to do it.
Should I just do another PR for this same issue or should I raise another issue?

pekkaklarck:
I think it's best to have just this one issue (already adjusted the title) and multiple PRs.

laurentbristiel:
I created 3 PR for this issue. That covers everything I found. If I did something the wrong way, or if I missed something, just tell me and I will adjust.

pekkaklarck:
I have merged PRs related to User Guide (#1899) and library docs (#1902) and fixed few problems I encountered in subsequent commits. I rejected the PR about acceptance test changes (#1903) because it didn't really provide any benefits and having references to ancient history there is kind of fun.

I consider this issue done. Thanks a lot @laurentbristiel! Next we need to find someone who would be interested to port the User Guide to [Sphinx](http://sphinx-doc.org).

PR #1899, opened 2015-01-23
Fix for #1897: clean-up reference to RF versions older than 2.7 in User Guide

laurentbristiel:


robotci:
Admin, can CI be enabled?

pekkaklarck:
Now that compiled User Guide locally, I noticed there's one small syntax error and also a reference to removed 'Listener interface versions' section. I'll fix those and a typo I noticed myself. User Guide changes should be tested by compiling it locally as explained at https://github.com/robotframework/robotframework/tree/master/doc/userguide

PR #1902, opened 2015-01-26
Fix for #1897: Clean-up reference to RF 2.6 and older from standard library docs

laurentbristiel:


robotci:
Admin, can CI be enabled?

PR #1903, opened 2015-01-27
Fix for #1897: clean-up reference to RF versions older than 2.7 in Acceptance Tests

laurentbristiel:


robotci:
Admin, can CI be enabled?

pekkaklarck:
I don't think this is worth doing. Actually I kind of like having references to ancient versions in acceptance tests. Similarly tests related to dates typically use the date when they were created. Kind of like a small internal joke.

PR #1909, opened 2015-02-05
Telnet: Keep track of telnet negotiation options

### LINKED ISSUE #1908: Telnet option negotiation loop
we are using the ser2net for a gateway between our network and the serial ports. If we use the telnet library to connect to ser2net we see endless IAC DONT/WONT/DO/WILL packets. Both processes run on the same machine and thus eats up the whole CPU.

I did a quick investigation and found out that this might the "nonterminating acknowledgment loop" which is mentioned in the telnet rfc854. I suspect this is because both sides (ser2net and the telnet library) are "dumb" implementations. Eg. they just answer by refusing/acknowledging the option. But the rfc says that:

> ```
>  b. If a party receives what appears to be a request to enter some
>  mode it is already in, the request should not be acknowledged.
>  This non-response is essential to prevent endless loops in the
>  negotiation.  It is required that a response be sent to requests
>  for a change of mode -- even if the mode is not changed.
> ```

Here is a short trace. Please note that this trace was done using the plain telnetlib, not the actual Telnet library.

> ```
>  # this is the initial sequence of ser2net
>  recv '\xff\xfb\x03\xff\xfb\x01\xff\xfe\x01\xff\xfd\x00'
>    IAC WILL 3  # this is the first command from the receive string from above
>  send DONT 3   # (1)
>    IAC WILL 1
>  send DONT 1
>    IAC DONT 1
>  send WONT 1
>    IAC DO 0
>  send WONT 0
>  recv '\xff\xfc\x03'  # (2) ser2net answers to (1)
>    IAC WONT 3
>  send DONT 3  # telnet library answers to (2), etc..
>  recv '\xff\xfc\x01\xff\xfe\x01\xff\xfe\x00'
>    IAC WONT 1
>  send DONT 1
>    IAC DONT 1
>  send WONT 1
>    IAC DONT 0
>  send WONT 0
>  recv '\xff\xfc\x03'
>    IAC WONT 3
>  send DONT 3
>  recv '\xff\xfc\x01\xff\xfe\x01\xff\xfe\x00'
>    IAC WONT 1
>  send DONT 1
>    IAC DONT 1
>  send WONT 1
>    IAC DONT 0
>  send WONT 0
>  recv '\xff\xfc\x03'
>    IAC WONT 3
>  send DONT 3
>  recv '\xff\xfc\x01\xff\xfe\x01\xff\xfe\x00'
>    IAC WONT 1
>  send DONT 1
>    IAC DONT 1
>  send WONT 1
>    IAC DONT 0
>  send WONT 0
>  recv '\xff\xfc\x03'
>    IAC WONT 3
>  send DONT 3
>  recv '\xff\xfc\x01\xff\xfe\x01\xff\xfe\x00'
>    IAC WONT 1
>  send DONT 1
>    IAC DONT 1
>  send WONT 1
>    IAC DONT 0
>  send WONT 0
>  [..]
> ```

mwalle:
To prevent endless negotiation loops with the telnet options, we must not
reply to already negotiated options.

### DIFF
--- a/src/robot/libraries/Telnet.py
+++ b/src/robot/libraries/Telnet.py
@@ -483,6 +483,7 @@ class TelnetConnection(telnetlib.Telnet):
         self._terminal_type = str(terminal_type) if terminal_type else None
         self.set_option_negotiation_callback(self._negotiate_options)
         self._set_telnetlib_log_level(telnetlib_log_level)
+        self._opt_responses = list()
 
     def set_timeout(self, timeout):
         """Sets the timeout used for waiting output in the current connection.
@@ -1051,6 +1052,15 @@ class TelnetConnection(telnetlib.Telnet):
             logger.write(msg, level or self._default_log_level)
 
     def _negotiate_options(self, sock, cmd, opt):
+        # We don't have state changes in our accepted telnet options.
+        # Therefore, we just track if we've already responded to an option. If
+        # this is the case, we must not send any response.
+        if cmd in (telnetlib.DO, telnetlib.DONT, telnetlib.WILL, telnetlib.WONT):
+            if (cmd, opt) in self._opt_responses:
+                return
+            else:
+                self._opt_responses.append((cmd, opt))
+
         # This is supposed to turn server side echoing on and turn other options off.
         if opt == telnetlib.ECHO and cmd in (telnetlib.WILL, telnetlib.WONT):
             self._opt_echo_on(opt)


robotci:
Admin, can CI be enabled?

mwalle:
unfortunately, i cannot provide a test case because this is too low level and doesn't fit into the "use a telnet server as test partner" method.

jussimalinen:
Looks good. Automatic test is too difficult probably, so we dont need to bother with that.

Could we just do this to simplify the code a bit:

```python

if cmd in (telnetlib.DO, telnetlib.DONT, telnetlin.WILL, telnetlib.WONT):
            if (cmd, opt) in self._opt_responses:
                return
            self._opt_responses.add((cmd, opt))

```

pekkaklarck:
@mwalle, any comments to comment by @jussimalinen? This looks trivial enough to still be included into RF 2.9 alpha 1.

mwalle:
sorry for the late response, i've some private matters atm.

anyway. yes, looks good, i've pushed a new version.

jussimalinen:
Excellent. I'm including this to alpha1 then after all

PR #1915, opened 2015-02-10
Exposing ERROR log level in the public logging api

jaloren:
Looking for the ability to log events at the ERROR level using robot's public logging api. However, that option didn't exist. This PR exposes this log level.

### DIFF
--- a/src/robot/api/logger.py
+++ b/src/robot/api/logger.py
@@ -33,9 +33,9 @@ using logger named ``RobotFramework``. This feature was added in RF 2.8.7.
 Log levels
 ----------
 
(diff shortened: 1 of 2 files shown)


robotci:
Admin, can CI be enabled?

pekkaklarck:
Logging errors has been designed only for internal usage. If you have good use cases, we can consider exposing it, though. In that case this PR would anyway need some tests and documentation in User Guide. The best way to continue is submitting an issue about this with explanation why this would be useful

jaloren:
Okay, not a problem. See issue 1916.

Issue #1818, opened 2014-10-08
Embedded arguments in keywords defined in test libraries


frishberg:
Following up on this:
https://groups.google.com/forum/#!msg/robotframework-users/J56wzoWU89I/Jxi3pL3G0-UJ
and this:
https://groups.google.com/forum/#!topic/robotframework-users/__rHImRK4aE

I think support can be added using the syntax proposed in the first thread.

Regarding this problem:
> A bigger problem would be how to support this with the static library
API. It's not possible to have a method with name like
"high_${number}" in the code so we needed to figure out some other
syntax. Any ideas?

Could we separate argument names with two underscores?
```
def method_with_argument__foo__:
    ...
```

----------
Note by @pekkaklarck: We have decided to implement this based on a custom name given to keywords using `robot_name` attribute that was implemented as #1835.

hellmanj:
@pekkaklarck, I spent some time trying to understand the inner workings I'd need to change but I think I could use that chat on IRC you offered if you have the time.  Can you let me know a few times that would be good for you?

pekkaklarck:
@hellmanj, like earlier, after 10pm my time (UTC+2) is probably best but before 4pm is often good too. If you don't have anything today or later this week, just try to ping me on IRC. Or should we agree some time beforehand so that others can also join?

I can understand that figuring out what's the best place to implement this functionality is not easy. The related code is old and bad in many ways. Luckily it's now good time to refactor it before a new major release.

hellmanj:
Others are certainly welcomed and encouraged, can we plan on Wednesday @ 10PM (UTC+2), 3PM (UTC-5) then?  It will give people time to ask us to reschedule.   Should we invite some people or is commenting here enough to draw attention?

pekkaklarck:
Wed 20:00 UTC sounds great. Not that many people is following this issue, so let's spread the word in case someone else is interested. Can you send a mail to robotframework-users and tell about the plan? I can mention it on IRC right away.

hellmanj:
Sure, will do.

pekkaklarck:
Yesterday's chat didn't go too well due to IRC problems @hellmanj had. Should we try again today same time?

hellmanj:
Yes, sorry again about that.  Let's do the same time today.

pekkaklarck:
Thanks for discussion today @hellmanj! Logs can be found from here:
http://koti.kapsi.fi/~peke/irclogs/robotframework/2015-01-29.log.html

[...]

PR #1917, opened 2015-02-10
library keyword with embedded arguments

hellmanj:
For Embedded arguments in keywords defined in test libraries #1818


robotci:
Admin, can CI be enabled?

pekkaklarck:
Crap, me reformatting error message when multiple keywords match a given name in a9968056a154a39c502305209280d9db3b16a3d6 apparently conflicts with this PR. Can you @hellmanj take a look is fixing that a big problem?

What's the status of this PR otherwise? I've been busy with issues targeted for 2.9 alpha 1 but on a quick look at the changes this looks pretty good already. Is there something still missing or should I do a bit more detailed review?

It would be great to get also this into the aforementioned alpha 1. I ought to get other remaining tasks done by next week. Does that sound like a good timeline for you?

hellmanj:
Fixed the merge conflict, no big deal.  I think it's ready for a detailed review whenever you're ready, I'd love to get it into the alpha.

src/robot/running/handlerstore.py
```diff
@@ -65,5 +65,5 @@ def _raise_no_single_match(self, name, found):
                             % (where, name))
         error = ["%s contains multiple keywords matching name '%s':"
                  % (where, name)]
-        names = sorted(handler.orig_name for handler in found)
+        names = seq2str([getattr(f, 'orig_name', None) or f.name for f in found])
```
pekkaklarck: If `f.name` always exists, could just pass it as the default value for `getattr`.

src/robot/running/testlibraries.py
```diff
@@ -182,7 +183,7 @@ def _create_handlers(self, libcode):
             if method:
                 handler = self._try_to_create_handler(name, method)
                 if handler:
-                    self.handlers.add(handler)
+                    self.handlers.add(handler, handler.__class__.__name__ == 'EmbeddedArgsTemplate')
```
pekkaklarck: 1. Wouldn't `isinstance(handler, EmbeddedArgsTemplate) ` work here instead of checking class name?

2. It would be a little cleaner to either to pass the Boolean argument with `embedded=` prefix or alternatively to extract a local variable containing the value.

src/robot/running/handlers.py
```diff
@@ -382,3 +382,26 @@ def _parse_arguments(self, handler_method):
...
         signatures = self._get_signatures(handler_method)
         return parser.parse(signatures, self.library.name)
+
+
+class EmbeddedArgsTemplate(object):
+
+    def __init__(self, embedded, orig_handler):
+        self.orig_handler = orig_handler
+        self._orig_run = orig_handler._run
+        self.orig_handler._run = self._run
```
pekkaklarck: This is a bit ugly but also very clever. A cleaner solution would likely require a bigger refactoring and I don't think that's worth the effort.

Only thing I'm a little worried about is that is there a change that the original handler would be used somehow without embedded arguments. In that case the original handler should use its normal `run` and `_run`. In practice it's very unlikely that would cause problems with existing keywords because normal keywords cannot contain `${}` in their name. Just deciding that such usage is not supported would probably be OK.

pekkaklarck:
Sorry for the delay reviewing this. Also alpha1 has been delayed a little so there's a good change to get this in.

I now only had time to really look at the code but it seems to be in very good shape. I was very pleasantly surprised how little changes were required.

I'd like to see some tests added for using embedded args with dynamic and also hybrid libraries. Based on the code they should work out-of-the-box, but it would be good to verify it. Adding some tests for Libdoc would be good too. I ought to have time for a bit more thorough test and doc review tomorrow.

pekkaklarck:
@robotci: Enable

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/70/
Test FAILed.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/71/
Test PASSed.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/72/
Test PASSed.

src/robot/running/handlers.py
hellmanj: I think not supporting that usage is fine for now, I'll be happy to help refactor if popular demand for that use case requires it.

src/robot/running/handlers.py
```diff
@@ -378,3 +378,26 @@ def _parse_arguments(self, handler_method):
...
+        self.embedded_name = embedded.name
+        self.embedded_args = embedded.args
+
+    def matches(self, name):
+        return self.embedded_name.match(name) is not None
+
+    def _run(self, context, args):
+        if args:
+            raise DataError("Positional arguments are not allowed when using embedded arguments")
+        return self._orig_run(context, [self] + list(self.embedded_args))
```
hellmanj: I added a test for the hybrid library which is failing due to a difference in # of expected arguments and I don't understand why.  Should I be somehow checking if the context is a hybrid library and not passing `self` if it is?

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/73/
Test FAILed.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/74/
Test FAILed.

src/robot/running/handlers.py
pekkaklarck: I don't actually understand why `[self]` should ever be passed there. I noticed functions used in tests had `self` as the first parameter but that seems just plain wrong.

It also seems old tests only had keywords implemented as functions in a module. Some tests for keywords implemented as methods in a class are needed too.

src/robot/running/handlers.py
```diff
@@ -378,3 +378,26 @@ def _parse_arguments(self, handler_method):
...
+
+    def create(self, name):
+        self.embedded_args = self.embedded_name.match(name).groups()
```
pekkaklarck: Re-setting `self.embedded_args` looks a bit weird because every call to `create` changes the state of the object. It may not actually cause problems, but it sure looks dangerous. It would be cleaner to return a new object similarly as `userkeywords.EmbeddedArgsTemplate` does. If that would require refactoring `_RunnableHandler` and/or other classes in this module, I'm willing to help.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/75/
Test PASSed.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/76/
Test PASSed.

pekkaklarck:
Looking really good! This is very close to be ready to be merged, but there are few more things I'd like to be tested:

1. Dynamic API. Ought to work out-of-the-box but would be good to add some tests to verify it.

2. Static API when library is implemented as a class. Same as above.

3. Defined embedded args not matching actual arguments. There should be a a clear error message (verified by tests) in cases like these:

```python
@keyword(name='Embedded ${args} are ${great}!')
def kw():
    # ...

class Example(object):
    @keyword(name='Embedded ${args} are ${great}!')
    def kw(self, arg):
        # ...
```

4. For consistency sake it would be good to test that it is OK for methods to accept more optional arguments than defined embedded arguments:

```python
@keyword(name='Embedded ${args} are ${great}!')
def kw(a=1, b=2, c=3):
    # ...

@keyword(name='Embedded ${args} are ${great}!')
def kw(*args):
    # ...
```

5. Library documentation. I think current implementation would include arguments of the method in docs generated by Libdoc. Should be tested and args somehow omitted if needed.

-----------

It's great that you squash commits to make the history cleaner, but it actually makes reviewing new changes harder. Would be easier to me to squash commits only just before merge.

hellmanj:
I think I addressed everything.  Good ideas all, but especially to test the dynamic library, it needed a change to work.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/77/
Test FAILed.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/78/
Test PASSed.

pekkaklarck:
Awesome! I'm on skiing holiday next week with a plan to do some work at evenings. Not sure how that will actually work out, so don't be surprised if it takes a little time before I have time to actually review this.

hellmanj:
No worries, enjoy your holiday :)

zephraph:
@pekkaklarck, have you had a chance to look at this anymore?

atest/testdata/keywords/embedded_arguments_library_keywords.robot
```diff
@@ -0,0 +1,122 @@
...
+    User Janne Selects x from webshop
+    User Janne Selects x fromwebshop
+
+Embedded Arguments Syntax is Underscore Sensitive
+    [Documentation]    FAIL No keyword with name 'User Janne Selects x from_webshop' found.
+    User Janne Selects x from webshop
+    User Janne Selects x from_webshop
+
+Keyword Matching Multiple Keywords In Library File
+    [Documentation]    FAIL Resource file 'embedded_args_in_lk_1' contains multiple keywords matching name 'foo+lib+bar-lib-zap':
```
pekkaklarck: Should say something like `Library ...` instead of `Resource file ...`.

atest/testdata/test_libraries/dynamic_libraries/EmbeddedArgs.py
```diff
@@ -0,0 +1,26 @@
...
+from robot.errors import DataError
+
+
+class EmbeddedArgs(object):
+
+    def callables(self):
+        return [getattr(self, f) for f in dir(self) if callable(getattr(self, f))]
+
+    def get_keyword_names(self):
+        return [f.robot_name for f in self.callables() if hasattr(f, 'robot_name')]
```
pekkaklarck: This is pretty complex. Couldn't this method just return `['Add ${count} Copies Of ${item} To Cart']`?

src/robot/running/testlibraries.py
```diff
@@ -23,7 +23,8 @@
 
 from .dynamicmethods import (GetKeywordArguments, GetKeywordDocumentation,
                              GetKeywordNames, RunKeyword)
-from .handlers import Handler, InitHandler, DynamicHandler
+from .handlers import Handler, InitHandler, DynamicHandler, EmbeddedArgsTemplate
+from .arguments import EmbeddedArguments
```
pekkaklarck: We try to keep imports sorted alphabetically.

src/robot/running/testlibraries.py
```diff
@@ -213,7 +214,12 @@ def _try_to_create_handler(self, name, method):
             self._report_adding_keyword_failed(name)
 
     def _create_handler(self, handler_name, handler_method):
-        return Handler(self, handler_name, handler_method)
+        handler = Handler(self, handler_name, handler_method)
+        if '$' in handler.name:
+            embedded = EmbeddedArguments(handler.name)
+            if embedded:
+                handler = EmbeddedArgsTemplate(embedded, handler)
+        return handler
```
pekkaklarck: This logic should be moved to `_create_handlers` (or a helper method used by it) in order to avoid duplicating the exact same logic in `_DynamicLibrary._create_handler`.

atest/robot/keywords/embedded_arguments_library_keywords.robot
```diff
@@ -0,0 +1,71 @@
+*** Settings ***
+Suite Setup     Run Tests  ${EMPTY}  keywords/embedded_arguments_library_keywords.robot
+Force Tags      regression  pybot
```
pekkaklarck: `jybot` tag missing.

pekkaklarck:
@jussimalinen and I are just now testing these changes locally. Assuming everything works fine, well merge changes to master and also fix problems I commented above.

pekkaklarck:
Changes merged and some enhancements done. Thanks a lot for your hard work @hellmanj!!

PR #1937, opened 2015-03-03
robot.utils: PYTHON2/3 checks and common types and type checks for future py3 support

userzimmermann:
As discussed in #1506 I started adding `PYTHON2/3` check vars and `is_<type>/<type>_like` check functions to `robot.utils`. I also added common `long`, `bytes` and `unicode` definitions and common `UserString/List/Dict` imports. If that stuff is ok this way I will add tests...

### DIFF
--- a/src/robot/utils/__init__.py
+++ b/src/robot/utils/__init__.py
@@ -51,7 +51,7 @@ from .match import eq, Matcher, MultiMatcher
 from .misc import (isatty, getdoc, plural_or_not, printable_name,
                    seq2str, seq2str2)
 from .normalizing import lower, normalize, NormalizedDict
-from .platform import IRONPYTHON, JYTHON, UNIXY, WINDOWS
+from .platform import PYTHON2, PYTHON3, IRONPYTHON, JYTHON, UNIXY, WINDOWS
 from .recommendations import RecommendationFinder
 from .robotenv import get_env_var, set_env_var, del_env_var, get_env_vars
 from .robotinspect import is_java_init, is_java_method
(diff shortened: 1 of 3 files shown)


robotci:
Admin, can CI be enabled?

pekkaklarck:
1. Do you have use cases for those new `is_xxx_like` methods and methods related to bytes? I think most important methods for Python 3 compatibility would be `is_string` and `is_integer`, but `is_unicode` and `is_number` might be useful too.

2. It would be good to also take the new methods into use to see what kind of effect they have and to see that they are actually useful.

3. Existing `is_str_like` should be renamed to `is_string_like` to be consistent with `is_string` and not to give an impression that it only checks against `str` type.

4. Why are `UserDict`, `UserList` and `UserString` exposed in `utills`? I don't think they should be generally used in the code base. The only current exception ought to be `UserDict` that is used as a base for `utils.NormalizedDict`, but also there we should use `collections.MutableMapping` now that we don't support Python 2.5 anymore. That's a good topic for a separate pull request if anyone is interested.

5. I thought a bit more about flags to use to indicate is Python 2 or 3 in use. I proposed `PYTHON2` and `PYTHON3` earlier to be compatible with existing `JYTHON` and `IRONPYTHON`, but actually the former are general language versions and the latter interpreter types. If/when Jython and IronPython start supporting Python 3, we should still be able to use same flags as with CPython and using `PYTHON2/3` would be a little weird. Perhaps using `PY2` and `PY3` with language versions similarly as `six` would be better. Any comments from others? If we aren't certain which to use, this particular PR can be implemented without addding "official" flags to `utils.platform` and by just using `PY3 = sys.version_info[0] == 3` in `utils.robottypes`.

userzimmermann:
I continued this topic in a separate branch and reopen as new PR #1988

pekkaklarck:
Great that you are continuing this work! Unfortunately I don't have time to look at the new PR right now. Should this one be closed nevertheless?

userzimmermann:
Yes. This one can be closed. I commented on your 5 points above in the new PR #1988

Issue #1938, opened 2015-03-03
remove obsolete robotframework-commit mailing list from user guide and .org site


laurentbristiel:
the mailing list [robotframework-commit](https://groups.google.com/forum/#!forum/robotframework-commit) is mentioned in:
* [User Guide](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#mailing-lists)
* [.org site Support and Contact](http://robotframework.org/#support-contact)

but this list seems to not be working anymore.
So we should probably remove references to this list.
(and maybe close the list itself).

I can do a PR for the cleaning. Quite straightforward.

pekkaklarck:
Yes, that list isn't used by any active project anymore. Go ahead.

laurentbristiel:
ok, I did 2 PR:
  * https://github.com/robotframework/robotframework/pull/1939
  * https://github.com/robotframework/robotframework.github.com/pull/6

pekkaklarck:
Both PRs merged. Thanks!

PR #1939, opened 2015-03-04
remove obsolete robotframework-commit mailing list from user guide (Fix for #1938)

laurentbristiel:


robotci:
Admin, can CI be enabled?

Issue #1945, opened 2015-03-10
Run Keyword If  True  MyKwWithReturnValue returns value when not executed


yahman72:
Is this correct behaviour?

Running this:
<pre>
foo
    ${var}=     Set Variable        defaultValue
    Log   ${var}   WARN
    ${var}=     Run Keyword If      False           Catenate  NonDefaultValue
    Log   ${var}   WARN
    ${var}=     Run Keyword If      True            Catenate  NonDefaultValue
    Log   ${var}   WARN
</pre>

Prints this:
<pre>
[ WARN ] defaultValue
[ WARN ] None
[ WARN ] NonDefaultValue
</pre>

I would have expected to get:
<pre>
[ WARN ] defaultValue
<b>[ WARN ] defaultValue</b>
[ WARN ] NonDefaultValue
</pre>

I would have expected that the `${var}` is not modified if the condition does not match.

If this is desired (or it's too late to change this for backwards compatibility) then the documentation should mention something about this.

laurentbristiel:
That's the way it works indeed.
To use "run keyword if" in variable assignments, you should always have a ELSE statement (or handle the fact that you will get a None when the IF does not pass).
Currently, the [keyword documentation](http://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Run%20Keyword%20If) does contain an example like that:
> Using ELSE and/or ELSE IF branches is especially handy if you are interested in the return value. This is illustrated by the example below that also demonstrates using ELSE IF and ELSE together:...

Do you think we should be more explicit and state that without ELSE we might get None?

yahman72:
yes, this needs to be documented better - I just spend quite a long time figuring out why my default value is replaced with "none" :-)

pekkaklarck:
I find it pretty natural that the return value of `Run Keyword If` is `None` if no keyword is actually executed. Documentation can still be enhanced to make that more explicit. Any suggestions how to improve it?

laurentbristiel:
how about PL https://github.com/robotframework/robotframework/pull/1946?

yahman72:
It's natural indeed after thinking about it for a while - at first I considered `Run Keyword If` as a control structure (i.e. "IF"-loop) not as a keyword (i.e. a function that has to return something). Now it sounds logical, because it is a keyword. 
Still there should be an example of this in the doc (like there are examples in the `Set Variable If`, `Get Variable Value` KWs).
[...]

PR #1946, opened 2015-03-11
Enhance doc to be more explicit on Run Keyword If return value. See #1945

laurentbristiel:


robotci:
Admin, can CI be enabled?

laurentbristiel:
will give this another try

PR #1948, opened 2015-03-12
check for detached stream buffer in utils.misc.isatty()

userzimmermann:
If the `.buffer` of an `io` module based stream was detached, the `stream.isatty()` method called in `utils.misc.isatty()` will raise a `ValueError`. As `utils.misc.isatty()` is usually only applied to `sys.stdout` and `sys.stderr` it most likely won't happen in __Python 2__. But who knows what other streams this function might be used with or what other projects might do with `sys.std...` before importing `robot`. And the buffer check is definitely needed for __Python 3__ compatibility. I encountered it recently on adding __Sphinx__ based docs to my __robotframework-tools__ project and running `setup.py build_sphinx` with __Python 3__, using a _conf.py_ which imports `robot`. And just imprting `robot` is enough, because the creation of `robot.output.logger.LOGGER` leads to the use of `isatty()`. Maybe my fix is not at the best place and streams should be checked earlier...

### DIFF
--- a/utest/utils/test_misc.py
+++ b/utest/utils/test_misc.py
@@ -1,8 +1,10 @@
+import io
+import sys
 import unittest
 
-from robot.utils.misc import getdoc, printable_name, seq2str
-from robot.utils.asserts import assert_equals
-from robot.utils import IRONPYTHON
+from robot.utils import IRONPYTHON, JYTHON
+from robot.utils.asserts import assert_equals, assert_false, assert_raises
+from robot.utils.misc import getdoc, isatty, printable_name, seq2str
 
 
 class TestMiscUtils(unittest.TestCase):
@@ -41,6 +43,31 @@ class TestMiscUtils(unittest.TestCase):
                          ('foo-bar', 'Foo-bar'),
                          ('', '')]:
             assert_equals(printable_name(inp, code_style=True), exp)
+
+    def test_isatty_with_stdout(self):
+        # file class based in PY2, io module based in PY3
+        assert_equals(isatty(sys.__stdout__), sys.__stdout__.isatty())
+
+    def test_isatty_with_io(self):
+        with io.StringIO() as stream:
+            assert_false(isatty(stream))
+            wrapper = io.TextIOWrapper(stream)
+            assert_false(isatty(wrapper))
+
+    def test_isatty_with_detached_io_buffer(self):
+        # make sure that isatty() checks for detached io stream buffers
+        # (otherwise it would raise an Exception)
+        with io.StringIO() as stream:
+            wrapper = io.TextIOWrapper(stream)
+            if sys.version_info >= (2, 7):
+                wrapper.detach() #==> wrapper.buffer is None
+                # Jython 2.7 behaves like CPython 2.6 in that case
+                exc_type = ValueError if not JYTHON else AttributeError
+            else: # no .detach and different behaviour if .buffer is None
+                wrapper.buffer = None
+                exc_type = AttributeError
+            assert_raises(exc_type, wrapper.isatty)
+            assert_false(isatty(wrapper))
 
 
 class TestGetdoc(unittest.TestCase):
--- a/src/robot/utils/misc.py
+++ b/src/robot/utils/misc.py
@@ -110,6 +110,9 @@ def getdoc(item):
 if not IRONPYTHON:
 
     def isatty(stream):
+        # first check if buffer was detached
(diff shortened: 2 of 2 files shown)


robotci:
Admin, can CI be enabled?

pekkaklarck:
1. The change itself looks good to me.

2. Is it possible to create a unit test for this?

3. Do you have references to docs about `io' based streams raising `ValueError`? I would like to study this topic a bit more myself.

4. @robotci: Enable

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/79/
Test PASSed.

userzimmermann:
I've added a utest in utils/test_misc

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/81/
Test FAILed.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/82/
Test FAILed.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/83/
Test PASSed.

utest/utils/test_misc.py
```diff
@@ -1,7 +1,11 @@
 import unittest
```
pekkaklarck: Nitpicking: No need to have an empty row between standard library imports.

utest/utils/test_misc.py
```diff
@@ -1,7 +1,11 @@
...
 
-from robot.utils.misc import getdoc, printable_name, seq2str
-from robot.utils.asserts import assert_equals
+import sys
+import io
+
+from robot.utils.misc import getdoc, printable_name, seq2str, isatty
```
pekkaklarck: Even more nitpicking: We prefer organizing imports alphabetically i.e. `isatty` should be after `getdoc`. Clearly this file doesn't fully follow that rule otherwise either, because `robot.utils.asserts` line should be before `robot.utils.misc`.

utest/utils/test_misc.py
```diff
@@ -42,6 +46,23 @@ def test_printable_name_with_code_style(self):
...
+        # make sure that isatty() checks for detached buffers
+        # (otherwise it would raise a ValueError)
+        if sys.version_info >= (2, 7):
+            wrapper.detach() #==> wrapper.buffer is None
+            assert_raises(ValueError, lambda: wrapper.isatty())
+        else: # no .detach and different .isatty behaviour if .buffer is None
+            wrapper.buffer = None
+            assert_raises(AttributeError, lambda: wrapper.isatty())
+        assert_false(isatty(wrapper))
+        stream.close()
```
pekkaklarck: +1 for a very thorough testing. It would be better to split this into multiple tests, one for each stream type. You can either keep them in the original test class and name them like `test_isatty_with_sys_stdout`, `test_isatty_with_stringio` or, preferably, create new `TestIsatty` class with individual tests like `test_sys_stdout`, `test_stringio`.

pekkaklarck:
Looks very good to me. Few more comments, though:

- As I already commented in a line note, it would be better to have individual tests for different stream types.
- Have you tested this using Jython and IronPython? The latter has a totally different `isatty` implementation that shouldn't be affected, but testing with Jython 2.7 would be great.
- Was there already an issue about this? Although this is a pretty minor issue, it would be good to have one for reference.

pekkaklarck:
@userzimmermann, have you had a change to look at this lately? I'm going through PRs today and would like to merge as many as I can. This one looks pretty good already.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/85/
Test PASSed.

userzimmermann:
- Did even further split and cleaned up imports
- Added Jython 2.7 support
- No existing issue

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/86/
Test PASSed.

pekkaklarck:
This PR looks good to be merged. Could you still [squash commits](http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit) and submit an issue about this?

userzimmermann:
As you like it ;) #2003

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/87/
Test PASSed.

pekkaklarck:
Thanks @userzimmermann!

PR #1947, opened 2015-03-11
Run Keyword If doc updates

yahman72:
initiated by #1945

PR #1949, opened 2015-03-13
Enhance doc to be more explicit on Run Keyword If return value. See #1945

laurentbristiel:
Trying to take the best of https://github.com/robotframework/robotframework/pull/1947 and https://github.com/robotframework/robotframework/pull/1946


robotci:
Admin, can CI be enabled?

yahman72:
looks good to me

Issue #1970, opened 2015-04-14
Enhance ROBOT_LIBRARY_LISTENER to accept a list of listeners


vkruoso:
It might be interesting to allow library listeners to register more than one listener instance with the ```ROBOT_LIBRARY_LISTENER``` interface. Please refer to issue #1969 for mode details.

vkruoso:
I've created pull request #1995 for this. Waiting for your feedback.

PR #1971, opened 2015-04-14
Support list of instances at ROBOT_LIBRARY_LISTENER

vkruoso:
This is a pull request for issue #1970.

Are missing tests and docs, but wanted to get this checked before moving on.

I've locally executed the tests at ```atests/robot/output/listener_interface/``` using Python (2.6.5) and Jython (was only able to run with 2.7rc2) to be sure this did not break anything.


robotci:
Admin, can CI be enabled?

PR #1986, opened 2015-05-04
update link to Quick Start Guide now that it moved to GitHub

laurentbristiel:


robotci:
Admin, can CI be enabled?

pekkaklarck:
Good catch!

Issue #1969, opened 2015-04-13
Allow giving listener and model modifier instances to `robot.run` and `TestSuite.run`


spikergit1:
As briefly discussed on IRC, I have a need to register an instance of a listener object at run-time, programmatically.  I am wrapping Robot in another application, running ``urwid``.  Robot is running in a thread, and I want to update a widget based on Robot's current status.  A listener would be the obvious choice for an interface between the applications.

Currently, most classes related to Listeners and logger proxies assume that the wrapped object is a string and must be imported.  I have overridden some of the core classes to allow instances to be passed in.

pekkaklarck:
Would it be enough that in addition of

```python
from robot import run
run('tests.robot', listener='listener.py')
```

you could also do

```python
from robot import run
run('tests.robot', listener=Listener())
```

I know I would have liked the above functionality few times. It shouldn't be too complicated to implement either.

vkruoso:
I'm also interested on a feature like this, but my use case is a little bit different. I've sent an email in the users list: "Add listener instance using robot.api".

The idea is to be able to register a listener instance at any point in robot execution. In my use case it would be in the initialization method of robot's dynamic library (that library works as a proxy).

pekkaklarck:
@vkruoso, libraries can register listeners since RF 2.8.5 (#811). Does that work for you?

vkruoso:
@pekkaklarck, I'm already using that interface. Let me be a little more clear.

```python
class Proxy(object):
    ROBOT_LISTENER_API_VERSION = 2
    def __init__(self):
        self.ROBOT_LIBRARY_LISTENER = self
```
So the Proxy library act as a listener. Great. Now, in the \_\_init\_\_ method (could be in any other), I create a bunch of instances. So what I have is something like:

```python
(instances, keywords) = Loader().load_libraries()
```

Now my wish was to register each instance as a listener calling a robot API function (and was also hopping it would check for the ROBOT* attributes). It could be a BuiltIn() method/keyword? ```register_listener``` for instance.

It might be interesting to restrict scope for those instances so it receives only events from the suite they have been registered.

pekkaklarck:
[...]

PR #1987, opened 2015-05-08
Allow to pass listener instances to robot.run

WojtekKowaluk:
implements https://github.com/robotframework/robotframework/issues/1969

### DIFF
--- a/src/robot/conf/settings.py
+++ b/src/robot/conf/settings.py
@@ -19,6 +19,7 @@ import time
 
 from robot import utils
 from robot.errors import DataError, FrameworkError
+from robot.utils import split_args_from_name_or_path
 from robot.output import LOGGER, loggerhelper
 from robot.result.keywordremover import KeywordRemover
 from robot.result.flattenkeywordmatcher import validate_flatten_keyword
@@ -75,7 +76,7 @@ class _BaseSettings(object):
     def _process_cli_opts(self, opts):
         for name, (cli_name, default) in self._cli_opts.items():
             value = opts[cli_name] if cli_name in opts else default
-            if default == [] and isinstance(value, basestring):
+            if default == [] and not utils.is_list_like(value):
                 value = [value]
             self[name] = self._process_value(name, value)
         self['TestNames'] += self['ReRunFailed'] or self['DeprecatedRunFailed']
@@ -116,9 +117,8 @@ class _BaseSettings(object):
             return utils.abspath(value)
         if name in ['SuiteStatLevel', 'MonitorWidth']:
             return self._convert_to_positive_integer_or_default(name, value)
-        if name in ['PreRunModifiers', 'PreRebotModifiers', 'Listeners',
-                    'VariableFiles']:
-            return [self._split_args_from_name_or_path(item) for item in value]
+        if name in ['PreRunModifiers', 'PreRebotModifiers', 'VariableFiles']:
+            return [split_args_from_name_or_path(item) for item in value]
         if name == 'ReportBackground':
             return self._process_report_background(value)
         if name == 'TagStatCombine':
@@ -278,30 +278,6 @@ class _BaseSettings(object):
 
     def _get_default_value(self, name):
         return self._cli_opts[name][1]
-
-    def _split_args_from_name_or_path(self, name):
-        if os.path.exists(name):
-            return os.path.abspath(name), []
-        index = self._get_arg_separator_index_from_name_or_path(name)
-        if index == -1:
-            return name, []
-        args = name[index+1:].split(name[index])
-        name = name[:index]
-        if os.path.exists(name):
-            name = os.path.abspath(name)
-        return name, args
-
-    def _get_arg_separator_index_from_name_or_path(self, name):
-        colon_index = name.find(':')
-        # Handle absolute Windows paths
-        if colon_index == 1 and name[2:3] in ('/', '\\'):
-            colon_index = name.find(':', colon_index+1)
-        semicolon_index = name.find(';')
-        if colon_index == -1:
-            return semicolon_index
-        if semicolon_index == -1:
-            return colon_index
-        return min(colon_index, semicolon_index)
 
     def _validate_remove_keywords(self, values):
         for value in values:
--- a/src/robot/output/listeners.py
+++ b/src/robot/output/listeners.py
@@ -64,14 +64,14 @@ class Listeners(object):
 
     def _import_listeners(self, listener_data):
         listeners = []
-        for name, args in listener_data:
+        for listener in listener_data:
             try:
-                listeners.append(ListenerProxy(name, args))
+                listeners.append(ListenerProxy(listener))
             except DataError as err:
-                if args:
-                    name += ':' + ':'.join(args)
+                if not isinstance(listener, basestring):
+                    listener = utils.type_name(listener)
                 LOGGER.error("Taking listener '%s' into use failed: %s"
-                             % (name, unicode(err)))
+                             % (listener, unicode(err)))
         return listeners
 
     def start_suite(self, suite):
@@ -222,8 +222,12 @@ class ListenerProxy(AbstractLoggerProxy):
                 'output_file', 'report_file', 'log_file', 'debug_file',
                 'xunit_file', 'close']
 
-    def __init__(self, name, args):
-        listener = self._import_listener(name, args)
+    def __init__(self, listener):
+        if isinstance(listener, basestring):
+            name, args = utils.split_args_from_name_or_path(listener)
+            listener = self._import_listener(name, args)
+        else:
+            name = utils.type_name(listener)
         AbstractLoggerProxy.__init__(self, listener)
         self.name = name
         self.version = self._get_version(listener)
--- a/src/robot/run.py
+++ b/src/robot/run.py
@@ -464,6 +464,9 @@ def run(*datasources, **options):
     passed as lists like `include=['tag1', 'tag2']`. If such option is used
     only once, it can be given also as a single string like `include='tag'`.
 
+    Additionally listener option allows passing object directly instead of
+    listener name, e.g. `run('tests.robot', listener=Listener())`.
+
     To capture stdout and/or stderr streams, pass open file objects in as
     special keyword arguments `stdout` and `stderr`, respectively.
 
--- a/src/robot/running/model.py
+++ b/src/robot/running/model.py
@@ -105,6 +105,9 @@ class TestSuite(model.TestSuite):
         If such an option is used only once, it can be given also as a single
         string like ``variable='VAR:value'``.
 
+        Additionally listener option allows passing object directly instead of
+        listener name, e.g. `run('tests.robot', listener=Listener())`.
+
         To capture stdout and/or stderr streams, pass open file objects in as
         special keyword arguments `stdout` and `stderr`, respectively. Note
         that this works only in version 2.8.4 and newer.
--- a/utest/api/test_run_and_rebot.py
+++ b/utest/api/test_run_and_rebot.py
@@ -14,6 +14,8 @@ from robot.utils.asserts import assert_equals, assert_true
 from robot.running import namespace
 from robot import run, rebot
 from resources.runningtestcase import RunningTestCase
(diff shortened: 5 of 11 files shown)


robotci:
Admin, can CI be enabled?

src/robot/output/listeners.py
```diff
@@ -221,15 +221,21 @@ class ListenerProxy(AbstractLoggerProxy):
...
+        listener_instance = self._get_listener(listener, args)
+        AbstractLoggerProxy.__init__(self, listener_instance)
+        self.name = str(listener)
+        self.version = self._get_version(listener_instance)
         if self.version == 1:
             LOGGER.warn("Listener '%s' uses deprecated API version 1. "
                         "Switch to API version 2 instead." % self.name)
 
+    def _get_listener(self, listener, args):
+        if isinstance(listener, str):
```
pekkaklarck: Should check against `basetring` to allow passing listener name as `str` and `unicode`.

src/robot/output/listeners.py
```diff
@@ -221,15 +221,21 @@ class ListenerProxy(AbstractLoggerProxy):
...
 
-    def __init__(self, name, args):
-        listener = self._import_listener(name, args)
-        AbstractLoggerProxy.__init__(self, listener)
-        self.name = name
-        self.version = self._get_version(listener)
+    def __init__(self, listener, args):
+        listener_instance = self._get_listener(listener, args)
+        AbstractLoggerProxy.__init__(self, listener_instance)
+        self.name = str(listener)
```
pekkaklarck: Does this really work? Don't you just get the string representation of the listener instance which, by default, is just something like `<module.Class object at 0x7f3519d63590>`?

src/robot/conf/settings.py
```diff
@@ -115,7 +115,12 @@ def _process_value(self, name, value):
             return utils.abspath(value)
         if name in ['SuiteStatLevel', 'MonitorWidth']:
             return self._convert_to_positive_integer_or_default(name, value)
-        if name in ['Listeners', 'VariableFiles']:
+        if name == 'Listeners':
+            if isinstance(value, list):
+                return [self._split_args_from_name_or_path(item) for item in value]
+            else:
+                return [(value, [])]
```
pekkaklarck: Do I get it right that it is possible to give only a single listener object but not a list of them? Fixing that would require changes to how individually given options are turned into a list. Currently there's a check

    if default == [] and isinstance(value, basestring):

on line 77 in this module. Changing that to

    if default == [] and not is_list_line(value):

would probably be enough. This obviously should also be tested.

src/robot/conf/settings.py
```diff
@@ -115,7 +115,12 @@ def _process_value(self, name, value):
...
+        if name == 'VariableFiles':
             return [self._split_args_from_name_or_path(item) for item in value]
```
pekkaklarck: Hmm, it would actually be useful to be able to pass also variable files as objects. Definitely a subject to a separate enhancement request, though.

pekkaklarck:
This looks very good! I especially like that you have added some unit tests for this functionality. There are, however, few issues still to be sorted out. I already added some line notes, but there are few additional things to do or at least consider:

- This functionality must be documented. Most logical place to document this is in the docstring of `robot.run` method.

- Should consider testing this also with `running.TestSuite.run` method. Testing this on higher level with `robot.run` is probably adequate as I can't see how this could be broken in `TestSuite.run`, but we have added many somewhat duplicate tests to `utest/runing/test_running.py` to make it certain both APIs work.

- Should also consider documenting this on `TestSuite.run` level. There it might actually be better to just add a note that options have same semantics as options passed to `robot.run`. That would avoid duplicate documentation as well as forgetting to document something similar in both places in the future.

pekkaklarck:
@robotci: Whitelist

WojtekKowaluk:
hello, I have improved code according to your comments. I have added test for TestSuite.run and seems there is a problem. When listeners are registered in one TestSuite.run call, they are still active in following calls. I am trying to figure out what is root cause of this behaviour.

WojtekKowaluk:
I just checked this with master branch (with .py file module listeners) and behaviour is the same. Seems listener are not unregistered correctly after TestSuite.run call. I will investigate it more unless you have some idea.

pekkaklarck:
Interesting. Sounds like a bug to me. We've had similar problems with preserved state when using programmatic APIs also earlier. If you have confirmed this, please submit a separate issue about it. If you have found a fix, please also submit a separate PR.

WojtekKowaluk:
I have added bit of documentation. Also created issue about listeners not being unregistered between TestSuite.run calls: https://github.com/robotframework/robotframework/issues/1992

src/robot/conf/settings.py
```diff
@@ -133,6 +134,12 @@ def _process_value(self, name, value):
             self._validate_flatten_keywords(value)
         return value
 
+    def _process_listener_value(self, value):
+        if isinstance(value, basestring):
+            return self._split_args_from_name_or_path(value)
+        else:
+            return (value, [])
```
pekkaklarck: I understand that you only wanted to support listeners, but I nevertheless think it would be OK to just modify `_split_args_from_name_or_path` to have this at the beginning:

```python
if not isinstance(value, basestring):
    return value, []
```

If someone passes non-string value to other settings, things will break anyway. A big benefit would be that I could add support for instances to model modifiers separately.

src/robot/output/listeners.py
```diff
@@ -64,14 +64,14 @@ def __nonzero__(self):
...
         listeners = []
-        for name, args in listener_data:
+        for listener, args in listener_data:
             try:
-                listeners.append(ListenerProxy(name, args))
+                listeners.append(ListenerProxy(listener, args))
             except DataError as err:
                 if args:
-                    name += ':' + ':'.join(args)
+                    listener += ':' + ':'.join(args)
```
pekkaklarck: This breaks if `listener` is isn't a s string. I'm not sure is it possible that creating `ListenerProxy` could actually fail in that case, though.

src/robot/run.py
```diff
@@ -464,6 +464,9 @@ def run(*datasources, **options):
     passed as lists like `include=['tag1', 'tag2']`. If such option is used
     only once, it can be given also as a single string like `include='tag'`.
 
+    Additionally listener option allows to pass object directly instead of
+    module name, e.g. `run('tests.robot', listener=Listener())`.
```
pekkaklarck: Listeners aren't typically implemented as a module. It would be better to change `module name` to `listener name`. Also `allows to pass` doesn't sound that good. Perhaps `allows passing` would be better.

src/robot/running/model.py
```diff
@@ -105,6 +105,9 @@ def run(self, settings=None, **options):
         If such an option is used only once, it can be given also as a single
         string like ``variable='VAR:value'``.
 
+        Additionally listener option allows to pass object directly instead of
+        module name, e.g. `run('tests.robot', listener=Listener())`.
```
pekkaklarck: Same comment here as in `run.py`.

utest/api/test_run_and_rebot.py
```diff
@@ -103,6 +105,19 @@ def test_listener_gets_notification_about_log_report_and_output(self):
                                      ('[log {}]'.format(LOG_PATH), 1),
                                      ('[listener close]', 1)])
 
+    def test_pass_listener_as_instance(self):
+        assert_equals(run(self.data, outputdir=TEMP, report='none', listener=Listener(1)), 1)
+        self._assert_outputs([("[from listener 1]", 1)])
```
pekkaklarck: Good tests! I'd like to enhance this one so that the listener would store some information into itself when it's methods are called and that changed state could be verified after execution. Perhaps the `Listener._log` could be changed to:

```python
    def _log(self, message):
        sys.__stdout__.write("[{}]\n".format(message))
        self.messages.append(message)
```

and that verified later with `assert_equals(listener.message, [...]`.

utest/api/test_run_and_rebot.py
```diff
@@ -103,6 +105,19 @@ def test_listener_gets_notification_about_log_report_and_output(self):
...
+
+    def test_pass_listener_as_string(self):
+        module_file = join(ROOT, 'utest', 'resources', 'Listener.py')
+        assert_equals(run(self.data, outputdir=TEMP, report='none', listener=module_file+":1"), 1)
+        self._assert_outputs([("[from listener 1]", 1)])
+
+    def test_pass_listener_as_list(self):
+        module_file = join(ROOT, 'utest', 'resources', 'Listener.py')
+        assert_equals(run(self.data, outputdir=TEMP, report='none', listener=[module_file+":1", Listener(2)]), 1)
+        self._assert_outputs([("[from listener 1]", 1), ("[from listener 2]", 1)])
```
pekkaklarck: Should have two empty lines between classes. #nitpick

utest/running/test_running.py
```diff
@@ -254,6 +255,12 @@ def test_reset_logging_conf(self):
 
 class TestListeners(RunningTestCase):
 
+    def test_listeners(self):
+        module_file = join(ROOTDIR, 'utest', 'resources', 'Listener.py')
+        suite = build('setups_and_teardowns.robot')
+        suite.run(output=None, log=None, report=None, listener=[module_file+":1", Listener(2)])
+        self._assert_outputs([("[from listener 1]", 1), ("[from listener 2]", 1)])
```
pekkaklarck: Also here it would be good to test also listener's changed state.

pekkaklarck:
This looks great! Please take a look at the line notes and comment if you disagree with them.

One thing I don't like so much anymore in the overall design is that in settings listeners etc. would be a tuple `(name, args)` or `(instance, [])`. I'm starting to think it would be better to just keep listener etc. there as it is given, and handle values where they are actually needed. To avoid duplication, we could move the logic currently in `_split_args_from_name_or_path` to a general utility function. After that `__init__` of `ListenerProxy` could be changed to something like this:

```python
def __init__(self, listener):
    if isinstance(listener, basestring):
        name, args = utils.split_args_from_name_or_path(listener)
        listener = self._import_listener(name, args)
    else:
        name = utils.type_name(listener)
    # ...
```

I think this would be better design than what we decided earlier. What do you think @WojtekKowaluk? If you agree, should we switch to that design immediately or first implement this using the original design and then refactor?

WojtekKowaluk:
I also think this current solution is not very elegant. I will attempt to refactor it like you say.

pekkaklarck:
Great! A good first step would be just creating the `split_args_from_name_or_path` utility function. It probably should go to `utils/text.py` where we recently added `split_tags_from_doc`. Old unit test class `TestSplitArgsFromNameOrPath` should also be moved, with needed modifications, from `test_settings.py` to `test_text.py`.

WojtekKowaluk:
I have refactored the code. I also went with your other suggestion except adding state to test listener (just skipped it for now, I'm not saying it's bad idea). What do you think?

src/robot/conf/settings.py
```diff
@@ -116,9 +117,10 @@ def _process_value(self, name, value):
             return utils.abspath(value)
         if name in ['SuiteStatLevel', 'MonitorWidth']:
             return self._convert_to_positive_integer_or_default(name, value)
-        if name in ['PreRunModifiers', 'PreRebotModifiers', 'Listeners',
-                    'VariableFiles']:
-            return [self._split_args_from_name_or_path(item) for item in value]
+        if name == 'Listeners':
+            return value
```
pekkaklarck: This isn't needed if nothing special is done the value.

WojtekKowaluk:
Fixed.

pekkaklarck:
Looks great! Please squash the commits and I'm happy to merge this. Do you want to continue w/ var files and modifiers or should I take care of them?

WojtekKowaluk:
Commits are squashed. I can continue with var files and visitors in new PR if you don't mind to wait for it a bit.

pekkaklarck:
PR merged. Thanks @WojtekKowaluk!

I'm currently busy with other tasks so it would be great if you could take a look at modifiers and variable files too. I'd like to get them included and the whole #1969 closed before alpha 3, tough. We try to get it out already next week, but if you don't have time for this before I can finish the work.

PR #1988, opened 2015-05-08
PY2/3 compatible is_<type>() and is_<type>_like() checks

userzimmermann:
This is a replacement for #1937 in a separate branch.
I also applied `is_integer`, `is_number`, `is_bytes`, `is_string` and `is_unicode` to the whole code base, replacing all __PY3__-incompatible `isinstance` checks.

Regarding the latest remarks from the original PR:

1. I just added them for completeness/consistency/whatever ;) Without actually thinking about the real need for all of them. But all of the `is_<type>` methods became actually useful.

2. Done.

3. There are both.

4. `UserDict` is used. And the rest... completeness again ;) `UserList` and `UserString` are used in the tests. But that could also be handled exclusively in the tests...

5. I removed `PYTHON2/3` again and just check locally in `utils.robottypes`


robotci:
Admin, can CI be enabled?

pekkaklarck:
I don't have time to review such a big PR right now, but I opened a separate issue #1990 tracking this and assigned it to RF 2.9 alpha 2.

pekkaklarck:
@robotci: Whitelist

src/robot/conf/settings.py
```diff
@@ -18,6 +18,7 @@
 import time
 
 from robot import utils
+from robot.utils import is_string
```
pekkaklarck: Other utils are used like `utils.function()` in this module. For consistency reasons it would be better to use also `utils.is_string()` instead of just `is_string()`.

Unfortunately different modules in our code base uses utils like `utils.is_string()` and `is_string()`. The reason is that we used to the former earlier and have been recently switching to latter style but haven't converted all modules to use the that style. Inconsistency between modules is obviously bad, but I consider inconsistency within a module is even worse.

What do you think, should we first convert all modules to use `is_string` style or would you be willing to change this PR to use `utils.is_string` style in modules that still use that style?

src/robot/utils/robottypes.py
```diff
@@ -33,6 +50,50 @@ def is_str_like(item):
...
+if PY3:
+    is_unicode = is_string
+    is_unicode_like = is_string_like
+else:
+    def is_unicode(item):
+        return isinstance(item, unicode)
+
+    is_unicode_like = is_unicode
+
+
```
pekkaklarck: This if/else logic looks overly complicated. What about adding separate `python2types` and `python3types` modules and having this in `robottypes`:

```python
if PY3:
    from python3types import is_string, is_integer, ...
else:
    from python2types import is_string, is_integer, ...
```

This approach has worked great when dealing with differences between CPython, Jython and IronPython.

src/robot/utils/robottypes.py
```diff
@@ -12,9 +12,26 @@
...
+    from UserDict import UserDict
+    from UserString import UserString, MutableString
+if PY3:
+    long = int
+    bytes = bytes
+    unicode = str
+else:
+    long = long
+    bytes = str if not IRONPYTHON else bytes
+    unicode = unicode
```
pekkaklarck: Why are these created and exposed? I don't see them used anywhere in the code base.

src/robot/utils/__init__.py
```diff
@@ -60,7 +60,13 @@
                         get_time, get_timestamp, secs_to_timestamp,
                         secs_to_timestr, timestamp_to_secs, timestr_to_secs,
                         parse_time)
-from .robottypes import is_dict_like, is_list_like, is_str_like, type_name
+from .robottypes import (long, bytes, unicode,
+                         UserString, UserList, UserDict,
```
pekkaklarck: Why are long, bytes, unicode, UserString, UserList, UserDict exposed? I didn't see them used anywhere in the code base except for `UserDict` as a base of `NormalizedDict`. `UserDict` can be conditionally imported in `normalizing` and should actually be replaced with `MutableMapping` now that we don't support Python 2.5 anymore.

pekkaklarck:
I really like these new utils. `is_string(thing)` and `is_number(thing)` read better and are shorted than `isinstance(thing, basestring)` and `isinstance(thing, (int, long, float))`. I'd be happy to merge these changes even without the added Python 3 compatibility.

See the line notes for additional comments.

pekkaklarck:
I looked at this with @jussimalinen and we agreed that `is_xxx` utils are so much cleaner than `isinstance` that we wanted to get this into RF 2.9. We merged this PR, fixed conflicts and fixed issues I commented earlier.

src/robot/utils/robottypes.py
userzimmermann: `long` and `unicode` are actually used in the code. And I often do `if PY3: long = int` and `from six import text_type as unicode` on demand in #1506 to make these cases work. I also changed some `str` uses to `bytes` for compatibility there.

src/robot/utils/__init__.py
userzimmermann: They are used in the tests...

userzimmermann:
Thanks for merging this stuff! Even with the code removals in `robottypes` ;) Unfortunately the Python 3 compatibility has gone due to the latter. But I did a recent merge of my Python 3 fork with 2.9rc1 for #1506 and found a simpler way for restoring it.

Regarding your question 3 above: I would just try to avoid it ;) I will check it anyway on doing merges to my Python 3 fork...

Issue #1954, opened 2015-03-28
New syntax: FOR ... IN ENUMERATE


edbrannin:
Suggested in the comments on #1952

It would be nice if Robot could use Python's `enumerate()` in for loops.

edbrannin:
About how long to I have to make a PR for this to get it considered for 2.9a2?

pekkaklarck:
Could you please elaborate the question?

edbrannin:
I meant: Obviously sooner is better, but I'm curious how soon I'd have to complete any pull-requests that make it into the next 2.9 alpha release.

pekkaklarck:
We don't have any hard deadline for alpha 2. Basically we'll create a release when all issues with alpha 2 label have been closed or postponed. I hope that would happen already this month because we'd like to get the final release ready in June. Missing alpha 2 shouldn't be a huge problem, though, because we'll continue with alpha 2 or beta 1 right after it.

pekkaklarck:
Now that PR #1989 is implementing also `FOR ... IN ENUMERATE`, should we merge this issue with #1952?

pekkaklarck:
Changed #1952 to cover both ZIP and ENUMERATE flavors. This can now be closed as its duplicate.

PR #1989, opened 2015-05-10
FOR ... IN ZIP/ENUMERATE support - fixes #1952, #1954

### LINKED ISSUE #1952: `FOR ... IN ZIP ...` and `FOR ... IN ENUMERATE`
It would be nice if we had a built-in way to [zip](https://docs.python.org/2/library/functions.html#zip)-and-iterate over a group of lists.

For example:

```
${numbers}=    Create List    1    2    3
${names}=    Create List    One    Two    Three
:FOR    ${number}    ${name}    IN ZIP    ${numbers}    ${names}
\    Log    ${number} is spelled ${name}
```

I haven't looked at the Robot parser at all yet, but it seems like this should be on par with `FOR ... IN RANGE` in terms of difficulty.

This would be useful shorthand for test cases where we build several lists in parallel and then want to check them against each other.  I could write an example of what this looks like today with `FOR ... IN RANGE` if you like.

edbrannin:
See #1952.


robotci:
Admin, can CI be enabled?

edbrannin:
The following test suites pass; I haven't run any others.

* `atest/run_atests.py python atest/testdata/running/for.robot atest/robot/running/for.robot`
* `utest/parsing/test_model.py`

pekkaklarck:
Great that you've had time to work with this! Unfortunately I don't have time to review this right now, but there's a change I can take a look at this later today or latest next week.

@robotci: Enable

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/84/
Test PASSed.

pekkaklarck:
Unfortunately our latest refactoring to how keywords and for loops are executed were incompatible with this PR. Can you @edbrannin take a look how big changes would be needed to fix the PR?

edbrannin:
At first glance, it shouldn't be too bad.  I started over lunch today, and will try to finish redoing this (and #1954) tomorrow morning.

src/robot/parsing/model.py
```diff
@@ -595,20 +595,33 @@ def __iter__(self):
...
+    - items:  Items to loop over (or otherwise parameters, like with INRANGE)
+    - comment: a comment, or None.
+    - steps:  A list of steps in the loop.
+    """
 
     def __init__(self, declaration, comment=None):
-        self.range, index = self._get_range_and_index(declaration)
+        self.flavor, index = self._get_flavor_and_index(declaration)
+        # self.range is preserved for backwards-compatibility
+        self.range = self.flavor == 'INRANGE'
```
pekkaklarck: Adding more generic `self.flavor` is a good idea. I hope we wouldn't need `self.range` anymore after that. I know removing it is a backwards incompatible change, but we are pretty deep in Robot internals here so such a change ought to be OK in a major release.

src/robot/running/keywords.py
```diff
@@ -173,21 +173,28 @@ def _report_failure(self, context):
         raise failure
 
 
-class ForLoop(_BaseKeyword):
+class BasicForLoop(_BaseKeyword):
+    """A basic For Loop.
+
+    To make a new kind of For loop, subclass this and override
+    _flavor_name() and _transform_items(items)."""
```
pekkaklarck: Nitpick: Would prefer closing quotes on a separate line like recommended by PEP-257: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

src/robot/running/keywords.py
```diff
@@ -307,6 +328,35 @@ def _to_number_with_arithmetics(self, item):
...
+FOR_LOOP_CLASS_FROM_TYPE = dict(
+    INRANGE=ForInRangeLoop,
+    INZIP=ForInZipLoop,
+    )
+
+def ForLoop(forstep, templated=False):
+    try:
+        answer_class = FOR_LOOP_CLASS_FROM_TYPE[forstep.flavor]
+    except KeyError:
+        answer_class = BasicForLoop
```
pekkaklarck: +1 for a factory method with dynamic dispatch. I would just inline the dict containing types and add `BasicForLoop` there too. If we want to protect against invalid flavors, we could use `{...}.get(foritem.flavor, BasicForLoop)`.

atest/testdata/running/for.robot
```diff
@@ -317,6 +317,13 @@ For In Range With Expressions Containing Floats
     \    @{result} =    Create List    @{result}    ${i}
     Should Be True    @{result} == [3.14, 1.14, -0.86]
 
+For In Zip
+    @{items}=    Create List    a    b    c    d
+    @{things}=    Create List    e    f    g    h
+    : FOR    ${item}    ${thing}    IN ZIP    ${items}    ${things}
+    \    @{result} =    Create List    @{result}    ${item}:${thing}
+    Should Be True    @{result} == ['a:e', 'b:f', 'c:g', 'd:h']
+
```
pekkaklarck: More tests needed. Some of these are likely to require enhancements to code too.

1. Handling lists with different number of elements. Probably should work like in Python here and just stop when the shortest list is empty?

2. Handling more than one list like `:FOR  ${a}  ${b}  ${c}  IN ZIP  ${x}  ${y}  ${z}`.

3. Handling non-lists. Should accept anything iterable except strings and fail with a clear error message if invalid data given. You can use `utils.is_list_like` to verify inputs.

4. Different number of variables than lists. Having just one variable works in Python (e.g. `for i in zip(x, y)`), but looking at the implementation it might be hard to support here. I'd be fine with a clear error if num(vars) != num(lists).

src/robot/parsing/model.py
```diff
@@ -595,20 +595,33 @@ def __iter__(self):
...
+    Properties:
+
+    - flavor: The value of the 'IN' item, with spaces removed.
+              Typically 'IN', 'INRANGE', 'INZIP', 'INEUMERATE', etc.(?)
+    - range:  True if flavor is 'INRANGE'
+    - vars:   Variables set per-iteration by this loop
+    - items:  Items to loop over (or otherwise parameters, like with INRANGE)
+    - comment: a comment, or None.
+    - steps:  A list of steps in the loop.
+    """
```
pekkaklarck: +1 for docstrings. We've generally used them too little especially with internal code.

I'd prefer docstrings to be in [Sphinx](http://sphinx-doc.org/) format, though, because that's what we use to document public APIs. I couldn't quickly find a better doc that would explain how to document params than http://sphinx-doc.org/domains.html#info-field-lists, but you can also see code e.g. in `robot.model` package for an example.

pekkaklarck:
Great that updating the PR shouldn't be that big a task. If I have time today, I may refactor the new `keywordrunner` module a bit more, but those changes are going to be much smaller than changes we did earlier.

I now had a change to actually look at your PR and it looked very good. Please take a look at the line notes, though, when you have time to work on this yourself. In addition to code and tests the PR will need updates to the docs in the User Guide too.

pekkaklarck:
I did some further clean-up to code related to executing for loops. Not planning to touch this code in the near future again so it ought to be safe to update this PR.

edbrannin:
Great, I'll try to tackle that sometime this week.

edbrannin:
About half done with the refactoring updates, then will update this PR for review and start on the additional tests & documentation.

Mind if I do for-in-enumerate with the same PR?

pekkaklarck:
Having these two very much related features in same PR is fine.

src/robot/running/keywords.py
edbrannin: I took your `get(flavor, BasicForLoopRunner)` suggestion but didn't inline the dict yet.

Would it be totally crazy to allow libraries to add new kinds of for-loops by defining their own `BasicForLoopRunner` subclass and calling a method that adds it to the factory's dict?

src/robot/running/keywords.py
edbrannin: Fixed.

atest/testdata/running/for.robot
edbrannin: I'm going to tackle this next, now that I've finally merged from master.

src/robot/parsing/model.py
edbrannin: I'll look into this after adding more tests and before updating the documentation.

edbrannin:
I "accidentally" included preliminary/untested ForInEnumerate support in that first merge.  I'll be sure to include tests for it when I flesh out the other atests.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/88/
Test FAILed.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/89/
Test PASSed.

src/robot/parsing/model.py
edbrannin: Done, though I'm not totally happy with how it turned out: I had to add another property (tentatively called `raw_flavor`) to preserve any spaces in the original name.

...actually, I'm going to drop the spaces-removed version of `flavor` and only remove those spaces in the dispatch-factory.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/90/
Test FAILed.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/91/
Test PASSed.

edbrannin:
OK, I'm calling it a night.

I've stubbed the acceptance test cases you requested (plus a few more), and started implementing them.

Obviously I don't expect this to be merged as-is, but if you get a chance to review my recent changes I'd appreciate any comments you have.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/92/
Test PASSed.

atest/robot/running/for.robot
```diff
@@ -254,6 +254,70 @@ For In Range With Multiple Variables
...
+    ${for_loop}=    Set Variable    ${tc.kws[3]}
+    Should Be For In Zip Keyword    ${for_loop}    4
+    Should Be For Item    ${for_loop.kws[0]}    \${item} = a, \${thing} = e, \${stuff} = 1
+    Should Be For Item    ${for_loop.kws[1]}    \${item} = b, \${thing} = f, \${stuff} = 2
+    Should Be For Item    ${for_loop.kws[2]}    \${item} = c, \${thing} = g, \${stuff} = 3
+    Should Be For Item    ${for_loop.kws[3]}    \${item} = d, \${thing} = h, \${stuff} = 4
+
+For In Zip With Other Iterables
+    [Documentation]    Handling non-lists. Should accept anything iterable except strings and fail with a clear error message if invalid data given. You can use utils.is_list_like to verify inputs.
+    Comment
```
pekkaklarck: Adding placeholder tests like this is a good idea. Using `Fail` or `Fail    Not implemented` as implementation instead of `Comment` would be a bit safer, though. If you are worried tests will fail on CI, you can use `not ready` tag to mark it non-critical. Obviously no need to change these now, I assume you'll actually implement this soonish.

atest/robot/running/for.robot
```diff
@@ -254,6 +254,70 @@ For In Range With Multiple Variables
...
+
+For In Zip With String "Lists"
+    [Documentation]    FAIL
```
pekkaklarck: No need to have this kind of documentation on the execution side. `FAIL Message` is used on the test data side to automatically validate messages.

atest/robot/running/for.robot
```diff
@@ -254,6 +254,70 @@ For In Range With Multiple Variables
...
+    ${tc} =    Check Test Case    ${TEST NAME}    status=FAIL    message=For-In-Zip Loop items must all be List-like (but not Strings); got <type 'unicode'> with value 'NotAListButStillIterable'
```
pekkaklarck: Typically better to have expected status and message on the test data side on documentation. Specifying them explicitly is useful when you want to run same tests multiple times with different configuration and expected results.

atest/testdata/running/for.robot
```diff
@@ -317,6 +317,63 @@ For In Range With Expressions Containing Floats
...
+For In Zip With Uneven Lists
+    [Documentation]    Handling lists with different number of elements. Probably should work like in Python here and just stop when the shortest list is empty?
```
pekkaklarck: Using my comments in documentation is clever. If you think docs are needed you can leave them, but works like `probably` don't belong there when we are ready with the design.

atest/testdata/running/for.robot
```diff
@@ -317,6 +317,63 @@ For In Range With Expressions Containing Floats
...
+    [Documentation]    Handling more than one list like :FOR ${a} ${b} ${c} IN ZIP ${x} ${y} ${z}.
+    @{items}=    Create List    a    b    c    d
+    @{things}=    Create List    e    f    g    h
+    @{stuffs}=    Create List    1    2    3    4    5
+    : FOR    ${item}    ${thing}    ${stuff}    IN ZIP    ${items}    ${things}    ${stuffs}
+    \    @{result} =    Create List    @{result}    ${item}:${thing}:${stuff}
+    Should Be True    @{result} == ['a:e:1', 'b:f:2', 'c:g:3', 'd:h:4']
+
+For In Zip With Other Iterables
+    [Documentation]    Handling non-lists. Should accept anything iterable except strings and fail with a clear error message if invalid data given. You can use utils.is_list_like to verify inputs.
```
pekkaklarck: If these non-list iterables are needed only once, it might be easiest to create them using `Evaluate` keyword.

atest/testdata/running/for.robot
```diff
@@ -317,6 +317,63 @@ For In Range With Expressions Containing Floats
...
+
+For In Zip With String "Lists"
+    [Documentation]    FAIL
+    ${items}=    Set Variable    NotAListButStillIterable
```
pekkaklarck: No need to create a variable here. Could just use this string in for loop values.

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +249,69 @@ def _to_number_with_arithmetics(self, item):
...
+        answer = list()
+        for item in items:
+            if not is_list_like(item):
+                # Checking here because _validate() doesn't have access
+                # to the actual values.
+                raise DataError("For-In-Zip Loop items must all be List-like (but not Strings); got %s with value '%s'" % (item.__class__, item))
+        for zipped_item in zip(*items):
+            answer.extend(zipped_item)
+        return answer
+
```
pekkaklarck: Two empty lines between classes expected. #nitpick

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +249,69 @@ def _to_number_with_arithmetics(self, item):
...
+    def _original_values_per_iteration(self, values):
+        """
+        The number of values per iteration;
+        used to check if we have (a multiple of this) values.
+
+        This is its own method to support loops like ForInEnumerate
+        which add/remove items to the pool.
+        """
+        if len(values) < 2:
+            raise DataError('FOR IN ENUMERATE expected 2+ loop variables, got %d.'
```
pekkaklarck: `2 or more` would be more consistent with other similar error messages than `2+`.

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +249,69 @@ def _to_number_with_arithmetics(self, item):
...
+
+# Defining this dict here in for potentially easier extension by libraries...
+# ...unless that's a terrible idea?
+_FOR_LOOP_FLAVORS = dict(
+        INRANGE=ForInRangeLoopRunner,
+        INZIP=ForInZipLoopRunner,
+        INENUMERATE=ForInEnumerateLoopRunner,
+        )
+def ForLoopRunner(context, templated=False, flavor='IN'):
+    flavor_key = flavor.replace(' ', '')
```
pekkaklarck: Is the given flavor uppercased already? If yes, spaces should be removed earlier too. If not, should add `.upper()` here.

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +249,69 @@ def _to_number_with_arithmetics(self, item):
...
+    runner_class = _FOR_LOOP_FLAVORS.get(flavor_key, BasicForLoopRunner)
```
pekkaklarck: Hmm, wouldn't it be better to just fail with a `DataError` if the given flavor is invalid? That's something that definitely needs to be tested too.

src/robot/running/keywordrunner.py
guykisel: @pekkaklarck for stuff like this maybe you could set up a static analysis and/or style checker to run in CI along with the tests. Not sure if it would be worth the effort for you though.

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +249,69 @@ def _to_number_with_arithmetics(self, item):
...
+            yield (idx,) + values
+
+
+# Defining this dict here in for potentially easier extension by libraries...
+# ...unless that's a terrible idea?
+_FOR_LOOP_FLAVORS = dict(
+        INRANGE=ForInRangeLoopRunner,
+        INZIP=ForInZipLoopRunner,
+        INENUMERATE=ForInEnumerateLoopRunner,
+        )
```
pekkaklarck: Do you have use cases for libraries or other code adding new for loops? That sounds a bit weird but I'm definitely not interested in preventing it. On the other hand, we cannot generally promise APIs like this are stable between major releases. The only way to make sure APIs don't break is adding unit or acceptance tests, but even then keeping API compatibility might be a burden.

pekkaklarck:
Looks great! More detailed comments in line notes.

Notice that we try to get alpha 3 out already next week. It would be great to get this in.

src/robot/running/keywordrunner.py
edbrannin: It has already been `upper()`d, but other classes need the spaces to still be there where they once said things like `['IN RANGE' if loop.range else 'IN']`

src/robot/running/keywordrunner.py
pekkaklarck: Good idea. Need to think about that more when I have time. Getting contribution guidelines explaining also coding conventions ready is also an important step. We won't release a3 without them!

src/robot/running/keywordrunner.py
pekkaklarck: OK. Adding `.upper()` here would probably be a good idea anyway. Makes it more explicit.

atest/robot/running/for.robot
edbrannin: That explains so much!  Thanks.

atest/robot/running/for.robot
edbrannin: Fixed.

atest/robot/running/for.robot
edbrannin: Done.

src/robot/running/keywordrunner.py
edbrannin: At one extreme, you could try running Robot through [yapf][] and recommend everyone do the same when contributing if you like the result.

(Disclaimer: I have not yet tried [yapf][] on my own code.)

[yapf]: https://github.com/google/yapf

src/robot/running/keywordrunner.py
edbrannin: Fixed.

src/robot/running/keywordrunner.py
edbrannin: Yes, it does sound a bit weird -- and I might have had a use-case for new for-loops until this PR adds for-in-zip/enumerate.

The only other FOR-ish control structures I can think of are:
* Looping over generator-functions
* (This is totally abusive) emulating a `with` statement

Actually, FOR-IN-KEYWORD could be a nice shortcut -- but probably not nice enough to _implement_ unless this doesn't work:

```
# AwesomeLibrary.py

def contrived_generator(arg):
    for x in arg:
        yield x
```

```
My Loop Keyword
    ${generator}=    Contrived Generator    Whatever
    : FOR    ${x}    IN    $(generator}
    ...    Log    ${x}
```

...That doesn't look like it should work.  I think I'll make an issue to discuss the merits of this idea.

src/robot/running/keywordrunner.py
edbrannin: Good point.

src/robot/running/keywordrunner.py
pekkaklarck: You actually should be able to do this already now:

```robotframework
*** Keywords ***
My Loop Keyword
    ${generator}=    Contrived Generator    Whatever
    : FOR    ${x}    IN    @{generator}
    \    Log    ${x}
```

Obviously the generator will be expanded to a list already when `@{generator}` is evaluated, but that shouldn't really matter unless the list it generates is really huge. In that case the overhead caused by Robot for loops (e.g. increased output.xml size) would probably be a much bigger problem.

I don't like FOR-IN-KEYWORD that much because you can so easily just run a keyword and use its return values as for loop values. That way you can then use IN, IN-RANGE, IN-ZIP or IN-ENUMERATE flavors depending on what your keyword returned. I definitely don't wat to add IN-KEYWORD, IN-RANGE-KEYWORD, IN-ZIP-KEYWORD and IN-ENUMERATE-KEYWORD.

edbrannin: Good points.  List expansion with `@{generator}` is what I was missing.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/93/
Test PASSed.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/94/
Test PASSed.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/95/
Test PASSed.

edbrannin:
With that push, I think I'm ready to switch to copypaper-on-demand (though the occasional DVRing would still be nice).

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/96/
Test PASSed.

edbrannin:
Documentation added, though Sphinx dislikes something about my tables.  Please advise.  And you'll probably find some paragraphs in need to style-corrections.

Pending new review-issues, I think my only task left is to apply Sphinx formatting to any docstrings I've written.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/97/
Test PASSed.

atest/testdata/running/for.robot
```diff
@@ -40,6 +40,13 @@ For Loop Over Empty List Variable Is Ok
     \    Fail    not executed
     Variable Should Not Exist    ${var}
 
+For Loop Over Generator
+    ${xrange}=    Evaluate    xrange(10)
```
pekkaklarck: Considering the future Python 3 support, it would be better to use, for example, `(i for i in 'abcdefg')` instead of `xrange`.

atest/robot/running/for.robot
```diff
@@ -33,6 +33,9 @@ For Without Value Fails
 For Loop Over Empty List Variable Is Ok
     Check Test Case    ${TEST NAME}
 
+For Loop Over Generator
+    Check Test Case    ${TEST NAME}
```
pekkaklarck: +1 for an explicit test for using generators.

atest/robot/running/for.robot
```diff
@@ -254,6 +257,80 @@ For In Range With Multiple Variables
...
+    ${tc} =    Check Test Case    ${TEST NAME}
+    ${for_loop}=    Set Variable    ${tc.kws[2]}
+    Should Be For In Zip Keyword    ${for_loop}    4
+    Should Be For Item    ${for_loop.kws[0]}    \${item} = a, \${thing} = e
+    Should Be For Item    ${for_loop.kws[1]}    \${item} = b, \${thing} = f
+    Should Be For Item    ${for_loop.kws[2]}    \${item} = c, \${thing} = g
+    Should Be For Item    ${for_loop.kws[3]}    \${item} = d, \${thing} = h
+
+For In Zip With Uneven Lists
+    [Documentation]    Handling lists with different number of elements. Probably should work like in Python here and just stop when the shortest list is empty?
```
pekkaklarck: This doc probably isn't needed anymore. Should at least be updated not to contain words like `probably`.

atest/testdata/running/for.robot
```diff
@@ -317,6 +324,120 @@ For In Range With Expressions Containing Floats
...
 
+For In Zip
+    @{items}=    Create List    a    b    c    d
+    @{things}=    Create List    e    f    g    h
+    : FOR    ${item}    ${thing}    IN ZIP    ${items}    ${things}
+    \    @{result} =    Create List    @{result}    ${item}:${thing}
+    Should Be True    @{result} == ['a:e', 'b:f', 'c:g', 'd:h']
+
+For In Zip With Uneven Lists
+    [Documentation]    Handling lists with different number of elements.  This will ignore any elements after the shortest list ends, just like with Python's zip()
```
pekkaklarck: Longish docs like this could be split to multiple lines by using `...`. The first sentence could also be removed because it tells pretty much what the test name tells.

atest/testdata/running/for.robot
```diff
@@ -317,6 +324,120 @@ For In Range With Expressions Containing Floats
...
+    @{items}=    Create List    a    b    c
+    @{things}=    Create List    d    e    f    g    h
+    : FOR    ${item}    ${thing}    IN ZIP    ${items}    ${things}
+    \    @{result} =    Create List    @{result}    ${item}:${thing}
+    Should Be True    @{result} == ['a:d', 'b:e', 'c:f']
+
+For In Zip With 3 Lists
+    [Documentation]    Handling more than one list like :FOR ${a} ${b} ${c} IN ZIP ${x} ${y} ${z}.
```
pekkaklarck: - This doc isn't very useful. 
- Would be nice to test also zipping more than three lists (e.g. 5).
- Tests could be simplified by constructing loops like this:

```robotframework
*** Test Cases ***
Example
    @{lower} =    Evaluate    list('abcde')
    @{upper} =     Evaluate    list('ABCDEFG')
    @{numbers} =    Evaluate    range(100)
    @{expected} =    Create List    a-A-0    b-B-1    c-C-2    d-D-3    e-E-4    not-used-anymore
    :FOR    ${low}    ${up}    ${num}    ${exp}    IN ZIP    ${lower}    ${upper}    ${numbers}    ${expected}
    \    Should Be Equal    ${low}-${up}-${num}    ${exp}
    Should Be Equal    ${exp}    e-E-4
```

atest/testdata/running/for.robot
```diff
@@ -317,6 +324,120 @@ For In Range With Expressions Containing Floats
...
+    @{things}=    Create List    e    f    g    h
+    @{stuffs}=    Create List    1    2    3    4    5
+    : FOR    ${item}    ${thing}    ${stuff}    IN ZIP    ${items}    ${things}    ${stuffs}
+    \    @{result} =    Create List    @{result}    ${item}:${thing}:${stuff}
+    Should Be True    @{result} == ['a:e:1', 'b:f:2', 'c:g:3', 'd:h:4']
+
+For In Zip With Other Iterables
+    [Documentation]    Handling non-lists. Should accept anything iterable except strings and fail with a clear error message if invalid data given. You can use utils.is_list_like to verify inputs.
+    @{xrange}=    Evaluate    xrange(10)
+    @{xrange2}=    Evaluate    xrange(10,20)
```
pekkaklarck: - `xrange` is not Python 3 compatible. Generator expression like `(i in range(10)` would be better in that regard.
- Other non-list iterable could just be a tuple.

atest/robot/running/for.robot
```diff
@@ -254,6 +257,80 @@ For In Range With Multiple Variables
...
+    ${tc} =    Check Test Case    ${TEST NAME}
+    ${for_loop}=    Set Variable    ${tc.kws[2]}
+    Should Be For In Zip Keyword    ${for_loop}    3
+    Should Be For Item    ${for_loop.kws[0]}    \${item} = a, \${thing} = d
+    Should Be For Item    ${for_loop.kws[1]}    \${item} = b, \${thing} = e
+    Should Be For Item    ${for_loop.kws[2]}    \${item} = c, \${thing} = f
+
+For In Zip With 3 Lists
+    [Documentation]    Handling more than one list like :FOR ${a} ${b} ${c} IN ZIP ${x} ${y} ${z}.
```
pekkaklarck: - Doc isn't very useful here. Especially when it should say "more than two lists" instead of "more than one".
- Test name should also contain "with more than two lists" rather than exact "with 3 lists".

atest/robot/running/for.robot
```diff
@@ -254,6 +257,80 @@ For In Range With Multiple Variables
...
+    Should Be For Item    ${for_loop.kws[3]}    \${item} = d, \${thing} = h, \${stuff} = 4
+
+For In Zip With Other Iterables
+    [Documentation]    Handling non-lists. Should accept anything iterable except strings and fail with a clear error message if invalid data given. You can use utils.is_list_like to verify inputs.
+    ${tc} =    Check Test Case    ${TEST NAME}
+
+For In Zip With String "Lists"
+    ${tc} =    Check Test Case    ${TEST NAME}
+    ${for_loop}=    Set Variable    ${tc.kws[1]}
+    Should Be For In Zip Keyword    ${for_loop}    0    # This seems like it should be 1, but I haven't messed with anything that would break it and other failing tests don't even check it...
```
pekkaklarck: I don't really get the comment here. Could you clarify? If it's a valid concern, we should fix it. Otherwise the comment should be removed.

atest/testdata/running/for.robot
```diff
@@ -317,6 +324,120 @@ For In Range With Expressions Containing Floats
...
+    : FOR     ${x}    ${y}    IN ZIP    ${xrange}    ${xrange2}
+    \    @{result}=    Create List    @{result}    ${x}:${y}
+    ${expected}=    Create List    0:10    1:11    2:12    3:13    4:14    5:15    6:16    7:17    8:18    9:19
+    Should Be Equal    ${result}    ${expected}
+
+
+For In Zip With String "Lists"
```
pekkaklarck: Name like "For In Zip does not accept string as iterable" would be better. Also unnecessary empty row above.

atest/robot/running/for.robot
```diff
@@ -254,6 +257,80 @@ For In Range With Multiple Variables
...
+    [Documentation]    Different number of variables than lists. Having just one variable works in Python (e.g. for i in zip(x, y)), but looking at the implementation it might be hard to support here. I'd be fine with a clear error if num(vars) != num(lists).
+    ${tc} =    Check Test Case    ${TEST NAME}
+
+For In Enumerate (with 4 items)
+    [Documentation]    This test is repeated with different lengths to expose subtle validation errors from development.
+    ${tc} =    Check Test Case    ${TEST NAME}
+
+For In Enumerate (with 5 items)
+    [Documentation]    This test is repeated with different lengths to expose subtle validation errors from development.
+    ${tc} =    Check Test Case    ${TEST NAME}
```
pekkaklarck: Testing enumerate with different number of items is a good idea, but having separate tests here on the execution side is not. There could just be one test on both sides or there could be two on the test date side and here something like this:

```robotframework
*** Test Cases ***
For In Enumerate
    Check Test Case    ${TEST NAME} (with 4 items)
    Check Test Case    ${TEST NAME} (with 5 items)
```

atest/robot/running/for.robot
```diff
@@ -254,6 +257,80 @@ For In Range With Multiple Variables
...
+
+For In Enumerate With 3 Variables
```
pekkaklarck: Good test! Name should be something like "with more than two variables" instead of exact "with 3 variables", and it would be good to test also with some other number than three.

atest/testdata/running/for.robot
```diff
@@ -317,6 +324,120 @@ For In Range With Expressions Containing Floats
...
+    [Documentation]    FAIL    Number of FOR loop values should be multiple of variables. Got 3 variables (2 of which matter for this error) but 7 values.
+    @{items}=    Create List    a    b    c    d    e    f    g
+    : FOR    ${index}    ${item}    ${another_item}    IN ENUMERATE    @{items}
+    \    Should Be Equal    @{items}[${index* 2 }]    ${item}    Loop value ${item} should be item ${index} of ${items}
+    \    Should Be Equal    @{items}[${index * 2 + 1}]    ${another_item}    Loop value ${another_item} should be item ${index} of ${items}
+    \    @{result}=     Create List    @{result}    ${index}:${item}:${another_item}
+    Should Be True    @{result} == ['0:a:b', '1:c:d', '2:e:f']
+
+For In Enumerate With Too Few Variables
+    [Tags]    Not Ready
```
pekkaklarck: These tags won't do anything on this side. I also assume these tests are already ready.

atest/testdata/running/for.robot
```diff
@@ -317,6 +324,120 @@ For In Range With Expressions Containing Floats
...
+    [Documentation]    FAIL    FOR IN ENUMERATE expected 2 or more loop variables, got 1.
+    : FOR    ${index}    IN ENUMERATE    a    b    c    d    e    f
+    \    Fail    Should not reach this line.
+
+For In Enumerate With Other Iterables
+    [Tags]    Not Ready
+    ${xrange}=    Evaluate    xrange(10)
```
pekkaklarck: `xrange` is not Python 3 compatible.

src/robot/running/keywordrunner.py
```diff
@@ -179,13 +188,54 @@ def _get_values_for_one_round(self, data):
...
-        if len(values) % len(data.variables) == 0:
+        values = self._transform_items(values)
+        values_per_iteration = self._original_values_per_iteration(data.variables)
+        if len(values) % values_per_iteration == 0:
             return values
+        error_addendum = ""
+        if len(data.variables) != values_per_iteration:
+            # This is a bit awkward, but I'm not sure how else to phrase
+            # the impact of a non-value variable like from ForInEnumerate.
+            error_addendum = " (%d of which matter for this error)" % values_per_iteration
```
pekkaklarck: I think it would be better to have a separate method that raises the exception (or returns the error message) that ForInEnumerate can override.

src/robot/running/keywordrunner.py
```diff
@@ -179,13 +188,54 @@ def _get_values_for_one_round(self, data):
...
+
+class ForInRangeLoopRunner(BasicForLoopRunner):
+    def __init__(self, context, templated=False):
+        super(ForInRangeLoopRunner, self).__init__(context, templated)
+
+    def _flavor_name(self):
+        return 'IN RANGE'
+
+    def _transform_items(self, items):
+        return self._get_range_items(items)
```
pekkaklarck: It's a bit unnecessary to call `_get_range_items` here. That method could just be renamed to `_transform_items` instead.

src/robot/running/keywordrunner.py
```diff
@@ -113,7 +114,12 @@ def _get_and_report_failure(self):
         return failure
 
 
-class ForLoopRunner(object):
+class BasicForLoopRunner(object):
```
pekkaklarck: I'd either rename this to `ForInRunner` or have separate abstract class `ForRunnerBase` that all concrete for loop runners extend.

src/robot/running/keywordrunner.py
```diff
@@ -113,7 +114,12 @@ def _get_and_report_failure(self):
...
+    """A basic For Loop.
+
+    To make a new kind of For loop, subclass this and override
+    _flavor_name() and _transform_items(items).
+    """
```
pekkaklarck: I don't think this doc provides much added value. If we wont' to make it explicit how to extend, it would be better to have `ForRunnerBase` with the abstract methods raising `NotImplementedError` or actually marked abstract using functionality in the `abc` module.

src/robot/running/keywordrunner.py
```diff
@@ -179,13 +188,54 @@ def _get_values_for_one_round(self, data):
...
+        The number of values per iteration;
+        used to check if we have (a multiple of this) values.
+
+        This is its own method to support loops like ForInEnumerate
+        which add/remove items to the pool.
+        """
+        return len(values)
+
+
+class ForInRangeLoopRunner(BasicForLoopRunner):
```
pekkaklarck: I'd like to drop `Loop` from the name and use `ForInRangeRunner`, `ForInZipRunner`, etc.

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +256,99 @@ def _to_number_with_arithmetics(self, item):
...
+
+    def _flavor_name(self):
+        return 'IN ZIP'
+
+    def _replace_variables(self, data):
+        """
+        Make sure we have as many loop-variables as zipped-iterables.
+
+        (Other than that, delegate to the superclass.)
+        """
```
pekkaklarck: The code tells this better than the doc does.

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +256,99 @@ def _to_number_with_arithmetics(self, item):
...
+        values = super(ForInZipLoopRunner, self)._replace_variables(data)
+        if len(data.variables) == len(data.values):
+            return values
+        raise DataError("Expected %d Loop variables, but found %d"
```
pekkaklarck: - `loop` instead of `Loop`
- `got` is used instead of `found` with other errors. I'd prefer that also here for consistency reasons.
- We generally have `.` at the end of the error when it is a full sentence.
- The error could be more explicit. Perhaps something like `'FOR IN ZIP expects equal number of variables and iterables. Got %d variable%s and %d iterable%s.' % (len(data.variables), s(data.variables), len(data.values), s(data.values)`.

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +256,99 @@ def _to_number_with_arithmetics(self, item):
...
+                % (len(data.values), len(data.variables)))
+
+    def _transform_items(self, items):
+        answer = list()
+        for item in items:
+            if not is_list_like(item):
+                # Checking here because _validate() doesn't have access
+                # to the actual values.
+                raise DataError("For-In-Zip Loop items must all be List-like (but not Strings); got %s with value '%s'" % (item.__class__, item))
```
pekkaklarck: - What does `_validate()` refer to in the comment? The main reason I don't like comments (or unnecessary docs) is that they get outdated so easily.
- Errors have used `FOR IN RANGE` style earlier. Should use that here too (or change those places to use `For-In-Range` style instead).
- Should use `utils.type_name(item)` instead of `item.__class__` for various reasons:
  - Get nicer name with known types.
  - Get consistent name with different Python interpreters (i.e. avoid problems with tests).
  - Avoid exception with types that don't actually have `__class__`.
- We try to keep line lengths below 80 chars. That's not so exact but this is too long.
- We haven't had a note like `but not strings` when talking about list-like objects elsewhere in the code base. No need to have it here either.
- As already noted, we generally have `.` at the end of the error.

Perhaps something simple like this would be good:

    "FOR IN ZIP items must all be list-like, got %s." % type_name(item)

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +256,99 @@ def _to_number_with_arithmetics(self, item):
...
+        return 'IN ENUMERATE'
+
+    def _original_values_per_iteration(self, values):
+        """
+        The number of values per iteration;
+        used to check if we have (a multiple of this) values.
+
+        This is its own method to support loops like ForInEnumerate
+        which add/remove items to the pool.
+        """
```
pekkaklarck: Considering how good and clear the code is, this kind of doc in a private method isn't very useful.

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +256,99 @@ def _to_number_with_arithmetics(self, item):
...
+    def _get_values_for_one_round(self, data):
+        for idx, values in enumerate(super(ForInEnumerateLoopRunner, self)
+                ._get_values_for_one_round(data)):
+            yield [idx] + values
+
+
+# Defining this dict here in for potentially easier extension by libraries...
+# ...unless that's a terrible idea?
+def ForLoopRunner(context, templated=False, flavor='IN'):
+    FOR_LOOP_FLAVORS = dict(
```
pekkaklarck: - Outdated comment.
- Should use lower case letters with a local variable.
- Should have this factory method above classes.

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +256,99 @@ def _to_number_with_arithmetics(self, item):
...
+    def __init__(self, context, flavor, pretty_flavor, expected):
+        super(InvalidForLoopRunner, self).__init__(context, False)
+        self.flavor = flavor
+        self.pretty_flavor = pretty_flavor
+        self.expected = repr(sorted(expected))
+
+    def _run(self, data, *args, **kwargs):
+        raise DataError(
+            "Unexpected For-loop type %s (%s); expected one of %s (but with spaces)" %
+            (self.flavor, self.pretty_flavor, self.expected))
```
pekkaklarck: I like dynamic code but here I'd just hard-code the types to make the error better. For example:

    "Invalid FOR loop type '%s'. Expected 'IN', 'IN RANGE', 'IN ZIP', or 'IN ENUMERATE'." % self.flavor

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +256,99 @@ def _to_number_with_arithmetics(self, item):
...
+
+
+class InvalidForLoopRunner(BasicForLoopRunner):
+    """
+    Used to send an error from ForLoopRunner() if it sees an unexpected error.
+
+    We can't simply throw a DataError from ForLoopRunner() because that happens
+    outside the "with StatusReporter(...)" blocks.
+
+    """
```
pekkaklarck: I consider even this doc redundant considering how good the code is. Tests would also warn if someone would change `ForLoopRunner` to raise `DataError` directly.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -399,6 +399,103 @@ lower limit, upper limit and step.
    \                                  Log          ${index}
    ================  ===============  ===========  =========  ===========  ========  =======
 
+For in Enumerate
```
pekkaklarck: Earlier title is `For in range` so this should be `For in enumerate`.

atest/robot/running/for.robot
edbrannin: I meant: The actual loop contains 1 line, but the loop-body doesn't seem to have been parsed because the loop itself failed to run.

Considering the loop failed because one of its variables had a bad value (and not a structural issue), I'm surprised the number of sub-lines should be nonzero.

atest/testdata/running/for.robot
edbrannin: I think all these have been resolved.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -399,6 +399,103 @@ lower limit, upper limit and step.
...
+~~~~~~~~~~~~~~~~
+
+Sometimes it is useful to loop over a list and also keep track of your location
+inside the list.  Robot Framework has a special
+`FOR index ... IN ENUMERATE ...` syntax for this situation.
+This syntax is derived from the
+`Python built-in function <https://docs.python.org/2/library/functions.html#enumerate>`.
```
pekkaklarck: I don't think this creates a link. Have you tested generating User Guide to see does it work?

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -399,6 +399,103 @@ lower limit, upper limit and step.
...
+
+:name:`:FOR` :name:`IN ENUMERATE` loops work just like regular :name:`:FOR` loops,
```
pekkaklarck: Why `FOR` and `IN ENUMERATE` separately and not `FOR IN ENUMERATE`?

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -399,6 +399,103 @@ lower limit, upper limit and step.
...
+   ================= ================ ============ ========= ========= ============= =======
+   For In Enumerate  [Documentation]  Two values   at once
+   \                 @{list}=         Create List  a         b         c             d
+   \                 :FOR             ${index}     ${item}   ${thing}  IN ENUMERAGE  @{list}
+   \                                  Log Many     ${index}  ${item}   $[thing}
+   ================= ================ ============ ========= ========= ============= =======
+
+The above test case will log `0,a,b` and `1,c,d`.
+
+For In Enumerate loops are new in Robot 2.9.
```
pekkaklarck: We prefer full `Robot Framework 2.9` over `Robot 2.9`.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -399,6 +399,103 @@ lower limit, upper limit and step.
...
+   ================= ================ =============== ============= =================== =======
+
+Just like with regular For-loops, you can loop over multiple values per loop iteration
+as long as the number of values in your list evenly is divisible by the number of
+loop-variables (excluding the first, `index` variable).
+
+.. table:: For in enumerate examples
+   :class: example
+   ================= ================ ============ ========= ========= ============= =======
+   Test Case         Action           Argument     Argument  arg       arg
```
pekkaklarck: Should use `Argument` in all argument headers.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -399,6 +399,103 @@ lower limit, upper limit and step.
...
+
+For In Enumerate loops are new in Robot 2.9.
+
+For in Zip
+~~~~~~~~~~
+
+Some tests build up several related lists, then loop over them together.
+Robot Framework has a shortcut for this case: `FOR ... IN ZIP ...`, which
+is derived from the
+`Python built-in zip function <https://docs.python.org/2/library/functions.html#zip>`.
```
pekkaklarck: This link seems to be broken too.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -399,6 +399,103 @@ lower limit, upper limit and step.
...
+   \                 @{list}=         Create List  a         b         c             d
+   \                 :FOR             ${index}     ${item}   ${thing}  IN ENUMERAGE  @{list}
+   \                                  Log Many     ${index}  ${item}   $[thing}
+   ================= ================ ============ ========= ========= ============= =======
+
+The above test case will log `0,a,b` and `1,c,d`.
+
+For In Enumerate loops are new in Robot 2.9.
+
+For in Zip
```
pekkaklarck: `For in zip`.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -399,6 +399,103 @@ lower limit, upper limit and step.
...
+the `IN` cell to read `IN ZIP`.
+
+Note that any lists or other iterables appearing after `IN ZIP` should usually be
+prefixed with `$` instead of `@` because For-In-Zip needs to keep track of each
+list separately.
+
+`IN ZIP` loops require the same number of loop variables as lists to loop over,
+and will stop after the shortest list is complete.
+
+For In Zip loops are new in Robot 2.9, and are named after a Python built-in function.
```
pekkaklarck: `Robot 2.9` -> `Robot Framework 2.9`. The latter part of the sentence is already explained above.

src/robot/parsing/model.py
pekkaklarck: I commented many other docs and said they aren't that useful, but on this level docstrings like this are good. Would still prefer using Sphinx style, though.

pekkaklarck:
This PR looks very good. I added plenty of line notes but most of them are about code/test/doc style and ought to be easy to fix. Let me know if my comments didn't make sense or you don't agree with them.

atest/robot/running/for.robot
pekkaklarck: That number tells how many times the loop is executed. If the loop construct is invalid, the loop isn't executed at all.

src/robot/running/keywordrunner.py
edbrannin: It feels a bit dirty, but the best response here may be to delete the docstring.

src/robot/running/keywordrunner.py
edbrannin: I'll try to rephrase the comment to better reflect this point: Usually one would expect validation to happen in _validate(), but this particular validation relies on information that is not available to _validate() so instead it's here.

src/robot/running/keywordrunner.py
edbrannin: I'm inclined to leave this one here if you don't mind, to save our future selves a few moments of "Why did we go through the trouble of adding a new class for this?"

edbrannin:
I just finished reading your notes, and have commented on the ones where I disagree and/or wanted to resolve confusion.

I'll be fixing the rest in a moment.

*Question:* I mentioned earlier that Sphinx doesn't like my tables.  The 3 I wrote each generated this error and I don't see what I did wrong that other tables did right:

`src/CreatingTestData/AdvancedFeatures.rst:465: (ERROR/3) Error in "table" directive: invalid option block.`

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/98/
Test PASSed.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
edbrannin: Probably just lack of sleep. :)

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
edbrannin: I fixed all instances of this, except for one spot in a table where I used "Version 2.9" to prevent widening a column excessively.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/99/
Test PASSed.

edbrannin:
I think this fixes everything you've pointed out so far.  Still no idea what I'm doing wrong with the tables in the docs.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/100/
Test PASSed.

src/robot/parsing/model.py
edbrannin: Whoops, I forgot to fix this last night.  Will tackle this shortly.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/101/
Test PASSed.

src/robot/parsing/model.py
```diff
@@ -597,20 +597,33 @@ def __iter__(self):
...
+    :param list declaration: The literal cell values that declare the loop
+                             (excluding ":FOR").
+    :param str comment: A comment, default None.
+    :ivar str flavor: The value of the 'IN' item, uppercased.
+                      Typically 'IN', 'IN RANGE', 'IN ZIP', or 'IN EUMERATE'.
+    :ivar list vars: Variables set per-iteration by this loop.
+    :ivar list items: Loop values that come after the 'IN' item.
+    :ivar str comment: A comment, or None.
+    :ivar list steps: A list of steps in the loop.
+    """
```
edbrannin: How's this?  I'm not sure how to render it to check the result.

edbrannin:
Any idea what's going wrong with Sphinx here, or should I ask Stack Overflow?

(I wasn't able to find any answers with Google.)

pekkaklarck:
You can generate API docs by running `doc/api/generate.py`. Did you have some specific issue with Sphinx?

pekkaklarck:
I haven't had change to follow this PR closely lately. What's the current status, are you working with something or is this ready for final review?

edbrannin:
I'm having an issue with generating the user guide -- clone this branch, give it try and you'll see the warnings right away.  There's something wrong with my 3 example-code tables.

I'll try generating the API docs to see how that one docstring looks.

Once the userguide issue is solved (again, I really need help there) and I've looked at the APIdocs output, this will be ready for final review.

edbrannin:
For context, here's the error I'm talking about ([earlier in this PR](https://github.com/robotframework/robotframework/pull/1989#issuecomment-112642811)):

`src/CreatingTestData/AdvancedFeatures.rst:465: (ERROR/3) Error in "table" directive: invalid option block.`

atest/robot/running/for.robot
```diff
@@ -254,6 +257,75 @@ For In Range With Multiple Variables
...
+    ${tc} =    Check Test Case    ${TEST NAME}
+    ${for_loop}=    Set Variable    ${tc.kws[3]}
+    Should Be For In Zip Keyword    ${for_loop}    4
+    Should Be For Item    ${for_loop.kws[0]}    \${item} = a, \${thing} = e, \${stuff} = 1
+    Should Be For Item    ${for_loop.kws[1]}    \${item} = b, \${thing} = f, \${stuff} = 2
+    Should Be For Item    ${for_loop.kws[2]}    \${item} = c, \${thing} = g, \${stuff} = 3
+    Should Be For Item    ${for_loop.kws[3]}    \${item} = d, \${thing} = h, \${stuff} = 4
+
+For In Zip With Other Iterables
+    ${tc} =    Check Test Case    ${TEST NAME}
```
pekkaklarck: Unnecessary to assign `${tc}`.

atest/robot/running/for.robot
```diff
@@ -289,6 +361,11 @@ Should Be For In Range Keyword
     Should Be For Keyword    ${kw}    ${subcount}
     Should Contain    ${kw.name}    IN RANGE    Not FOR IN RANGE keyword
 
+Should Be For In Zip Keyword
+    [Arguments]    ${kw}    ${subcount}
+    Should Be For Keyword    ${kw}    ${subcount}
+    Should Contain    ${kw.name}    IN ZIP    Not FOR IN RANGE keyword
```
pekkaklarck: The configured error message is wrong. It's not that useful here but definitely should be correct if it is used.

atest/testdata/running/for.robot
```diff
@@ -8,7 +8,7 @@ Variables   binary_list.py
 ${NO VALUES}      FOR loop has no loop values.
 ${NO KEYWORDS}    FOR loop contains no keywords.
 ${NO VARIABLES}   FOR loop has no loop variables.
-${WRONG VALUES}   Number of FOR loop values should be multiple of variables.
+${WRONG VALUES}   Number of FOR loop values should be multiple of its variables.
```
pekkaklarck: +1 for enhanced error message.

atest/testdata/running/for.robot
```diff
@@ -317,6 +324,128 @@ For In Range With Expressions Containing Floats
...
+    : FOR    ${item}    ${thing}    ${stuff}    IN ZIP    ${items}    ${things}    ${stuffs}
+    \    @{result} =    Create List    @{result}    ${item}:${thing}:${stuff}
+    Should Be True    @{result} == ['a:e:1', 'b:f:2', 'c:g:3', 'd:h:4']
+
+For In Zip With Other Iterables
+    [Documentation]    Handling non-lists. Should accept anything iterable
+    ...    except strings and fail with a clear error message if invalid
+    ...    data given. You can use utils.is_list_like to verify inputs.
+    @{range}=    Evaluate    (i for i in range(10))
+    @{range2}=    Evaluate    (10, 11, 12, 13, 14, 15, 16, 17, 18, 19)
```
pekkaklarck: In the current for this test actually doesn't test that it should. When you assign a list variable like `@{range}`, Robot will actually convert the value to a real list. You should assign these iterables to a scalar variable instead. It would also be a little cleaner to use more explicit variable names like `${generatot}` and `${tuple}` in this context.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -371,9 +371,9 @@ lower limit, upper limit and step.
 .. table:: For in range examples
    :class: example
 
-   ================  ===============  ===========  =========  ===========  ========  =======
-      Test Case          Action        Argument     Argument    Arg          Arg       Arg
-   ================  ===============  ===========  =========  ===========  ========  =======
+   ================  ===============  ===========  =========  ===========  ========  ========
+      Test Case          Action        Argument     Argument    Argument   Argument  Argument
+   ================  ===============  ===========  =========  ===========  ========  ========
```
pekkaklarck: +1 for cleaning up example headers. @laurentbristiel is actually working to get all these examples converted to the plain text format. We are creating a bit more work for him with these new examples but I think it's still better to use the old style until his PR is merged.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -397,7 +397,104 @@ lower limit, upper limit and step.
...
+`Python built-in function <https://docs.python.org/2/library/functions.html#enumerate>`_.
+
+:name:`:FOR IN ENUMERATE` loops work just like regular :name:`:FOR` loops,
+except the cell after its loop variables must say :name:`IN ENUMERATE`,
+and they must have an additional index variable before any other loop-variables.
+That index variable has a value of `0` for the first loop, `1` for the second, etc.
+
+For example, the following two test cases do the same thing:
+
+
```
pekkaklarck: Unnecessary empty line. #nitpick

src/robot/running/keywordrunner.py
```diff
@@ -17,7 +17,8 @@
                           HandlerExecutionFailed)
 from robot.result.keyword import Keyword as KeywordResult
 from robot.utils import (format_assign_message, frange, get_error_message,
-                         get_timestamp, plural_or_not as s, type_name)
+                         get_timestamp, plural_or_not as s, type_name,
+                         is_list_like, type_name)
```
pekkaklarck: Importing `type_name` twice. Also, `is_list_like` should be placed after `get_timestamp` to keep imports in alphabetical order.

src/robot/running/keywordrunner.py
```diff
@@ -113,8 +114,22 @@ def _get_and_report_failure(self):
...
+            IN=ForInRunner,
+            INRANGE=ForInRangeRunner,
+            INZIP=ForInZipRunner,
+            INENUMERATE=ForInEnumerateRunner,
+            )
+    flavor_key = flavor.upper().replace(' ', '')
+    if flavor_key in for_loop_flavors:
+        runner_class = for_loop_flavors[flavor_key]
+        return runner_class(context, templated)
+    else:
```
pekkaklarck: Unnecessary `else:`.

src/robot/running/keywordrunner.py
```diff
@@ -179,19 +197,54 @@ def _get_values_for_one_round(self, data):
...
+        if len(values) % values_per_iteration == 0:
             return values
-        raise DataError('Number of FOR loop values should be multiple of '
-                        'variables. Got %d variables but %d value%s.'
-                        % (len(data.variables), len(values), s(values)))
+        raise DataError(self._replace_variables_error_message(data, values))
+
+    def _replace_variables_error_message(self, data, values):
+        return ('Number of FOR loop values should be multiple of '
+                + 'its variables. Got %d variables but %d value%s.'
```
pekkaklarck: `+` isn't actually needed here because Python implicitly concatenates strings like `"foo" "bar"`. That is especially handy when doing string formatting because

```python
("A: %s\n"
 "B: %s" % (1, 2))
```
works but
```python
("A: %s\n"
 + "B: %s" % (1, 2))
```
does not.

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +259,81 @@ def _to_number_with_arithmetics(self, item):
...
+                    len(data.values), s(data.values)))
+
+    def _transform_items(self, items):
+        answer = list()
+        for item in items:
+            if not is_list_like(item):
+                # Usually one would expect validation to happen in _validate(),
+                # but this particular validation relies on information that is
+                # not available to _validate().
+                raise DataError("FOR IN ZIP items must all be List-like, got %s."
```
pekkaklarck: I'd prefer `list-like` instead of `List-like`.

src/robot/running/keywordrunner.py
```diff
@@ -206,16 +259,81 @@ def _to_number_with_arithmetics(self, item):
...
+
+
+class InvalidForRunner(ForInRunner):
+    """
+    Used to send an error from ForRunner() if it sees an unexpected error.
+
+    We can't simply throw a DataError from ForRunner() because that happens
+    outside the "with StatusReporter(...)" blocks.
+
+    """
```
pekkaklarck: The first line should follow the leading quotes like `"""Used to ...` as recommended in [PEP-257](https://www.python.org/dev/peps/pep-0257/). The unnecessary empty line should also be removed.

src/robot/parsing/model.py
```diff
@@ -597,20 +597,33 @@ def __iter__(self):
 
 
 class ForLoop(_WithSteps):
+    """
+    py:class:: ForLoop(declaration, comment=None)
```
pekkaklarck: This isn't needed nor is it parsed anyway separately. I think you would use something like this if you were writing a separate API doc using Sphinx, but we are using `autodoc` extension which can figure out the types of documented items automatically.

A short plain-text description for the class would be nice, though. It should be placed right after the opening quotes to follow PEP-257.

src/robot/parsing/model.py
```diff
@@ -597,20 +597,33 @@ def __iter__(self):
...
+
+    :param list declaration: The literal cell values that declare the loop
+                             (excluding ":FOR").
+    :param str comment: A comment, default None.
+    :ivar str flavor: The value of the 'IN' item, uppercased.
+                      Typically 'IN', 'IN RANGE', 'IN ZIP', or 'IN EUMERATE'.
```
pekkaklarck: Typo in `EUMERATE`.

src/robot/parsing/model.py
pekkaklarck: I think I already commented somewhere else that you can generate API docs by running `doc/api/generate.py`. I checkout out this PR and was able to run it just fine with these changes and the result looks good too. Notice the two comments above, though.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
edbrannin: Unless you can figure out why Sphinx hates my tables, I'm all for switching to a plaintext format.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -397,7 +397,104 @@ lower limit, upper limit and step.
...
+.. table:: For in enumerate comparison
+   :class: example
+   =================  ================  ===============  =============  ===================  ========
```
pekkaklarck: There should be an empty row after the table declaration and the table itself:

```restructuredtext
.. table:: For in enumerate comparison
   :class: example

=================  ================  ===============  =============  ===================  ========
```

After adding empty row to all examples I was able to generate User Guide without problems.

pekkaklarck:
1. I added few more comments. Can you take a look at them and fix those you consider fixing?

2. As the last comment shows, I found the problem you had with generating User Guide.

3. The conflict preventing automatic merge seems to be in AUTHORS.txt and thus trivial to fix.

4. Please squash commits once you are done with changes and I'll merge this. This you do that today, this can still get to beta 1 we plan to release tomorrow.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
pekkaklarck: User Guide actually uses standard reStructuredText, not Sphinx. As I already commented elsewhere, the problem you had was not having an empty row after the table declaration.

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
edbrannin: YAY!

edbrannin:
By "Today" do you mean "before tomorrow morning in Finland", or "before today ends in Finland"?

pekkaklarck:
Preferably today in Finnish time. That way all test would be run on CI on various interpreters overnight and inclusion for b1 would be safer. It's already 10pm here, but I'm still going to be working few hours. It plan to have b2 or rc1 next week so missing b1 is not a biggie.

atest/robot/running/for.robot
edbrannin: False.  It's used on the next line (unless `Check Test Case` sets a test variable?).

atest/robot/running/for.robot
edbrannin: Fixed.

src/robot/running/keywordrunner.py
edbrannin: Fixed.

src/robot/running/keywordrunner.py
edbrannin: Fixed.

src/robot/running/keywordrunner.py
edbrannin: Fixed.

src/robot/parsing/model.py
edbrannin: Fixed.

src/robot/parsing/model.py
edbrannin: Is this about just the `py:class::` line, or also the `param/ivar` lines below?

edbrannin:
The last thing for me to do is squash -- which I've never done before.  I'll clone a backup of my repo, find a reasonable-looking HOWTO and cross my fingers.

atest/robot/running/for.robot
pekkaklarck: I think this test just had this one line. It might be that I added a comment to a wrong test, though. No need to worry about this anyway.

src/robot/parsing/model.py
pekkaklarck: That was just about the `py:class::` line.

pekkaklarck:
1. Squashing shouldn't be hard but a backup is always a good idea. Here's a HOWTO I've used successfully:
http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit

2. You apparently still need to merge with master to get AUTHORS.txt fixed.

3. You seem to have already fixed most of my latest comments. Fixing the test that accidentally converts non-list iterables to lists by assigning them to list variables would be good.

edbrannin:
1. I'll try that, thanks.
2. ...[but I did that already](https://github.com/edbrannin/robotframework/commit/26b7005961c4d40d0a0e39626d3ac242aa04863f).
3. I'm pretty sure I did that, too.

pekkaklarck:
The PR still says "We can’t automatically merge this pull request.".

edbrannin:
Rebase: I may have messed this up by merging instead of rebasing during development (and earlier today).  I just tried `git rebase -i origin/master` with my branch checked out and it gave me a list of ALL commits.

edbrannin:
I think this might work:
I'm doing a fresh clone from github, and then I'll make a new branch off master and rebase-squash this branch onto it.

edbrannin:
I need to stop now and go to a meeting; I'll try again in a few hours.

My best attempt so far:
I started an interactive rebase in SourceTree, squashed everything I've done, and then it started hitting merge conflicts on early (pre-refactoring) commits.  I'm not sure how best to handle these (trying to cleanly handle all these intermediate versions seems lie it would be a ton of work), or if I should just make a patch of my combined changes and apply it to a fresh branch.  What do you think?

I see now that somehow my AUTHORS.TXT merge didn't quite make it in.  I'll make sure that's resolved when I do the squash.

pekkaklarck:
I tried squashing the commits myself but run into so many conflicts that I eventually gave up. There ought to be some easier way to convert all changes in a branch into a single commit. If it turns out to be too complicated, just merging this PR in the current form (with AUTHORS fixed) is not that bad.

edbrannin:
(It turns out I had done my last review-fixes on the wrong branch earlier.  Whoops.)

It's starting to look like the only way to easily squash would be to make a patch from `upstream/master` to `origin/1952-for-in-zip` and then apply it to a fresh branch from `upstream/master`.

I'm having a bit of trouble with that from here (yay Windows), so I'm just going to merge one last time and push now.   aaaand done.  NOW I'm going.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/102/

pekkaklarck:
Squashing commits seems to be too problematic due to big changes to running keywords during development of this PR. We'll merge this as-is and also fix the conflicts caused by our latest changes.

edbrannin:
Oh, new conflicts?  I'll take care of this in an hour or two.

pekkaklarck:
No worries, we'll fix commits when we merge this.

pekkaklarck:
Merged. Thanks for a great PR @edbrannin!

Issue #1992, opened 2015-05-13
Listeners are not unregistered when using `TestSuite.run` API


WojtekKowaluk:
let _a_ and _b_ be instances of TestSuite.
If i run `a.run(listeners='listener.py')` and then after i run `b.run()` then listener will still be used during execution of TestSuite _b_.

WojtekKowaluk:
Seems this fixes the problem:
https://github.com/robotframework/robotframework/pull/1993

pekkaklarck:
Have you tested does this happen also with `robot.run` API or only with `TestSuite.run`?

WojtekKowaluk:
I will check that.

WojtekKowaluk:
I have tried and this does not happen with `robot.run`.

pekkaklarck:
Interesting. Have you analyzed why not?

WojtekKowaluk:
So far i was not able to figure that out.

WojtekKowaluk:
I found the reason.
When I call `robot.run`, `execute' method use `_logging` contextmanager. This manager calls `close` method on listener object. This method clears `self._loggers` in logger object. When using `TestSuite.run` non of that is called.
Anyway I think I am on the track with my fix to unregister listeners in `close` method for output object. Let me know what do you think.

pekkaklarck:
Great that you found the root cause! Your fix sounds good too. It would great if you could create a PR with unit tests and your fix. I can then review it and look at the related code too.

WojtekKowaluk:
It is already done: https://github.com/robotframework/robotframework/pull/1993
Or do you mean we need something more?

pekkaklarck:
Sorry, forgot about that. Hopefully I have time to look at it and some other PRs tomorrow.

pekkaklarck:
PR #1993 was merged in 0886b61. This issue can be considered done. Thanks @WojtekKowaluk!!

pekkaklarck:
Unfortunately the fix in PR #1993 broke listener interface slightly. Now that listeners are closed already after execution has ended, they do not get notification when log and report are created. There are tests that catch this error but unfortunately our CI has been down lately and the PR wasn't automatically tested.

Good news is that I have a new fix that looks promising. The idea is to use `LOGGER` as a context manager so that it can be nested, and automatically call `LOGGER.close()` when the context manager on top exits.

WojtekKowaluk:
I'm sorry to hear that. :(

[...]

PR #1993, opened 2015-05-13
Unregister listeners on Output.close

WojtekKowaluk:
Fix for https://github.com/robotframework/robotframework/issues/1992

### DIFF
--- a/src/robot/output/output.py
+++ b/src/robot/output/output.py
@@ -26,15 +26,16 @@ class Output(AbstractLogger):
     def __init__(self, settings):
         AbstractLogger.__init__(self)
         self._xmllogger = XmlLogger(settings['Output'], settings['LogLevel'])
+        self._registered_loggers = []
         self._register_loggers(settings['Listeners'], settings['DebugFile'])
         self._settings = settings
 
     def _register_loggers(self, listeners, debugfile):
         LOGGER.register_context_changing_logger(self._xmllogger)
-        for logger in (Listeners(listeners), LibraryListeners(),
-                       DebugFile(debugfile)):
+        for logger in (Listeners(listeners), LibraryListeners(), DebugFile(debugfile)):
             if logger:
                 LOGGER.register_logger(logger)
+                self._registered_loggers.append(logger)
         LOGGER.disable_message_cache()
 
     def register_error_listener(self, listener):
@@ -45,6 +46,9 @@ class Output(AbstractLogger):
         self._xmllogger.close()
         LOGGER.unregister_logger(self._xmllogger)
         LOGGER.output_file('Output', self._settings['Output'])
+        for logger in self._registered_loggers:
+            logger.close()
+            LOGGER.unregister_logger(logger)
 
     def start_suite(self, suite):
         LOGGER.start_suite(suite)
@@ -70,4 +74,3 @@ class Output(AbstractLogger):
     def set_log_level(self, level):
         pyloggingconf.set_level(level)
         return self._xmllogger.set_log_level(level)
-
--- a/utest/api/test_run_and_rebot.py
+++ b/utest/api/test_run_and_rebot.py
@@ -90,6 +90,14 @@ class TestRun(RunningTestCase):
         assert_equals(run_without_outputs(self.data, exclude='fail'), 0)
         self._assert_outputs([('FAIL', 0)])
 
+    def test_listeners_unregistration(self):
+        module_file = join(ROOT, 'utest', 'resources', 'Listener.py')
+        assert_equals(run_without_outputs(self.data, listener=[module_file+":1"]), 1)
+        self._assert_outputs([("[from listener 1]", 1), ("[listener close]", 1)])
+        self._clear_outputs()
+        assert_equals(run_without_outputs(self.data), 1)
+        self._assert_outputs([("[from listener 1]", 0), ("[listener close]", 0)])
+
 
 class TestRebot(RunningTestCase):
     data = join(ROOT, 'atest', 'testdata', 'rebot', 'created_normal.xml')
--- a/utest/resources/Listener.py
+++ b/utest/resources/Listener.py
@@ -1,0 +1,14 @@
+import sys
+
+
+class Listener:
+    ROBOT_LISTENER_API_VERSION = 2
+
+    def __init__(self, name):
+        self.name = name
+
+    def start_suite(self, name, attrs):
+        sys.__stdout__.write("[from listener {}]\n".format(self.name))
+
+    def close(self):
+        sys.__stdout__.write("[listener close]\n")
--- a/utest/running/test_running.py
+++ b/utest/running/test_running.py
@@ -11,7 +11,8 @@ from resources.runningtestcase import RunningTestCase
 
 
 CURDIR = dirname(abspath(__file__))
-DATADIR = normpath(join(CURDIR, '..', '..', 'atest', 'testdata', 'misc'))
+ROOTDIR = dirname(dirname(CURDIR))
+DATADIR = join(ROOTDIR, 'atest', 'testdata', 'misc')
 
 
 def run(suite, **kwargs):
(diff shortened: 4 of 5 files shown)


robotci:
Admin, can CI be enabled?

utest/running/test_running.py
```diff
@@ -251,5 +252,16 @@ def test_reset_logging_conf(self):
...
 
+class TestListeners(RunningTestCase):
+
+    def test_listeners(self):
+        module_file = join(ROOT, 'utest', 'resources', 'Listener.py')
+        suite = build('setups_and_teardowns.robot')
+        suite.run(output=None, log=None, report=None, listener=module_file+":1")
+        self._assert_outputs([("[from listener 1]", 1)])
+        suite.run(output=None, log=None, report=None)
+        self._assert_outputs([("[from listener 1]", 1)])
```
pekkaklarck: Do I get it right that the verified `"[from listener 1]"` was already created by the first run and here it is tested that the second run didn't produce new output? I think it would be better to add a method like `RunningTestCase._clear_outputs` that could be called before the second run and then verified that outputs are empty after the second run.

src/robot/output/output.py
```diff
@@ -31,10 +32,12 @@ def __init__(self, settings):
 
     def _register_loggers(self, listeners, debugfile):
         LOGGER.register_context_changing_logger(self._xmllogger)
-        for logger in (Listeners(listeners), LibraryListeners(),
-                       DebugFile(debugfile)):
+        loggers = (Listeners(listeners), LibraryListeners(),
+                    DebugFile(debugfile))
+        for logger in loggers:
```
pekkaklarck: Why was `loggers` created? It seems to be used only in the for loop so it could still be inlined.

pekkaklarck:
I looked at the PR and the related code and have two general comments:

- The related code is horrible.
- The PR looks good.

Looking at the code, it seems `Logger.close()` that is used when using `robot.run` closes all registered loggers by calling `close()` of each logger. If I got it right, the `Logger.unregister_logger` call you added doesn't close loggers. If that's the case, listeners' `close` won't be called and also possible debug file will be left open.

Could you verify my analysis by adding `close` to the listener you used in tests and verifying is it called or not? If it's not called, we need to figure out a good place to close unregistered loggers. It would be convenient if `LoggerCollection.unregister_logger` would do that automatically, but I'm not sure do loggers it gets always have `close`.

src/robot/output/output.py
WojtekKowaluk: ahh, that is leftover after some debug code I had.

WojtekKowaluk:
Yes, you are right. After my change `close` of listener is not called anymore. I will look into it. I will also improve unit tests.

pekkaklarck:
Did I get it right that abbf1eb didn't yet fix problems with `close`?

@robotci: Whitelist

WojtekKowaluk:
Yes, I'm still thinking how to do that right. Maybe you have some idea?

src/robot/output/output.py
```diff
@@ -22,6 +22,7 @@
 
 
 class Output(AbstractLogger):
+    _registered_loggers = []
```
pekkaklarck: Is there a reason to have `_registered_logger` as a class attribute? Is there a need to share a list of loggers? If not, this definitely should be an instance attribute.

pekkaklarck:
There seems to be two easy solutions how to close loggers:

1. `Output.close` can all `close` for all its `_registered_loggers`. This is pretty much what `Output.close` does before resetting all loggers. 

2. `LoggerCollection.unregister_logger` that actually unregisters loggers calls `close` for the `logger` it gets. This has a problem that all loggers don't necessarily have `close`. There could also be cases where the caller wants to call close separately.

WojtekKowaluk:
Thank you. I have decided to use first way. Seems to be best way for now. I also did other changes you have requested.

I have one extra question:
In test listener I had to use `sys.__stdout__` for printing because it seems nothings else works with `RunningTestCase`. Is this OK? Is there better way?

pekkaklarck:
Using `sys.__stdout__` ought to be fine. I guess Robot has `sys.stdout` intercepted at that point. Would be a lot easier to test this if APIs allowed passing listeners as objects. :)

src/robot/output/output.py
```diff
@@ -45,6 +46,9 @@ def close(self, result):
         self._xmllogger.close()
         LOGGER.unregister_logger(self._xmllogger)
         LOGGER.output_file('Output', self._settings['Output'])
+        for logger in self._registered_loggers:
+            logger.close()
+        LOGGER.unregister_logger(*self._registered_loggers)
```
pekkaklarck: It would perhaps be a little more clear to format the above like this:

```python
for logger in self._registered_loggers:
    logger.close()
    LOGGER.unregister_logger(logger)
```

utest/running/test_running.py
```diff
@@ -12,6 +12,7 @@
 
 CURDIR = dirname(abspath(__file__))
 DATADIR = normpath(join(CURDIR, '..', '..', 'atest', 'testdata', 'misc'))
+ROOT = dirname(dirname(dirname(abspath(__file__))))
```
pekkaklarck: `PROJECT_ROOT` would be a bit more explicit name than `ROOT`. Regardless the name, `DATADIR` could be defined based on it like

```python
DATADIR = join(ROOT, 'atest', 'testdata', 'misc')
```

pekkaklarck:
This PR looks very good. For consistency reasons new `RunningTestCase.close_output` should be renamed to `_close_outputs`. Other two line notes are probably matter of taste and not really that important.

After you are done, please squash the commits and I'll merge this. If you can do it during Tuesday, we'll still get it to 2.9a2.

jussimalinen:
@WojtekKowaluk, do you think you get this PR finalised today? We try to create 2.9a2 on Wednesday and are now descoping issues we cannot get into it. This is pretty much ready and we'd like to get it closed.

WojtekKowaluk:
I am working on it right now. I will do my best to finish it today.

WojtekKowaluk:
@pekkaklarck Are you sure you want this method to be named `_close_outputs` and not `_clear_outputs`. It makes more sense for me.

WojtekKowaluk:
@pekkaklarck @jussimalinen commits are squashed now, let me know if anything else should be done.

utest/api/test_run_and_rebot.py
```diff
@@ -90,6 +90,14 @@ def test_multi_options_as_single_string(self):
         assert_equals(run_without_outputs(self.data, exclude='fail'), 0)
         self._assert_outputs([('FAIL', 0)])
 
+    def test_listeners_unregistration(self):
+        module_file = join(ROOT, 'utest', 'resources', 'Listener.py')
+        assert_equals(run(self.data, outputdir=TEMP, report='none', listener=[module_file+":1"]), 1)
```
pekkaklarck: Is `run` needed here instead of `run_without_outputs` that's used by above tests?

utest/running/test_running.py
```diff
@@ -11,7 +11,9 @@
 
 
 CURDIR = dirname(abspath(__file__))
-DATADIR = normpath(join(CURDIR, '..', '..', 'atest', 'testdata', 'misc'))
+ROOTDIR = dirname(dirname(CURDIR))
+DATADIR = join(ROOTDIR, 'atest', 'testdata', 'misc')
+
```
pekkaklarck: Extra line added here.

utest/resources/Listener.py
```diff
@@ -0,0 +1,14 @@
...
+    ROBOT_LISTENER_API_VERSION = 2
+
+    def __init__(self, name):
+        self.name = name
+
+    def start_suite(self, name, attrs):
+        sys.__stdout__.write("[from listener {}]\n".format(self.name))
+
+    def close(self):
+        sys.__stdout__.write("[listener close]\n")
```
pekkaklarck: Is calling this `close` method actually tested somewhere? Should it be added to both of the newly added tests?

pekkaklarck:
Yeah, definitely `_clear_outputs` and not `_close_outputs`.

In the final review I noticed that calling listeners' close method doesn't seem to be tested. Can you still take a look at that and the other two line notes? Sorry about not noticing these earlier.

WojtekKowaluk:
Fixed.

src/robot/output/output.py
```diff
@@ -45,6 +46,9 @@ def close(self, result):
         self._xmllogger.close()
         LOGGER.unregister_logger(self._xmllogger)
         LOGGER.output_file('Output', self._settings['Output'])
+        for logger in self._registered_loggers:
+            logger.close()
+            LOGGER.unregister_logger(logger)
```
pekkaklarck: Should also reset `self._registered_loggers`. I can add that.

PR #1995, opened 2015-05-15
Support list of instances at ROBOT_LIBRARY_LISTENER

vkruoso:
This is a pull request for issue #1970.


robotci:
Admin, can CI be enabled?

src/robot/running/testlibraries.py
```diff
@@ -91,13 +91,17 @@ def doc(self):
...
-            return self._get_listener(self.get_instance())
+            return self._get_listeners(self.get_instance())
         return None
 
-    def _get_listener(self, inst):
-        return getattr(inst, 'ROBOT_LIBRARY_LISTENER', None)
+    def _get_listeners(self, inst):
+        try:
+            listeners = getattr(inst, 'ROBOT_LIBRARY_LISTENER')
+            return listeners if isinstance(listeners, list) else [listeners]
```
pekkaklarck: Testing just against Python list has a problem that Java libraries cannot register a list of listeners. We have `utils.is_list_like` that could be used here instead, but it has a potential problem that a listener itself could be iterable. I guess iterable listeners are rare enough that this isn't a big problem. If someone actually faces that problem, they can wrap the iterable listener into a list.

src/robot/running/testlibraries.py
```diff
@@ -91,13 +91,17 @@ def doc(self):
...
+        except AttributeError:
```
pekkaklarck: I always find it a bit risky to catch `AttributeError` when there's more than one line of code. It's so easy to trigger this exception by just having a typo or changing APIs elsewhere so that an attribute is removed. That's not likely here, but I'd still prefer having `try/except AttributeError` just around `getattr`. It might be even better to use `hasattr`, though:

```python
if not hasattr(inst, 'ROBOT_LIBRARY_LISTENER'):
    return None
listeners = inst.ROBOT_LIBRARY_LISTENER
# ...
```

pekkaklarck:
1. The code looks good except that only accepting Python lists is not desirable.

2. It seems that you have forgotten to add new test files. At least I couldn't see `import_library_multiple_listeners.robot` anywhere. Related to that, I'm not sure is there a need to test this functionality with `Import Library` keyword. Tests should be added to `as_listener.robot` and `as_listener_java.robot`, though.

3. New functionality needs to be documented in the User Guide.

vkruoso:
Thanks for your valuable inputs @pekkaklarck. Will try to improve the PR over this weekend.

vkruoso:
I've updated the PR with changes regarding your comments. The commits are a little mess, but I can rebase before the merge anyway. I have two questions:

* Is the User Guide information enough?
* Why do you keep Java compiled classes in the repository?

pekkaklarck:
1. Code looks great. Please squash it and we are ready for merge.

2. User Guide info looks good too, but it would be even better if the fact that any "list like" object is OK would be somehow mentioned.

3. Compiled Java test code is in the repository mainly because we haven't bothered to automate compilation when running tests on Jython. Recompilation when running tests would also mean that JRE wouldn't be enough to run tests.

vkruoso:
1 & 2. Will try to improve the User Guide to make that clear.

3. Forgive me as I'm mostly a Java newbie. But it matter how do I compile those classes (compiler, and version)? My goal is to push a `.class` file that is compatible with the ones that are already in the repository.

pekkaklarck:
There's `atest/testresources/compile_java.sh` that you can use to compile all `*.java` files under `atest/testresources/testlibs` and `atest/testresources/listeners`. It ought to set needed compiler options correctly (and should actually be updated now that Jython 2.7 that we require works only on Java 7 (a.k.a. 1.7) or newer). If you are on Windows where `*.sh` doesn't really work, you can check out the options from the script anyway. If you do use the script, please revert changes to `*.class` files related to `*.java` files you haven't actually changed.

vkruoso:
Compiling (using jython2.7) with `-target 1.5 -source 1.5` I get differences in five files, and with `-target 1.7 -source 1.7` all of them differs (or close to it). Would you like me to create a separate PR to update the script and the class files? Anyway, I think for now in this PR it would be best to leave it with the 1.5.

pekkaklarck:
Using 1.5 is definitely OK in this PR. If you want, you can create a separate PR updating version to 1.7, but we can just do that ourselves too at some point.

Let me know when this PR is ready for a final review. We try to get RF 2.9a2 released early next week.

vkruoso:
I've fixed the java files, updated the user guide and squashed commits. I thinks this is now ready for a final review. Did this ran over your CI environment? 

Once this is merged, I'll be happy to provide a PR updating all java compiled as mentioned before.

pekkaklarck:
Our CI is down right now. We'll review and hopefully merge this on Monday.

pekkaklarck:
The PR looks great! Would you @vkruoso still like to add yourself to AUTHORS.txt in project root? No need to squash again if you don't want to. I can also add you there myself after merging this.

vkruoso:
I wont be able to do it in the following days, so if you can do it after the merge would be appreciated. :)

pekkaklarck:
Ok, I'll update AUTHORS. Thanks for the great PR @vkruoso!

vkruoso:
Thanks for the merge. I'll try to to provide the PR with updated java compiled code as soon as I have a little time for it.

PR #2008, opened 2015-06-01
Recompiled java code with 1.7 (support for Jython 2.7)

vkruoso:
This is a recompilation of Java code with 1.7 version as discussed in PR #1995.


robotci:
Admin, can CI be enabled?

Issue #1985, opened 2015-04-29
String: New `Get Regexp Matches` keyword


synsun:
I want find multiple match result in a regexp. But 'Should Match Regexp' does not support multiple result.

``` python
import re
p = re.compile(ur'(?P<name>a)b', re.MULTILINE)
test_str = u"abcdeabababbbaaba"

re.findall(p, test_str)
```

```
${result}=  Should Match Regexp abcdeabababbbaababc (?P<name>ab)c
log ${result}       

INFO : ${result} = [u'abc', u'ab']
```

I don't know what is the good form to get multiple named result. (maybe ['abc','ab','abc','ab'] or [['abc','ab'],['abc','ab']])
I think add an option is good to apply that. Thanks.

pekkaklarck:
Do you have a concrete idea how to enhance `Should Match Regexp` without breaking backwards compatibility?

MoonSungHoon:
```
Should Match Regexp | ${string} | ${regexp} | global=false
```
If global option is true, return all matches. ( Don't return on first matches. )
Python regexp default value is 'global=on'. But not to break backwards compatibility, above option should be needed.

MoonSungHoon:
Suggest new built-in keyword like below.
```
    def get_matches_regexp(self, string, regexp, group_name=None):
        """Returns match result of regexp

        Example:
        | ${results} =         | Get Matches Group Regexp | ${string}         | ${regexp} | ${group_name} |
        """
        import re
        c = re.compile(regexp, re.MULTILINE)
        if group_name:
            m = c.match(string)
            results = [m.groupdict()[group_name] for m in c.finditer(string)]
        else:
            results = [''.join(m) for m in c.findall(string)]
        return results
```

pekkaklarck:
Could you explain, preferably with examples, how that global option would change the behavior? How would the implementation look like?

What is the use case for the new keyword you propose? Is that somehow related to the earlier enhancement you proposed?

MoonSungHoon:
Okay, the new keyword is another way to get regexp results with global option.

This is an example why global option is needed.
I want find some longitude and latitude in a big text messages. That include some kind of texts and longitude, latitude ( I cannot define how many longitude, latitude ). After that using results of regexp filtering it using conditions to printout.

(Below text is just an example, not actual value)
```
Amenu    Bmenu    Logitude    Latitude
amanda  safe         W'114.54   N'36.10
bob         unsafe     W'401.41   N'61.20
[...]

PR #2023, opened 2015-06-15
Added new keywords to String library: Get Regexp Matches #1985

MoonSungHoon:
As discussed in #1985 here is pull request to add new keywords to String library.

I closed and delete previous pull request, because that contain unnecessary merge commits.

Please check this. Thanks.


robotci:
Admin, can CI be enabled?

Issue #2022, opened 2015-06-10
Document that preformatted text with spaces in Robot data requires escaping


yahman72:
The generated documentation doesn't include the "extra" spaces from the source.
e.g.
<pre>
My Keyword
    [Documentation]   Doc with preformatted text
    ...
    ...     Here comes the preformatted text:
    ...     | First line without any extra spaces
    ...     | Second line with   some    extra    spaces
    ...     |      Third line intended
</pre>
libdoc and testdoc remove the extra spaces i.e.
![image](https://cloud.githubusercontent.com/assets/8724383/8080055/88b7a01a-0f69-11e5-8504-6bf1adafdbc3.png)
generated source:
`"<p>Doc with preformatted text\x3c/p>\n<p>Here comes the preformatted text:\x3c/p>\n<pre>\nFirst line without any extra spaces\nSecond line with some extra spaces\nThird line intended\n\x3c/pre>"`

RF version: 2.8.6

pekkaklarck:
The problem here is that whitespace is stripped at parsing time. You unfortunately need to escape it either with a backslash or `${SPACE}`. For example:

```robotframework
*** Keywords ***
My Keyword
    [Documentation]   Doc with preformatted text
    ...
    ...     Here comes the preformatted text:
    ...     | First line without any extra spaces
    ...     | Second line with \ \ some \ \ \ extra \ \ \ spaces
    ...     | ${SPACE * 5}Third line intended
```

I know this is far from ideal, but I don't think we can do anything to it in the near future.

yahman72:
Okay, thanks - Maybe update the docs to match the above?
Now the user guide writes:
"Starting from Robot Framework 2.7, it is possible to embed blocks of preformatted text in the documentation. Preformatted block is created by starting lines with '| ', one space being mandatory after the pipe character except on otherwise empty lines. The starting '| ' sequence will be removed from the resulting HTML, but **all other whitespace is preserved**.
"

yahman72:
PS. 
I could create a PR for the Doc change, but I'm not 100% certain how the user guide is managed as it seems to be possible to edit the docs also online.
Is it okay to do the changes (like I would do for a code-change) in my fork and then submit a PR? ... or is there another way that should be used?

pekkaklarck:
You are right that there should be a note that in Robot test data whitespace isn't preserved but needs to escaped.

[...]

PR #2025, opened 2015-06-16
Preformatted Text Doc fix

yahman72:
Corrected documentation for preformatted text.
See #2022


robotci:
Admin, can CI be enabled?

pekkaklarck:
It needs to be explained that escaping is only needed in Robot data. I'll add a note about that myself. Thanks for the issue and the original PR anyway.

Issue #1936, opened 2015-03-02
Detect unused imports in dry-run mode


pekkaklarck:
Detecting and reporting unused library and resource imports in dry-run mode would ease maintaining test data. Resource file should be considered used also if it contains any additional imports.

Unfortunately we cannot really detect unused variable file imports because dry-run doesn't validate variables in general. Similarly resource files containing only variables will be incorrectly reported as unused. Need to think are these false positives so big a problem that it should be possible to somehow disable this functionality.

This functionality has been requested by Zhang Yu from NSN.

stasiek:
@pekkaklarck how do You propose could we crunch it? Since @jussimalinen closed #1998 We could use that just to create set of libraries from which all parsed keywords originated and later compare it with all imports, right?

What I'm thinking is we need three things:
1. place to create set of all used imports
2. add items to the set witch from each keyword
3. compare the set with *** Settings ***

Should I think about it as a new listener, or should this happen some other place?

pekkaklarck:
Assuming listeners actually work in dry-run, using them to collect information about used libs and resources is a good idea. The listener should also get information about imports somehow. I haven't looked at that, but assume that the info is somehow available through namespace that ought to be accessible through context (see e.g. BuiltIn).

pekkaklarck:
One thing to decide is do we want to enable this functionality automatically or should we have a separate command line option like `--detectunusedimports` or `--unusedimportwarning`. If we have an option, it would be good to have it available also when running tests normally (which is probably also easier than having it available only in dry-run). We then should also decided should the option be enabled automatically in dry-run or not. Because there can be false positives (i.e. a resource containing only other imports and/or variables but no keywords), having it off by default might actually be a good idea.

pekkaklarck:
We have decided to implement this using listeners. They already get a notification when a keyword is called, but they don't know what libraries or resource files are imported. Issue #2030 covers adding a new listener method to get that information and this issue is basically blocked until it has been implemented.

jussimalinen:
We discussed this further with @pekkaklarck and realized that there are a lot of open questions on how exactly this should work. 

Obviously if a test suite file contains an import for a library or a resource which is not used at all, there should be a warning.
But what about these cases:

- What if a resource does not implement any keywords itself, but imports other resources, from which keywords are used? (This is actually a fairly common use case and I dont there should be a warning.)
- What if two resource files both dont directly provide a keyword that is used, but both import the same library, which is used. Should there be a warning?
[...]

PR #2026, opened 2015-06-16
Detect unused imports in dry-run mode #1936

stasiek:
Peke take a look at this and please comment.

### DIFF
--- a/src/robot/output/importvalidator.py
+++ b/src/robot/output/importvalidator.py
@@ -1,0 +1,57 @@
+#  Copyright 2008-2015 Nokia Solutions and Networks
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


robotci:
Admin, can CI be enabled?

pekkaklarck:
1. The overall approach looks good.

2. Adding a new listener method that gets events when something is imported definitely needs a separate issue and pull request. It will also need these individual enhancements:
  - It must be possible to separate library and resource import and also get notification about variables. We could have separate `library`, `resource` and `variables` methods, but it's probably better to have `imported` method that gets the import type as the first argument. The listener API already has `output_file` method that works similarly.
  - Acceptance tests.
  - Documentation in User Guide.

3. We need to think how to report unused imports. Just listing them on the console isn't enough because people are generally looking at produced log.html. Perhaps reporting unused imports as errors is the best solution.

4. We need to think how to enable this functionality. Should it be enabled automatically, only when using dry-run, or by using a separate command line option?

5. If possible, library should not be considered unused if it contains a listener. This is especially important if this inspection is automatic.

6. This functionality will need tests and documentation.

7. Commented code should just be removed.

pekkaklarck:
It's better to add `imported` method to listener interface (#2030) first. After that this PR will be outdated.

Issue #2030, opened 2015-06-17
Notify listeners about library, resource and variable file imports


stasiek:
A new listener method that gets events when something is imported is needed.

> It must be possible to separate library and resource import and also get notification about variables. We could have separate library, resource and variables methods, but it's probably better to have imported method that gets the import type as the first argument. The listener API already has output_file method that works similarly.
> copied from: https://github.com/robotframework/robotframework/pull/2026#issuecomment-112553868

pekkaklarck:
Have you had time to look at this? We are going to release beta 1 on Friday and there's no time to get this included into it anymore. We try to get beta 2 or perhaps even rc 1 out next week. We thus need to get this implemented early next week to be able to also get #1936 in. I can also take a look at this myself after we've got other issues done.

jussimalinen:
In order to track transitive imports from resource files, the imported-method will need as an argument the origin of the import (the file in which the import happens as a path) and then probably also it would be a good idea to include the path to the imported file as a source. With that information it should be possible to map a tree about the imports.

pekkaklarck:
@jussimalinen, I reviewed the final version of code and tests and everything looked good. I did a little clean-up for tests in the previous commit, though, and also added docs a little earlier. Please review my changes and we pretty close to cal this issue done.

When writing docs, I noticed keys in attribute dicts consisting of multiple words have earlier used `wordstogether` style like `starttime` and `totaltests`. Should we be consistent and change new `original_name` to `originalname` or `origname`?

jussimalinen:
Consistency is good, so I it should be changed. I would prefer `originalname`  over `origname`.

pekkaklarck:
Yeah, let's use `originalname`.

PR #2050, opened 2015-06-29
Extend listener interface with imported() method #2030

stasiek:
imported() consumes both name and type of import.

What else should I do with this?


robotci:
Admin, can CI be enabled?

src/robot/output/listeners.py
```diff
@@ -74,6 +74,10 @@ def _import_listeners(self, listener_data):
                              % (listener, unicode(err)))
         return listeners
 
+    def imported(self, name, import_type):
+        for listener in self._listeners:
+            listener.call_method(listener.imported, name, import_type)
```
pekkaklarck: 1. Should have `import_type` argument first and `name` second to be consistent with `output_file` method (where `name` argument should really be called `file_type`).

2. Should call method with third argument `attrs` that could be an empty dictionary at this point. It would allow us adding more information to the call by adding items to the `attrs` dictionary similarly as we've done with `start_xxx` and `end_xxx` method. Such information could include arguments to the imported thing, source where it was imported from, etc.

src/robot/running/runner.py
```diff
@@ -63,6 +63,10 @@ def start_suite(self, suite):
                                          self._settings.skip_teardown_on_exit)
         ns = Namespace(self._variables, result, suite.resource.keywords,
                        suite.resource.imports)
+
+        for imported_item in suite.resource.imports:
+            self._output.imported(imported_item.name, imported_item.type)
+
```
pekkaklarck: Unnecessary empty lines. We generally avoid them inside functions/methods.

pekkaklarck:
1. Changes to listener API itself look pretty good. Arguments to the method should still be changed as commented, though.

2. The current code only notifies about resource imports, not about libraries or variable files. It also only notifies about resource imports in test case file itself, not about resource imports in resources or done dynamically using `Import Resource` keyword.

3. Problems in 2. ought to be at least mostly fixed by calling listener in `Namespace` where libraries, resources and variable files actually are imported. Notice that listener should only be called if an import succeeds. Listeners get notifications about errors separately.

4. Tests are totally missing. At least all cases discussed above should be tested.

5. Documentation in the User Guide is missing too.

src/robot/output/listeners.py
stasiek: When talking about attrs as additional parameter. Should it look like def imported(self, import_type, name, attrs) or shall I just pass the whole suite and attrs?
grumps
I've tried such approach before, but YAGNId it. Take a look: https://github.com/stasiek/robotframework/blob/import-validation/src/robot/output/importvalidator.py
grumps
There we'd have source, the rest of stuff and in suite.resource.variables - variables so we'll be able to proceed with import validation.

stasiek:
For now I've pushed new commit. @pekkaklarck the imported() method seams to work for Resources and Libraries both! 

Here's a directory with files I've used for testing: http://dl.dropbox.com/u/1286991/bloodrop/rf_test_listener_api.zip

in importvalidator.py You'll see it actually gets the type of import, passed parameters and name! Haven't checked with variable files yet, but the rest works fine for me.

I'll try writing some 'Solid' tests in the evening.

You can check how it works by running it - running command I've used is in HOWTO_RUN.sh and resulting screen in same dir.

Now I gotta go afk. Laters!

pekkaklarck:
It seems that you haven't updated this PR with your latest commits. No worries with that at this point though, we'll continue with this issue ourselves.

Issue #1869, opened 2014-12-11
Variable errors should not exit `Wait Until Keyword Succeeds`, `Run Keyword And Expect Error`, etc.


bjornfor:
I realize that whether or not this is considered a bug depends on how the documentation for "Wait Until Keyword Succeeds" is interpreted:

```
All normal failures are caught by this keyword. Errors caused by invalid syntax,
test or keyword timeouts, or fatal exceptions (caused e.g. by Fatal Error) are not caught.
```

(Is AttributeError a fatal exception?) But even if the behaviour is currently documented, I think it is really unfortunate that AttributeError is not a normal error.

Imagine that you have a test library that retrieves some object. On error, an exception is raised and None is returned instead. And if this is in a Teardown context, errors are ignored and robot continues to the next keyword. This keyword then runs into the AttributeError because a None object was received instead of the real/expected object. Now the "Wait Until ..." is stopped before the timeout.

How to get out of the above situation, unless AttributeError is handled like a normal error?

Here are two test cases to play around with (at least for my own reference):

```
*** Test Cases ***
"Wait until ..." with normal error
    # Keyword is run multiple times, until timeout. Each run gives an exception
    # traceback.
    Wait Until Keyword Succeeds    1 sec    0.5    Keyword With Normal Error

"Wait until ..." with AttributeError
    # Keyword is run only once, even if there is time left until the timeout.
    # There is no exception traceback like above.
    Wait Until Keyword Succeeds    1 sec    0.5    Keyword With AttributeError

*** Keywords ***
Keyword With Normal Error
    ${obj} =    Evaluate    "foo"
    Should Be Equal As Strings    ${obj}    "bar"

Keyword With AttributeError
    # In real life, this would get an object and use some of its (valid) attributes.
    # In case of an error, and in Teardown context (continue-on-failure), a None object
    # is returned instead causing the next keyword to create an AttributeError.
    ${obj} =    Evaluate    "foo"
    Should Be Equal As Strings    ${obj.bad_attr}    "foo"
```

pekkaklarck:
After thinking this a bit more, I don't think changing the type of variable related errors is worth the effort. DataErrors are raised and handled in so many places that changing them, and making sure nothing breaks, would just be too big a task.

We could consider changing WUKS so that it catches also DataErrors, though. What do you think? As discussed above, WUKS will eventually fail anyway, even if the error would be caused by a syntax error. Timeouts definitely should be uncatchable also in the future.

[...]

PR #2051, opened 2015-06-29
Fix for issue : AttributeError stops "Wait Until Keyword Succeeds" #1869

jcdevil:
@pekkaklarck : PR concerning the approach you proposed (creation of a new VariableError, based on DataError). Although I'; interested in investigating the approach to separate DataError and VariableError, I've not finished that at the moment :(  But WDYT about this first approach ?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,7 +7,7 @@
   <artifactId>robotframework</artifactId>
   <packaging>jar</packaging>
   <name>Robot Framework</name>
-  <version>2.9b2</version>
+  <version>2.9.dev20150702</version>
   <description>High level test automation framework</description>
   <url>http://robotframework.org</url>
   <licenses>
--- a/src/robot/version.py
+++ b/src/robot/version.py
@@ -17,7 +17,7 @@ import sys
 
 # Version number typically updated by running `invoke set_version <version>`.
 # Run `invoke --help set_version` or see tasks.py for details.
-VERSION = '2.9b2'
+VERSION = '2.9.dev20150702'
 
 
 def get_version(naked=False):


robotci:
Admin, can CI be enabled?

pekkaklarck:
1. The approach itself looks fine. I doubt trying to separate `DataError` and `VariableError` otherwise would be worth it.

2. `VariableError` should be used every time a variable does not exist, not only when the extended variable evaluation fails.

3. Variable related failures should not have `VariableError:` prefix. The prefix is dropped from the error in `robot.utils.error.ErrorDetails`, and I refactored in 01c591e71f76df374a1b039dda8f45bec96e2462 and now that ought to be done automatically.

4. You should merge with master to get the change mentioned above and also other recent changes.

5. If you want this enhancement to be included in RF 2.9, it should be ready by Wednesday to be included into beta 2 or latest by Monday to be included in the first release candidate.

jcdevil:
@pekkaklarck : thx for your feedback ;)
1- OK for this approach
2- OK then `VariableError` will be raised when variable doesn't exist (including case of errors with `identifier`)
3- Good point that name of the error had indeed not a great interest for the user - thx for the update
4- Done updating the PR today ;)

src/robot/errors.py
```diff
@@ -53,6 +53,14 @@ class DataError(RobotError):
     this exception with care.
     """
 
+class VariableError(DataError):
+    """Used when variable syntax or evaluation is invalid.
+
+    VariableErrors are caught by keywords that run other keywords
+    (e.g. `Run Keyword And Expect Error`). Libraries should thus use
+    this exception with care.
```
pekkaklarck: I think the last sentence can be removed. Libraries should never use these exceptions. If we want to make that more explicit, it could be documented either on module level or in RobotError base class.

src/robot/variables/finders.py
```diff
@@ -117,12 +117,12 @@ def find(self, name):
         try:
             variable = self._find_variable('${%s}' % base_name)
         except DataError as err:
-            raise DataError("Resolving variable '%s' failed: %s"
+            raise VariableError("Resolving variable '%s' failed: %s"
                             % (name, unicode(err)))
```
pekkaklarck: Indentation not fixed here or with the next VariableError. It was fixed above.

atest/robot/standard_libraries/builtin/wait_until_keyword_succeeds.robot
```diff
@@ -70,6 +70,9 @@ Invalid Keyword Inside Wait Until Keyword Succeeds
 Keyword Not Found Inside Wait Until Keyword Succeeds
     Check Test Case    ${TESTNAME}
 
+Advanced Variable Syntax Inside Wait Until Keyword Succeeds
```
pekkaklarck: The term we use is "extended variable syntax", not "advanced variable syntax".

pekkaklarck:
Looks good to me. Few more things I'd like to see:

- Test also normal variable errors with WUKS, not only extended variable error.
- Add similar test(s) also to Run Keyword And Expect Error.
- Add some kind of a note to the affected keywords that variable related errors were not caught prior to RF 2.9.
- Add yourself to AUTHORS.txt.

src/robot/errors.py
jcdevil: OK. Removed on my local copy. PR will be updated soon

src/robot/variables/finders.py
jcdevil: OK. Removed on my local copy. PR will be updated soon

atest/robot/standard_libraries/builtin/wait_until_keyword_succeeds.robot
jcdevil: OK. Fixed on my local copy (2 occurrences). PR will be updated soon

jcdevil:
@pekkaklarck 
1- OK. I also added tests with list and dictionaries
2- In progress : I will probably split the files `run_keywords_with_errors` in `run_keyword_and_expect_error` and `run_keyword_and_ignore_error` : WDYT ? - will be pushed tomorrow
3- I suppose you're speaking about keywords documentation (docstring) : - will be pushed tomorrow
4-  will be pushed tomorrow

pekkaklarck:
What do you mean with splitting `run_keywords_with_errors` and other files? I don't think this functionality needs to be tested thoroughly with each keyword. It would probably be enough to add just one test for others and test this more thoroughly with WUKS.

Yes, I meant keyword documentation (i.e. method doctrings).

We try to get beta 2 out on Wednesday or latest Thursday so there's not much time for that. First release candidate is planned for early next week, and there's more time for that deadline.

jcdevil:
I just wanted to create 1 file by keyword (`run_keyword_with_errors` is dealing with 2 keywords at the moment ;) ) but this can indeed be done later.
I will just add 1 test for other keywords then and push this asap then.

pekkaklarck:
Ok. There are lot of too big suites and splitting them is generally a good idea. I'd prefer to get them as separate PRs, though. Smallish cleanup is fine also as part of other PRs, but it gets harder to review PRs if there's lot of unrelated noise.

jcdevil:
For information :
I was about to push but finally got an unexpected error in my tests on `Run Keyword and Expect Error` :
```robot
*** Test Cases ***
Expect Error When Access To Nonexisting Variable
    [Documentation]    FAIL  STARTS: Variable '\${nonexisting}' not found.
    Run Keyword And Expect Error    *    Access To Nonexisting Variable

Expect Error When Access To Nonexisting Extended Variable
    [Documentation]    FAIL  STARTS: Resolving variable '\${list[2]}' failed: IndexError: list index out of range
    Run Keyword And Expect Error    *    Access To Nonexisting Extended Variable
*** Keywords ***
Access To Nonexisting Variable
    Log   ${nonexisting}

Access To Nonexisting Extended Variable
    ${list} =    Create list    1    2
    Log   ${list[2]}
```

Launching these 2 tests, I get the error :
```
FAIL: Run Keyword With Errors.Expect Error When Access To Nonexisting Extended Variable
Test was expected to FAIL but it PASSED. Expected message:
STARTS: Resolving variable '${list[2]}' failed: IndexError: list index out of range
```
If you have an idea about the reason where does the problem come from... :)
NB : using it outside the test framework seems however OK ;)

pekkaklarck:
Those tests catch all errors by using `*` as the pattern. Earlier they failed because variable related exceptions were never caught. Now they are caught and thus those tests pass. They should be updated so that the expected error is the actually error, not `*`.

This also reminds me about the probably reason why we decided not to catch variable related errors. Earlier they exit also `Run Keyword And Ignore Error`, but now they are going to be ignored. That can hide errors that should not be hidden, but that's a general problem with `Run Keyword And Ignore Error`.

jcdevil:
OK thx for this info. However, what remains unclear for me is the reason why I get an error in the scalar variable case (so that the test PASS) but NOT in the Extended variable case (so that the test FAIL).
Did I miss anything ? WDYT ?

pekkaklarck:
My guess is that some code is still raising `DataError` and not `VariableError`.

jcdevil:
OK all is explained and fixed... works fine ;) - problem was in `variables/notfound.py`
All remarks taken into account, tests OK (with update of one expected results in `running/test_templates.robot`)

atest/testdata/running/test_template.robot
```diff
@@ -169,7 +169,7 @@ Templates, timeouts, and for loops
     \    0.05 seconds
 
 Templates and syntax errors
-    [Documentation]  FAIL  Variable '\${nonexisting}' not found.
+    [Documentation]  FAIL Several failures occurred:\n\n 1) Variable '${nonexisting}' not found.\n\n 2) not compared != anymore
     [Template]  Should be equal
     ${nonexisting}      something
     not compared  anymore
```
pekkaklarck: The idea of this test is checking that execution of a templated test stops if there is a syntax error. Now that non-existing variable isn't a syntax error anymore, the test doesn't make much sense in this format. It should be updated to use something that truly is a syntax error. It probably would be easiest to use the template keyword with wrong number of arguments:

```robotframework
*** Test Cases ***
Templates and syntax errors
    [Documentation]    FAIL xxx
    [Template]    Should be equal
    we    have    way    too    many    arguments    here
    not compared    anymore
```

It would actually be a good idea to add a new test here verifying that the whole execution isn't stopped if a variable doesn't exist.

atest/testdata/standard_libraries/builtin/run_keyword_and_continue_on_failure.robot
```diff
@@ -57,6 +57,15 @@ Run Keyword And Continue On Failure With Timeout
...
     Fail    This Should Not Be Executed!
 
+Run Keyword And Continue On Failure With Nonexisting Variable
+    [Documentation]    FAIL   Variable '\${nonexisting}' not found.
+    Run keyword And Continue On Failure    Log    ${nonexisting}
+
+Run Keyword And Continue On Failure With Nonexisting Extended Variable
+    [Documentation]    FAIL   Resolving variable '\${list[2]}' failed: IndexError: list index out of range
+    ${list} =  Create list    1    2
+    Run keyword And Continue On Failure    Log  ${list[2]}
```
pekkaklarck: With these tests it's not possible to know did the failure occur because the error escaped `Run Keyword And Continue On Failure` or because test ended and the continued failure was the only error that occurred. You could, for example, add something like `Fail    Normal failure after continuable failure` as the second step in these tests.

atest/testdata/standard_libraries/builtin/run_keyword_with_errors.robot
```diff
@@ -134,6 +140,13 @@ Expect Error When Syntax Error In For Loop
     [Documentation]    FAIL  Variable '\${nonexisting}' not found.
     Run Keyword And Expect Error    *    For Loop With Syntax Error
 
+Expect Error When Access To Nonexisting Variable
+    Run Keyword And Expect Error    Variable '\${nonexisting}' not found.    Access To Nonexisting Variable
+
+Expect Error When Access To Nonexisting Extended Variable
+    ${expected_err_msg} =  Set Variable  Resolving variable '\${list[2]}' failed: IndexError: list index out of range
+    Run Keyword And Expect Error  ${expected_err_msg}  Access To Nonexisting Extended Variable
```
pekkaklarck: It's a bit risky to tests the exact error message from Python because it can be different on different Python interpreters (Python, Jython, IronPython) and they sometimes even change between versions. It would be a bit safer to change `IndexError: list index out of range` to `IndexError: *` in the verified message.

Also, please use four spaces as the separator. We have lot of old tests suites that still use two spaces and for consistency reasons using two is fine on them. This suite uses the nowadays recommended four-space-separator, though.

atest/testdata/standard_libraries/builtin/wait_until_keyword_succeeds.robot
```diff
@@ -104,6 +104,28 @@ Keyword Not Found Inside Wait Until Keyword Succeeds
...
+Extended Variable Syntax Using List Inside Wait Until Keyword Succeeds
+    [Documentation]    FAIL Keyword 'log' failed after retrying for 1 second. The last error was: Resolving variable '${list[2]}' failed: IndexError: list index out of range
+    ${list} =    Create List    1    2
+    wait until keyword succeeds    1 second    0.1s    log  ${list[2]}
+
+Extended Variable Syntax Using Dictionary Inside Wait Until Keyword Succeeds
+    [Documentation]    FAIL Keyword 'log' failed after retrying for 1 second. The last error was: Resolving variable '${dict['key_nonex']}' failed: KeyError: 'key_nonex'
+    ${dict} =  Create Dictionary  key1=val1  key2=val2
+    wait until keyword succeeds    1 second    0.1s    log  ${dict['key_nonex']}
+
```
pekkaklarck: 1. These tests have the same problem as `Run Keyword And Continue On Failure` tests above that the error would be same regardless does the exception exit WUKS or not.

2. I'm actually fairly certain that the three first test fail immediately. You are using non-existing variable as a keyword and I don't think WUKS will run any rounds in that case.

3. It would probably be easiest to verify that WUKS is actually executed several times in test execution side. Let me know if you need help with that. You could also try constructing WUKS where the variable first doesn't exist in the beginning but will be available later. Even in that case it would be a good idea to verify that WUKS has been executed more than once on the execution side.

4. Please use shorter timeout than 1s if it is going to be exhausted. It unnecessarily extends suite execution time.

pekkaklarck:
Few more comment in addition to the line notes I already added:

1. I looked at `DataError` usages under `src/robot/variables` and noticed that `replacer.py` still uses it when it probably should use `VariableError`. I didn't test it, but it looks like that accessing list or dict variable items using syntax `@{list}[index]` and `&{dict}[key]`, respectively, raises `DataError` if index/key is invalid. Shouldn't be hard to fix but tests should also be added.

2. Please merge with master and resolve conflict in AUTHORS.txt.

jcdevil:
Will create a new PR with squashed commits.

atest/testdata/running/test_template.robot
jcdevil: OK right.
Test updated locally.

atest/testdata/standard_libraries/builtin/run_keyword_and_continue_on_failure.robot
jcdevil: Right. Thx
Test updated locally.

atest/testdata/standard_libraries/builtin/run_keyword_with_errors.robot
jcdevil: OK remark taken into account on my local copy

atest/testdata/standard_libraries/builtin/wait_until_keyword_succeeds.robot
jcdevil: working on this topic

PR #2057, opened 2015-07-03
Fix for issue : AttributeError stops "Wait Until Keyword Succeeds" #1869

jcdevil:
@pekkaklarck 
New PR squashing previous commits and taking into account remarks of previous reviews.
Thx for your comments ;)
About : 
```It would probably be easiest to verify that WUKS is actually executed several times in test execution side. Let me know if you need help with that. You could also try constructing WUKS where the variable first doesn't exist in the beginning but will be available later. Even in that case it would be a good idea to verify that WUKS has been executed more than once on the execution side``` : 
I added a test that initially has a variable affected with None... as you'll see.
Cheers

### CODE REVIEW

**Comment by robotci** [comment]
Admin, can CI be enabled?

File: atest/testdata/running/test_template.robot [hunk 1]
```diff
@@ -169,10 +169,11 @@ Templates, timeouts, and for loops
     \    0.05 seconds
 
 Templates and syntax errors
-    [Documentation]  FAIL  Variable '\${nonexisting}' not found.
+    [Documentation]  FAIL Keyword 'BuiltIn.Should Be Equal' expected 2 to 4 arguments, got 5.
     [Template]  Should be equal
-    ${nonexisting}      something
-    not compared  anymore
+    odd    number    of     compared    arguments
+    not    not    ok    ok
```
pekkaklarck: This actually is valid. Should Be  Equal takes 2-4 arguments.

atest/robot/standard_libraries/builtin/run_keyword_with_errors.robot
```diff
@@ -111,6 +126,21 @@ Expect Error When Syntax Error In Setting Variables
 Expect Error When Syntax Error In For Loop
     Check Test Case    ${TEST NAME}
 
+Expect Error When Access To Nonexisting Variable
+    Check Test Case    ${TEST NAME}
+
+Expect Error When Access To List Variable Nonexisting Index Syntax 1
+    Check Test Case    ${TEST NAME}
+
+Expect Error When Access To List Variable Nonexisting Index Syntax 2
```
pekkaklarck: It's fine to have multiple tests on the test data side for one feature named like `Something 1`, `Something 2`, etc. They should, however, be combined on the execution side:

```robotframework
*** Test Cases ***
Something
    Check Test Case   ${TEST NAME} 1
    Check Test Case   ${TEST NAME} 2
```

atest/testdata/standard_libraries/builtin/run_keyword_with_errors.robot
```diff
@@ -134,6 +149,29 @@ Expect Error When Syntax Error In For Loop
     [Documentation]    FAIL  Variable '\${nonexisting}' not found.
     Run Keyword And Expect Error    *    For Loop With Syntax Error
 
+Expect Error When Access To Nonexisting Variable
+    Run Keyword And Expect Error    Variable '\${nonexisting}' not found.    Access To Nonexisting Variable
+
+Expect Error When Access To List Variable Nonexisting Index Syntax 1
+    ${expected_err_msg} =  Set Variable    Resolving variable '\${list[2]}' failed: IndexError:*
+    Run Keyword And Expect Error    ${expected_err_msg}    Access To List Variable Nonexisting Index Syntax 1
+#    Run Keyword     Access To List Variable Nonexisting Index Syntax 1
```
pekkaklarck: Commented out lines should be just removed.

atest/testdata/standard_libraries/builtin/run_keyword_with_errors.robot
```diff
@@ -159,3 +197,22 @@ For Loop With Syntax Error
...
+    ${list} =    Create list    1    2
+    Log    @{list}[2]
+
+Access To Dictionary Variable Nonexisting Key Syntax 1
+    ${dict} =    Create dictionary    a=1    b=2
+    Log    ${dict['c']}
+
+Access To Dictionary Variable Nonexisting Key Syntax 2
+    ${dict} =    Create dictionary    a=1    b=2
+    Log    &{dict}['c']
```
pekkaklarck: There should be no quotes when accessing dict variable times like this.

atest/testdata/standard_libraries/builtin/wait_until_keyword_succeeds.robot
```diff
@@ -115,3 +125,13 @@ Wait Until Inside User Keyword
 Timeouted UK with Wait Until KW
     [Timeout]    ${timeout}
     Wait Until Keyword Succeeds    100ms    10ms    Fail    Error in timeouted UK
+
+Get Length On Returned Variable Works On Last Attempt
+    ${myvar} =    return_value_after_retries    Hello
+    Get Length    ${myvar}
+    Fail    Test PASSES : variable is no more None
```
pekkaklarck: This keyword doesn't test what it is supposed to test. The variable `${myvar}` exists every time, so accessing it never causes a variable related failure.

You have two good ways to actually test that WUKS first fails when a variable doesn't exist and then succeeds when it does:

1. Just use a variable that doesn't exist. Verify that WUKS was executed multiple times on the test execution side.

2. Dynamically create a variable after first accessing it when it didn't yet exist. The keyword that is used with WUKS could be something like this:

    ```robotframework
    *** Keywords ***
    Example
        Log    ${dynamically set variable}    # doesn't exist when this keyword is executed first time
        Set Test Variable    ${dynamically set variable}    value
    ```

    Also in this case I'd like to verify on the execution side that the first round really failed.

src/robot/variables/replacer.py
```diff
@@ -142,12 +142,12 @@ def _get_list_variable_item(self, splitter):
         try:
             index = int(index)
         except ValueError:
-            raise DataError("List variable '%s' used with invalid index '%s'."
+            raise VariableError("List variable '%s' used with invalid index '%s'."
                             % (name, index))
```
pekkaklarck: Second line should be re-indented. Same issue with other VariableErrors in this file too.

pekkaklarck:
Code looks great but few more things to fix in tests.

atest/robot/standard_libraries/builtin/run_keyword_with_errors.robot
jcdevil: OK modified as proposed

atest/testdata/standard_libraries/builtin/run_keyword_with_errors.robot
jcdevil: Oups..Indeed.. done

atest/testdata/standard_libraries/builtin/run_keyword_with_errors.robot
jcdevil: OK. removing confusing ' character.

atest/testdata/running/test_template.robot
jcdevil: OK just kept this then :
```
Templates and syntax errors
    [Documentation]  FAIL Keyword 'BuiltIn.Should Be Equal' expected 2 to 4 arguments, got 5.
    syntax    error    makes    test    end
```

src/robot/variables/replacer.py
jcdevil: OK done

atest/testdata/standard_libraries/builtin/wait_until_keyword_succeeds.robot
pekkaklarck: I looked at the existing tests in this suite and was surprised that they didn't verify anything on the running side. I decided to add that for few tests because I think that's a good idea in general and because they can act as examples for you.

Tests were enhanced in 12b646aeec291565ce6b1378223722fd2356f0f6. I hope that doesn't cause merge conflicts.

jcdevil: OK Thx. I'll have a look then ;)
Notice that, at the moment, using the proposed keyword in WUKS :
```robot
*** Keywords ***
Example
    Log    ${dynamically set variable}    # doesn't exist when this keyword is executed first time
    Set Test Variable    ${dynamically set variable}    value
```
the line `Set Test Variable    ${dynamically set variable}    value` is never executed.

Still analysing this

pekkaklarck: Ooops, you are correct! Hmm, perhaps that `Set Test Variable` could be used in keyword teardown?

jcdevil: :+1:  Yep good point ! I'll try this. (never used this feature in my tests till now :) )

jcdevil:
PR updated with last remarks

atest/testdata/running/test_template.robot
```diff
@@ -169,10 +169,8 @@ Templates, timeouts, and for loops
     \    0.05 seconds
 
 Templates and syntax errors
-    [Documentation]  FAIL  Variable '\${nonexisting}' not found.
-    [Template]  Should be equal
-    ${nonexisting}      something
-    not compared  anymore
+    [Documentation]  FAIL Keyword 'BuiltIn.Should Be Equal' expected 2 to 4 arguments, got 5.
+    syntax    error    makes    test    end
```
pekkaklarck: Should keep the second iteration here to test that it is not executed. Unless I found something more serious here, I'll add it back myself after the merger.

src/robot/errors.py
```diff
@@ -53,6 +53,13 @@ class DataError(RobotError):
     this exception with care.
     """
 
+class VariableError(DataError):
+    """Used when variable syntax or evaluation is invalid.
```
pekkaklarck: Isn't this error only used when variable doesn't exist? I'll update docs myself.

PR #2059, opened 2015-07-07
Fix StringIO attribute error

shoonoise:
We should take care about cases when `StringIO` not implements needed `sys.stdout` methods.

### DIFF
--- a/src/robot/running/outputcapture.py
+++ b/src/robot/running/outputcapture.py
@@ -53,6 +53,15 @@ class OutputCapturer(object):
         return stdout, stderr
 
 
+class StreamIO(StringIO):
+
+    def __getattr__(self, name):
+        # Take care about missed attributes
+        if hasattr(sys.__stdout__, name):
+            return getattr(sys.__stdout__, name)
+        raise AttributeError(name)
+
+
 class PythonCapturer(object):
 
     def __init__(self, stdout=True):
@@ -62,7 +71,7 @@ class PythonCapturer(object):
         else:
             self._original = sys.stderr
             self._set_stream = self._set_stderr
-        self._stream = StringIO()
+        self._stream = StreamIO()
         self._set_stream(self._stream)
 
     def _set_stdout(self, stream):


robotci:
Admin, can CI be enabled?

pekkaklarck:
1. Could you explain the problem this PR fixes in a bit more detail?

2. I'm not sure is it a good idea to get attributes from `sys.__stdout__` if the stream doesn't have them. I see at least two problems:
  - The same object is used to wrap also stderr.
  - Someone may really want to avoid getting anything to `sys.__stdout__`.

shoonoise:
@pekkaklarck hi! 
The problem is that a library which I use in my tests and the library uses `sys.stdout.encoding` attribute to get current encoding. So as robot redefines `sys.stdout` and `StringIO` has no such method `AttributeError` raises. 

>The same object is used to wrap also stderr.
You totally right, I should care about it too. 

>Someone may really want to avoid getting anything to `sys.__stdout__`
Why someone uses `sys.stdout` but wants to get attributes from `StringIO`? Aren't `sys.stdout` and `sys.__stdout__` same?

pekkaklarck:
I see. I would say that your library is buggy if it assumes `sys.stdout` always has `encoding` attribute and especially if it assumes it always gets a correct value from it. By default `sys.stdout` and `sys.__stdout__` are the same object, but it is totally OK for Python programs to replace `sys.stdout` with a file like object such as `StringIO` that doesn't have `encoding`.

`sys.__stdout__` should never be altered and it should thus always contain the original stdout object. Notice that even then you don't always get what you expect from `encoding`:

    $ python -c "import sys; print sys.__stdout__.encoding"
    UTF-8
    $ python -c "import sys; print sys.__stdout__.encoding" | cat
    None

Detecting the correct encoding in Python is annoyingly hard. See here what kind of hack we do in Robot to get it:
https://github.com/robotframework/robotframework/blob/master/src/robot/utils/encodingsniffer.py

PR #2069, opened 2015-07-21
Add slack badge and fix broken badges

zephraph:


robotci:
Admin, can CI be enabled?

zephraph:
I'll squash everything and then it'll be ready for merge.

zephraph:
@pekkaklarck, after you get back from holiday take a look. =D

README.rst
```diff
@@ -32,18 +32,22 @@ for the standalone JAR distribution that is on `Maven central`_.
...
-   :alt: Latest version
-
-.. image:: https://pypip.in/download/robotframework/badge.png
-   :target: https://pypi.python.org/pypi/robotframework/
-   :alt: Number of downloads
-
-.. image:: https://pypip.in/license/robotframework/badge.png
+.. image:: https://img.shields.io/pypi/v/robotframework.svg
+   :target: https://pypi.python.org/pypi/robotframework
+   :alt: Version badge
```
pekkaklarck: Why was this text changed? Doesn't matter much because I don't think modern browsers show alt text as a tool tip anymore. It would be nice to have such a tool tip, but I'm not sure does reST support adding title for images.

pekkaklarck:
Adding a Slack badge is a good idea and thanks also for fixing the existing. It would be great if we could add titles to the badges to explain them, but I'm not sure is that possible with reST.

I'd like Slack and also robotframework-users be mentioned in the README also explicitly. I can add them myself but don't have time for that right now. I'll merge this PR at the same time.

pekkaklarck:
That Slack looks quite different to shields.io badges (rauchg/slackin#55). It is pretty convenient so I think I'll merge this PR anyway.

zephraph:
So, the black badge thing is intermitted. 

I see this on my screen

screen shot 2015-07-27 at 8 36 43 pm

Also, if you want me to change the text, I will. I can submit that in another PR if you'd like.

Issue #2115, opened 2015-08-23
Add 'id' to attrs of 'start_keyword' and 'end_keyword' on listeners v2


vkruoso:
The `start_keyword` and `end_keyword` does not provide the keyword `id` field in the attributes dictionary. As discussed in #1208 this may be added to the version 2 of the listener API.

pekkaklarck:
Yeah, it's a good idea to add ids also to keyword related listener methods. On a quick look PR #2116 looks so good that I don't see any problems getting this into 2.9.1. I'll take a bit closer look at the PR later.

pekkaklarck:
As I already commented PR #2116, keywords don't have enough context information at run time to get correct ids. Changing that is too big and risky task for RF 2.9.1. If you are interested, we can return to this in RF 3.0 that we plan to start developing already next month.

vkruoso:
That is OK. I'm definitely interested to work more on this. Just to be sure, in RF 3.0 there will be a new listener API version that takes real objects as arguments. Will this issue also affect the new API? At a high level view if seems that they complement each other.

pekkaklarck:
Yes, the plan is that RF 3.0 contains listener API v3 that gets real model and/or result objects as arguments (#1208). Adding ids to information passed to listener API v2 makes sense anyway, as long as the needed information is available somewhat easily. Having that information available is actually a requirement for API v3 if we want it contain `start/end_keyword` methods.

pekkaklarck:
The listener v2 API still doesn't add `id` to `start/end_keyword` methods, but this information is available via the listener v3 API that finally got `start/end_keyword` methods in RF 7.0 (#3296). The `id` attribute is dynamic and although generating it is fast, there's still small overhead. I believe having the `id` available only via the more powerful v3 API is enough.

PR #2116, opened 2015-08-23
Add 'id' to attributes on keyword methods to listeners

vkruoso:
This aims to fix issue #2115. I was assuming that the `id` field would be calculated when calling `start_keyword` and `end_keyword`, but that does not seems to be the case. I was not able to figure it out why at this point. Do you have any thoughts about it?

Another comment: it does not look like the id field is tested at the `end_suite`, `end_teste` and `end_keyword`. Do you think this should be added?


robotci:
Admin, can CI be enabled?

pekkaklarck:
As I already commented in #2115, this PR looks really good. @robotci: enable

I would assume suite/test ids are tested somewhere too. I can take a look do they exist somewhere when I have time or you can check that too. If they don't exist, we can add them to same place as the new kw id tests. If they do, perhaps we should add some kw id tests into that location.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/103/

pekkaklarck:
There seem to be some failures in listener tests on CI. My guess is that tests themselves just need to be updated. Listener tests are _very_ brittle and annoying to fix after smallest changes. We definitely should rewrite them to be easier to maintain.

vkruoso:
Tests aren't passing because the keyword `id`property is always being calculated as `k1`, as if there is no parent relation established when calling the listener methods. I did not had the time to investigate this more deeply.

It looks like there is tests related to the id only in the `start_*` methods. I need to confirm this as my walk through the code was brief.

pekkaklarck:
Yeah, getting id there correctly requires changes to code where listeners are called. I can take care of that later today or tomorrow.

pekkaklarck:
Unfortunately keywords don't have enough context information at run time to allow calculating ids and getting such context information there would require pretty big changes. That seems to be both too much work and too risky in RF 2.9.x minor releases but can be done in RF 3.0. Good news is that we are planning to start 3.0 development already next monthn.

vkruoso:
@pekkaklarck Its been a while now since this implementation. With the new listener interface do we get the keyword ``id`` attribute?

pekkaklarck:
Unfortunately not. The new listener API doesn't have `start_keyword` and `end_keyword` methods exactly for the same reason that not enough context information is available. Definitely should fix that but requires changes are big. In RF 3.1 I may need to look at kw execution againg and will keep this in mind.

vkruoso:
Will close this as the implementation was challenging, and maybe this is available in 3.x. I'm not as active using the tool right now as well. Thanks for the great work.

PR #2118, opened 2015-08-26
Telnet Library: allow control over connection timeout

vkruoso:
This still missing specific tests, but I should be able to create them soon. I've just fixed current tests to work with the new argument. Tests passed on my local machine.


robotci:
Admin, can CI be enabled?

Issue #2079, opened 2015-07-29
Telnet Library: Allowing control over connection timeout


vkruoso:
When using the `open_connection` method of the Telnet library we can provide a `timeout` parameter. That value is used to control the timeout when executing commands (basically read functions). That timeout is only applied after the connection has been established:

```python
    telnetlib.Telnet.__init__(self, host, int(port) if port else 23)
    self._set_timeout(timeout)
```

As you can see, there is no way to provide a timeout to the connection establishment in the ``__init__`` call. Looks like the telnetlib will use the operating system global timeout ``socket._GLOBAL_DEFAULT_TIMEOUT`` that seems to be 2 minutes in my system.

That support should be easy to implement and I can provide a PR if you see that this can be merged.

pekkaklarck:
Sounds good to me. Any objections by @jussimalinen? 

Remote library had same issue. You can take a look at how it was fixed.

jussimalinen:
Sounds good.

MoonSungHoon:
Sounds good. :+1:

vkruoso:
Does it sounds reasonable to use the same `timeout` parameter as the connection timeout too? Creating a new `connection_timeout` parameter does not sound good, and also may cause confusion about the purpose of the `timeout` parameter itself. What to you think about it?

pekkaklarck:
I'm worried that the current `timeout` is a bit short, only 3 seconds, to be used for connection. This is especially true if it would be used also when waiting replies in general, not only when connecting, similarly as Remote does. This change would also be backwards incompatible without an easy to keep using system's default timeout.

vkruoso:
That is true and if that was the approach chosen I would suggest increasing the default timeout. Anyway, how does a `connection_timeout` parameter sounds? Can you think on a better name? If you can't, I can go ahead on implementing, improve docs and add tests.

pekkaklarck:
Increasing the default timeout would be backwards incompatible and I'm afraid we needed to increase it quite a lot.

`connection_timeout` sounds good. I doubt there's a need for a separate keyword for setting it, but I think the library itself should accept the default value and `Open Connection` should allow overriding it.

jussimalinen:
Implemented in #2119

PR #2119, opened 2015-08-26
Telnet Library: allow control over connection timeout

vkruoso:
Aims to implement issue #2079. This still missing specific tests, but I should be able to create them soon. I've just fixed current tests to work with the new argument. Tests passed on my local machine.

### DIFF
--- a/src/robot/libraries/Telnet.py
+++ b/src/robot/libraries/Telnet.py
@@ -15,6 +15,7 @@
 from contextlib import contextmanager
 import inspect
 import re
+import socket
 import struct
 import telnetlib
 import time
@@ -105,6 +106,14 @@ class Telnet(object):
     Timeout defines how long is the maximum time to wait when reading
     output. It is used internally by `Read Until`, `Read Until Regexp`,
     `Read Until Prompt`, and `Login` keywords. The default value is 3 seconds.
+
+    == Connection Timeout ==
+
+    Connection Timeout defines how long is the maximum time to wait when
+    opening the telnet connection. It is used internally by `Open Connection`.
+    The default value is the system global default timeout.
+
+    New in Robot Framework 2.9.2.
 
     == Newline ==
 
@@ -295,7 +304,8 @@ class Telnet(object):
                  encoding='UTF-8', encoding_errors='ignore',
                  default_log_level='INFO', window_size=None,
                  environ_user=None, terminal_emulation=False,
-                 terminal_type=None, telnetlib_log_level='TRACE'):
+                 terminal_type=None, telnetlib_log_level='TRACE',
+                 connection_timeout=None):
         """Telnet library can be imported with optional configuration parameters.
 
         Configuration parameters are used as default values when new
@@ -320,6 +330,7 @@ class Telnet(object):
         | Library     | Telnet    | telnetlib_log_level=NONE |                      |                     | # disable logging messages from the underlying telnetlib |
         """
         self._timeout = timeout or 3.0
+        self._set_connection_timeout(connection_timeout)
         self._newline = newline or 'CRLF'
         self._prompt = (prompt, prompt_is_regexp)
         self._encoding = encoding
@@ -373,7 +384,8 @@ class Telnet(object):
                         encoding=None, encoding_errors=None,
                         default_log_level=None, window_size=None,
                         environ_user=None, terminal_emulation=None,
-                        terminal_type=None, telnetlib_log_level=None):
+                        terminal_type=None, telnetlib_log_level=None,
+                        connection_timeout=None):
         """Opens a new Telnet connection to the given host and port.
 
         The ``timeout``, ``newline``, ``prompt``, ``prompt_is_regexp``,
@@ -391,6 +403,9 @@ class Telnet(object):
         `Close All Connections` keyword.
         """
         timeout = timeout or self._timeout
+        connection_timeout = (timestr_to_secs(connection_timeout)
+                              if connection_timeout
+                              else self._connection_timeout)
         newline = newline or self._newline
         encoding = encoding or self._encoding
         encoding_errors = encoding_errors or self._encoding_errors
@@ -413,7 +428,8 @@ class Telnet(object):
                                           environ_user,
                                           is_truthy(terminal_emulation),
                                           terminal_type,
-                                          telnetlib_log_level)
+                                          telnetlib_log_level,
+                                          connection_timeout)
         return self._cache.register(self._conn, alias)
 
     def _parse_window_size(self, window_size):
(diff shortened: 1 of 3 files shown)


robotci:
Admin, can CI be enabled?

pekkaklarck:
1. New arguments to keywords and library `__init__` must be last (and have a default value) to avoid backwards compatibility issues. That rule can be broken in major releases if there are strong reasons, but even then it would require a deprecation period.

2. @jussimalinen, can you take care of reviewing and testing this PR when it gets more ready?

3. Updates to Telnet documentation may affect Libdoc tests that use Telnet as their test data. It would be great if you @vkruoso could run those tests and verify did something broke. I can help fixing tests if that turns out to be lot of work. It's not really related to Telnet anyway.

4. We plan to release RF 2.9.1 tomorrow. I doubt there's time to get this in, but we'll probably create 2.9.2 still before 3.0.

vkruoso:
I've updated this PR to include tests and fix the issues pointed by you. The libdoc test suites were fixed to consider the new parameter. I think this is complete, but let me know if something is missing or we need more improvements.

jussimalinen:
Looks very good. A couple of comments:

Adding a default timeout to 2 minutes is a backwards incompatible change. We could break stuff from existing tests which rely on existing timeout behaviour and that is bad. So this should be changed so that by default is no timeout given to telnetlib.Telnet._init__. (So that underlying socket timeout is used in the end.)

Also, please add yourself to the AUTHORS.txt file!

vkruoso:
You are right! I've updated the PR. Code in the init method looks ugly to me, though. Can you think on a better way to do it?

I'm already in the AUTHORS file! :)

src/robot/libraries/Telnet.py
```diff
@@ -106,6 +107,12 @@ class Telnet(object):
     output. It is used internally by `Read Until`, `Read Until Regexp`,
     `Read Until Prompt`, and `Login` keywords. The default value is 3 seconds.
 
+    == Connection Timeout ==
+
+    Connection Timeout defines how long is the maximum time to wait when
+    opening the telnet connection. It is used internally by `Open Connection`.
+    The default value is the system global default timeout.
```
jussimalinen: Note that this is new in RF 2.9.2.

src/robot/libraries/Telnet.py
```diff
@@ -491,8 +531,17 @@ def __init__(self, host=None, port=23, timeout=3.0, newline='CRLF',
...
+        if connection_timeout is None:
+            telnetlib.Telnet.__init__(self, host, int(port) if port else 23)
+        else:
+            try:
+                telnetlib.Telnet.__init__(self, host,
+                                          int(port) if port else 23,
+                                          connection_timeout)
+            except socket.timeout:
+                raise AssertionError("Connection timeout after %s."
+                                     % secs_to_timestr(connection_timeout))
```
jussimalinen: Some users do use the TelnetLibrary programmatically. Raising different exception types (socket.timeout, AssertionError) based on what timeout was used is a bit unclean. Also on my linux machine I get socket.error when the system refuses the connection and this fails the test "Set Connection Timeout Keyword".

If socket.timeout is caught then we should catch also it in the case where the timeout occurs with default timeout or then alternatively, we should rise the same exeption (socket.timeout). I discussed this with @pekkaklarck and we were thinking that maybe easiest would be not to catch any exceptions here and just let the original socket exception to go through.

I assume that in your environment that test passes, so presumably you get socket.timeout instead of socket.error. Does you network have this machine "172.16.254.254" or are you running on some other platform than I do? (Linux mint, python 2.7.3)

jussimalinen: We managed to get socket.timeout by binding a server to localhost port, but not starting to listen:
python -c "import socket; ss = socket.socket(socket.AF_INET, socket.SOCK_STREAM); ss.bind(('127.0.0.1',2424)); raw_input()"

Then opening a connection to 127.0.0.1, 2424 results in socket.timeout on our machines.

src/robot/libraries/Telnet.py
```diff
@@ -430,6 +444,32 @@ def _get_connection(self, *args):
...
+        Example:
+        | ${old} =          | `Set Connection Timeout` | 2 minutes 30 seconds |
+        | `Open Connection` | ${host}
+
+        See `Configuration` section for more information about global and
+        connection specific configuration.
+        """
+        old = self._connection_timeout
+        self._set_connection_timeout(connection_timeout)
+        return secs_to_timestr(old) if old is not None else None
```
jussimalinen: This keyword seems a bit unnecessary. We can set default timeout in library init and also for individual "Open connection" calls, and I would assume that should be enough?

jussimalinen:
Thanks for the pull request! We went through it with @pekkaklarck and I added a few line comments.

We are not getting the socket.timeout with the address used in the tests. The trick of starting a python process that binds to a localhost port could maybe used to test the timeout functionality? It hopefully works consistently.

src/robot/libraries/Telnet.py
vkruoso: Makes sense to let the socket exception go through. In my environment the test passes as trying to connect to that address causes a timeout because the address does not exists. I'm running in a Ubuntu 14.04 VM (must confirm the python version). I had issues getting the timeout when passing a hostname to the init function, as the DNS lookup does not respect the timeout and issue another exception when the name is not found.

Anyway, I'll try to remove this and try to use you approach for testing.

src/robot/libraries/Telnet.py
vkruoso: I think it is enough. I added this to maintain consistency with the "normal" timeout keywords. I'll remove it.

vkruoso:
Thanks for the feedback @jussimalinen and @pekkaklarck!  I'll try to provide an update on this until the end of this weekend!

vkruoso:
I'm taking a look at this and have some observations. I've tried your approach with no success in my local machine: instead of a socket.timeout exception I'm getting socket.error (connection refused) error. This seems reasonable as there is no call to listen on the socket so the server should deny any connection attempt. It seems odd that the behavior is the opposite in your machine. I'm running python 2.7.6, maybe that is the reason?

The initial idea is really not good because when trying to connect to a local address, packets will be sent to the gateway and behavior should differ depending on router configurations.

pekkaklarck:
I thought both Jussi and I were able to get `socket.timeout` but now that I test it I get `socket.error`. If we cannot figure out how to test this reliably, it is fine to leave it untested. The change itself is pretty trivial anyway.

vkruoso:
I've updated the PR removing the trap server. Will try to think about a way to test it reliably, but in any case it sounds reasonable if we can not figure it out. After the next release this will for sure be tested in a daily basis.

jussimalinen:
This looks good as it is. Thanks! I will add this feature to you contributions in AUTHORS

vkruoso:
Thanks for the merge. Sorry I did not had the time to investigate further. Looking forward to use it in the next release tough.

PR #2122, opened 2015-08-29
Fix typo in ListenerInterface.rst

mkorpela:
typo fix


robotci:
Admin, can CI be enabled?

mkorpela:
Sorry robotci I speek only to humans.

pekkaklarck:
Thanks, good catch! Think it should read "as a list", though.

pekkaklarck:
Fixed this in 5597602 with `as a list`.

Issue #519, opened 2014-06-29
Given/When/Then should support other languages than English


spooning:
> _Originally submitted to [Google Code](http://code.google.com/p/robotframework/issues/detail?id=519) by Andreas.EbbertKarroum on 30 Mar 2010_

When <a class="closed_ref" href="https://code.google.com/p/robotframework/issues/detail?id=500" title="Possibility to specify 'Test Template'  for all tests to ease data-driven testing"> issue 500 </a> is implemented, it would be nice to supply with the
[Template] keyword not only the test case that should be used, but also, in
what language that template is written. 

When I say

[Template]  User Login  en

RF will ignode prefixes "Given", "When", "Then", and "And" from keywords.
When I specify another language, like "de", then "Gegeben", "Wenn", "Dann",
and "Und". 

This has already been briefly discussed in the user group:
<a href="http://groups.google.com/group/robotframework-users/browse_thread/thread/488c5bc04ab2315d/7f81d5b118e7aa7a?lnk=gst&amp;q=german#7f81d5b118e7aa7a" rel="nofollow">http://groups.google.com/group/robotframework-users/browse_thread/thread/488c5bc04ab2315d/7f81d5b118e7aa7a?lnk=gst&amp;q=german#7f81d5b118e7aa7a</a>

keskiju:
This would be a really useful feature. BDD tests/scenarios are basically software specifications that should be written in a domain specific language that all stakeholders understand very well. English is not always the best choice for that. Many other BDD tools already support multiple languages (cucumber, behat, etc). Hopefully we get language support also for robot framework soon.

pekkaklarck:
Yes, this would be a nice feature. Getting it in just requires someone needing it badly enough to either provide a pull request or sponsoring the developement.

felipecaputo:
I believe I can help with it. 

Anyone defined how it should work? maybe like [this](https://github.com/robotframework/robotframework/issues/519#issuecomment-47478667) or maybe support all of the Gherkin translations defined in any configuration

pekkaklarck:
@felipecaputo Awesome! It's been some time since I've been thinking about this, but the comments you referred sounded pretty good still.

The most convenient solution would be able to use some command line option to specify the language to use. This would obviously require including various languages in Robot's code base, but I doubt that's a big issue. Most likely we could just use prefixes defined by Cucumber with attribution.

Open design questions:

[...]

PR #2125, opened 2015-09-01
Extract BDD-style ignored prefixes to a constant

neithere:
As a Developer who uses Robot Framework,
I want the BDD prefixes as a constant
In order to easily monkey-patch it and introduce my own prefixes.

### DIFF
--- a/src/robot/running/namespace.py
+++ b/src/robot/running/namespace.py
@@ -30,6 +30,8 @@ from .runkwregister import RUN_KW_REGISTER
 
 IMPORTER = Importer()
 
+BDD_STYLE_IGNORED_PREFIXES = ('given ', 'when ', 'then ', 'and ', 'but ') 
+
 
 class Namespace(object):
     _default_libraries = ('BuiltIn', 'Reserved', 'Easter')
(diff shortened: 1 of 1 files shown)


robotci:
Admin, can CI be enabled?

pekkaklarck:
I don't see any harm accepting this PR, but there's a very serious risk that the related code will be at least partly rewritten in RF 3.0. That obviously wouldn't be good for the possible monkeypatches or other extensions.

Would you be interested in making BDD prefixes officially configurable? We already have issue #519 about that and it would be a great enhancement in RF 3.0.

neithere:
Monkey-patching is generally unreliable, so this PR was intended to enable a moderate hack in the absence of a proper solution. Currently we have to monkey-patch the whole `Namespace._get_bdd_style_handler()` method to enable things like "When I click button" directly via "Click button" keyword. If the next release brings configurable prefixes, I would be happy to remove the hack from our project-specific library.

Concerning #519: it sounds very promising but as I understand it still awaits design decisions, right? I don't see how I would estimate total effort at this point.

pekkaklarck:
If you are interested, I'd be happy to discuss design related to #519. We can use issue comments or perhaps chat on IRC or Slack.

bhirsz:
@pekkaklarck I believe we can close this PR since 5.1 will allow using BDD in other languages (and it's kind of connected to this PR).

pekkaklarck:
True, this is covered by #519 that's already closed.

PR #2126, opened 2015-09-02
Add Windows Presentation Foundation (WPF) support in dialogs_ipy.py

### LINKED ISSUE #1235: IronPython support for `Dialogs` library
> *Originally submitted to [Google Code](http://code.google.com/p/robotframework/issues/detail?id=1235) by @pekkaklarck on 18 Sep 2012*

<a href="https://code.google.com/p/robotframework/wiki/DialogsLibrary" title="A standard library providing dialogs for interacting with users.">DialogsLibrary</a> currently only works with Python and Jython. Implementing IronPython support using Winforms ought to be relatively simple.

moto-timo:
Resolves #1235

Uses more modern Windows Presentation Foundation (WPF) rather than the older WinForms.

This allows for:
* better flow of long, wrapped text
* better automatic sizing of controls


robotci:
Admin, can CI be enabled?

src/robot/libraries/Dialogs.py
```diff
@@ -16,15 +16,14 @@
...
+IronPython or Jython but they provide the same functionality.
 
 Long lines in the provided messages are wrapped automatically since
 Robot Framework 2.8. If you want to wrap lines manually, you can add
 newlines using the ``\\n`` character sequence.
 
-The library has following two limitations:
-- It is not compatible with IronPython.
+The library has the following limitation:
 - It cannot be used with timeouts on Python.
```
pekkaklarck: This could be rephrased. Having a list with one item is a bit odd. Also, a note that IronPython support is new in 2.9.2 is needed somewhere here.

src/robot/libraries/dialogs_ipy.py
```diff
@@ -12,32 +12,183 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import wpf
+from System.Windows import Application, GridLength, SizeToContent, TextWrapping, \
+                           Thickness, Window, WindowStartupLocation, WindowState
+from System.Windows.Controls import Button, ColumnDefinition, Grid, Label, ListBox, \
+                                    PasswordBox, RowDefinition, TextBlock, TextBox
```
pekkaklarck: We've used parenthesis to cut long import lines like

```python
from System.Windows import (Application, GridLength, SizeToContent, TextWrapping,
                            Thickness, Window, WindowStartupLocation, WindowState)
```

but that's definitely not a big issue.

src/robot/libraries/dialogs_ipy.py
```diff
@@ -12,32 +12,183 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import wpf
```
pekkaklarck: This module doesn't seem to be used. Is the import required to initialize something on .NET or could it be removed? If it is required, adding a comment briefly explaining it would be nice.

src/robot/libraries/dialogs_ipy.py
```diff
@@ -12,32 +12,183 @@
...
+        self._close()
+
+    def _get_right_button_value(self):
+        return None
+
+    def show(self):
+        if Application.Current == None:
+            parent = _mainWindow()
+        else:
+            parent = Application.MainWindow
```
pekkaklarck: - Should use `is None`, not `== None`.
- `parent` doesn't seem to be used for anything. If there even need to access it?

src/robot/libraries/dialogs_ipy.py
```diff
@@ -12,32 +12,183 @@
...
 
+import wpf
+from System.Windows import Application, GridLength, SizeToContent, TextWrapping, \
+                           Thickness, Window, WindowStartupLocation, WindowState
+from System.Windows.Controls import Button, ColumnDefinition, Grid, Label, ListBox, \
+                                    PasswordBox, RowDefinition, TextBlock, TextBox
 
-class _AbstractWinformsDialog:
 
+class _mainWindow(Window):
```
pekkaklarck: Should be named `_MainWindow` or just `MainWindow`. But is this really needed? This seems to be used only once and that usage could perhaps be inlined.

pekkaklarck:
Both code and the created dialogs look great! I added some line notes, but they are pretty nitpicky.

I also noticed a minor missing feature. When an input dialog is opened, possible default value should be selected to make it easier to overwrite it. This isn't that important feature, and it actually seems to be broken on Windows with Python, and we can easily live without it.

src/robot/libraries/dialogs_ipy.py
moto-timo: We can use any naming you prefer.

This is critical, because without the Application.MainWindow defined the
Dialog itself becomes the MainWIndow and we cannot run multiple
applications in the same AppDomain. In other words, only the first 'atest'
will run, and the rest will all fail with an error about the AppDomain.

src/robot/libraries/dialogs_ipy.py
moto-timo: dialogs_jy.py used \, dialogs.py used () and pylint complained about the ()...

I actually think the parentheses look better.

moto-timo:
I was about to figure out the code for selecting the default value on input dialogs and then checked the behavior of the Tk version and saw it didn't do that :)

I will figure out how to select the default value, because that is really the expected behavior :)

src/robot/libraries/dialogs_ipy.py
moto-timo: ```parent``` is just a placeholder for the ```Application.MainWindow```. On the first 'atest', the application is launched with a minimized ```MainWindow```. On subsequent tests, the ```Application``` already exists and it's existing ```MainWindow``` is used... this was how I got around the AppDomain issue mentioned in the comments on ```class _mainWindow(Window):```

src/robot/libraries/dialogs_ipy.py
moto-timo: Good catch. It's not actually used. It only provides [loading of XAML files] (https://github.com/IronLanguages/main/blob/ipy-2.7-maint/Languages/IronPython/IronPython.Wpf/wpf.cs) and my very first round of code used a XAML file rather than doing the layout programmatically as it is now.

We can drop it.

moto-timo: Turns out that ```import wpf``` is pulling in some required assemblies behind the scenes. It can be done in other ways, but they are all uglier and much more verbose. I'll add some kind of comment.

pekkaklarck:
Thanks for the great PR! I merged it and then did few minor cleanups in 2d341e531973901a610d1cd93954c063024d3cfb.

Issue #2093, opened 2015-08-07
Supress docutils errors/warnings with reST format


vkruoso:
When using reST format for writing tests. I'm using the same source file to generate Sphinx documentation. This allows me to have one file for both purposes. The problem is that Sphinx define some non-standard directives that docutils complains about. 

Example: 

```
tests/resources/settings.rst:2: (ERROR/3) Unknown directive type "highlight". 

.. highlight:: robotframework 

bfd_test.rst:15: (ERROR/3) Unknown interpreted text role "ref". 
bfd_test.rst:276: (ERROR/3) Unknown directive type "automodule". 

.. automodule:: bfd_test 
    :members: 
    :undoc-members: 
    :show-inheritance: 
```

As the test grows the amount of errors we get on the output increases.

This has been discussed in the robotframework-users list, and I've also talked about this at the sphinx-dev list (where I'm waiting for more answers). Here are the links for reference:
https://groups.google.com/forum/?hl=pt-BR#!topic/robotframework-users/0V-q3xJgcZI
https://groups.google.com/forum/?hl=pt-BR#!topic/sphinx-dev/97e5xDcfsl8

I'm proposing that we create a simple script that finds all directives created by Sphinx from its source, and integrate them at the robot by adding them to docutils when reading source files. The script would generate code that would them be committed as part of robot's source. Would that be a reasonable approach? If so, I can try to provide a PR for it.

Do we need to worry about other reST files directives defined by other tools? Sphinx seems popular enough in the python world, but I'm not really sure.

pekkaklarck:
This is definitely something we should fix somehow. I'm not sure is having a list of ignored directives and roles a good idea, though. We needed to get them from various Sphinx and docutils plugins too, and maintaining the list would be an annoying task.

I'd prefer disabling warnings from all non-existing directives/roles. Did you already investigate is that possible somehow? Should probably ask that on docutils forums. If that's not possible, disabling all warnings might be the next best solution. It could be argued that you see warnings when generating the docs and thus they aren't important when running tests, but it might still be safer to make that configurable.

vkruoso:
I've sent a message at the docutils-users list. Taking a quick look at docutils code I tend to say that it does not provide this flexibility about error messages.

Disabling all warnings is not enough as unknown directives/roles are considered errors. An option to disable all errors sound strong. Looks like there is a `severe` message type as well, tough.

One could argue that if you do not use that file to compile docs you should not use it to run tests.

PR #2152, opened 2015-09-19
reST format: changed docutils report level to severe

vkruoso:
This is relative to issue #2093. The argument is that errors outside robot code blocks should be ignored when running robot. Still need to find a way to test this.

### DIFF
--- a/src/robot/parsing/restreader.py
+++ b/src/robot/parsing/restreader.py
@@ -27,7 +27,10 @@ def RestReader():
         def read(self, rstfile, rawdata):
             doctree = publish_doctree(
                 rstfile.read(), source_path=rstfile.name,
-                settings_overrides={'input_encoding': 'UTF-8'})
+                settings_overrides={
+                    'input_encoding': 'UTF-8',
+                    'report_level': 4  # only severe messages
+                })
             store = RobotDataStorage(doctree)
             if store.has_data():
                 return self._read_text(store.get_data(), rawdata)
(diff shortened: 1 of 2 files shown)


robotci:
Admin, can CI be enabled?

src/robot/parsing/restreader.py
```diff
@@ -27,7 +27,10 @@ class RestReader(object):
         def read(self, rstfile, rawdata):
             doctree = publish_doctree(
                 rstfile.read(), source_path=rstfile.name,
-                settings_overrides={'input_encoding': 'UTF-8'})
+                settings_overrides={
+                    'input_encoding': 'UTF-8',
+                    'report_level': 4  # only severe messages
```
pekkaklarck: Looking at docutils sources, it seems string `SEVERE` could be used instead of integer `4`. It would be easier to understand and remove the need for the comment.

pekkaklarck:
Sorry, for missing this PR earlier. Ignoring other than severe errors sounds like a decent behavior for me. I don't think we need to make that configurable. Some further comments:

- It ought to be possible to test this by validating that there are no errors in STDERR after `sample.rst` is executed.
- As I commented in a line note, report level should probably be set also when specifying test cases in reST tables that are converted to HTML internally.
- Should also add a note to the User Guide that lower level errors are ignored when running tests to avoid noise about non-standard directives. They can be seen when processing docs otherwise if needed.

How urgent is this feature for you and how fast you think you could finish it? We are planning to release RF 2.9.2 very soon, preferably already on Thursday. Because it was my mistake to miss this PR, we can delay it a little but not further than than early next week. The planned next release after 2.9.2 is 3.0 in December.

jussimalinen:
We decided to make the release Tomorrow Friday. We will do the tests and userguide changes needed to finish this. Thanks!

vkruoso:
Thanks for the review! I would not have time to finish this until early next week, so it is good to hear that it is merged and you are doing improvements! Looking forward to 2.9.2.

Issue #2153, opened 2015-09-21
`Copy/Move File` fails if source and destination are same


pekkaklarck:
For example this example fails with `IOError: No such file or directory`:

```robotframework
*** Test Cases ***
Copy File Bug
    Create File    ${CURDIR}/example.txt
    Copy File    ${CURDIR}/example.txt    ${CURDIR}/example.txt
```

Same problem occurs also with `Move File` and if the specified destination is a directory (e.g. just `${CURDIR}` in the above example). To make matters worse, a temporary directory is left into the directory where the file exists.

We should handle this that source and destination being same is not a failure at all. Preferably there should be a special log message telling that file was not copied/moved because destination was same.

markwinspear:
Copy should be allowed to have the same source and destination folder. If the filename is also the same then either overwrite or throw a sensible error

pekkaklarck:
This will be implemented so that if source and destination point to the same file, there will be no copying or moving whatsoever and the keyword will pass.

PR #2159, opened 2015-09-24
Ommiting copying and moving when source and destination are the same …

stasiek:
Solution for issue #2153 
Now before copying/moving we're checking if the source and directory are the same file.


robotci:
Admin, can CI be enabled?

Issue #1882, opened 2014-12-16
Add BackgroundLogger to robot.api.logger


jussimalinen:
The robot.api.logger should provide a helper class to log from background threads. https://github.com/robotframework/robotbackgroundlogger provides such functionality and it should be included in our api.

jussimalinen:
Reverted the BackgroundLogger from 2.8.7 (02fcc63). We were not sure what the API should be to support use cases in #1791 and #1887 so we decided remove it for now because we dont want to commit to an API that might change.

pekkaklarck:
We decided to descope this from RF 2.9 because there still are open question related to how this should actually work. We also want to get RF 2.9 is released ASAP and don't thus have time to investigate this at the moment.

pekkaklarck:
Unfortunately no time for this in 3.0.

PR #2182, opened 2015-10-22
Adding support to logger.py for logging from child processes and threads

blooblah123:
I'm raising this pull-request in relation to issue [#2014 output.xml child process log meshed with parent log] (https://github.com/robotframework/robotframework/issues/2014)

I have a number of tests that run in child processes. Robot's default behaviour is to ignore any log messages generated by these processes (which is not useful). I did try using the backgroundlogger which worked OK for threaded processes but not for multiprocessor or poolworker processes. I also started seeing corruption of output.xml.

The changes I'm proposing here add support for logging from child processes to the public logging API library logger.py. Logging is supported for threaded, multiprocessor and poolworker children.

To maintain backwards compatibility with existing behaviour, the logging of messages from child processes is **disabled** by default, and must be explicitly enabled by calling`logger.enable_background_logging()`. Conversely, logging can be disabled by calling`logger.disable_background_logging()`.

When enabled,`librarylogger.write()`stores each message created by a child process in a shared dictionary (keyed by process name).
* Stored messages are only written to the log by calling`logger.write_background_messages()`
* This can only be done from the main thread/process (doing this from elsewhere raises an exception).
* The function`logger.write_background_messages()`can be called to write all messages (the default behaviour) or write messages for just a single process.
* As messages are written to the log, there are removed from the dictionary.

The function`logger.clear_background_messages()`can be called to clear stored background messages without writing them to the log.


robotci:
Admin, can CI be enabled?

blooblah123:
CI can be enabled.
robotci: enable

pekkaklarck:
Interesting. As you probably know, we've been thinking about adding background logger to the core framework for some time (#1882) but haven't had time to design it properly. Unfortunately both @jussimalinen and I are busy with training courses and holidays (well, perhaps not that busy with the latter...) this and next week, but we'd definitely like to see this in RF 3.0. Is this PR close enough to #1882 that we don't need a separate issue?

@robotci: enable

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/104/

blooblah123:
Good to hear that you think this is worth including in RF3.0 :smile:

As far as my understanding of https://github.com/robotframework/robotframework/issues/1882 goes, I believe this PR pretty much meets those requirements so shouldn't need a separate issue raising. I'm away on vacation as well next week, so maybe we could chat more about this when we're all back?

pekkaklarck:
Sounds great!

blooblah123:
Hi...Just wondering what the current state of this pull-request is what the next step is? Need any further info from me?

pekkaklarck:
Umfortunately it looks like we won't have time to include backgroundlogger into the core in RF 3.0. There's still a small change we can take a look at it, but it it would require new listener API to be ready tomorrow. This is very unfortunate, but we have made a promise to our sponsor that RF 3.0 is released still this year.

Issue #2188, opened 2015-11-02
`${TEST_MESSAGE}` is not updated by `Set Test Message` keyword


osjc:
When you use BuiltIn.SetTestMessage to set or amend the message in the test, the ${TEST_MESSAGE} variable still shows the original test message. Additionally, the BuiltIn.SetTestMessage is confusing because in one sentence it states that "the original test message is available as a built-in variable ${TEST_MESSAGE}" but then proceed to show an example where the test message is altered and then ${TEST_MESSAGE} is checked to see that the alteration is in place.

Occurs in 2.9 and also found in the latest tip version.

I created a patch to fix this (the cause looks like omission in the code).

pekkaklarck:
It really looks like we've just forgot to set the variable. Thank's for the report and a PR. Tests in the PR require still some work, but I comment the PR separately.

pekkaklarck:
Fixed by PR #2191. Thanks @osjc both for reporting the issue and for the PR.

PR #2189, opened 2015-11-02
Fixed TEST_MESSAGE not reacting to SetTestMessage

osjc:
Fixes issue https://github.com/robotframework/robotframework/issues/2188

### DIFF
--- a/atest/robot/standard_libraries/builtin/set_test_message_and_test_message_variable.robot
+++ b/atest/robot/standard_libraries/builtin/set_test_message_and_test_message_variable.robot
@@ -1,0 +1,17 @@
+*** Settings ***
+Documentation     Unit test suite to Utils.Report_Failure_Due_To_Bug keyword.
+
+*** Test cases ***
+Message_Contains_Reported_Bug
+    Pass_Execution    Initial Test Message
+    [Teardown]    Check_Test_Message_Behavior
+
+*** Keywords ***
+Check_Test_Message_Behavior
+    Should_Be_Equal    ${TEST_MESSAGE}    Initial Test Message
+    BuiltIn.Set_Test_Message    First    True
+    Should_Be_Equal    ${TEST_MESSAGE}    Initial Test Message First
+    BuiltIn.Set_Test_Message    Other
+    Should_Be_Equal    ${TEST_MESSAGE}    Other
+    BuiltIn.Set_Test_Message    Second    True
+    Should_Be_Equal    ${TEST_MESSAGE}    Other Second
--- a/src/robot/libraries/BuiltIn.py
+++ b/src/robot/libraries/BuiltIn.py
@@ -2789,8 +2789,10 @@ class _Misc(_BuiltInBase):
 
         In test teardown this keyword can alter the possible failure message,
         but otherwise failures override messages set by this keyword. Notice
-        that in teardown the initial message is available as a built-in variable
-        ``${TEST MESSAGE}``.
+        that in teardown the test message is available as a built-in
+        variable ``${TEST MESSAGE}`` which will be equal to the initial test
+        message at the start of the test teardown (before the first use of
+        this keyword).
 
         It is possible to use HTML format in the message by starting the message
         with ``*HTML*``.
@@ -2812,6 +2814,7 @@ class _Misc(_BuiltInBase):
                                "suite setup or teardown.")
         test.message = self._get_possibly_appended_value(test.message, message,
                                                          append)
+        self._variables.set_suite("${TEST_MESSAGE}", test.message)
         message, level = self._get_logged_test_message_and_level(test.message)
         self.log('Set test message to:\n%s' % message, level)
 


robotci:
Admin, can CI be enabled?

osjc:
I added a test into CI checking against the bug but the test might need some refactoring as it is separate from the rest of the SetTestMessage testing. I studied the SetTestMessage test but realized it is not going to be easy to understand it and I was not sure if it is going to be a good idea to add my test there so I created a separate test suite.

osjc:
The test suite was designed to use just a plain vanilla Robot infrastructure. No fancy Python library enhancements or something.

osjc:
I also fixed the documentation to make it self consistent. Namely the sentence mentioned in the issue was fixed to emphasize that the test message is available in that variable only at the beginning of the test teardown, before the first use of BuiltIn.SetTestMessage.

src/robot/libraries/BuiltIn.py
```diff
@@ -2812,6 +2814,7 @@ def set_test_message(self, message, append=False):
                                "suite setup or teardown.")
         test.message = self._get_possibly_appended_value(test.message, message,
                                                          append)
+        self._variables.set_suite("${TEST_MESSAGE}", test.message)
```
pekkaklarck: Should use `variables.set_test`, not `set_suite`.

atest/robot/standard_libraries/builtin/set_test_message_and_test_message_variable.robot
```diff
@@ -0,0 +1,17 @@
+*** Settings ***
+Documentation     Unit test suite to Utils.Report_Failure_Due_To_Bug keyword.
```
pekkaklarck: ??

atest/robot/standard_libraries/builtin/set_test_message_and_test_message_variable.robot
```diff
@@ -0,0 +1,17 @@
...
+
+*** Test cases ***
+Message_Contains_Reported_Bug
+    Pass_Execution    Initial Test Message
+    [Teardown]    Check_Test_Message_Behavior
+
+*** Keywords ***
+Check_Test_Message_Behavior
+    Should_Be_Equal    ${TEST_MESSAGE}    Initial Test Message
+    BuiltIn.Set_Test_Message    First    True
```
pekkaklarck: No need to use `_` in test and keyword names nor `BuiltIn.` prefix when calling `Set Test Message`.

src/robot/libraries/BuiltIn.py
```diff
@@ -2789,8 +2789,10 @@ def set_test_message(self, message, append=False):
 
         In test teardown this keyword can alter the possible failure message,
         but otherwise failures override messages set by this keyword. Notice
-        that in teardown the initial message is available as a built-in variable
-        ``${TEST MESSAGE}``.
+        that in teardown the test message is available as a built-in
+        variable ``${TEST MESSAGE}`` which will be equal to the initial test
+        message at the start of the test teardown (before the first use of
+        this keyword).
```
pekkaklarck: I agree the original explanation is not good but the proposed one is also pretty confusing. Would it be enough to just drop `initial` from the original doc?

pekkaklarck:
I already added some line notes, but here are comments about the overall testing approach:

1. You should add tests into the existing `set_test_message.robot` suite.

2. Robot's tests have two sides. Tests under `atest/robot` run tests defined under `atest/testdata`. Test you creates belongs to the latter side, but you also need to verify it on the former side. For more details see README at https://github.com/robotframework/robotframework/tree/master/atest

pekkaklarck:
Why did you close this PR? You could have just committed changes to your branch and the PR would have been automatically updated.

osjc:
Because this is the first time I tried to make a pull request and I still don't understand the quirks of this interface. I tried to replace the commit with a better one (or, as you would say, squash the changes I was making with the original commit) and it just closed the pull request.

osjc:
Ok. You can continue watching this work at https://github.com/robotframework/robotframework/pull/2191.

pekkaklarck:
It is often a good idea to squash commits when the PR is ready to be merged, but when you are still working with it you can just push new changes to your branch. Opening a new one is fine too. Unfortunately I won't have time to look at #2191 today and probably not tomorrow either.

osjc:
No problem. This mess was caused by myself working with Gerrit which handles squashing commits completely differently. And also I suffer from a medical condition that causes me not being able to productively use point-and-click interfaces so I got pretty lost when this happened. And now I think that GitHub shall at least ask me for permission to close the pull request if it sees the branch with the commit disappearing because the disappearing commit usually means a squashed version is on its way.

Now I have to figure out how to squash commits together when you say "ok to be merged". Maybe forced push?

osjc:
Answer to the question was found at the end of CONTRIBUTING.rst

PR #2191, opened 2015-11-02
Fixed TEST_MESSAGE not reacting to SetTestMessage

osjc:
This is a resurrection of https://github.com/robotframework/robotframework/pull/2189 which was killed by GitHub's reaction to an attempt to amend the commit based on the comments made in the original pull request.

### DIFF
--- a/atest/robot/standard_libraries/builtin/set_test_message.robot
+++ b/atest/robot/standard_libraries/builtin/set_test_message.robot
@@ -57,6 +57,9 @@ Set Message In Setup
 Check Message From Previous Test
     Check Test Case    ${TEST NAME}
 
+Test Message Variable Reacts On Set Test Message
+    Check Test Case    ${TEST NAME}
+
 Not Allowed In Suite Setup or Teardown
     ${error}=    Catenate    SEPARATOR=\n
     ...    Suite setup failed:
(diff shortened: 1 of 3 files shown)


robotci:
Admin, can CI be enabled?

osjc:
Yes.

osjc:
Line 2817 was fixed to use the correct variable setting method. I also found out that the variable should be set only when the test is in the teardown phrase because that is the time when it gets created. The test was merged into the CI properly (now when I finally managed to decipher the message from atest/README.rst with the help of pekkaklarck, maybe I shall take a look at this as well, the "Test data" needs a bit of expanding as the tests here work quite differently from what a seasoned Robot Framework user would normally expect but I understand why given the fact that behavior during test failures needs to be tested as well). This merging was where I found the bug with the missing IF. Now the acceptance tests run the new test as well and they are passing.

jussimalinen:
robotci: once

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/105/

jussimalinen:
Hmm, there seems to be some configuration problem on the CI.

jussimalinen:
The tests are passing and they look good to me with the changes.

jussimalinen:
Thanks!

jussimalinen:
I will add you to contributors if that is okay.

osjc:
To jussimalinen: Of course it is okay to add me to contributors.

PR #2214, opened 2015-11-13
String: add strip keyword

### LINKED ISSUE #2213: String: provide a strip keyword
Similar to the python [l,r]strip() method. This will eliminate hacks like:

    ${output}=  Evalutate  '${output}'.strip()

mwalle:
Just like the python strip(), lstrip() and rstrip() functions.

This fixes #2213.

Signed-off-by: Michael Walle <michael.walle@kontron.com>


robotci:
Admin, can CI be enabled?

mwalle:
mh, left and right are not documented. will do so in the next iteration after any feedback.

src/robot/libraries/String.py
```diff
@@ -530,6 +530,39 @@ def get_substring(self, string, start, end=None):
...
+        characters in the string will be stripped in the string. Please note,
+        that this is not a substring to be removed but a list of characters,
+        see the example below.
+
+        Examples:
+        | ${stripped}=  | Strip String | ${SPACE}Hello${SPACE} | |
+        | Should Be Equal | ${stripped} | Hello | |
+        | ${stripped}=  | Strip String | mode=left | ${SPACE}Hello${SPACE} |
+        | Should Be Equal | ${stripped} | Hello${SPACE} | |
+        | ${stripped}=  | Strip String | characters=abe | aabaHelloeee |
```
pekkaklarck: Arguments seem to be in wrong order.

src/robot/libraries/String.py
```diff
@@ -530,6 +530,39 @@ def get_substring(self, string, start, end=None):
...
+        """
+        try:
+            method = {'BOTH': string.strip,
+                      'LEFT': string.lstrip,
+                      'RIGHT': string.rstrip,
+                      'NONE': lambda characters: string}[mode.upper()]
+        except KeyError:
+            raise ValueError("Invalid mode '%s'." % mode)
+
+        return method(characters)
```
pekkaklarck: Empty row above isn't needed. Not a big problem, but our general code style is not to have them inside method or functions.

atest/testdata/standard_libraries/string/string.robot
```diff
@@ -86,3 +86,27 @@ Get Substring With Invalid Start
 Get Substring With Invalid End
     [Documentation]    FAIL ValueError: Cannot convert 'end' argument 'invalid' to an integer.
     Get Substring    Hello Robot    2    invalid
+
+Strip String
+    ${result} =    Strip String  ${SPACE}${SPACE}Hello${SPACE}
```
pekkaklarck: Four spaces between keywords and arguments would be a good. I know existing tests don't use that style consistently, but we are trying to move to that direction.

PR #2235, opened 2015-11-25
[RFC] Telnet: Add support for Binary Transmission

mwalle:
This is specified in RFC 856. In a nutshell, it disables any special
interpretation for all characters but FFh (IAC). Useful, if you have a
serial-line-to-telnet service and want to transmit binary data on the
serial line.

Signed-off-by: Michael Walle michael.walle@kontron.com

### DIFF
--- a/src/robot/libraries/Telnet.py
+++ b/src/robot/libraries/Telnet.py
@@ -530,7 +530,9 @@ class TelnetConnection(telnetlib.Telnet):
         self._terminal_type = self._encode(terminal_type) if terminal_type else None
         self.set_option_negotiation_callback(self._negotiate_options)
         self._set_telnetlib_log_level(telnetlib_log_level)
-        self._opt_responses = list()
+        self._opt_responses = set()
+        self._allow_binary = False
+        self._binary_enabled = False
 
     def set_timeout(self, timeout):
         """Sets the timeout used for waiting output in the current connection.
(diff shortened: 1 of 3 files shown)


robotci:
Admin, can CI be enabled?

src/robot/libraries/Telnet.py
```diff
@@ -528,7 +528,9 @@ def __init__(self, host=None, port=23, timeout=3.0, newline='CRLF',
         self._terminal_type = self._encode(terminal_type) if terminal_type else None
         self.set_option_negotiation_callback(self._negotiate_options)
         self._set_telnetlib_log_level(telnetlib_log_level)
-        self._opt_responses = list()
+        self._opt_responses = set()
```
mwalle: convert to set(), so we can use the discard() method instead of `try: l.remove(); except ValueError: pass`

src/robot/libraries/Telnet.py
```diff
@@ -723,6 +725,28 @@ def close_connection(self, loglevel=None):
...
+            self.sock.sendall(telnetlib.IAC + telnetlib.DO + telnetlib.BINARY)
+            self.sock.sendall(telnetlib.IAC + telnetlib.WILL + telnetlib.BINARY)
+        else:
+            self.sock.sendall(telnetlib.IAC + telnetlib.DONT + telnetlib.BINARY)
+            self.sock.sendall(telnetlib.IAC + telnetlib.WONT + telnetlib.BINARY)
+
+        # XXX: this is rather hacky. Because we may have a state transision for
+        # the BINARY option, we have to remove it from our option responses
+        for cmd in (telnetlib.DO, telnetlib.DONT, telnetlib.WILL, telnetlib.WONT):
+            self._opt_responses.discard((cmd, telnetlib.BINARY))
```
mwalle: the other way would be a full-blown state tracking of these options.

src/robot/libraries/Telnet.py
```diff
@@ -1180,6 +1215,74 @@ def _check_terminal_emulation(self, terminal_emulation):
         return TerminalEmulator(window_size=self._window_size,
                                 newline=self._newline, encoding=self._encoding)
 
+    def process_rawq(self):
```
mwalle: this is the copy of the process_rawq() from python with two lines (1226 and 1228) changed

src/robot/libraries/Telnet.py
```diff
@@ -1124,23 +1148,34 @@ def _negotiate_options(self, sock, cmd, opt):
...
                 return
             else:
-                self._opt_responses.append((cmd, opt))
+                self._opt_responses.add((cmd, opt))
 
         # This is supposed to turn server side echoing on and turn other options off.
         if opt == telnetlib.ECHO and cmd in (telnetlib.WILL, telnetlib.WONT):
             self._opt_echo_on(opt)
+        elif opt == telnetlib.BINARY and cmd in (telnetlib.WILL, telnetlib.WONT):
+            self._binary_enabled = self._allow_binary and cmd == telnetlib.WILL
```
mwalle: remember if the peer has binary mode enabled or not.

mwalle:
next patch will include support for python 3 and a small fix, both are backwards compatible to python 2.6 and  2.7 (they now use byte arrays instead of strings). apart from that, the process_rawq() wasn't changed since python (at least) 2.3.

pekkaklarck:

PR #2236, opened 2015-11-25
Adds the byte string flag to a test condition.

HelioGuilherme66:
This fixes tests in Windows  when username preceded by \ would result in escape sequence, for example starting with 'x'.

This is a very low priority PR because it only affects RF acceptance tests. However it could serve as an example for a fix in a particular situation. After this fix, there are only three failed tests (one of then due to lowercase/uppercase comparison in a Windows path).


robotci:
Admin, can CI be enabled?

pekkaklarck:
That `r` actually turns the name to a raw string, not a byte string. In a raw string backslash has no special meaning and it thus works well with Windows paths (`r'c:\temp`) and regexps (`r'\s*\d+'`).

Although adding the `r` fixes the problem, a better solution here is using `$outputname` in the expression. This way the code works correctly also if `${outputname}` is `None`. I can do that myself.

PR #2262, opened 2015-12-15
Supporting time for Repeat Keyword in BuiltIn Lib

joongh:
"Repeat Keyword" repeats the keyword specified times basically.
This commit is to support time string (in robot's time format) as the
the first argument of the "Repeat Keyword".
When a time string passed to the first argument, "Repeat Keyword"
repeats the keyword for the specified time regardless of how many times
it repeats.
A number like 42 is interpreted as a count instead of 42 seconds for
backward compatibility. Also floating point numbers like 1.5 would
raise exception just like before for same reason.

### DIFF
--- a/atest/robot/standard_libraries/builtin/repeat_keyword.robot
+++ b/atest/robot/standard_libraries/builtin/repeat_keyword.robot
@@ -25,10 +25,17 @@ Zero And Negative Times
     ${tc} =    Check Test Case    ${TEST NAME}
     Check Repeated Messages    ${tc.kws[0]}    0
     Check Repeated Messages    ${tc.kws[2]}    0
+    Check Repeated Messages    ${tc.kws[3]}    0
 
 Invalid Times
     Check Test Case    Invalid Times 1
     Check Test Case    Invalid Times 2
+
+Repeat Keyword With Time String
+    ${tc} =    Check Test Case    ${TEST NAME}
+    Check Repeated Messages With Time    ${tc.kws[0]}    This is done for 00:00:00.003
+    Check Repeated Messages With Time    ${tc.kws[1]}    This is done for 3 milliseconds
+    Check Repeated Messages With Time    ${tc.kws[2]}    This is done for 3ms
 
 Repeat Keyword Arguments As Variables
     ${tc} =    Check Test Case    ${TEST_NAME}
@@ -72,6 +79,15 @@ Check Repeated Messages
     \    Check Log Message    ${kw.kws[${i}].msgs[0]}    ${msg}
     Run Keyword If    ${count} == 0    Check Log Message    ${kw.msgs[0]}    Keyword 'This is not executed' repeated zero times.
     Run Keyword If    ${count} != 0    Should Be Equal As Integers    ${kw.msg_count}    ${count}
+
+Check Repeated Messages With Time
+    [Arguments]    ${kw}    ${msg}=${None}
+    Should Be True    ${kw.kw_count} > 1
+    : FOR    ${i}    IN RANGE    ${kw.kw_count}
+    \    Check Log Message    ${kw.msgs[${i}]}
+    \    ...    Repeating keyword, round ${i+1}, *remaining.    pattern=yes
+    \    Check Log Message    ${kw.kws[${i}].msgs[0]}    ${msg}
+    Should Be Equal As Integers    ${kw.msg_count}    ${kw.kw_count}
 
 Check Repeated Keyword Name
     [Arguments]    ${kw}    ${count}    ${name}=${None}
--- a/atest/testdata/standard_libraries/builtin/repeat_keyword.robot
+++ b/atest/testdata/standard_libraries/builtin/repeat_keyword.robot
@@ -22,6 +22,7 @@ Zero And Negative Times
     Repeat Keyword    0 times    This is not executed
     ${name} =    Set Variable    This is not executed
     Repeat Keyword    ${-1}    ${name}    ${nonex}
+    Repeat Keyword    0 secs    This is not executed
 
 Invalid Times 1
     [Documentation]    FAIL STARTS: '1.3' cannot be converted to an integer: ValueError:
@@ -30,6 +31,11 @@ Invalid Times 1
 Invalid Times 2
     [Documentation]    FAIL STARTS: 'notaninteger' cannot be converted to an integer: ValueError:
     Repeat Keyword    Not an integer    No Operation
+
+Repeat Keyword With Time String
+    Repeat Keyword    00:00:00.003    Log    This is done for 00:00:00.003
+    Repeat Keyword    3 milliseconds    Log    This is done for 3 milliseconds
+    Repeat Keyword    3ms    Log    This is done for 3ms
 
 Repeat Keyword Arguments As Variables
     ${kw}    ${arg} =    Set Variable    Should Be Equal    Hello, world!
--- a/src/robot/libraries/BuiltIn.py
+++ b/src/robot/libraries/BuiltIn.py
@@ -1662,26 +1662,56 @@ class _RunKeyword(_BuiltInBase):
         return unic(error)
 
     @run_keyword_variant(resolve=2)
-    def repeat_keyword(self, times, name, *args):
+    def repeat_keyword(self, repeat, name, *args):
         """Executes the specified keyword multiple times.
 
         ``name`` and ``args`` define the keyword that is executed
-        similarly as with `Run Keyword`, and ``times`` specifies how many
-        the keyword should be executed. ``times`` can be given as an
-        integer or as a string that can be converted to an integer. If it is
-        a string, it can have postfix ``times`` or ``x`` (case and space
-        insensitive) to make the expression more explicit.
+        similarly as with `Run Keyword`.
 
-        If ``times`` is zero or negative, the keyword is not executed at
+        ``repeat`` specifies how many times (count) or how long time
+        (timeout) the keyword should be executed.
+
+        If ``repeat`` is given as count, it specifies how many times the
+        keyword should be executed. ``repeat`` can be given as an integer or
+        as a string that can be converted to an integer. If it is a string,
+        it can have postfix ``times`` or ``x`` (case and space insensitive)
+        to make the expression more explicit.
+
+        If ``repeat`` is given as timeout, it must be in Robot Framework's
+        time format (e.g. ``1 minute``, ``2 min 3 s``). Using a number alone
+        (e.g. 1 or 1.5) does not work in this context.
+
+        If ``repeat`` is zero or negative, the keyword is not executed at
         all. This keyword fails immediately if any of the execution
         rounds fails.
 
         Examples:
         | Repeat Keyword | 5 times | Go to Previous Page |
         | Repeat Keyword | ${var}  | Some Keyword | arg1 | arg2 |
+        | Repeat Keyword | 2 mins  | Some Keyword | arg1 | arg2 |
         """
-        times = self._get_times_to_repeat(times)
-        self._run_keywords(self._yield_repeated_keywords(times, name, args))
+        try:
+            count = self._get_times_to_repeat(repeat)
+        except RuntimeError as err:
+            timeout = self._get_timeout_to_repeat(repeat)
+            if timeout is None:
+                raise err
+            keywords = self._yield_repeated_keywords_until(timeout, name, args)
+        else:
+            keywords = self._yield_repeated_keywords(count, name, args)
+        self._run_keywords(keywords)
+
+    def _get_timeout_to_repeat(self, timestr):
+        try:
+            float(timestr)
+        except ValueError:
+            pass
+        else:
+            return None
+        try:
+            return timestr_to_secs(timestr)
+        except ValueError:
+            return None
 
     def _get_times_to_repeat(self, times, require_postfix=False):
         times = normalize(str(times))
@@ -1698,6 +1728,18 @@ class _RunKeyword(_BuiltInBase):
             self.log("Keyword '%s' repeated zero times." % name)
         for i in range(times):
             self.log("Repeating keyword, round %d/%d." % (i+1, times))
+            yield name, args
+
+    def _yield_repeated_keywords_until(self, timeout, name, args):
+        repeat_round = 0
+        if timeout <= 0:
+            self.log("Keyword '%s' repeated zero times." % name)
+        maxtime = time.time() + timeout
+        while time.time() < maxtime:
+            repeat_round += 1
+            self.log("Repeating keyword, round %d, %s remaining."
+                     % (repeat_round,
+                        secs_to_timestr(maxtime - time.time(), compact=True)))
             yield name, args
 
     @run_keyword_variant(resolve=3)
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -69,3 +69,4 @@ Jean-Charles Deville           Make variable errors not exit `Runner keywords` (
 Laurent Bristiel               Convert examples in User Guide to plain text format (2.9, #1972)
 Tim Orling                     IronPython support for `Dialogs` library (2.9.2, #1235)
 Jozef Behran                   Fix ${TEST_MESSAGE} to reflect current test message (3.0, #2188)
+Joong-Hee Lee                  Extend 'Repeat Keyword' to support timeout (3.0, #2245)


robotci:
Admin, can CI be enabled?

atest/testdata/standard_libraries/builtin/repeat_keyword.robot
```diff
@@ -31,6 +32,12 @@ Invalid Times 2
     [Documentation]    FAIL STARTS: 'notaninteger' cannot be converted to an integer: ValueError:
     Repeat Keyword    Not an integer    No Operation
 
+Repeat Keyword With Time String
+    Repeat Keyword    00:00:03    Keyword Executing For 1 second    00:00:03
+    Repeat Keyword    3 seconds    Keyword Executing For 1 second    3 seconds
+    Repeat Keyword    3 secs    Keyword Executing For 1 second    3 secs
+    Repeat Keyword    3s    Keyword Executing For 1 second    3s
```
pekkaklarck: Repeating keywords for 3 seconds in all these iterations consumes quite a bit time altogether. Please change it to something smaller. If you are only testing time string handling, it ought to be OK to have just 3 milliseconds. If you need to test that the keyword was executed several times, 0.3 seconds ought to be good.

atest/testdata/standard_libraries/builtin/repeat_keyword.robot
```diff
@@ -89,3 +96,8 @@ First Continuable Failure And Then Pass Execution
     Pass Execution If    "${PASS EXECUTION}" == "yes"    Message
     Set Suite Variable    ${PASS EXECUTION}    yes
     Run Keyword And Continue On Failure    Fail    Continuable
+
+Keyword Executing For 1 second
+    [Arguments]     ${TIME FORMAT}
+    Log    This is done for ${TIME FORMAT}
+    Sleep    1 seconds
```
pekkaklarck: Related to the above comment, 1 second is too long. I don't actually see why sleeping is needed altogether.

Also, we want to have newlines at the end of files. I recommend configuring your editor/IDE to add them automatically.

src/robot/libraries/BuiltIn.py
```diff
@@ -1662,26 +1662,55 @@ def run_keyword_and_expect_error(self, expected_error, name, *args):
...
+        similarly as with `Run Keyword`.
+
+        If ``repeat`` is given as count, it specifies how many the keyword
+        should be executed. ``repeat`` can be given as an
         integer or as a string that can be converted to an integer. If it is
         a string, it can have postfix ``times`` or ``x`` (case and space
         insensitive) to make the expression more explicit.
 
-        If ``times`` is zero or negative, the keyword is not executed at
+        If ``retry`` is given as timeout, it must be in Robot Framework's
```
pekkaklarck: `repeat`, not `retry`.

src/robot/libraries/BuiltIn.py
```diff
@@ -1662,26 +1662,55 @@ def run_keyword_and_expect_error(self, expected_error, name, *args):
...
         Examples:
         | Repeat Keyword | 5 times | Go to Previous Page |
         | Repeat Keyword | ${var}  | Some Keyword | arg1 | arg2 |
+        | Repeat Keyword | 2 mins  | Some Keyword | arg1 | arg2 |
         """
-        times = self._get_times_to_repeat(times)
-        self._run_keywords(self._yield_repeated_keywords(times, name, args))
+        try:
+            count = self._get_times_to_repeat(repeat, require_postfix=False)
+        except RuntimeError, e:
```
pekkaklarck: This is not Python 3 compatible.

src/robot/libraries/BuiltIn.py
```diff
@@ -1662,26 +1662,55 @@ def run_keyword_and_expect_error(self, expected_error, name, *args):
...
+                # originally occurred by _get_times_to_repeat() for
+                # backward compatibility
+                # Without it, "Invalid Times" in acceptance test fails.
+                raise e
+            else:
+                kwiter = self._yield_repeated_keywords_until(maxtime, name,
+                                                             args)
+        else:
+            kwiter = self._yield_repeated_keywords(count, name, args)
+        self._run_keywords(kwiter)
```
pekkaklarck: The above code is pretty complicated with nested exception handlers etc. and doesn't fully follow our coding guidelines otherwise either. Please try to clean it up. Things to take into account:

- Try to separate handling repeat as count and maxtime more clearly.
- You should be able to write the code so cleanly that it needs no comments.
- The code needs to be Python 3 compatible. The current exception handling syntax is not.
- Avoid one char variables except in tight loops. Use e.g. `err` or `error` instead of `e`.

If you need more help, please first check code conventions found from CONTRIBUTING.rst file. If that's not enough, I can comment more when I have time.

pekkaklarck:
Looks pretty good! As I already commented in a line-note, I'd like the code to be cleaned up a little and tests to be made faster.

Please also add yourself to `AUTHORS.txt` and squash commits after the final review (no need to do it now). For more details about these topics, coding conventions, etc. see https://github.com/robotframework/robotframework/blob/master/CONTRIBUTING.rst.

atest/testdata/standard_libraries/builtin/repeat_keyword.robot
joongh: To check that "Repeat Keyword" actually repeats the keyword, I needed to be sure that the keyword runs several times. Because just "several times" cannot be the criterion for the test, I wanted to repeat keyword as number of times as I intended.
If the keyword sleep for 1 second at once, and "Repeat Keyword" repeats it for 3 seconds, the keyword should run 3 times approximately.

I will modify that the keyword sleeping 0.1 second and repeating it for 0.3 second.
Is it OK? Any other suggestions?

atest/robot/standard_libraries/builtin/repeat_keyword.robot
```diff
@@ -25,11 +25,19 @@ Zero And Negative Times
     ${tc} =    Check Test Case    ${TEST NAME}
     Check Repeated Messages    ${tc.kws[0]}    0
     Check Repeated Messages    ${tc.kws[2]}    0
+    Check Repeated Messages With Time    ${tc.kws[3]}    0
```
pekkaklarck: Shouldn't you be able to use the same keyword here as the earlier steps?

It would actually be a good idea to add a separate test for not executing keyword at all if time is 0 or smaller. The old test can be left for testing the same behavior when count <= 0.

atest/robot/standard_libraries/builtin/repeat_keyword.robot
```diff
@@ -73,6 +81,18 @@ Check Repeated Messages
...
+    [Arguments]    ${kw}    ${expect_count}    ${msg}=${None}
+    # Cannot test exactly how many times kw is run because it depends on
+    # interpreter speed.
+    Should Be True    ${kw.kw_count} <= ${expect_count}
+    : FOR    ${i}    IN RANGE    ${kw.kw_count}
+    \    Check Log Message    ${kw.msgs[${i}]}
+    \    ...    Repeating keyword, round ${i+1}, remaining*    pattern=yes
+    \    Check Log Message    ${kw.kws[${i}].kws[0].msgs[0]}    ${msg}
+    Run Keyword If    ${kw.kw_count} == 0    Check Log Message    ${kw.msgs[0]}    Keyword 'This is not executed' repeated zero times.
+    Run Keyword If    ${kw.kw_count} != 0    Should Be Equal As Integers    ${kw.msg_count}    ${kw.kw_count}
```
pekkaklarck: This is pretty complicated. Wouldn't it be enough to test that the keyword has been executed at least once? Trying to guess how many times it is actually run is hard and creates complex and/or fragile tests.

atest/testdata/standard_libraries/builtin/repeat_keyword.robot
```diff
@@ -31,6 +32,12 @@ Invalid Times 2
     [Documentation]    FAIL STARTS: 'notaninteger' cannot be converted to an integer: ValueError:
     Repeat Keyword    Not an integer    No Operation
 
+Repeat Keyword With Time String
+    Repeat Keyword    00:00:00.3    Keyword Executing For 0.1 second    00:00:00.3
+    Repeat Keyword    0.3 seconds    Keyword Executing For 0.1 second    0.3 seconds
+    Repeat Keyword    0.3 secs    Keyword Executing For 0.1 second    0.3 secs
+    Repeat Keyword    0.3s    Keyword Executing For 0.1 second    0.1s 200x
```
pekkaklarck: These times could still be shorted if it would be just verified that keywords are executed at least once. Also, what does that `200x` mean above?

src/robot/libraries/BuiltIn.py
```diff
@@ -1662,26 +1662,58 @@ def run_keyword_and_expect_error(self, expected_error, name, *args):
...
+        try:
+            count = self._get_times_to_repeat(repeat)
+        except Exception as err:
+            if not maxtime:
+                raise err
+        if count:
+            kwiter = self._yield_repeated_keywords(count, name, args)
+        else:
+            kwiter = self._yield_repeated_keywords_until(maxtime, name, args)
+        self._run_keywords(kwiter)
```
pekkaklarck: This is still pretty complicated and using plain `except:` is *very* strongly against PEP-8 and our coding guidelines.

I would expect that something like this would work:

```python
try:
    count = self._get_times_to_repeat(repeat)
except RuntimeError:
    timeout = timestr_to_secs(repeat)
    keywords = self._yield_repeated_keywords_until(timeout, name, args)
else:
    keywords = self._yield_repeated_keywords(count, name, args)
self._run_keywords(keywords)
```

joongh: @pekkaklarck, Thanks for your suggestion.
I implemented this just like your code for the first time. The code was becoming complicated for the backward compatibility issues.
With your code, the test case "Invalid Times" in acceptance test (repeat_keyword.robot) fails.
- Invalid Times 1
  : 1.3 does not make failure. It is interpreted into 1.3 sec
- Invalid Times 2
  : FAIL message is different, because the exception is raised by timestr_to_secs().

The issues cannot be a problem with the modification of the existing test case "Invalid Times".
Is it OK with you?

atest/testdata/standard_libraries/builtin/repeat_keyword.robot
joongh: @pekkaklarck Thanks for your opinion.
OK, I will change the test code just verifying only more than one executing to simplify and shorten the test.
I wanted to test 'x' as millisecond not a count if it is in time string, so that I wrote 0.1second and 200 millisecond (totally 0.3 second)

src/robot/libraries/BuiltIn.py
pekkaklarck: I see. It would be confusing if `1.0` is interpreted as maxtime and `1` as count. To make things even more confusing, the current code actually accepts float `1.0` but not string `'1.0'`. There are two options:

- Inside `except RuntimeError` validate that `count` is not `'1.0'` or similar and reraise the exception if it is. Validation is best done with a separate helper method.

- Decide that strings `1.0` and `1.3` are valid counts. This would be consistent with how counts given as floats would be handled. It would basically require changing `_get_times_to_repeat` to have `return int(self._convert_to_number(times))`. Should then also decide do we want `1.9` to be `1` or `2`. I guess rounding down like `int` does is fine. If we would like to round to the nearest integer instead, it wouldn't require anything else than passing `precision=0` to `_convert_to_number`, though. (See also #2268 about a related bug I just spotted and fixed.)

The latter solution would obviously require changing the related tests too. The test that validates that `1.3` is invalid should basically be changed to a test that validates that it is valid and same as `1`.

The other problem with the error message is likely to require a change to tests as well as small adjustment to the code I posted:

```python
try:
    timeout = timestr_to_secs(repeat)
except ValueError:
    raise RuntimeError("Given repeat value '%s' is invalid." % repeat)

Comment (joongh) [inline src/robot/libraries/BuiltIn.py]: It seems the first suggestion which is count validation inside ```except RuntimeError``` looks fine. I guess the second suggestion would bring a confusion for users.
For the second problem, it seems also a reasonable solution, but it cannot avoid the nested exception handling.

Putting two solution together, the code will be just like the code I pushed for the first time.

Comment (pekkaklarck) [inline src/robot/libraries/BuiltIn.py]: I'm OK both with `'1.3'` being invalid and it being same as `'1'`. If it is invalid, the error should be same as if the value would be totally wrong like `'foo'`.

Nested exception handling isn't itself a problem, but the code gets pretty complicated if exception handling blocks are long. That's typically easy to avoid by using a helper method. Also, plain `except:` should never be used unless the exception handling code propagates exceptions like `SystemExit`.

Anyway, hope you can create one more version. I can do simple cleanup after merging if I think it is necessary too.

File: src/robot/libraries/BuiltIn.py [hunk 10]
```diff
@@ -1662,26 +1662,58 @@ def run_keyword_and_expect_error(self, expected_error, name, *args):
...
-    def repeat_keyword(self, times, name, *args):
+    def repeat_keyword(self, repeat, name, *args):
         """Executes the specified keyword multiple times.
 
         ``name`` and ``args`` define the keyword that is executed
-        similarly as with `Run Keyword`, and ``times`` specifies how many
-        the keyword should be executed. ``times`` can be given as an
+        similarly as with `Run Keyword`.
+
+        If ``repeat`` is given as count, it specifies how many the keyword
```
pekkaklarck: 'how many the keyword' -> 'how many times the keyword'

src/robot/libraries/BuiltIn.py
```diff
@@ -1662,26 +1662,58 @@ def run_keyword_and_expect_error(self, expected_error, name, *args):
...
+        should be executed. ``repeat`` can be given as an
         integer or as a string that can be converted to an integer. If it is
         a string, it can have postfix ``times`` or ``x`` (case and space
         insensitive) to make the expression more explicit.
 
-        If ``times`` is zero or negative, the keyword is not executed at
+        If ``repeat`` is given as timeout, it must be in Robot Framework's
+        time format (e.g. ``1 minute``, ``2 min 3 s``) except a number like
+        42 is interpreted as count, so that a floating point number like
+        4.2 would raise exception.
```
pekkaklarck: This sounds a bit complicated. Perhaps something like this would be better:

> If ``repeat`` is given as timeout, it must be in Robot Framework's  time format (e.g. ``1 minute``, ``2 min 3 s``). Using a number alone (e.g. `1` or `1.5`) does not work in this context.

src/robot/libraries/BuiltIn.py
```diff
@@ -1662,26 +1662,58 @@ def run_keyword_and_expect_error(self, expected_error, name, *args):
...
+        You can use ``x`` as millisecond only with ``repeat`` including other
+        string that can be interpreted as timeout such as ``1 s 10 x``. Time
+        string only with ``x`` is interpreted as count.
```
pekkaklarck: Is this relevant? Being able to use `x` as milliseconds is actually a bug.

src/robot/libraries/BuiltIn.py
```diff
@@ -1662,26 +1662,58 @@ def run_keyword_and_expect_error(self, expected_error, name, *args):
...
 
     @run_keyword_variant(resolve=2)
-    def repeat_keyword(self, times, name, *args):
+    def repeat_keyword(self, repeat, name, *args):
         """Executes the specified keyword multiple times.
 
         ``name`` and ``args`` define the keyword that is executed
-        similarly as with `Run Keyword`, and ``times`` specifies how many
-        the keyword should be executed. ``times`` can be given as an
+        similarly as with `Run Keyword`.
```
pekkaklarck: Should add something like this at the end:

> ``repeat`` specifies how many times (count) or how long time (timeout) the keyword should be executed.

src/robot/libraries/BuiltIn.py
joongh: Can you explain why it is a bug?
I thought that ```x``` in the time string can clearly be distinguished against ```x``` as count postfix.
As you know, only number and ```x``` such as ```10x``` is interpreted in the count, but ```x``` in time string seems to be interpreted as millisecond.

src/robot/libraries/BuiltIn.py
```diff
@@ -1662,26 +1662,58 @@ def run_keyword_and_expect_error(self, expected_error, name, *args):
...
+        else:
+            keywords = self._yield_repeated_keywords(count, name, args)
+        self._run_keywords(keywords)
+
+    def _get_maxtime_to_repeat(self, timestr, accept_numeric=True):
+        if not accept_numeric:
+            if is_string(timestr) and timestr.replace('.', '', 1).isdigit():
+                return None
+            elif is_number(timestr):
+                return None
```
pekkaklarck: `accept_numeric` seems to always be `False`. Couldn't the above code be written just like:

```python
try:
    float(timestr)
except ValueError:
    pass
else:
    return None
```

src/robot/libraries/BuiltIn.py
```diff
@@ -1662,26 +1662,58 @@ def run_keyword_and_expect_error(self, expected_error, name, *args):
...
+        try:
+            timeout = timestr_to_secs(timestr)
+            maxtime = time.time() + timeout
+        except ValueError:
+            return None
+        return maxtime
```
pekkaklarck: I would rather return `timeout` from here and let later methods calculate `maxtime`. That would allow changing the code to this:

```python
try:
    return timestr_to_secs(timestr)
except ValueError:
    return None
```

src/robot/libraries/BuiltIn.py
```diff
@@ -1700,6 +1732,17 @@ def _yield_repeated_keywords(self, times, name, args):
             self.log("Repeating keyword, round %d/%d." % (i+1, times))
             yield name, args
 
+    def _yield_repeated_keywords_until(self, maxtime, name, args):
+        repeat_round = 0
+        if time.time() >= maxtime:
```
pekkaklarck: With the current implementation there is a possibility that if timeout is very small, `time.time() >= maxtime` is true even before the keyword is executed once. It would be safer (and in my opinion cleaner) to pass timeout, not maxtime, here and change the condition to `timeout <= 0`. After that maxtime can be set like `maxtime = time.time() + timeout`.

src/robot/libraries/BuiltIn.py
```diff
@@ -1700,6 +1732,17 @@ def _yield_repeated_keywords(self, times, name, args):
...
+            self.log("Keyword '%s' repeated zero times." % name)
+        while time.time() < maxtime:
+            repeat_round += 1
+            self.log("Repeating keyword, round %d, remaining %s."
```
pekkaklarck: `round %d, %s remaining` would perhaps read a bit better in logs.

pekkaklarck:
The current implementation and tests look good. I added few lines notes about the docs and about possible cleanups to the code. Remaining tasks before we can merge:

- Take a look at the aforementioned line notes.
- Resolve conflict.
- Squash commits (see CONTRIBUTING.rst for details)

We try to get RF 3.0 beta 1 released tomorrow. It would be great to get this into it, but there's going to still be at least one release candidate before the final release.

pekkaklarck:
@robotci: Enable

src/robot/libraries/BuiltIn.py
pekkaklarck: `x` in time strings being interpreted as milliseconds is a bug in the time string parsing algorithm.

robotci:
http://robot.radiaatto.ri.fi//job/RobotFramework-pull-requests/106/

pekkaklarck:
There seems to be something wrong on CI. No need to care about the results there.

pekkaklarck:
The latest version looks great. I forgot to mention that docs should mention what specifying repeat as a timeout is new in RF 3.0 but I can add that myself.

pekkaklarck:
I merged this PR from the command line and apparently messed something there because the merge wasn't automatically detected here. It could be that the reason was that I did `git pull --rebase` before pushing changes. Doesn't really matter, I can close this PR manually.

Thanks for the PR @joongh! If you are interested, you can see small enhancements I did after merge in 64468cdf51ebc55f867ce05753961805e1b2f862.

Issue #2216, opened 2015-11-16
New `robot` start-up script to replace `pybot`, `jybot` and `ipybot`


jussimalinen:
Now that we start supporting more interpreter types, like python 3 and perhaps pypy in future, it starts to get quite a bit much to create a new starter script for each. (pybot, pybot3, pypybot...?) And what when there will be 3-versions of ironpython and jython?

It seems like a good time to go for the generic `robot` starter script, which would be for the version that was last installed. It is rather rare to need to run robot on multiple interpreters on the same machine. And if that is needed, there is always the option `jython -m robot.run`.

For now we should just install the new script along with the old scripts. Then maybe in future versions we can start deprecating the old scripts.

UPDATE: We also decided to use `rebot` with all interpreters. `jyrebot` and `ipyrebot` will continue to work as earlier for now.

pekkaklarck:
Code changes look good and updating `r?bot.bat` wasn't as hacky as I was afraid. Doc changes are to be reviewed separately, but I still want to test the new Windows scripts with virtual environments.

jussimalinen:
Also, now we tested this on all sorts of platforms, including virtualenv and everything works like a charm. The new `robot` script also now sets the complete path to actual interpreter, unlike `pybot`, `jybot` and others, which simply call `python` or `jython` and assume that the interpreter in path is the correct one.

pekkaklarck:
Tested with virtual environments both on Linux and Windows. Worked great! I consider this issue done.

PR #2314, opened 2016-02-09
Added idom to allow for multiprocessing spawn to work correctly.

cleopold:
Typically only needed on windows, but with python 3.4, you can spawn instead of fork with multiprocessing on Linux/Posix operating systems making this necessary.

### DIFF
--- a/src/bin/ipybot
+++ b/src/bin/ipybot
@@ -3,4 +3,5 @@
 import sys
 from robot import run_cli
 
-run_cli(sys.argv[1:])
+if __name__ == '__main__':
+    run_cli(sys.argv[1:])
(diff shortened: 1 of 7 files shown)


robotci:
Admin, can CI be enabled?

pekkaklarck:
Sounds good. Now that the official start-up scripts are `robot` and `rebot` (#2216), I think adding this to `pybot` is not worth it. If it's added there, it should then logically be added variants too.

Could you submit an issue about this too so that we can include that in RF 3.0.1 scope?

cleopold:
Went ahead and updated all the variants, they might as well work in this case if they are there.

pekkaklarck:
Added guards in 2473247c442ddfbd68590c817830e1c52f3a450c. Only wanted to add them to robot and rebot scripts in the end and didn't want to bother you with such a trivial change. Thanks for submitting the PR and the issue anyway!

PR #2329, opened 2016-03-03
Fix a bug in the JS model builder tests

philsc:
The test intends to make sure that the TestSuite class produces the expected
behaviour. However, the test does not account for potentially compressed
strings that are produced inside the TestSuite class.

This patch compresses those strings to match what TestSuite would do.

Example to showcase the failure:
```bash
#!/bin/bash

readonly TEMP_DIR="$(mktemp -d)"
CLONE_DIR="$TEMP_DIR"
for i in {1..30}; do
    CLONE_DIR+="/longdirname"
done
readonly CLONE_DIR

mkdir -p "$CLONE_DIR"

git clone https://github.com/robotframework/robotframework.git "$CLONE_DIR"/

pushd "$CLONE_DIR"
./utest/run.py
popd

rm -rf "$TEMP_DIR"
```

### DIFF
--- a/utest/reporting/test_jsmodelbuilders.py
+++ b/utest/reporting/test_jsmodelbuilders.py
@@ -1,7 +1,10 @@
+import base64
 import unittest
+import zlib
 from os.path import abspath, basename, dirname, join
 
 from robot.utils.asserts import assert_equal, assert_true
+from robot.utils.platform import PY2
 from robot.result import Message, Keyword, TestCase, TestSuite
 from robot.result.executionerrors import ExecutionErrors
 from robot.model import Statistics
@@ -17,9 +20,17 @@ except NameError:
 CURDIR = dirname(abspath(__file__))
 
 
+def decode_string(string):
+    string = string if PY2 else string.encode('ASCII')
+    return zlib.decompress(base64.b64decode(string)).decode('UTF-8')
+
+
 def remap(model, strings):
     if isinstance(model, StringIndex):
-        return strings[model][1:]
+        if strings[model].startswith('*'):
+            # Strip the asterisk from a raw string.
+            return strings[model][1:]
+        return decode_string(strings[model])
     elif isinstance(model, (int, long, type(None))):
         return model
     elif isinstance(model, tuple):


robotci:
Admin, can CI be enabled?

philsc:
Here's my system information. I can reliably reproduce this on several machines.
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 8.3 (jessie)
Release:	8.3
Codename:	jessie
$ python --version
Python 2.7.9
```

pekkaklarck:
Good catch and a great fix!

utest/reporting/test_jsmodelbuilders.py
```diff
@@ -19,7 +19,10 @@
 
 def remap(model, strings):
     if isinstance(model, StringIndex):
-        return strings[model][1:]
+        if strings[model].startswith('*'):
+            # Strip the asterisk from a raw string.
+            return strings[model][1:]
+        return strings[model]
```
pekkaklarck: Would it actually be enough to decompress the string here? It ought to be easier than compressing expected results. It would also be easier to analyse errors if values wouldn't be compressed.

philsc: It would be, I agree. It was actually my first thought, but I couldn't find a decompress method in any of the libraries. I would have to write an inverse operation of what StringCache does. I'm totally up for doing that. Would you prefer that?

pekkaklarck: Yeah, there's no decompress method in the code base simply because it isn't needed. It ought to be easy to implement here. I'm fairly sure the end result would be better.

philsc: OK, let me rework that then.

philsc:
I've uploaded a new version. Let me know what you think.

pekkaklarck:
Looks great! `decompress_text` would perhaps be a bit better name for the introduced helper method but that doesn't really matter in unit tests.

PR #2330, opened 2016-03-03
Avoid a recursive call to List files

alkhatib:
Tried running the example code locally and got an error "Maximum limit of started keywords exceeded."
It is because List files calls itself


robotci:
Admin, can CI be enabled?

pekkaklarck:
Good catch! Nice to also see someone is really studying these examples. :)

PR #2353, opened 2016-04-13
Telnet.py: Add missing shutdown()

jonivanhamaki:
Telnet connections might be left open. Detected with Ubuntu 14.04 and is most likely valid for other distributions as well.

Cannot provide test case for this at the moment.

### DIFF
--- a/src/robot/libraries/Telnet.py
+++ b/src/robot/libraries/Telnet.py
@@ -14,6 +14,7 @@
 
 from contextlib import contextmanager
 import inspect
+import socket
 import re
 import struct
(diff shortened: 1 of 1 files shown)


src/robot/libraries/Telnet.py
```diff
@@ -14,6 +14,7 @@
 
 from contextlib import contextmanager
 import inspect
+import socket
```
pekkaklarck: Nitpick: Imports are ordered alphabetically. This import should be before `import struct`.

PR #2368, opened 2016-05-09
Adding working Docker build for the robot framework

danielwhatmuff:
This Docker build is based on Alpine Linux and is just 64MB. Found it useful when running tests from Jenkins, also providing an image for new Test Engineers means consistent execution across Windows/Mac etc.

REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
robot-docker        latest              42837806b30a        45 minutes ago      64.7 MB

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM gliderlabs/alpine:3.3
+
+MAINTAINER "Daniel Whatmuff" <danielwhatmuff@gmail.com>
+
+LABEL name="Docker image for the Robot Framework http://robotframework.org/"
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Is there a reason this should be part of the main project? What benefits would it have compared to hosting it as a separate independent project?

stasiek:
We've got something similar made with Vagrant, but for Robot Framework development.
@danielwhatmuff Main repo isn't a place for such generic container addition.
First thing is this container holds only robot and selenium 2. We got tens of thousands of users doing test automation other than web pages.
Secondly shouldn't this go on Docker Hub?

pekkaklarck:
It seems this should really be a separate project instead. Closing.

PR #2370, opened 2016-05-09
Better sort ordering in Log and Report, (Tags, Test Suites, Test Cases)

### LINKED ISSUE #2369: [rebot] [enhancement] Improved sorting of in Statistics tables, Log and Report
Attached are log.html and report.html, with the following changes in the Statistics Tables:

- It is now possible to sort in "natural" order instead of numeric (alphanumeric). 
- Clicking in the table headings selects Ascending, Descending or Initial sort order.
[new_tag_sort.zip](https://github.com/robotframework/robotframework/files/255232/new_tag_sort.zip)

Uses original TableSorter 2.7.8 but also works with newest, 2.26.0 (should it be upgraded?).

See similar functionality in [RIDE/#1579](https://github.com/robotframework/RIDE/issues/1579)

HelioGuilherme66:
Changes TableSorter to sort in "natural" order instead of numeric (alphanumeric).

Clicking in the table headings selects Ascending, Descending or Initial sort order.

Uses original TableSorter 2.7.8 but also works with newest, 2.26.0.

Closes #2369

### DIFF
--- a/src/robot/htmldata/rebot/view.js
+++ b/src/robot/htmldata/rebot/view.js
@@ -97,18 +97,17 @@ function addTooltipsToElapsedTimes() {
 function enableStatisticsSorter() {
     $.tablesorter.addParser({
         id: 'statName',
-        type: 'numeric',
-        is: function(s) {
-            return false;  // do not auto-detect
-        },
-        format: function(string, table, cell, cellIndex) {
-            // Rows have class in format 'row-<index>'.
-            var index = $(cell).parent().attr('class').substring(4);
-            return parseInt(index);
+        type: 'text',
+        format: function(s, table, cell, cellIndex) {
+            // Ignore Case
+            return s.toLowerCase();
         }
     });
+    //$.tablesorter.defaults.sortList = [[0,0]];
     $(".statistics").tablesorter({
         sortInitialOrder: 'desc',
+        sortReset: false,
+        sortRestart: false,
         headers: {0: {sorter:'statName', sortInitialOrder: 'asc'},
                   5: {sorter: false}}
     });


pekkaklarck:
Could you clarify how functionality is actually changed?

HelioGuilherme66:
The content of view.js is included in both generated log.html and report.html. All tables with the class "statistics" will use the [TableSorter fork](https://github.com/Mottie/tablesorter) with the parser 'statName'.
The modification was to change the sorting type from 'numeric' to 'text', which by default uses the "natural" sort algorithm. It was also modified the parser to process each cell by using lowercase text.
The added options:
      -  usNumberFormat: false, > _Indicates how tablesorter should deal with a numerical format_  (in this case) | German: 1.234.567,89 | French:   1 234 567,89 |
      -  sortReset: true, > _Setting this option to `true` will allow you to click on the table header a third time to reset the sort_
      -  sortRestart: true, > _Setting this option to `true` will start the sort with the `sortInitialOrder` when clicking on a previously unsorted column._

Images of Tags Statistics table: 
- Initial and ascending order of the official RF 3.0:

original_sort_3 0

- Initial and ascending order with this PR (equal to the official RF 3.0):

original_and_resettable_sort_3dev

- New ordering "Natural" ascending order with this PR (but resettable to the original sort):

ascending_and_resetable_sort_3dev

pekkaklarck:
What's the difference between the current and the "natural" sort order? How does the proposed three state sorting work with numbers?

HelioGuilherme66:
You can try the file [new_tag_sort.zip](https://github.com/robotframework/robotframework/files/255232/new_tag_sort.zip) attached in #2369.
Sample for comparing examples "current" vs "natural" (in **bold** wrong positioned values, in _italic_ new positioned values):
          current         natural
        -----------     -----------
        48-test          48-test
        **4Another_13**      _62-test_
        62-test          _-5-7_
        ^a               ^a
        ab-01+01         ab-01+01
        **ab-08+01**         _ab-1+02_
        **ab-08-09**         _ab-08+01_
        ab-1+02          ab-8+01
        ab-8+01          ab-8-1
        ab-8-1           ab-08-09
        abcd             abcd
        Another_10       _Another_2_
        Another_11       Another_10
        Another_18       Another_11
        **Another_2**        Another_18
        Another_21       Another_21

The three state cycles just the same for Text or Numbers columns:
->Initial -> Natural Ascending -> Natural Descending ->
\<---------<-------------------------<---------------------------<-/

The sorting icons change from:
^ _ ^ _ **^**
v _ **v** _ v

pekkaklarck:
First of all, it would make reviewing issues and pull requests easier if you would introduce the feature/fix a bit more. At least I did't know what term "natural sorting" meant before I read this informative Wikipedia article: https://en.wikipedia.org/wiki/Natural_sort_order

As I commented #2369, I'm in general fine using natural sort order instead of pure alphabetical order. I don't like how this implementation supports both sort orders, though:

- It is confusing that the initial sort order is not natural but clicking the header changes it to natural and later back to the initial order.
- UI doesn't make it clear what order is used.
- In many cases tags contain no multi digit groups and thus order is actually same regardless the ordering. This means that two of the possible three sorting states have actually same end results.

In other words, I'd prefer the initial order to be natural too. That might be best to change on Python side, though, and care must be taken to avoid more complex sorting slowing down anything.

I'd also like all other sorting logs/reports to use natural sorting for consistency. Does tablesorter use natural sorting by default or could it be easily enabled?

Finally, I'm confused what `Closes #2369` means in the description. #2369 is an issue and this is a pull request. We generally prefer having an issue about enhancements and then separate pull request about implementation. That allows closing PRs with implementation doesn't work without affecting the original issue.

HelioGuilherme66:
Squashed and ready.

Now it will appear "natural" sorted by default, and with only two sort states for all sortable columns in Log and Report.

pekkaklarck:
There are lot of problems in the current implementation:

- Critical/All Tests in Total Statistics are in wrong order. Critical Tests are first on purpose.
- Suite Statistics are in wrong order. They should be in execution order, not sorted by name. This is *very* important.
- Using tablesorter to set default sorting slows down opening larger reports and logs. Items should be ordered correctly already on Python side.
- When viewing a suite in Test Details, tests should initially be in execution order.
- Tags listed on per tests should be in natural order for consistency.

I'm actually not sure are changes needed at all on the Javascript side. Sorting tag statistics on Python side (and sorting tags tests have) might be enough.

HelioGuilherme66:
Ready to merge (I think ;)-
>     Critical/All Tests in Total Statistics are in wrong order. Critical Tests are first on purpose.
**Done**
>     Suite Statistics are in wrong order. They should be in execution order, not sorted by name. This is very important.
**Done (not sorted by default)**
>     Using tablesorter to set default sorting slows down opening larger reports and logs. Items should be ordered correctly already on Python side.
**Not Changed (no changes in Python code)**
>     When viewing a suite in Test Details, tests should initially be in execution order.
**Done (not sorted by default)**
>     Tags that tests contain (shown in Test Details and in Log) should also be in natural order for consistency.
**Not Changed (no changes in JS/Python code)** - I think that Tags should be shown in the order they are set in Tests.

pekkaklarck:
There are some inconsistencies here still:

- The initial tag order in statistics is non-natural and turns into natural only if you sort tags. 
- As I already commented, tags tests have are in non-natural sort order. Because tags come from various sources, there's no way to keep them in order they are defined.

I really think sorting should be changed on the Python side and not on the JS side.

pekkaklarck:
Do you @HelioGuilherme66 still have interest to look at this?

HelioGuilherme66:
Yes, I still have interest, but need to study the Python code.
If you have some tips, I'd appreciate. Thanks.

pekkaklarck:
The problem is that items are sorted in many places and if we want to use natural ordering we should try to find all those places and see should they be updated. Basically that involves finding all occurrences of `sorted(x)` and `x.sort()` and studying are they sorting something that we want to be naturally sorted (tests, tags, etc.). Relevant occurrences should then be changed to use `sorted(x, key=natural_sort_key`) or `x.sort(key=natural_sort_key)` instead. Implementing `natural_sort_key` itself shouldn't be too complicated.

Another problem is that using a custom sort key will certainly slow down sorting. I doubt there's big enough difference for that to matter in the real life, but we definitely need to measure it using large data sets.

HelioGuilherme66:
@pekkaklarck There are many implementations of "natural" or "human" sort algorithms. There is this [natsort](https://pypi.python.org/pypi/natsort) very well documented, and addresses performance concerns (by adding other dependencies). Although it need better testing under Python 3.4/3.5, should we consider adding this package?

pekkaklarck:
@HelioGuilherme66 `natsort` looks cool but is too big dependency for Robot. I want to avoid external dependencies in general to keep installation easy in all environments. We also don't need the customization support provided by `natsort`. We can implement our own utility exactly like we want.

As I've already commented, sorting should be done on Python side. I close this PR and we can continue discussion in issue #2369.

PR #2381, opened 2016-05-27
Switch to entry_points in setup.py

alexandrul:


pekkaklarck:
This is a topic we should discuss first. The main issue is compatibility with Jython and IronPython. Do you know are entry points supported by them? Should a separate issue be opened about that discussion so that this PR can be left for discussion about the actual implementation?

alexandrul:
No idea about Jython and IronPython.
IMHO, the actual implementation is pretty simple, so we can continue here with everything.

pekkaklarck:
1. We actually need an issue anyway if we merge this PR. We have a script for creating release note templates based on issues with a certain milestone. Can obviously add an issue before merging with just a reference to this PR.

2. Situation with Jython, IronPython and also PyPy needs to be tested.

3. Could you clarify what actual benefits entry points have compared to the current system?

alexandrul:
3:

- Simplify setup.py
- On windows:
 - native executables (.exe) instead of .bat files
 - when interrupting an execution, I won't be asked if I want to terminate the current script (side effect of using .bat files)

alexandrul:
I have only tested this with Python 2.7.11 and 3.5.1 on Windows.

Is it possible to setup Travis CI for this repo?

alexandrul:
Created an issue for this, which includes a link to valid reasons for this switch: #2415

Issue #2383, opened 2016-05-31
[ ERROR ] Taking listener 'MyListener' into use failed:  Listener 'MyListener' does not have mandatory 'ROBOT_LISTENER_API_VERSION' attribute


Tset-Noitamotua:
Getting an error while running a listener example from [RF Userguide (Version 3.0)](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#listener-examples)

Content of MyListener.py:
```python
import os.path
import tempfile


class PythonListener:
	ROBOT_LISTENER_API_VERSION = 2
	def __init__(self, filename='listen.txt'):
		outpath = os.path.join(tempfile.gettempdir(), filename)
		self.outfile = open(outpath, 'w')

	def start_suite(self, name, attrs):
		self.outfile.write("%s '%s'\n" % (name, attrs['doc']))

	def start_test(self, name, attrs):
		tags = ' '.join(attrs['tags'])
		self.outfile.write("- %s '%s' [ %s ] :: " % (name, attrs['doc'], tags))

	def end_test(self, name, attrs):
		if attrs['status'] == 'PASS':
			self.outfile.write('PASS\n')
		else:
			self.outfile.write('FAIL: %s\n' % attrs['message'])

	def end_suite(self, name, attrs):
		self.outfile.write('%s\n%s\n' % (attrs['status'], attrs['message']))

	def close(self):
		self.outfile.close()
```

Command to start test in TESTS folder:
```shell
robot --listener MyListener TESTS
```

Output:
```shell
$ robot --listener MyListener TESTS
[ ERROR ] Taking listener 'MyListener' into use failed: Listener 'MyListener' does not have mandatory 'ROBOT_LISTENER_API_VERSION' attribute.
==============================================================================
TESTS
...
```

pekkaklarck:
The reason is that you take the module to use as a listener, not the class. Change the module (i.e. file) name to match the class name and Robot ought to realize that you want to use the class as a listener. Alternatively you could import the listener by including both the module and class name in the listener name. Please let us know if you see that related docs could be enhanced somehow.

Tset-Noitamotua:
@pekkaklarck Thank you so much. Once again you brought me back on the right track. Did PR #2388 to enhance related docs.

pekkaklarck:
Cool, thanks for the doc PR. I comment it separately.

PR #2388, opened 2016-06-01
Update ListenerInterface.rst

Tset-Noitamotua:
After running into issue #2383 I suggest this little enhancement of corresponding user guide section.

### DIFF
--- a/doc/userguide/src/ExtendingRobotFramework/ListenerInterface.rst
+++ b/doc/userguide/src/ExtendingRobotFramework/ListenerInterface.rst
@@ -508,6 +508,35 @@ The same example could also be implemented also using the newer
            print 'Test "%s" failed: %s' % (result.name, result.message)
            raw_input('Press enter to continue.')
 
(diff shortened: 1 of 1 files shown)


pekkaklarck:
You are right that docs about taking listeners into use could be enhanced, but I don't think a new listener example is needed. Wouldn't it be enough to just explain, with an example, how to take the next listener example into use? Notice also that there's a separate "Taking listeners into use" section that should probably be referenced from these examples and possibly enhanced too.

Tset-Noitamotua:
- [ ] review PR @me

Tset-Noitamotua:
Is it not possible to assign a task to oneself?

pekkaklarck:
What did you try to do?

Tset-Noitamotua:
@pekkaklarck : I wanted to assign the task 'to review my own PR' to myself here on Github so that I dont forget it later but that is not possible nor is it possible to mark a notification as 'unread' to keep track of it.

Issue #2120, opened 2015-08-26
New `Should Contain Any` and `Should Not Contain Any` keywords


jongu82:
I am searching through a list and would like to pass if any one of my values are found.  It appears that the List Should Contain Value keyword only accepts one value.  I have tried using Value1 or Value2 but it does not appear to be searching for either one.  I would suggest to modify this keyword to accept multiple values separated by 'or'.

Furthermore, I would like to have the output display the matches, and fail if no matches are found

pekkaklarck:
Yes, there doesn't seem to be any keyword that supports this in Collections. We cannot change `List Should Contain Value` because basically any change would be backwards incompatible. Adding separate `List Should Contain Any Value` is possible, but we could also consider adding more generic `Should Contain Any` to BuiltIn. Are you interested to provide a pull request?

jongu82:
Yes I am interested.  I'm fairly new to github and the use of pull requests but I'd be willing to give it a shot.

pekkaklarck:
GitHub has pretty good documentation about pull requests, and our contributing guidelines have links to them along with some other valuable information.

Would you be interested to implement `List Should Contain Any Value` or `Should Contain Any`? I think the latter would be more valuable because it would be more generic. If the former is implemented, then we should add related dict keywords too. Adding all of them is possible too but perhaps too much work compared to benefits. Regardless what is implemented, `Not` version (e.g. `Should Not Contain Any`) should be added too.

pekkaklarck:
After thinking this a bit more, I believe it's best to implement `Should Contain Any` and `Should Not Contain Any`. They are generic and can be used with strings or anything that is list-like. This functionality isn't needed too often, so adding separate `List Should Contain Any Value`, `Dictionary Should Contain Any Key`, etc. wouldn't be worth the effort in my opinion.

Anyone interested to implement `Should (Not) Contain Any`? There's plenty of time until RF 3.0 which is the next planned release and I can help if needed. If you are new to providing pull requests and contributing in general, you can start by reading these guidelines:
https://github.com/robotframework/robotframework/blob/master/CONTRIBUTING.rst

pekkaklarck:
Nobody has shown any interest to implement the new keywords, but I consider them so useful that I plan to include them in RF 3.0.1 anyway. If someone is interested in contributing to RF, this would be a pretty easy issue to get started.

jimni:
I would like to have a go at this issue. 
[...]

PR #2393, opened 2016-06-05
added `Should (Not) Contain Any` keywords #2120

jimni:


src/robot/libraries/BuiltIn.py
```diff
@@ -838,6 +838,62 @@ def should_contain(self, container, item, msg=None, values=True):
...
+
+        Note that possible equal signs in ``*items`` must be escaped
+        with a backslash (e.g. ``name\=value``) to avoid them to be passed in
+        as ``**configuration``.
+
+        Examples:
+        | Should Contain Any | ${some list} | item 1 | item 2 |
+        | Should Contain Any | ${string} | substring | letter | punctuation! |
+        | Should Contain Any | ${container} | item 1 | ... | item n | msg=Keyword Failed | values=False |
+        """
```
pekkaklarck: This kind of note needed:

> New in Robot Framework 3.0.1.

src/robot/libraries/BuiltIn.py
```diff
@@ -838,6 +838,62 @@ def should_contain(self, container, item, msg=None, values=True):
...
+        msg = configuration.pop('msg', None)
+        values = configuration.pop('values', True)
+        if configuration:
+            raise RuntimeError("Unsupported configuration parameters were used: %s."
+                               % unic(seq2str(configuration.keys())))
```
pekkaklarck: Both `unic` and `.keys()` are unnecessary.

src/robot/libraries/BuiltIn.py
```diff
@@ -838,6 +838,62 @@ def should_contain(self, container, item, msg=None, values=True):
...
+
```
pekkaklarck: We prefer not to have empty lines inside functions or methods. #nitpick

atest/testdata/standard_libraries/builtin/verify.robot
```diff
@@ -293,6 +297,44 @@ Should Contain With Non-String Values
     Should Contain    ${dict}    a
     Should Contain    ${dict}    c
 
+Should Contain Any
+    [Documentation]    FAIL Message: 'abcdefg' does not contain any of ''A', 'X' or 'Ф''
```
pekkaklarck: There are extra quotes around the list of items but that's due to the `_get_string_msg` helper unconditionally adding them. I guess the helper could get a new config parameters like `quote_str1=True` and `quote_str2=True` to control that, and the new keywords could call it like `quote_str2=False`. What do you think?

Somewhat related to this, I'd prefer the `_get_string_msg` to add a full stop at the end of the default message. That's totally unrelated to this PR (the change require changing lot of tests) and, unless you want to help with that too, I can take care of that myself.

pekkaklarck:
Looks very good. I added some line notes about things I think should still be looked at. In addition to that, please add yourself to AUTHORS.txt file.

pekkaklarck:
See also our contribution guidelines: https://github.com/robotframework/robotframework/blob/master/CONTRIBUTING.rst

jimni:
Thank you very much for your feedback. I will have time later this week to look into all your comments and modify my fix accordingly. I hope such delay is acceptable for this issue.

pekkaklarck:
No worries. I've been busy with other tasks myself and haven't had time to concentrate on 3.0.1 too much either.

jimni:
I believe I have fixed everything you mentioned in the line notes. As per your helper suggestion, I started another issue #2403, please take a look when it's convenient for you.

Is there anything else you would like me to add to the current PR?

pekkaklarck:
The PR looks very good. I could merge it right away, but it would be a bit better if `_get_string_msg` was enhanced to allow dropping extra quotes first. Squashing the commits when everything is done would be nice too.

jimni:
You want `_get_string_msg` fix to be in this branch, correct?

pekkaklarck:
Depends on the changes to `_get_string_msg`. If you only enhance it to allow dropping quotes, that can be part of this PR. Adding `.` to the end should be a separate PR because it is totally unrelated to this PR and requires updating so many tests. That `.` is *very* low priority anyway.

jimni:
Ready to squash once you accepted the changes to `_get_string_message`. 
Please let me know if the code needs anything else or if my fix doesn't comply with the project's coding practice.

jimni:
Will be mostly unable to contribute for the next couple weeks, sorry for the possible inconvenience.

pekkaklarck:
Been busy with holidays, house moves,and lot of training courses, but now back working with RF 3.0.1. This PR seems to be in a really good shape and ready to be merged. The only issue I saw is that the error message when there are no `*items` is likely not too good. I'll check how it looks in practice and possibly improve it myself.

jimni:
Thanks a lot for all your input. It's been a pleasure working on this fix with you.

Issue #2201, opened 2015-11-11
BuiltIn Library: add "Run Keyword If Keyword" variations


vkruoso:
When user keyword fails and a keyword teardown is implemented, it is necessary to add the following boilerplate code if the teardown wants to run keywords when the user keyword itself has failed.

``` robotframework
*** Keywords ***
Some Keyword
    Fail  Just fails
    [Teardown]  Some Keyword Teardown

Some Keyword Teardown
    Run Keyword If  '${KEYWORD STATUS}' == 'FAIL'
    ...   Log  'The Keyword Failed'
```

It would be nice to have `Run Keyword If Keyword Failed` and `Run Keyword If Keyword Passed` keywords to handle that case.

pekkaklarck:
Sounds good to me. Have you already looked at implementation?

vkruoso:
Not really. Was waiting on the feedback, but I won't be able to look into it in the short term tough. It does not sound hard as there is implementation on the test case level. Will try to give it a try this weekend.

pekkaklarck:
Cool. There's still good time to get this into RF 3.0 that we ought to get ready by the end of the year. Next preview release is planned for next week or week after.

This is related to #2078 about making status variables and passed/failed keywords to be available also outside teardowns. The new proposed keywords could be used, for example, after a continuable failure. No need to take that into account at this point, though.

pekkaklarck:
Have you @vkruoso had time to look at this?

vkruoso:
Sorry @pekkaklarck I did not. Will try to take a look at this soon, but no say it for sure when it will be.

pekkaklarck:
We are going to create RF 3.0 in few days and do not plan to do any more code changes to it. Depending on the implementation, this can be considered in 3.0.x or latest in 3.1.

pekkaklarck:
We are staring with RF 3.0.1 development with a target to get it released in June. Do you have time and interest to get this one included? Or should we just close the whole issue?

vkruoso:
I'll take a look at implementing this in until the end of next week. I think those keywords can be very useful. Sorry about the delay.

pekkaklarck:
This is a bit too big change for a minor release. Targeting for 3.1.

pekkaklarck:
RF 3.1 getting ready and need to make decision related to this issue.

Reasons I think this issue should still be included:
+ Functionality is consistent with existing `Run Keyword If Test Failed/Passed`.
[...]

PR #2396, opened 2016-06-10
BuiltIn: Created "Run Keyword If Keyword {Failed,Passed}" keywords

vkruoso:
That is a incomplete PR to issue #2201. There is still some work to be done regarding testing.


vkruoso:
I've added some tests by one is not passing. This is due to the fact that when executing the keyword teardown the keywords status for PASS or FAIL is no correctly set. I could no find the area to make that change, can you provide some guidelines?

vkruoso:
@pekkaklarck do you have any thoughts on that? Is it acceptable to use the ``KEYWORD STATUS`` variable?

vkruoso:
@pekkaklarck Any thoughts on that?

vkruoso:
Closing this due to the amount of time without feedback. I'm also not using the framework so actively now. Thanks for the great work!

Issue #1966, opened 2015-04-11
Inverted --suite and --test flags to exclude tests from being run


bjornfor:
(This is a feature request.)

I'm looking at the various flags to include/exclude test cases to run:

```
--suite (includes by suite name)
--test (includes by test name)
--include (includes by tag name)
--exclude (excludes by tag name)
```

It seems that tests can only be _excluded_ by tag. I for one would have use for being able to exclude by suite name. And possibly test name too.

It feels a bit odd that we're able to include tests in 3 "dimmensions" (suite, test, tag), but can only exclude in 1 of them (tag).

pekkaklarck:
The main reason the are no negative versions of --test and --suite options is that nobody has needed them bad enough to implement them. It's more common to unselect a group of tests and tags with --exclude work great for that. If someone would provide a pull request with implementation, tests, and documentation, I would be happy go merge it, though.

pekkaklarck:
There doesn't seem to be interest for this. Closing.

PR #2402, opened 2016-06-28
Added a --removetestcases command line argument. 

snyderr:
This keyword behaves similar to the `--removekeywords` command line argument. 

--removetestcases all|passed|name:<pattern>|tag:<pattern> *
                          Remove test case data from all generated outputs.
                          Test cases containing warnings are not removed except
                          in `all` mode.
                          all:     remove data from all test cases
                          passed:  remove data only from test cases in passed
                                   test cases and suites
                          name:<pattern>:  remove data from test cases that match
                                   the given pattern. The pattern is matched
                                   against the full name of the test case (e.g.
                                   'MyLib.Keyword', 'resource.Second Keyword'),
                                   is case, space, and underscore insensitive,
                                   and may contain `*` and `?` as wildcards.
                                   Examples: --removetestcases name:Lib.HugeKw
                                             --removetestcases name:myresource.*
                          tag:<pattern>:  remove data from test cases that match
                                   the given pattern. Tags are case and space
                                   insensitive and it is possible to use
                                   patterns with `*` and `?` as wildcards.
                                   Tags and patterns can also be combined
                                   together with `AND`, `OR`, and `NOT`
                                   operators.
                                   Examples: --removetestcases foo
                                             --removetestcases fooANDbar*


pekkaklarck:
I understand the need to remove tests but I'm not sure is this the best way. My reasoning:

- Removing tests by tags is already possible using `--exclude` (and `--include`). I don't think it is a good idea to add another mechanism for exactly same purpose.
- I don't see any reason why anyone would remove all tests.
- I don't really see use cases for removing all passed or failed tests either.
- Removing by name would be a useful feature and it has been requested earlier (e.g. #1966). I'd prefer it getting a separate option like `--notest`, though.

On a very quick look the PR itself looks really good.

snyderr:
@pekkaklarck Thanks for the quick response.

When originally designing this I wanted to make it as much like `--removekeywords` as possible. On that note I completely agree with you that there would be no reason to remove all of the test cases. 

If I am understanding how `--removekeywords` works is that when passed as a command line argument for `robot/pybot` all test cases in the suite will still be ran and during the creation of the report/log the the keywords will then be removed. So if I am correct I would consider this different than using `--exclude` because the test cases will still be ran but removed from the report as opposed to not being run at all (I think this has a better use case with `rebot` rather than during runtime with `pybot/robot`). 

The ability to remove passing test cases was one of the main reasons that I decided to create this pull request. On my company's current setup we have well over a thousand different test suites that will run during a full regression. On an average day the majority of these tests will pass, and as you could imagine the log can get pretty big in size if you have a lot of tests to report. By removing all of the passed test cases the size of the logs shrunk significantly and also made reading through the logs a much better experience when you're only looking at the failings tests. Another use case that we found for this keyword now that we know all of the tests that are failing we can look through the logs and remove unnecessary screen shots that may have been saved during testing.

pekkaklarck:
`--removekeywords` naturally can remove executed keywords only after execution, but it typically makes more sense to remove tests already before execution. That's also how `--suite`, `--test`, `--include` and `--exclude` options work nowadays. Rebot can be used if you want to create reports with only a subset of tests afterwards.

As I already commented, I'd be fine adding new `--nosuite` and `--notest` options for dropping out named suites or tests.

Your use case for removing passed tests is very good and I actually think we've been discussing also that earlier. One way to implement it would be enhancing `--include` and `--exclude` to accept statuses like `--include STATUS:FAIL` when used with Rebot.

Finally, `--splitlog` and `--prerebotmodifier` options might work pretty well in your setup.

snyderr:
@pekkaklarck `--prerebotmodifier` was exactly what I needed, thank you for the help!

pekkaklarck:
Yeah, pre run/rebot modifiers are *very* handy in many advanced and/or special cases. Common cases are best handled with dedicated options, and I think both removing named suites/tests and removing all passed tests are use cases that would benefit from them. I don't think the proposed `--removetestcases` solution is the best approach, though.

I guess all this means that this PR should be closed. Do you @snyderr still have interest to implement other dedicated solutions that would solve your issues? I think `--notest` and `--exclude STATUS:PASS` would both be nice.

snyderr:
@pekkaklarck 
This is a feature that I would like to see implemented in RobotFramework. As of right now I have enough work ahead of me that it wouldn’t make sense for me to take this on at this time. If anyone else doesn’t pick this up I am all for doing this when my workload clears up.

pekkaklarck:
Sounds good. I guess we should nevertheless close this PR and submit issues (or reopen existing) about excluding tests (and suites) by name and by status. Would `--notest/--nosuite` and `--include/--exclude STATUS:PASS/FAIL` I've proposed earlier work in your context or do you see better ways to accomplish them? If you are fine with those proposals, I submit/reopen issues related to them.

snyderr:
@pekkaklarck I believe that those soultion would work but I was able to solve my problem by using some of the code I wrote in `testcaseremover.py` and making it into a prerebomodifier.
The code below will remove all passing test cases.

```
from robot.api import SuiteVisitor

class RemovePassingTestCases(SuiteVisitor):
    _message = 'Testcase data removed'

    def __init__(self):
        self._removal_message = RemovalMessage(self._message)

    def _clear_content(self, tc):
        tc.keywords = []
        tc.message = []
        self._removal_message.set(tc)

    def _failed_or_warning_or_error(self, item):
        return not item.passed or self._warning_or_error(item)

    def _warning_or_error(self, item):
        finder = WarningAndErrorFinder()
        item.visit(finder)
        return finder.found

    def visit_suite(self, suite):
        for test in suite.tests:
            if not self._failed_or_warning_or_error(test):
                self._clear_content(test)

class WarningAndErrorFinder(SuiteVisitor):

    def __init__(self):
        self.found = False

    def start_suite(self, suite):
        return not self.found

    def start_test(self, test):
        return not self.found

    def start_keyword(self, keyword):
        return not self.found

    def visit_message(self, msg):
        if msg.level in ('WARN', 'ERROR'):
            self.found = True

class RemovalMessage(object):

    def __init__(self, message):
        self._message = message

    def set_if_removed(self, kw, len_before):
        removed = len_before - len(kw.keywords)
        if removed:
            self.set(kw, self._message % (removed, plural_or_not(removed)))

    def set(self, kw, message=None):
        kw.doc = ('%s\n\n_%s_' % (kw.doc, message or self._message)).strip()
```

snyderr:
I would also like to add that I think adding `--include/--exclude STATUS:PASS/FAIL` might not be the best approach. If I am understanding how you would like that to work, `--include` would now sometimes work with `rebot` and `pybot/robot` but in this new case would only work with `rebot`. Seems to me to be kind of inconsistent.

Issue #2346, opened 2016-03-23
Feedback needed on a new design for reports and logs created by rebot


akshbn:
It is time the reports and logs created by rebot get a makeover. I have made a few tweaks to rebot in this regard. I have attached the screenshots of a report and log created by rebot for one of my test suites. I request the community to give me feedback on this. If a majority of people like it i will send a pull request.

Like you may have noticed, its based on the material design philosophy.

![success](https://cloud.githubusercontent.com/assets/13907699/13977624/bbafb70a-f0f0-11e5-8df0-c4a3b6f95a88.PNG)
Report when all test cases pass.

![error](https://cloud.githubusercontent.com/assets/13907699/13977647/f43eb85a-f0f0-11e5-9a29-9b1101ecc62c.PNG)
Report  when test cases fail.

![log](https://cloud.githubusercontent.com/assets/13907699/13977665/11c99282-f0f1-11e5-8f86-b30756016581.PNG)
Log

akshbn:
#2281

Tattoo:
Throwing my two cents in, have you tested your designs with huge (ie. maybe even 1-2GB) [as this has been problem previously](https://github.com/robotframework/robotframework/issues/360)?

akshbn:
@Tattoo I havent tested it with such huge log files. I feel it will not affect user experience at all because all changes made are pure css. So the experience will stay consistent with the existing log.html, ie, if there is a noticeable lag with the existing log.html, there maybe some lag in my version of it as well.

pekkaklarck:
Looks pretty nice. I'd prefer having all info visible in one view without tabs, though. Final versions should also be fine tuned so also color blind people can separate red and green.

It would be great if you could make your changes available as a PR or otherwise so that others can test them with their own data.

pekkaklarck:
Any updates to this?

akshbn:
@pekkaklarck  Will drop a PR by the end of June.

pekkaklarck:
Cool! With bigger changes we need wider discussing with the community, but a PR is a great way to get started. It allows both reviewing the changes and running the code to test them for real.

zastress:
Any update on this issue? A new interface change on the report page might be cool with the new release.

pekkaklarck:
[...]

PR #2412, opened 2016-07-08
Redesign of Log and Report UI

akshbn:
I have designed the logs and report UI as per material design. #2346


pekkaklarck:
Unfortunately this PR wasn't reviewed back when it was created. Most likely that has been due to lack of time and the PR introducing lot of changes. Nowadays there are so many conflicts due to changes like added SKIP and removed criticality that review is pretty much impossible and I can only close this PR.

The proposed changes to logs/reports looked really nice and something like that could be done in the future. This kind of changes are so big that it would be best do discuss them first on our Slack or elsewhere.

PR #2413, opened 2016-07-11
Add ":IF" identifier to run a conditional block logic, just like ":FOR".

jnhyperion:
When we have a branch logic in our cases. I usually use "run keyword if". But when when branch logic is pretty long, I have to use a lot of "run keyword if". So I add this new feature to make cases look more logical and beautiful.


jnhyperion:
Use like this:
image

Here's the test report:

image

joaonc:
Would **love** for this to get merged.

pekkaklarck:
IF/ELSE support will be added in RF 4.0. The new parser introduced in RF 3.2 has made this PR totally invalid.

PR #2416, opened 2016-07-16
added some prerun modifier examples to the documentation that cover s…

jaloren:
…some common use cases

modified prerun modifier code example so that setup and teardown are set directly to none

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ dist
 MANIFEST
 atest/results
 utest/jasmine-results
+*~
 .idea
 .*project
 *.pyc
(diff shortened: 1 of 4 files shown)


jaloren:
@pekkaklarck  this should be reopened and merged in. I didn't mean to close the issue.

pekkaklarck:
It seems this PR cannot be reopened because you have removed the branch its based on.

jaloren:
I apologize for being late to address this and that i opened a separate PR but I figured it was just easiest to create a new branch. Please let me know if there are any other issues.

PR #2418, opened 2016-07-18
String and OS: adding string strip and generate uuid keywords

kkhalasi:


src/robot/libraries/String.py
```diff
@@ -462,6 +462,24 @@ def split_string_to_characters(self, string):
...
+        return string.split(delimiter)
+
+    def remove_leading_spaces(self, string):
+        """Removes any and only leading spaces in a string. Does not remove any spaces in between strings such as "Hello Hello"
+
+        Example:
+        | ${removed_leading_spaces}= | Remove Leading Spaces | ${string}
+
+        """
+        return string.split(string)
```
bkhouri: The docstring does not match what the implementation does.

This code will basically return a list with two empty items in it and not remove the leading whitespace characters.

```
>>> f = "   \t   foo   "
>>> f.split(f)
['', '']
```

As a result, I think this should be `return string.lstrip()`

```
>>> f = "   \t   foo   "
>>> f.lstrip()
'foo   '
>>> f = "d   \t   foo   "
>>> f.lstrip()
'd   \t   foo   '
```

src/robot/libraries/OperatingSystem.py
```diff
@@ -1052,6 +1052,15 @@ def log_environment_variables(self, level='INFO'):
...
         return variables
 
+    def generate_uuid(self):
+        """Generates a random UUID.
+
+        Example:
+        | ${uuid} = | Generate UUID
+
+        """
+        return str(uuid.uuid4())
```
bkhouri: This will likely fail as the `uuid` module does not appear to be imported.

PR #2423, opened 2016-08-04
De-flake test_get_current_timetuple_excluding_millis

philsc:
Looking at the source for Python 3.4, it appears that time.localtime()
is implemented such that it calls time(2) while time.time() calls
clock_gettime(2).

The phenomenon is such that performing the following reads:
1. time.localtime()
2. time.time()  - inside of _get_timetuple()
3. time.localtime()
   the timestamps 1. and 3. return a time.struct_time with the tm_sec
   field set to X. Occasionally, timestamp 2. returns a value of X+1.
   This only appears to happen on a second boundary. The fraction
   component of time.time() is in the sub-millisecond range.

Forcing all three timestamps to be sourced in the same way seems to
fix the problem.

On those occasional failures, I see the following printed:
AssertionError: (2016, 3, 13, 10, 40, 21) != (2016, 3, 13, 10, 40, 20)

### DIFF
--- a/utest/utils/test_robottime.py
+++ b/utest/utils/test_robottime.py
@@ -22,10 +22,10 @@ class TestTime(unittest.TestCase):
 
     def test_get_current_timetuple_excluding_millis(self):
         while True:
-            expected = time.localtime()
+            expected = time.localtime(time.time())
             actual = _get_timetuple()
             # make sure got same times and _get_timetuple() did not round millis
-            if expected == time.localtime() and actual[-1] > 0:
+            if expected == time.localtime(time.time()) and actual[-1] > 0:
                 break
         assert_equal(actual[:-1], expected[:6])
 


philsc:
Simple script to reproduce the failure (I grabbed snippets throughout the code base):

```Python
#!/usr/bin/python3

import time

IRONPYTHON = False

def assert_equal(actual, expected):
    if (actual != expected):
        raise Exception(str(actual) + ' != ' + str(expected))

def _get_timetuple(epoch_secs=None):
    if epoch_secs is None:  # can also be 0 (at least in unit tests)
        epoch_secs = time.time()
    secs, millis = _float_secs_to_secs_and_millis(epoch_secs)
    timetuple = time.localtime(secs)[:6]  # from year to secs
    return timetuple + (millis,)

def _float_secs_to_secs_and_millis(secs):
    isecs = int(secs)
    millis = roundup((secs - isecs) * 1000)
    return (isecs, millis) if millis < 1000 else (isecs+1, 0)

def roundup(number, ndigits=0, return_type=None):
    """Rounds number to the given number of digits.

    Numbers equally close to a certain precision are always rounded away from
    zero. By default return value is float when ``ndigits`` is positive and
    int otherwise, but that can be controlled with ``return_type``.

    With the built-in ``round()`` rounding equally close numbers as well as
    the return type depends on the Python version.
    """
    sign = 1 if number >= 0 else -1
    precision = 10 ** (-1 * ndigits)
    if not return_type:
        return_type = float if ndigits > 0 else int
    quotient, remainder = divmod(abs(number), precision)
    # https://github.com/IronLanguages/main/issues/1236
    if (not (IRONPYTHON and (quotient * precision + remainder > abs(number)))
        and remainder >= precision / 2):
        quotient += 1
    return sign * return_type(quotient * precision)


def test_get_current_timetuple_excluding_millis():
    while True:
        expected = time.localtime()
        actual = _get_timetuple()
        # make sure got same times and _get_timetuple() did not round
        # millis
        if expected == time.localtime() and actual[-1] > 0:
            break

    assert_equal(actual[:-1], expected[:6])

if __name__ == '__main__':
    test_get_current_timetuple_excluding_millis()
```

If you run this in a busy loop then I can reproduce the failure pretty quickly:
```ShellSession
$ while ./foo_test.py; do :; done
  File "./foo_test.py", line 57, in <module>
    test_get_current_timetuple_excluding_millis()
  File "./foo_test.py", line 54, in test_get_current_timetuple_excluding_millis
    assert_equal(actual[:-1], expected[:6])
  File "./foo_test.py", line 9, in assert_equal
    raise Exception(str(actual) + ' != ' + str(expected))
Exception: (2016, 8, 4, 23, 29, 55) != (2016, 8, 4, 23, 29, 54)
```

If you add a call to `time.time()` into `time.localtime()` then the failure goes away.

philsc:
I should mention that I tested this on Debian jessie 64bit with Python 3.4.

pekkaklarck:
Looks good to me. Sorry for missing this earlier.

PR #2425, opened 2016-08-09
fix For-in-range loop Documentation in AdvancedFeatures.rst

caimaoy:
The Documentation of the case, Float parameters should be "Loops over values 3.14, 4.34, and 5.54".


pekkaklarck:
*Very* good catch. Nice to see examples are studied in this detail. =)

PR #2427, opened 2016-08-15
add Simplified Chinese prefix of Gherkin language

mebusw:


pekkaklarck:
Support for Given/When/Then prefixes using other languages than English was added in RF 6.0 (#519) along with other translation related enhancements. This PR is thus outdated.

PR #2444, opened 2016-09-19
Updated documentation on Varargs with user keywords, added testcase example

StefanIGit:
to clarify this "at the end the list variable gets all the leftover
arguments that do not match
other arguments."

You do not need the crate a list for multiple arguments.

### DIFF
--- a/doc/userguide/src/CreatingTestData/CreatingUserKeywords.rst
+++ b/doc/userguide/src/CreatingTestData/CreatingUserKeywords.rst
@@ -259,6 +259,20 @@ at the end the list variable gets all the leftover arguments that do not match
 other arguments. The list variable can thus have any number of items, even zero.
 
 .. sourcecode:: robotframework
+    *** Test Cases ***
+    Varargs with user keywords
+       @{argList}    Create List        argOne    argTwo    argThree   argFour    argTen
+       Any Number Of Arguments
+       Any Number Of Arguments    argOne
+       Any Number Of Arguments    argOne    argTwo
+       Any Number Of Arguments    argOne    argTwo    argThree   argFour    argTen
+       Any Number Of Arguments    @{argList}
+       One Or More Arguments    justOneArg
+       One Or More Arguments    argOne    argTwo    argThree   argFour    argTen
+       One Or More Arguments    oneArgAnd    @{argList}
+       Required, Default, Varargs    rqArg    @{argList}
+       Required, Default, Varargs    rqArg    ${opt}=optional    @{argList}
+       Required, Default, Varargs    rqArg    argOne    argTwo    argThree   argFour    argTen
 
    *** Keywords ***
    Any Number Of Arguments
--- a/doc/userguide/src/CreatingTestData/CreatingUserKeywords.rst
+++ b/doc/userguide/src/CreatingTestData/CreatingUserKeywords.rst
@@ -259,20 +259,6 @@ at the end the list variable gets all the leftover arguments that do not match
 other arguments. The list variable can thus have any number of items, even zero.
 
 .. sourcecode:: robotframework
-    *** Test Cases ***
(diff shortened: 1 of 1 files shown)


pekkaklarck:
It's a good idea to add an example test to illustrate how these keywords can be used, but I'd rather have it after the keyword definitions. There could be a small paragraph with text such as "The keywords in the examples above could be used, for example, like this:" and then the example itself afterwards. What do you think?

In addition to the above, I'd like some changes to be done to the example itself. I'll add line notes to the example separately.

doc/userguide/src/CreatingTestData/CreatingUserKeywords.rst
```diff
@@ -259,6 +259,20 @@ at the end the list variable gets all the leftover arguments that do not match
 other arguments. The list variable can thus have any number of items, even zero.
 
 .. sourcecode:: robotframework
+    *** Test Cases ***
+    Varargs with user keywords
+       @{argList}    Create List        argOne    argTwo    argThree   argFour    argTen
```
pekkaklarck: We haven't used `camelCase` format with arguments or other code (except for Java examples) in the User Guide. In here `@{argList}` could be `@{arg list}` or preferably just `@{args}`. With the actual arguments, you could use `arg1` or `arg one`.

We've also used the optional `=` when assigning variables in the examples. In other words, this line should start with `@{args} =`.

Finally, examples should use four spaces as a separator consistently unless arguments are explicitly aligned.

doc/userguide/src/CreatingTestData/CreatingUserKeywords.rst
```diff
@@ -259,6 +259,20 @@ at the end the list variable gets all the leftover arguments that do not match
...
+       Any Number Of Arguments
+       Any Number Of Arguments    argOne
+       Any Number Of Arguments    argOne    argTwo
+       Any Number Of Arguments    argOne    argTwo    argThree   argFour    argTen
+       Any Number Of Arguments    @{argList}
+       One Or More Arguments    justOneArg
+       One Or More Arguments    argOne    argTwo    argThree   argFour    argTen
+       One Or More Arguments    oneArgAnd    @{argList}
+       Required, Default, Varargs    rqArg    @{argList}
+       Required, Default, Varargs    rqArg    ${opt}=optional    @{argList}
```
pekkaklarck: `${opt}=optional` doesn't work, it should be `opt=optional`. I'm not sure is this example really needed when discussing about varargs, though.

doc/userguide/src/CreatingTestData/CreatingUserKeywords.rst
```diff
@@ -259,6 +259,20 @@ at the end the list variable gets all the leftover arguments that do not match
...
+       @{argList}    Create List        argOne    argTwo    argThree   argFour    argTen
+       Any Number Of Arguments
+       Any Number Of Arguments    argOne
+       Any Number Of Arguments    argOne    argTwo
+       Any Number Of Arguments    argOne    argTwo    argThree   argFour    argTen
+       Any Number Of Arguments    @{argList}
+       One Or More Arguments    justOneArg
+       One Or More Arguments    argOne    argTwo    argThree   argFour    argTen
+       One Or More Arguments    oneArgAnd    @{argList}
+       Required, Default, Varargs    rqArg    @{argList}
```
pekkaklarck: If the above example is included, there should be also this kind of example:

    Required, Default, Varargs    @{args}

StefanIGit:
I think I committed the changes, still getting to know how to use git.

pekkaklarck:
Sorry for forgetting about this. The example looks pretty good now. I think it is a bit long and some of the combinations could be removed. It also seems to have three space indentation when it should have four. I can easily fix these myself. Thanks for the PR!

PR #2447, opened 2016-09-22
First 2 keywords for case-insensitive string comparison

chriscallan:
…rings" perform string compares case-insensitively; (#1)

Modified the Acceptance tests affected by the above 2 keywords;
Added an Acceptance test to prove out the case-insensitive string comparison;

### DIFF
--- a/atest/testdata/standard_libraries/builtin/verify.robot
+++ b/atest/testdata/standard_libraries/builtin/verify.robot
@@ -68,15 +68,16 @@ Should Not Be Equal
     ${STR1}    ${INT1}
     ${STR1}    1
 
+Should Not Be Equal With Case Insensitivity
+    [Template]      Should Not Be Equal
+    test value      TEST VALUE1      ignore_case=True
+    HYVÄÄ YÖTÄ      hyvää yötä1     ignore_case=True
+
 Should Not Be Equal with bytes containing non-ascii characters
     [Documentation]    FAIL ${BYTES WITH NON ASCII} == ${BYTES WITH NON ASCII}
     Should Not Be Equal    ${BYTES WITH NON ASCII}    ${BYTES WITHOUT NON ASCII}
     Should Not Be Equal    ${BYTES WITH NON ASCII}    unicode
     Should Not Be Equal    ${BYTES WITH NON ASCII}    ${BYTES WITH NON ASCII}
-
-Should Be Equal With Case Insensitivity
-    Should Be Equal     "test value"     "TEST VALUE"     ignore_case=True
-    Should Be Equal As Strings  "test value"     "TEST VALUE"     ignore_case=True
 
 Should Be Equal
     [Documentation]    FAIL Error message: A != B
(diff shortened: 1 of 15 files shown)


atest/testdata/standard_libraries/builtin/verify.robot
```diff
@@ -74,6 +74,10 @@ Should Not Be Equal with bytes containing non-ascii characters
     Should Not Be Equal    ${BYTES WITH NON ASCII}    unicode
     Should Not Be Equal    ${BYTES WITH NON ASCII}    ${BYTES WITH NON ASCII}
 
+Should Be Equal With Case Insensitivity
+    Should Be Equal     "test value"     "TEST VALUE"     ignore_case=True
+    Should Be Equal As Strings  "test value"     "TEST VALUE"     ignore_case=True
```
pekkaklarck: - Quotes in the verified text are not necessary and mainly confuse reader.
- I'd prefer having case-insensitive test after normal Should Be Equal test. 
- Should test also with non-ASCII characters. For example that `HYVÄÄ YÖTÄ` (good night in Finnish) equals `hyvää yötä` case-insensitively. Could consider using `[Template]` like many other tests here.
- Should Be Equal As Strings should get its own test after normal Should Be Equal As Strings test.

src/robot/libraries/BuiltIn.py
```diff
@@ -613,9 +613,16 @@ def should_be_equal(self, first, second, msg=None, values=True):
         for example, string ``false`` or ``no values``. See `Boolean arguments`
         section for more details.
 
+        ``ignore_case`` is False by default.  It is a boolean value, and if True
+        indicates that 'first' and 'second' should be compared case-insensitively,
+        provided that 'first' and 'second' are string types.
```
pekkaklarck: - Docs should refer to `Boolean arguments` section about possible values for `ignore_case`. You can check how other keywords here do it. Values should also be handled according to that section, but more about that below.
- Should use double-backticks style also with arguments `first` and `second`.
- Should add a note that this is a new feature in Robot Framework 3.0.1.

src/robot/libraries/BuiltIn.py
```diff
@@ -613,9 +613,16 @@ def should_be_equal(self, first, second, msg=None, values=True):
...
+
         If both arguments are multiline strings, the comparison is done using
         `multiline string comparisons`.
         """
+        if ignore_case and is_string(first) and is_string(second):
```
pekkaklarck: Should use `is_truthy(ignore_case)` similarly as other keywords accepting Boolean values. See also the previous comment about docs referring to `Boolean arguments` section.

src/robot/libraries/BuiltIn.py
```diff
@@ -613,9 +613,16 @@ def should_be_equal(self, first, second, msg=None, values=True):
...
+            first = str(first).lower()
+            second = str(second).lower()
```
pekkaklarck: Must not use `str()` here because it won't handle non-ASCII characters in Python 2. If we know values are strings, we can simply do `first = first.lower()`.  Adding a test for non-ASCII stuff  like I proposed in another comment is a good idea.

PR #2450, opened 2016-09-23
Case insensitive String comparisons

chriscallan:
Added input parameter to all the keywords that commonly compare strings, and could use that feature; 
Modified the function logic to trigger off the input parameter and compare strings case-insensitively;


src/robot/libraries/BuiltIn.py
```diff
@@ -745,87 +761,154 @@ def should_be_equal_as_numbers(self, first, second, msg=None, values=True,
...
+    def should_not_contain(self, container, item, msg=None, values=True, ignore_case=False):
         """Fails if ``container`` contains ``item`` one or more times.
 
         Works with strings, lists, and anything that supports Python's ``in``
         operator. See `Should Be Equal` for an explanation on how to override
         the default error message with ``msg`` and ``values``.
 
+        ``ignore_case`` is False by default.  If True, it indicates that
+        ``container`` and ``item`` should be compared case-insensitively,
+        provided that ``container`` and ``item`` are string types.  See `Boolean
```
pekkaklarck: If a container is list or list-like, shouldn't we actually convert items in it into lower case? More about that in a separate comment below, but most likely these docs need to be updated too.

src/robot/libraries/BuiltIn.py
```diff
@@ -745,87 +761,154 @@ def should_be_equal_as_numbers(self, first, second, msg=None, values=True,
...
         | Should Not Contain | ${output}    | FAILED |
         | Should Not Contain | ${some list} | value  |
         """
-        if item in container:
+        prv_container = container
+        prv_item = item
+        if is_truthy(ignore_case) and is_string(container) and is_string(item):
+            prv_container = container.lower()
+            prv_item = item.lower()
+        if prv_item in prv_container:
```
pekkaklarck: As I already wrote in the previous comment, we should also support a case that the container is a list or list-like. In my opinion we should convert all strings in the container to lower case first. Something like this:

```python
tested_container = container
if is_truthy(ignore_case) and is_string(item):
    item = item.lower()
    if is_list_like(container):    # needs to be imported from `robot.utils`
        tested_container = set([c.lower() if is_string(c) else c for c in container])
if item in test_container:
    # ...
```

The above implementation converts all list-like containers (including dictionaries) to sets (could also use lists, and could use set literals if we still didn't need to support Python 2.6). Need to preserve the original container too to be able to show the original structure in possible error message.

In theory `item` could also be a list of strings and `container` could be a nested list of strings, but recursively lowercasing everything is probably not a good idea. It would add quite a bit of complexity, including making sure we don't change the type of the `item`.

src/robot/libraries/BuiltIn.py
```diff
@@ -745,87 +761,154 @@ def should_be_equal_as_numbers(self, first, second, msg=None, values=True,
...
             raise AssertionError(self._get_string_msg(container, item, msg,
                                                       values, 'contains'))
 
-    def should_contain(self, container, item, msg=None, values=True):
+    def should_contain(self, container, item, msg=None, values=True, ignore_case=False):
         """Fails if ``container`` does not contain ``item`` one or more times.
 
+        ``ignore_case`` is False by default.  If True, it indicates that
+        ``container`` and ``item`` should be compared case-insensitively,
+        provided that ``container`` and ``item`` are string types.  See `Boolean
```
pekkaklarck: Same discussion about docs and implementation as with `should_not_contain` applies also to this keyword.

src/robot/libraries/BuiltIn.py
```diff
@@ -834,22 +917,33 @@ def should_contain(self, container, item, msg=None, values=True):
...
+    def should_contain_x_times(self, item1, item2, count, msg=None, ignore_case=False):
         """Fails if ``item1`` does not contain ``item2`` ``count`` times.
 
         Works with strings, lists and all objects that `Get Count` works
         with. The default error message can be overridden with ``msg`` and
         the actual count is always logged.
 
+        ``ignore_case`` is False by default.  If True, it indicates that
+        ``item1`` and ``item2`` should be compared case-insensitively,
+        provided that ``item1`` and ``item2`` are string types.  See `Boolean
```
pekkaklarck: Same discussion about docs and implementation as with `should_not_contain` applies also to this keyword.

src/robot/libraries/BuiltIn.py
```diff
@@ -876,31 +970,51 @@ def get_count(self, item1, item2):
...
+    def should_not_match(self, string, pattern, msg=None, values=True, ignore_case=False):
         """Fails if the given ``string`` matches the given ``pattern``.
 
         Pattern matching is similar as matching files in a shell, and it is
         always case-sensitive. In the pattern ``*`` matches to anything and
         ``?`` matches to any single character.
 
+        ``ignore_case`` is False by default.  If True, it indicates that
+        ``first`` and ``second`` should be compared case-insensitively,
+        provided that ``first`` and ``second`` are string types.  See `Boolean
```
pekkaklarck: Similarly as with other keywords that should *always* be used with strings, `provided that ... are string types` is unnecessary.

src/robot/libraries/BuiltIn.py
```diff
@@ -876,31 +970,51 @@ def get_count(self, item1, item2):
...
         See `Should Be Equal` for an explanation on how to override the default
         error message with ``msg`` and ``values``.
         """
-        if self._matches(string, pattern):
+        prv_string = string
+        prv_pattern = pattern
+        if is_truthy(ignore_case) and is_string(string) and is_string(pattern):
+            prv_string = string.lower()
+            prv_pattern = pattern.lower()
+        if self._matches(prv_string, prv_pattern):
```
pekkaklarck: No need to explicitly check are `string` and `pattern` strings or not. Simply using

```python
if is_truthy(ignore_case):
    string = string.lower()
    pattern = pattern.lower()
```

ought to be enough.

src/robot/libraries/BuiltIn.py
```diff
@@ -876,31 +970,51 @@ def get_count(self, item1, item2):
...
+    def should_match(self, string, pattern, msg=None, values=True, ignore_case=False):
         """Fails unless the given ``string`` matches the given ``pattern``.
 
         Pattern matching is similar as matching files in a shell, and it is
         always case-sensitive. In the pattern, ``*`` matches to anything and
         ``?`` matches to any single character.
 
+        ``ignore_case`` is False by default.  If True, it indicates that
+        ``string`` and ``pattern`` should be compared case-insensitively,
+        provided that ``string`` and ``pattern`` are string types.  See `Boolean
```
pekkaklarck: `provided that ... are string types` is unnecessary here too.

src/robot/libraries/BuiltIn.py
```diff
@@ -876,31 +970,51 @@ def get_count(self, item1, item2):
...
         See `Should Be Equal` for an explanation on how to override the default
         error message with ``msg`` and ``values``.
         """
-        if not self._matches(string, pattern):
+        prv_string = string
+        prv_pattern = pattern
+        if is_truthy(ignore_case) and is_string(string) and is_string(pattern):
+            prv_string = string.lower()
+            prv_pattern = pattern.lower()
+        if not self._matches(prv_string, prv_pattern):
```
pekkaklarck: No need to check are `string` and `pattern` strings here either.

pekkaklarck:
This is quite a big PR. It would be easier to review changes if each keyword and its "not" variant would be in a separate pull request.

src/robot/libraries/BuiltIn.py
```diff
@@ -613,9 +613,17 @@ def should_be_equal(self, first, second, msg=None, values=True):
...
+        ``first`` and ``second`` should be compared case-insensitively,
+        provided that ``first`` and ``second`` are string types.  See `Boolean
+        arguments` section for more details.  (New in RF 3.0.1)
+
         If both arguments are multiline strings, the comparison is done using
         `multiline string comparisons`.
         """
+        if is_truthy(ignore_case) and is_string(first) and is_string(second):
+            first = first.lower()
+            second = second.lower()
```
chriscallan: I'd prefer to not add workarounds for Python engines that have bugs which need to be resolved.

src/robot/libraries/BuiltIn.py
```diff
@@ -745,87 +761,154 @@ def should_be_equal_as_numbers(self, first, second, msg=None, values=True,
...
+        ``first`` and ``second`` should be compared case-insensitively,
+        provided that ``first`` and ``second`` are string types.  See `Boolean
+        arguments` section for more details.  (New in RF 3.0.1)
+
         See `Should Be Equal` for an explanation on how to override the default
         error message with ``msg`` and ``values``.
         """
         self._log_types_at_info_if_different(first, second)
         first, second = [self._convert_to_string(i) for i in (first, second)]
+        if is_truthy(ignore_case) and is_string(first) and is_string(second):
```
chriscallan: Good point.

src/robot/libraries/BuiltIn.py
```diff
@@ -745,87 +761,154 @@ def should_be_equal_as_numbers(self, first, second, msg=None, values=True,
...
         See `Should Be Equal` for an explanation on how to override the default
         error message with ``msg`` and ``values``.
         """
-        if str1.startswith(str2):
+        prv_str1 = str1
+        prv_str2 = str2
+        if is_truthy(ignore_case) and is_string(str1) and is_string(str2):
+            prv_str1 = str1.lower()
+            prv_str2 = str2.lower()
+        if prv_str1.startswith(prv_str2):
```
chriscallan: I was trying to avoid having unhandled exceptions when someone sends in something besides a string and blindly calling ".lower()" on whatever was passed in.
I added the part with assigning to private variables so that if there was an output message (i.e. things weren't equal) that the output message matched up to what the user sent in to the keyword rather than referencing the lower-cased version.  I found it to be a bit confusing even during my testing.

src/robot/libraries/BuiltIn.py
pekkaklarck: Adding workarounds like that is annoying, but the framework working incorrectly on some platform is worse. It's especially bad if tests fail on that platform when run on CI. In this particular case we can argue that people ought to use newish IronPython versions where this problem is fixed. Like we don on our CI.

src/robot/libraries/BuiltIn.py
pekkaklarck: I understand the point of avoiding exceptions, but there's anyway going to be an exception when `.startswith()` is run unless strings are used. The extra protection for `.lower()` thus doesn't help at all but adds a little complexity. If we wanted to add a guard to avoid invalid usage, we should check argument types in the beginning of the keyword and raise an exception if types are invalid. I don't think that's worth the effort with a keyword clearly expecting strings.

I also see the point to use the original strings in the error message, but that should then be done consistently with all keywords. Now, for example, `should_be_equal` just does `first = first.lower()`. This is something that we need to talk separately rather than in code comments.

pekkaklarck:
As already discussed in code comments, some keywords currently show the lower cased strings in the error message and others show the original. Probably showing the original is better, but that needs to be done consistently. If we agree it is a good idea, we should enhance `_get_string_msg` to add `(case-insensitively)` at the end of the default error message it creates when `case_insensitive=True`. Otherwise we will get confusing errors like `'foo' != 'FOO'` or `'foo' starts with 'F'`.

On the code level we have two general possibilities to preserve the original values. Either we can just keep a reference to the original values and use them in the error message like this:

```python
orig_first = first
orig_second = second
if is_truthy(case_insensitive):
    first = first.lower()
    second = second.lower()
if not first.startswith(second):
    # raise exception using `orig_first` and `orig_second`
```

Alternatively we could do the actual validation differently depending on case-sensitivity:

```python
if not is_truthy(case_insensitive):
    passed = first.startswith(second)
else:
    passed = first.lower().startswith(second.lower())
if not passed:
    # raise exception using `first` and `second`
```

Personally I generally prefer the latter but perhaps the former is more explicit.

src/robot/libraries/BuiltIn.py
chriscallan: So, are you advocating for the stance that users of IronPython 2.7.x and below should upgrade to fix the bug with "string.lower()" in their interpreter  or modify all the "string.lower()" calls to "utils.lower(string)" to workaround that bug?
My opinion says to leave the string conversions and have the users upgrade since bugs with "string.lower()".

src/robot/libraries/BuiltIn.py
chriscallan: I like the idea of changing 'container' to a set, if it's 'list-like', but the points you bring up for 'item' are very good.

chriscallan:
I just revisited all of the above, and I'm more convinced that your suggestion to handle this update as multiple bite-size PR's is much wiser idea.  I'll go ahead and close this PR and then reintroduce them as a single PR focused on one logical grouping of keywords (i.e. "Should Be Equal"/"Should Not Be Equal"), keeping all your above feedback in mind of course :-)
Sound good?

pekkaklarck:
Sounds great!

ShirleyPerkins:
Learn more about how Robot Framework can simplify your tests: http://blog.testproject.io/2016/11/22/robot-framework-introduction/

PR #2461, opened 2016-10-14
Case-insensitivity support to "Should (Not) Be Equal (As Strings)" keywords

chriscallan:
Implementation of "Should (Not) Be Equal" keywords, along with the additional and modified Acceptance Tests;

### DIFF
--- a/atest/testdata/standard_libraries/builtin/verify.robot
+++ b/atest/testdata/standard_libraries/builtin/verify.robot
@@ -68,6 +68,11 @@ Should Not Be Equal
     ${STR1}    ${INT1}
     ${STR1}    1
 
+Should Not Be Equal With Case Insensitivity
+    [Template]  Should Not Be Equal
+    test value      TEST VALUE1      ignore_case=True
+    HYVÄÄ YÖTÄ      hyvää yötä1     ignore_case=True
+
 Should Not Be Equal with bytes containing non-ascii characters
     [Documentation]    FAIL ${BYTES WITH NON ASCII} == ${BYTES WITH NON ASCII}
     Should Not Be Equal    ${BYTES WITH NON ASCII}    ${BYTES WITHOUT NON ASCII}
@@ -81,6 +86,11 @@ Should Be Equal
     ${INT1}    ${1}
     ${BYTES WITHOUT NON ASCII}    ${BYTES WITHOUT NON ASCII}
     A    B    Error message    values=yes
+
+Should Be Equal With Case Insensitivity
+    [Template]  Should Be Equal
+    test value      TEST VALUE      ignore_case=True
+    HYVÄÄ YÖTÄ      hyvää yötä      ignore_case=True
 
 Should Be Equal fails with values
     [Documentation]    FAIL Several failures occurred:\n\n 1) 3: 1 != 2\n\n 2) c: a != b\n\n 3) z: x != y
--- a/atest/testdata/standard_libraries/builtin/verify.robot
+++ b/atest/testdata/standard_libraries/builtin/verify.robot
@@ -70,8 +70,9 @@ Should Not Be Equal
 
 Should Not Be Equal With Case Insensitivity
     [Template]  Should Not Be Equal
-    test value      TEST VALUE1      ignore_case=True
+    test value      TEST VALUE1     ignore_case=True
     HYVÄÄ YÖTÄ      hyvää yötä1     ignore_case=True
+    foo             FOO             ignore_case=False
 
 Should Not Be Equal with bytes containing non-ascii characters
     [Documentation]    FAIL ${BYTES WITH NON ASCII} == ${BYTES WITH NON ASCII}
@@ -91,6 +92,7 @@ Should Be Equal With Case Insensitivity
     [Template]  Should Be Equal
     test value      TEST VALUE      ignore_case=True
     HYVÄÄ YÖTÄ      hyvää yötä      ignore_case=True
+
 
 Should Be Equal fails with values
     [Documentation]    FAIL Several failures occurred:\n\n 1) 3: 1 != 2\n\n 2) c: a != b\n\n 3) z: x != y
--- a/atest/testdata/standard_libraries/builtin/verify.robot
+++ b/atest/testdata/standard_libraries/builtin/verify.robot
@@ -223,12 +223,22 @@ Should Not Be Equal As Strings
     False    ${True}
     bar    bar    These strings most certainly should not be equal    False
 
+Should Not Be Equal As Strings With Case Insensitivity
+    [Template]    Should Not Be Equal As Strings
+    1    ${1.1}     ignore_case=True
+    False    ${True}    ignore_case=True
+
 Should Be Equal As Strings
     [Documentation]    FAIL foo != bar
     [Template]    Should Be Equal As Strings
     ${1}    1
     ${None}    None
     foo    bar
+
+Should Be Equal As Strings With Case Insensitivity
+    [Template]  Should Be Equal As Strings
+    test value      TEST VALUE      ignore_case=True
+    HYVÄÄ YÖTÄ      hyvää yötä      ignore_case=True
 
 Should Be Equal As Strings Multiline
     [Documentation]    FAIL Multiline strings are different:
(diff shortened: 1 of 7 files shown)


chriscallan:
Hopefully this one keyword pair is more easier to work with than all the keywords at once.
I also added some list comprehensions for list-like objects which contain string collections, or even a mix, but only one-level deep.  I think that covers the major use case for the keyword.

atest/testdata/standard_libraries/builtin/verify.robot
```diff
@@ -68,6 +68,11 @@ Should Not Be Equal
     ${STR1}    ${INT1}
     ${STR1}    1
 
+Should Not Be Equal With Case Insensitivity
+    [Template]  Should Not Be Equal
+    test value      TEST VALUE1      ignore_case=True
+    HYVÄÄ YÖTÄ      hyvää yötä1     ignore_case=True
```
pekkaklarck: It would be good to test that e.g. `foo == FOO` fails when `ignore_case=True`. The current values are unequal also with `ignore_case=False`.

There's also different number of spaces before `ignore_case` above.

src/robot/libraries/BuiltIn.py
```diff
@@ -613,10 +613,21 @@ def should_be_equal(self, first, second, msg=None, values=True):
         for example, string ``false`` or ``no values``. See `Boolean arguments`
         section for more details.
 
+        If ignore_case is True, it indicates that ``first`` and ``second`` should be
+        compared case-insensitively.  See `Boolean  arguments` section for more details.
+        (This option is new in Robot Framework 3.0.1)
```
pekkaklarck: `This option is new in Robot Framework 3.0.1.` without parenthesis would be more consistent with other similar notes.

src/robot/libraries/BuiltIn.py
```diff
@@ -613,10 +613,21 @@ def should_be_equal(self, first, second, msg=None, values=True):
...
         `multiline string comparisons`.
         """
         self._log_types_at_info_if_different(first, second)
+        if is_truthy(ignore_case):
+            if is_string(first) and is_string(second):
+                first = first.lower()
+                second = second.lower()
+            elif is_list_like(first) and is_list_like(second):
+                first = [x.lower() if type(x) is str else x for x in first]
+                second = [x.lower() if type(x) is str else x for x in second]
```
pekkaklarck: I don't think it is a good idea to compare list-like items case-insensitively with this keyword in general. Even if it would be, the current implementation considers a list `['foo', 'bar']` equal to a  tuple `('foo', 'bar')` and even to a dict `{'foo': 1, 'bar': 2}`. Additionally, the current implementation wouldn't handle nested lists like `[['foo']]` and `[['FOO']]`.

Better to simply remove the `elif` block and just use `if is_truthy(ignore_case) and is_string(first) and is_string(second)`. If we want to make it explicit that `ignore_case=True` only applies to strings, the docstring could be changed to something like this:

> If ``ignore_case`` is given a true value (see `Boolean arguments` for details) and ``first`` and ``second`` are strings, the comparison is done case-insensitively.

src/robot/libraries/BuiltIn.py
```diff
@@ -646,13 +657,24 @@ def _raise_multi_diff(self, first, second):
...
-    def should_not_be_equal(self, first, second, msg=None, values=True):
+    def should_not_be_equal(self, first, second, msg=None, values=True, ignore_case=False):
         """Fails if the given objects are equal.
 
         See `Should Be Equal` for an explanation on how to override the default
         error message with ``msg`` and ``values``.
+
+        If ignore_case is True, it indicates that ``first`` and ``second`` should be
+        compared case-insensitively.  See `Boolean  arguments` section for more details.
+        (This option is new in Robot Framework 3.0.1)
```
pekkaklarck: The above comments about the docs apply here too.

src/robot/libraries/BuiltIn.py
```diff
@@ -646,13 +657,24 @@ def _raise_multi_diff(self, first, second):
...
         """
         self._log_types_at_info_if_different(first, second)
+        if is_truthy(ignore_case):
+            if is_string(first) and is_string(second):
+                first = first.lower()
+                second = second.lower()
+            elif is_list_like(first) and is_list_like(second):
+                first = [x.lower() if type(x) is str else x for x in first]
+                second = [x.lower() if type(x) is str else x for x in second]
```
pekkaklarck: The above comments about handling list-like objects applies here too.

src/robot/libraries/BuiltIn.py
chriscallan: I guess I was just trying to get to fancy here.  Thanks for pulling me back to reality.  
I've updated the doc string and removed the code per your suggestion.

atest/testdata/standard_libraries/builtin/verify.robot
chriscallan: I think I see what you're getting at, but not sure.  If I modify the test case to look at the case: 
   ...
    foo                      FOO                ignore_case=True

then ignoring case would make the strings equal, which would in-turn fail this Acceptance Test.

Putting the extra character on the second argument was the best that I could think of, since just modifying case isn't enough to keep the strings from being unequal.
Am I missing something right in front of my nose here?

chriscallan:
Latest commit should have all the changes you requested.

src/robot/libraries/BuiltIn.py
```diff
@@ -613,10 +613,18 @@ def should_be_equal(self, first, second, msg=None, values=True):
...
+        are string types.  See `Boolean  arguments` section for more details.
+        This option is new in Robot Framework 3.0.1.
+
         If both arguments are multiline strings, the comparison is done using
         `multiline string comparisons`.
         """
         self._log_types_at_info_if_different(first, second)
+        if is_truthy(ignore_case) and is_string(first) and is_string(second):
+                first = first.lower()
+                second = second.lower()
```
pekkaklarck: Too much indentation. Same also with the other keyword.

src/robot/libraries/BuiltIn.py
```diff
@@ -613,10 +613,18 @@ def should_be_equal(self, first, second, msg=None, values=True):
         for example, string ``false`` or ``no values``. See `Boolean arguments`
         section for more details.
 
+        If ignore_case is True, it indicates that ``first`` and ``second`` should be
```
pekkaklarck: Should have double-backticks around `ignore_case` similarly as with other arguments. Also with the other keyword.

atest/testdata/standard_libraries/builtin/verify.robot
```diff
@@ -82,6 +89,12 @@ Should Be Equal
     ${BYTES WITHOUT NON ASCII}    ${BYTES WITHOUT NON ASCII}
     A    B    Error message    values=yes
 
+Should Be Equal With Case Insensitivity
+    [Template]  Should Be Equal
+    test value      TEST VALUE      ignore_case=True
+    HYVÄÄ YÖTÄ      hyvää yötä      ignore_case=True
+
+
```
pekkaklarck: Extra empty row.

atest/testdata/standard_libraries/builtin/verify.robot
```diff
@@ -68,6 +68,13 @@ Should Not Be Equal
     ${STR1}    ${INT1}
     ${STR1}    1
 
+Should Not Be Equal With Case Insensitivity
+    [Documentation]     FAIL foo == foo
```
pekkaklarck: We need to decide should this and other similar errors be reported like this (i.e. `foo == foo`) or should we rather show the original values and mention that comparison was case-insensitive (e.g. `foo == FOO (case-insensitively)`). Probably better to discuss that in the PR comments or perhaps even in the comments of the original issue.

pekkaklarck:
Changes in this PR have been merged after rebase on the command line and committed as 29cf1bc1e93d18184ca63324897b9ddd59f9f915. GitHub didn't detect that automatically so I'll just close this PR now. See also c6a8ceb46d53f66bf311d05304958798812b101e which contains some clean-up to this PR.

PR #2464, opened 2016-10-25
Add telnet.write_until_expected_output_and_read_all keyword which returns all received output instead of last chunk of data

winter-triangle:
Hi!
First and foremost I do apologize for my somewhat poor English, but I'm hope I describe my idea as clear as I can.
Please consider to add a keyword for 'Telnet' library which I called 'write_until_expected_output_and_read_all'.
**Why**: write_until_expected_output returns just last chunk of output. It may be useful to receive all the command output.
**Use case**: consider network devices configuration testing. There are a lot of continuable commands, such as 'show ?' command in Cisco CLI. It can be useful to got all the result of such continuable command, in my case it was a internal logging buffer.
**Why not to use existing keywords**: I found this solution kind of ugly, because it conains a lot of programming logic within and IMO better have place in the library itself.

### DIFF
--- a/src/robot/libraries/Telnet.py
+++ b/src/robot/libraries/Telnet.py
@@ -859,6 +859,49 @@ class TelnetConnection(telnetlib.Telnet):
                 pass
         raise NoMatchError(expected, timeout)
 
+    def write_until_expected_output_and_read_all(self, text, expected, timeout,
+                                    retry_interval, loglevel=None):
+        """Writes the given ``text`` repeatedly, until ``expected`` appears in the output.
+        Returns back all output received.
+
+        ``text`` is written without appending a newline and it is consumed from
+        the output before trying to find ``expected``. If ``expected`` does not
+        appear in the output within ``timeout``, this keyword fails.
+
+        ``retry_interval`` defines the time to wait ``expected`` to appear before
+        writing the ``text`` again. Consuming the written ``text`` is subject to
+        the normal [#Configuration|configured timeout].
+
+        Both ``timeout`` and ``retry_interval`` must be given in `time string
+        format`. See `Logging` section for more information about log levels.
+
+        Example:
+        | Write Until Expected Output | ps -ef| grep myprocess\\r\\n | myprocess |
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Could you clarify what benefits this would have over just using existing `Write Until Expected Output` and then `Read` afterwards?

winter-triangle:
Sorry for the kinda late answer.
I found this solution (to implement this functionality in terms of Robot Framework instead of in library) ugly because it is a lot of logic in resources while telnet library already have a routine for similar task - `write_until_expected_output`.
Your solution have a point, but let's assume this case:
1. We execute a "continuable" command in Cisco CLI. Let'assume this is `# show running-config` executed using `Write` and continued by `Write Until Expected Output`. We awaiting a privileged prompt such as "hostname#"
2. Cisco responds with somewhat like `too long blah-blah, press <Space> to show next page and <q> to quit`
3. Prompt doesn't found in this output, so `Write Until Expected Output` perform a `Write` action with sequence we supplied
4. Cycle continues until the prompt was reached. At this point we've got a last chunk of text returned.
Note that we can't call `Read` while `Write Until Expected Output` getting work done. 

Of course it is possible to implement custom keyword in terms of Robot Framework using loops with checks and `Read`/`Write` within, but it, eh, smells IMHO. Maybe (most likely) my solution even smells more, and better solution will just add some argument to `Write Until Expected Output` like
```python
write_until_expected_output( _here_goes_arguments_, Return_All=False )
```
Instead of another routine

vkruoso:
For this use case it seems you can probably disable paging at the CLI session before running the command, unless you are actually testing the paging feature. There might be other useful use cases tough. I think the keyword name is confusing.

winter-triangle:
Thank you for advise, @vkruoso! Heck, I didn't know about this feature. Even though I'm working with custom network hardware (Cisco was just a sample) it also have similar command. Thanks! 👍

pekkaklarck:
Good that there was a workaround! The proposed keyword was somewhat complex and seemed to have only narrow usage.

PR #2467, opened 2016-10-27
Should (not) be equal as strings

chriscallan:
This should be the implementation of the "Should (Not) Be Equal As Strings" keyword pair.
I tried my best to get the "Should (Not) Be Equal" changes to not be included, but I don't think I was successful.


chriscallan:
Closing this PR to fix the inclusion of the previous keyword pair in this fork.  I'll open another without those changes very soon.

PR #2469, opened 2016-11-04
Update OperatingSystem.py(IP address to decimal and hexa conversion )

vasujammula:
IP address to decimal and hexa conversion 
this helps while writing test cases requires IP address conversions.

### DIFF
--- a/src/robot/libraries/OperatingSystem.py
+++ b/src/robot/libraries/OperatingSystem.py
@@ -21,7 +21,10 @@ import shutil
 import sys
 import tempfile
 import time
-
+import binascii
+import socket
+import netaddr
+from binascii import hexlify
 from robot.version import get_version
 from robot.api import logger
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Thanks for the PR. Unfortunately it cannot be merged. The main reason is that I don't think these two IP related keywords belong to the OperatingSystem in general. It would be better to implement a separate library for handling IP addresses and possibly also downloads and other network related stuff. The first step towards that direction would be submitting an issue about that.

There are also several issues with the code itself that would prevent merging this PR. These don't matter too much now, but please take them into account if you plan to submit new PRs in the future:

- Documentation missing.
- Tests missing.
- Not Python 3 compatible.
- Doesn't follow PEP-8.

For more details about the listed issues see CONTRIBUTING.rst file in the project root.

Issue #2483, opened 2016-11-28
Copying model objects doesn't always work


krizex:
When I want to copy test case, robot framework throw an exception as below:
[ ERROR ] Executing model modifier 'RepeatTestCase' failed: TypeError: type object argument after * must be a sequence, not Timeout
Traceback (most recent call last):
 File "/home/dqian/dev/cafe/calix/src/caferobot/robotmodifier/prerunmodifier/repeatcase.py", line 27, in start_suite
   self._create_repeat_testcase(suite, case)
 File "/home/dqian/dev/cafe/calix/src/caferobot/robotmodifier/prerunmodifier/repeatcase.py", line 50, in _create_repeat_testcase
   new_case = copy.copy(case)
 File "/opt/ActivePython-2.7/lib/python2.7/copy.py", line 96, in copy
   return _reconstruct(x, rv, 0)
 File "/opt/ActivePython-2.7/lib/python2.7/copy.py", line 346, in _reconstruct
   setattr(y, key, value)
 File "/opt/ActivePython-2.7/lib/python2.7/site-packages/robot/utils/setter.py", line 34, in __set__
   setattr(instance, self.attr_name, self.method(instance, value))
 File "/opt/ActivePython-2.7/lib/python2.7/site-packages/robot/running/model.py", line 107, in timeout
   return Timeout(*timeout) if timeout else None

My execution environment is:
OS: CentOS 6.7
Python: 2.7
Robot framework: 3.0

I think it's necessary to enable `copy` and `deepcopy` feature with these model objects.

pekkaklarck:
It can be considered a bug that `copy.copy(test)` fails. The actual problem show in the traceback seems to be related to timeouts that are handled somewhat inconsistently. Need to investigate more closely what's the best approach to fix this. PR #2484 has one alternative, but could also look at fixing timeouts.

Related to this, it would be great to have a convenience method `test.copy()` and possibly also `test.deepcopy()`. Even better if it would be possible to set same attributes as in `__init__` directly like `test.copy(name='New name', tags=['new', 'tags'])`.

krizex:
The root cause of this issue is we cannot set the `timeout` attribute by a `Timeout` type argument. There may be more issues like this when we invoke `copy` with sub class of `ModelObject`. So I implement the `__setstate__` to make it compatible with `copy` module.
We can customize the `copy` method of these classes by calling the `copy.copy(self)` in the implementation.
For example:
[...]

PR #2484, opened 2016-11-28
Model object should support copy and deepcopy

krizex:
Enable the `copy` and `deepcopy` feature of model objects.
This PR is the fix for that I mentioned in #2483 .

### DIFF
--- a/src/robot/model/modelobject.py
+++ b/src/robot/model/modelobject.py
@@ -26,4 +26,33 @@ class ModelObject(with_metaclass(SetterAwareType, object)):
     def __repr__(self):
         return repr(str(self))
 
+    def __setstate__(self, state):
+        """Customize the attribute updating when using `copy.copy` or `copy.deepcopy`
+        """
+
+        # We should consider the state format per pickle protocol version
+        # Refer to: https://www.python.org/dev/peps/pep-0307
+        if isinstance(state, tuple) and len(state) == 2:
+            dictstate, slotstate = state
+        else:
+            dictstate = state
+            slotstate = None
+
+        if dictstate is not None:
+            self.__dict__.update(dictstate)
+
+        def get_setter_name(x):
+            return '_setter__' + x
+
+        if slotstate is not None:
+            for k, v in slotstate.iteritems():
+                setter_name = get_setter_name(k)
+                # If we have defined the attribute both in __slots__ and with @setter,
+                # we just need to set the setter attribute.
+                if setter_name in slotstate:
+                    continue
+
+                setattr(self, k, v)
+
+
 
--- a/utest/model/test_keyword.py
+++ b/utest/model/test_keyword.py
@@ -1,3 +1,4 @@
+import copy
 import unittest
 from robot.utils.asserts import (assert_equal, assert_none, assert_true,
                                  assert_raises, assert_raises_with_msg)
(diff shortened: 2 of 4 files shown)


krizex:
Per https://docs.python.org/2/library/copy.html?highlight=copy.copy#copy.copy we can see:

> Classes can use the same interfaces to control copying that they use to control pickling. See the description of module pickle for information on these methods. The copy module does not use the copy_reg registration module.

Inspect to the source code, `copy.copy` and `copy.deepcopy` will call `_reconstruct` and in this method it will call the `__setstate__` method if the instance has it. When we implement `__setstate__` we should consider the argument as a tuple as described on https://www.python.org/dev/peps/pep-0307/ 

> Case 3: pickling new-style class instances using protocol 2

pekkaklarck:
Did you investigate implementing `__copy__` and `__deepcopy__` instead of using the somewhat magical looking `__setstate__`?

pekkaklarck:
It's good to see a reference to `pickle` that is documented to use `__setstate__` in `copy` module docs. And supporting `pickle` could be useful in some cases too.

krizex:
Per the implementation of `copy.copy` and `copy.deepcopy`, we don't have to implement methods of `__copy__` and `__deepcopy__`, and we just need to implement the `__setstate__` as I did in this PR.

src/robot/model/modelobject.py
```diff
@@ -26,4 +26,33 @@ def __unicode__(self):
...
+        """Customize the attribute updating when using `copy.copy` or `copy.deepcopy`
+        """
+
+        # We should consider the state format per pickle protocol version
+        # Refer to: https://www.python.org/dev/peps/pep-0307
+        if isinstance(state, tuple) and len(state) == 2:
+            dictstate, slotstate = state
+        else:
+            dictstate = state
+            slotstate = None
```
pekkaklarck: We have `__slots__` and can thus know, based on the referred PEP, that state is two-tuple.

src/robot/model/modelobject.py
```diff
@@ -26,4 +26,33 @@ def __unicode__(self):
...
+        def get_setter_name(x):
+            return '_setter__' + x
+
+        if slotstate is not None:
+            for k, v in slotstate.iteritems():
+                setter_name = get_setter_name(k)
+                # If we have defined the attribute both in __slots__ and with @setter,
+                # we just need to set the setter attribute.
+                if setter_name in slotstate:
+                    continue
```
pekkaklarck: Could you clarify why is this needed? I tested that all unit tests pass also without it.

src/robot/model/modelobject.py
```diff
@@ -26,4 +26,33 @@ def __unicode__(self):
...
+            dictstate, slotstate = state
+        else:
+            dictstate = state
+            slotstate = None
+
+        if dictstate is not None:
+            self.__dict__.update(dictstate)
+
+        def get_setter_name(x):
+            return '_setter__' + x
```
pekkaklarck: This function isn't really needed.

src/robot/model/modelobject.py
```diff
@@ -26,4 +26,33 @@ def __unicode__(self):
...
+            for k, v in slotstate.iteritems():
+                setter_name = get_setter_name(k)
+                # If we have defined the attribute both in __slots__ and with @setter,
+                # we just need to set the setter attribute.
+                if setter_name in slotstate:
+                    continue
+
+                setattr(self, k, v)
+
+
```
pekkaklarck: Our code conventions say empty rows shouldn't generally be used inside functions/methods and there shouldn't be empty rows at the end.

utest/model/test_keyword.py
```diff
@@ -33,6 +34,22 @@ def test_id_with_keyword_parents(self):
...
 
+    def test_copy_keyword(self):
+        kw = Keyword()
+        kw_new = copy.copy(kw)
+        self.assertEqual(kw.name, kw_new.name)
+
+        kw_new.name = kw.name + '1'
+        self.assertNotEqual(kw.name, kw_new.name)
+
+        self.assertEqual(id(kw.tags), id(kw_new.tags))
```
pekkaklarck: Good tests. Two nitpick comments:

- We have our own `assert_equal` and friends.
- We don't use empty lines inside functions (unless functions are long but we also avoid long functions).

pekkaklarck:
OK, I think I figured out why that `if setter_name in slotstate: continue` thing is needed: Without it copying a `running.TestCase` that has a timeout fails. There were some unit tests for copying these tests, but those tests didn't have timeout so unit tests passed also without the guarding if. I try to enhance tests so that they'll detect this problem too.

pekkaklarck:
One more thing, do you @krizex want to be credited in `AUTHORS.txt`? Please add yourself there if you want to. You definitely deserve it after this PR.

krizex:
@pekkaklarck , thanks for your kindly reminder about the code convention of this project. I have created a PR #2511  to modify the AUTHORS.txt, please merge it at your convenience.
Happy new year!

pekkaklarck:
Thank you for the PR! Not sure had I found `__setstate__` myself or learned how to use it.

Issue #2117, opened 2015-08-24
Re-run test suite instead of tests


cdeokar:
Is there any way which will facilitate to re-run failed test suites?

pekkaklarck:
No. Are you proposing to add something like that? Are you interested to implement it?

satish88:
I had the same requirement where test cases in a test suite were dependent on each other.
I achieved it with a hack where I processed the output.xml to mark all the test cases as failed, if even a single test case failed in a test suite. 
Anyway, I want to offer me to implement it in more cleaner way, if it is yet to be implemented

pekkaklarck:
Would something like `--rerunfailedsuites` be a good idea? If someone is interested to implement it, I'd be happy to merge a pull request and also help if needed. 

Implementation ought to be pretty easy otherwise, but handling non-critical failures may require a little extra work. Only suites with critical failures are considered failed by Robot, and I think this functionality should follow the same logic. The problem is that the result model parsed from the output.xml file won't have any information about criticality. This hopefully can be fixed pretty easily by configuring the result model with same criticality information (`--critical` and `--noncritical` options) as the suite that is about to be executed.

pekkaklarck:
Anyone interested about `--rerunfailedsuites` enough to provide a pull request? If not, I'll close this issue. If you are but are new to pull requests and contributing, you can start by reading these guidelines:
https://github.com/robotframework/robotframework/blob/master/CONTRIBUTING.rst

jussimalinen:
There seems to be no interest for this. Can be reopened if someone feels like doing this.

antechrestos:
I started implementing it. I made a *WIP* [merge requests](https://github.com/robotframework/robotframework/pull/2487)

pekkaklarck:
We now got PR #2487 with a very promising implementation. Reopening and assigning to RF 3.0.1 scope.

pekkaklarck:
Finally looking at PR #2487 by @antechrestos now and it looks pretty good. Started to think should we anyway run suites that have *any* failed tests without thinking criticality? I know I proposed taking criticality into account above, but now that I think about it running all suites with failed tests somehow sounds better. It would also simplify implementation a little.

antechrestos:
@pekkaklarck I will change my implementation by tomorrow. I agree with you, that might be confusing if non critical failed are not re-run...

pekkaklarck:
[...]

PR #2487, opened 2016-12-05
Add --rerunfailedsuites option

antechrestos:
Based on [this issue](https://github.com/robotframework/robotframework/issues/2117) (I do need this functionnality too), I started implementing it.

:warning: this do not support the `--critical/noncritical`as I **need some advice**:
- shall I filter from first output (where for example `--noncritical`was set) : then I need to update the model to get the information when output is processed
- shall I only filter from second execution parameters

Can you also give me some comment about my work, what I need to change/correct if any.

Thanks!

### DIFF
--- a/atest/robot/cli/runner/rerunfailedsuites.robot
+++ b/atest/robot/cli/runner/rerunfailedsuites.robot
@@ -1,0 +1,44 @@
+*** Settings ***
+Suite Setup       Suite initialization
+Suite Teardown    Remove Directory    ${RERUN SUITE DIR}    recursive
+Resource          atest_resource.robot
+
+*** Variables ***
+${ORIG DIR}           ${DATADIR}/cli/runfailed
+${RERUN SUITE DIR}    %{TEMPDIR}/rerunsuites-dir
+${SUITE DIR}          ${RERUN SUITE DIR}/suite
+${RUN FAILED FROM}    ${RERUN SUITE DIR}/rerun-suites-output.xml
+
+*** Test Cases ***
+Passing suite is not re-executed
+    Test Should Not Have Been Executed    Passing
+
+Failing suite is re-executed
+    Test Should Have Been Executed   Failing
+    Test Should Have Been Executed   Not Failing
+
+Suite teardown failures are noticed
+    Test Should Have Been Executed   Test passed but suite teardown fails
+
+Excluded failing is not executed
+    Test Should Not Have Been Executed    Failing with tag
+
+Non-existing failing from output file is not executed
+    Test Should Not Have Been Executed    Only in one suite
+
+*** Keywords ***
+Suite initialization
+    Copy Directory    ${ORIG DIR}/suite    ${SUITE DIR}
+    Copy File    ${ORIG DIR}/runfailed1.robot     ${SUITE DIR}/runfailed.txt
+    Run Tests    ${SUITE DIR}
+    Copy File    ${OUTFILE}    ${RUN FAILED FROM}
+    Copy File    ${ORIG DIR}/runfailed2.robot     ${SUITE DIR}/runfailed.txt
+    Run Tests    --rerunfailedsuites ${RUN FAILED FROM} --noncritical non_critical --exclude excluded_tag ${SUITE DIR}
+
+Test Should Have Been Executed
+    [Arguments]    ${name}
+    Check Test Case    ${name}
+
+Test Should Not Have Been Executed
+    [Arguments]    ${name}
+    Run Keyword And Expect Error    No test '${name}' found*    Check Test Case    ${name}
--- a/doc/userguide/src/Appendices/CommandLineOptions.rst
+++ b/doc/userguide/src/Appendices/CommandLineOptions.rst
@@ -22,6 +22,8 @@ Command line options for test execution
   -R, --rerunfailed <file>  `Selects failed tests`_ from an earlier `output file`_ to be re-executed.
   --runfailed <file>      Deprecated since Robot Framework 2.8.4.
                           Use :option:`--rerunfailed` instead.
+  -S, --rerunfailedsuites <file>  `Selects failed test suite`_ from an earlier `output file`_ to be re-executed.
+
   -i, --include <tag>     `Selects the test cases`_ by tag.
   -e, --exclude <tag>     `Selects the test cases`_ by tag.
   -c, --critical <tag>    Tests that have the given tag are `considered critical`_.
@@ -148,6 +150,7 @@ Command line options for post-processing outputs
 .. _Sets the tag(s): `Setting tags`_
 .. _Selects the test cases by name: `By test suite and test case names`_
 .. _Selects the test suites: `Selects the test cases by name`_
+.. _Selects failed tests suites: `Re-executing failed test suites`_
 .. _Selects failed tests: `Re-executing failed test cases`_
 .. _Selects the test cases: `By tag names`_
 .. _considered critical: `Setting criticality`_
--- a/doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
+++ b/doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
@@ -140,6 +140,20 @@ is same as not specifying this option at all.
 
 __ `Merging outputs`_
 
+Re-executing failed test suites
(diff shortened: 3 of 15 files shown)


doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
```diff
@@ -165,6 +165,23 @@ running tests.
 
 .. note:: :option:`--ProcessEmptySuite` option was added in Robot Framework 2.7.2.
 
+Re-executing failed test suites
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Command line option :option:`rerunfailedsuites (-S)` can be used to select all failed
+suites from an earlier `output file`_ for re-execution. Like  :option:`--rerunfailed (-R)`, this option is useful when
+full test execution takes a lot of time. Note that all tests from a failed test suite will be re-executed, even
+passing ones. This option may be useful when you missed developing tests of a test suite as independent from each other.
```
laurentbristiel: "when you missed developing tests of a test suite as independent from each other"
=> bit unclear

antechrestos:
@laurentbristiel I replaced it with 
>>> This option is useful when the tests of a test suite depends on each other.

Do you have any idea about the advice I need for the `critical/noncritical`. It is not the way to do it, it is more what is expected.

Thanks!

antechrestos:
@pekkaklarck : I implemented it supposing the `--noncritical` tags had to be given at the second execution, and not guessed from the first output:
1. `pybot    ${SUITE DIR}`
2. `pybot --rerunfailedsuites ./output.xml --noncritical non_critical ${SUITE DIR}`
This will exclude tests bearing the *non_critical* tag from re execution (as described in `rerunfailedsuites.robot`).
 
Thinking this way, I found that the `critical` property of a `TestCase` was [a calculated property](https://github.com/robotframework/robotframework/blob/master/src/robot/result/model.py#L152) that bases its calculus upon parent (suite) `criticality` field. This later is set by the `Runner` ... I took a look at the `output.xml`and found no trace of the critical/noncritical tags. 

Consequently, I faced either overriding the `TestCase` property (which I found a little ugly), or building a `Criticability` at visit time. I then faced the last problem: when visiting, we are not sure that all settings property are set, specialy `critical` and `noncritical`. Hence, I implemented it as a trigger that get calculated at the end of `_BaseSettings._process_cli_opts`execution

Looking forward to reading you.

src/robot/conf/gatherfailed.py
```diff
@@ -32,6 +32,21 @@ def visit_keyword(self, kw):
         pass
 
 
+class GatherFailedSuites(SuiteVisitor):
+
+    def __init__(self, critical_tags, non_critical_tags):
+        self.suites = set()
+        self._criticality = Criticality(critical_tags,
+                                        non_critical_tags)
```
pekkaklarck: This ought to work, but it might be cleaner to call `suite.set_crticality()` when `start_suite` is called for the first time. It would especially simplify `visit_test` implementation.

src/robot/conf/settings.py
```diff
@@ -92,6 +94,9 @@ def __setitem__(self, name, value):
     def _process_value(self, name, value):
         if name == 'ReRunFailed':
             return gather_failed_tests(value)
+        if name == 'ReRunFailedSuites':
+            # will be calculated at the end
+            return lambda: gather_failed_suites(value, self.critical_tags, self.non_critical_tags)
```
pekkaklarck: Returning a lambda that is called in `_process_cli_opts()` is a bit ugly but works. Alternatively we could just let the normal logic set `ReRunFailedSuites` option and in `_process_cli_opts()` do something like this:

```python
if self['ReRunFailedSuites']:
    self['SuiteNames'] += gather_failued_suites(self['ReRunFailedSuites'])
```

For consistency reasons the logic with `ReRunFailed` should be changed to follow this logic too.

What do you think?

pekkaklarck: Returning a lambda that is later called in `_process_cli_opts` is a bit ugly but works. Alternative option would be letting the normal option setting logic to set `ReRunFailedSuites` and then in `_process_cli_opts` doing something like this:

```python
if self['ReRunFailedSuites']:
    self['SuiteNames'] += gather_failed_suites(self['ReRunFailedSuites'], self.critical_tags, self.non_critical_tags)
```

For consistency reason `ReRunFailed` should be changed to follow the same logic. What do you think?

pekkaklarck:
Looks pretty good to me. Unfortunately I don't have time for a full review, but briefly looked and commented changes under `src`. I don't know when exactly I have time to look at this more closely (currently returning from a short holiday with too much tasks waiting for me...), but I'll anyway open the original issue and target it to RF 3.0.1. Thanks!

pekkaklarck:
Two more general comments before I need to close my laptop:

- Note(s) like "Added in Robot Framework 3.0.1" needed in docs.
- Add yourself to AUTHORS.txt if you want.

For more details see CONTRIBUTING.rst in the project root.

src/robot/conf/gatherfailed.py
antechrestos: @pekkaklarck yeah I first implemented this way but had a failure with my tests:
`Processing output failed: IOError: [Errno 2] No such file or directory: u'/tmp/robottests/Python-2.7-Linux/output/output.xml'`
looking closer at the documentation of showed me I missed the necessity to only set it on the root suite :face_with_thermometer:

antechrestos:
@pekkaklarck I took into account your notes:
- I added a note in the *ConfiguringExecution.rst* file
 good luck for your return :smile:

doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
```diff
@@ -165,6 +165,25 @@ running tests.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Command line option :option:`rerunfailedsuites (-S)` can be used to select all failed
+suites from an earlier `output file`_ for re-execution. Like  :option:`--rerunfailed (-R)`, this option is useful when
+full test execution takes a lot of time. Note that all tests from a failed test suite will be re-executed, even
+passing ones. This option is useful when the tests of a test suite depends on each other.
+
+::
+
+  robot tests                                  # first execute all tests
+  robot --rerunfailedsuites output.xml tests    # then re-execute failing
```
pekkaklarck: Do we really need an example here when there's one with `--rerunfailed` and these features work exactly same way?

atest/robot/cli/runner/rerunfailedsuites.robot
```diff
@@ -0,0 +1,44 @@
+*** Settings ***
+Suite Setup       Suite initialization
+Suite Teardown    Remove Directory    ${RERUN SUITE DIR}    recursive
+Resource          atest_resource.robot
+
+*** Variables ***
+${ORIG DIR}           ${DATADIR}/cli/runfailedsuites
+${RERUN SUITE DIR}          %{TEMPDIR}/rerunsuites-dir
```
pekkaklarck: Strange amount of spaces here. Or are there tabs?

doc/userguide/src/Appendices/CommandLineOptions.rst
```diff
@@ -22,6 +22,8 @@ Command line options for test execution
   -R, --rerunfailed <file>  `Selects failed tests`_ from an earlier `output file`_ to be re-executed.
   --runfailed <file>      Deprecated since Robot Framework 2.8.4.
                           Use :option:`--rerunfailed` instead.
+  -S, --rerunfailedsuites <file>  `Selects failed test suite`_ from an earlier `output file`_ to be re-executed.
```
pekkaklarck: This link doesn't match the target. I'll fix it myself.

pekkaklarck:
Reviewing the this PR was a pleasure. I can be pretty nitpicky in my comments, but had absolutely nothing to say about core or tests in the final version. The was one broken link in docs but that was easy to fix. Thanks @antechrestos!

antechrestos:
@pekkaklarck no problem. In fact I like this way because it allows to share knowledge.  Coding this feature brought me the opportunity to look deeper into the beast :smirk:. 
Being nitpicky as you say is useful as it maintains a good quality of code and I respect that.

PR #2488, opened 2016-12-06
Should (not) be equal as strings

chriscallan:
Pull Request to extend the case-sensitivity functionality to the "Should (Not) Be Equal As Strings" keyword pair


pekkaklarck:
PR #2461 contains these keywords too. Which should be merged? Are `Should (Not) Be Equal` implemented somewhere else than in that other PR?

pekkaklarck:
I'm going to merge #2461 which already covers these keywords. Closing this one.

PR #2494, opened 2016-12-15
Implementation of Case Insensitive flag for "Should contain x times"

chriscallan:
Will lowercase the items, if they're strings, or lowercase what it can from the first level of the list-like object.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -71,3 +71,4 @@ Tim Orling                     IronPython support for `Dialogs` library (2.9.2, 
 Jozef Behran                   Fix ${TEST_MESSAGE} to reflect current test message (3.0, #2188)
 Joong-Hee Lee                  Extend 'Repeat Keyword' to support timeout (3.0, #2245)
 Anton Nikitin                  Should (Not) Contain Any (3.0.1, #2120)
+Chris Callan                   Case Insensitive 'Should Contain X Times' (3.0.1, #2494)
--- a/atest/testdata/standard_libraries/builtin/verify.robot
+++ b/atest/testdata/standard_libraries/builtin/verify.robot
@@ -527,7 +527,9 @@ Should Contain X Times With Case Insensitivity
     hello    L    2           ignore_case=True
     hello    eLLo    1        ignore_case=True
     ["This", "is", "it"]    IS      1   ignore_case=True
+    ["This", "is", "it", 12, 22, {"a": "value", "b": 2}]    ["IS", "IT"]      1   ignore_case=True
     ${LIST}    cEE    1       ignore_case=True
+    ${LIST}    12       0     ignore_case=True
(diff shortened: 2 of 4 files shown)


atest/testdata/standard_libraries/builtin/verify.robot
```diff
@@ -522,6 +522,18 @@ Should Contain X Times Failing With Defined Message
     [Documentation]    FAIL My message
     Should Contain X Times    hello    nonex    3    My message
 
+Should Contain X Times With Case Insensitivity
+    [Template]      Should Contain X Times
+    hello    L    2           ignore_case=True
+    hello    eLLo    1        ignore_case=True
+    ["This", "is", "it"]    IS      1   ignore_case=True
+    ["This", ["is", "it"], 12, 22, {"a": "value", "b": 2}]    ["IS", "IT"]      1   ignore_case=True
+    ["This", "is it"]   ["IT", "iS"]    0       ignore_case=True
```
pekkaklarck: This doesn't actually compare lists, only strings.

src/robot/libraries/BuiltIn.py
```diff
@@ -908,18 +908,28 @@ def should_not_contain_any(self, container, *items, **configuration):
...
         Examples:
         | Should Contain X Times | ${output}    | hello  | 2 |
         | Should Contain X Times | ${some list} | value  | 3 |
         """
         count = self._convert_to_integer(count)
+        if is_truthy(ignore_case):
+            item1 = item1.lower() if is_string(item1) else \
+                [x.lower() if is_string(x) else x for x in item1]
+            item2 = item2.lower() if is_string(item2) else \
+                [x.lower() if is_string(x) else x for x in item2]
```
pekkaklarck: Only `item1` can be container. (And should be renamed to `container` like it is with `should_contain`). Should also check is the item list-like before iterating over it.

AUTHORS.txt
```diff
@@ -71,3 +71,4 @@ Tim Orling                     IronPython support for `Dialogs` library (2.9.2,
 Jozef Behran                   Fix ${TEST_MESSAGE} to reflect current test message (3.0, #2188)
 Joong-Hee Lee                  Extend 'Repeat Keyword' to support timeout (3.0, #2245)
 Anton Nikitin                  Should (Not) Contain Any (3.0.1, #2120)
+Chris Callan                   Case Insensitive 'Should Contain X Times' (3.0.1, #2494)
```
pekkaklarck: No need to add AUTHORS info in all PRs related to same issue. It just causes conflicts.

pekkaklarck:
Too many problems in this PR to make it worth merging it. I want to get this ready today so I'll implement this myself. Thanks anyway for reminding that `Should Contain X Times` also needs case-insensitivity support!

Issue #2496, opened 2016-12-19
Enhance documentation of programmatic execution entry points


aaltat:
When my code looks like this:
```
from robot import run

run(
    '/path/to/tests',
    pythonpath='/path/to/libs'
)
```
Then I get error: 
Error in file '/path/to/tests/common/common_library_imports.robot': Importing test library 'foo.Bar' failed: ImportError: No module named foo.Bar

But if my code looks like this: 
```
import sys

from robot import run

sys.path.append('/path/to/libs')
run(
    '/path/to/tests'
)
```
All works fine. 

In user groups someone has asked a similar question[1] but it looks like issue was not raised and therefore it never got fixed. At least could not find an issue which would be closed and says that issue is fixed.

[1] https://groups.google.com/forum/#!searchin/robotframework-users/pythonpath$20robot.run%7Csort:date/robotframework-users/0D__p4amqRU/lK0GK5TgRTMJ

pekkaklarck:
Yeah, `--pythonpath` is set as part command line argument processing and thus omitted when using `robot.run`. Do you really need that when you can control `sys.path` easily yourself? Should at least document what options are handled by which APIs better, though.

aaltat:
At least doc update would be nice to explain what arguments are supported. Is there other arguments that are not supported by the `robot.run` and could be fixed in the PR.

pekkaklarck:
`ArgumentParser` handles these options:

    --pythonpath
    --argumentfile
    --help
    --version
    --escape

Mentioning that these options don't work with `robot.run`, and perhaps also mentioning that all supported options work with `robot.run_cli`, would be nice. Then there's also `TestSuite.run` that handles only options related to actual test execution, but there ought to be some kind of a note about that already.

aaltat:
I can create PR for this. Should we also document the workarounds for the not supported arguments, I think it could be useful? 

The workaround for the `pythonpath` argumentis easy to document that use instead `sys.path.append` or similar methods. But what about the others? I think `help`and `version` do not need workaround methods. Most likely the `argumentfile` and `escape` are not needed either. But what you think?

aaltat:
Created #2500 to fix the issue.

pekkaklarck:
PR #2500 which lists options that are't supported by `robot.run` is merged. On a closer look at the entry point docs I noticed that there's room for other enhancements also. I updated the title of this issue accordingly and will list bigger enhancements separate once I've updated the docs.

[...]

PR #2500, opened 2016-12-25
Updates the run method documention by listing not supported arguments.

aaltat:
The --pythonpath, --argumentfile, --help, --version and --escape
arguments are not supported from the run method. Updated method
documentation to declare the exceptions.


pekkaklarck:
On a closer look the docs of the execution entry points could be enhanced also otherwise. I'll merge this in first and will do additional enhancements to docs afterwards. Don't think these enhancements require a separate issue but I'll update #2496 accordingly.

Issue #2350, opened 2016-04-06
Extend Libdoc to contains information about line and source, where keyword is defined


adam-wypych:
It will be very useful to get in libdoc generated by `robot.libdoc` information about files and lines, where library keyword is defined.

pekkaklarck:
Could you clarify use cases for this kind of functionality? I doubt normal users are so much interested in the source code in general, and adding this kind of information for every keyword would take quite a bit of space.

Adding a link to the library code into the header along with library version and other generic information would be OK, but linking to the local library wouldn't help much docs are hosted online. For example, [BuiltIn](http://robotframework.org/robotframework/latest/libraries/BuiltIn.html) containing a link to the local library version on the machine where the docs were generated wouldn't be useful. Having a link to the [source code online](https://github.com/robotframework/robotframework/blob/master/src/robot/libraries/BuiltIn.py) would be nice, but I don't think there's any way to add it automatically. Libdoc accepting free metadata in name:value format would make that possible, though.

adam-wypych:
Hello Pekka,

Originally the request was made by testers, which are more or less experience with Python. Currently in the RED we are supporting for source code libraries just jump to library file without move just to line, where keyword exists, so I wonder if is possible to extend current libdoc generated to have information `file_as_URI:line_number` per keyword. Mostly users interested in got information about own libraries and I do not see such requests for RF or `Remote` libraries.

Thanks in advance!

pekkaklarck:
I still don't see how this would work in practice. If you have a better idea, please try implementing it and providing a PR. Things to take into account:

- Where to show the information in HTML without wasting too much space?
- How to enhance internal model to contain this information?
- Should this information be written also to XML files produces by Libdoc? If not, Libdoc cannot anymore use those XML files as inputs like it can nowadays.
- How to handle libraries using the dynamic API?
- How to handle libraries implemented using Java?

Finally, this functionality should not be on by default to avoid adding information that may not be useful and will only work if the relative path to the library source code doesn't change.

Notice also that instead of enhancing Libdoc, you could consider using normal Python API doc generators like Sphinx or Epydoc. They ought to support linking to source out-of-the-box or via plugins already.

adam-wypych:
Thanks for Your point. The request was made for libraries means Java classes or Python files, so I agree that made it for resource files is not economical and waste of time (since even RED support it nowdays). The output should be only for libspec made in XML as I see even RIDE only supports description made in XML format. 

[...]

Issue #2507, opened 2016-12-29
Enhanced support for reStructuredText format in library documentation


pekkaklarck:
[reStructuredText](http://docutils.sourceforge.net/rst.html) has been supported as library documentation format since RF 2.7.5 (#489) and RF 3.0.1 (#2448) adds support syntax highlighting using code blocks. It is still not that convenient to use reST, especially if you'd like to use the same docs also with other tools consuming reST like [Sphinx](http://www.sphinx-doc.org). Features that should still be added:

- [ ] Highlight literal blocks created by using `::` automatically in Robot Framework syntax. This should be
  done so that there's no need to add test case table header or other boilerplate. Supporting this would
  be *very* convenient:

      Example::

          ${result} =    My Keyword    argument
          Another Keyword    ${result}

- [ ] Support reST's normal link syntax with a trailing `_` when referring to other keywords or section titles. In other words, keywords and titles should create implicit link targets that can be used everywhere in the docs.

- [ ] Support using explicit link targets created in introduction everywhere in the docs. Related to the above.

- [ ] Format arguments documented using Sphinx supported syntax (at least one of them) nicely.

If the above are implemented, reST would become a very attractive format for documenting larger test libraries like Robot's own standard library and SeleniumLibrary (see robotframework/SeleniumLibrary#389). The problem implementing the above is that `docutils` APIs aren't particularly easy to use externally. Sphinx is able to support similar features in its domain so this shouldn't be impossible for us either.

pekkaklarck:
This would be a great addition, but based on my earlier prototypes I know this is too big task for RF 3.1. Getting help from someone who knows doctutils or Sphinx better that I, or would be interested to study them, would obviously help getting this done.

PR #2514, opened 2016-12-31
#2350 Source lines 

adam-wypych:
#2350 proof-of-concept

### DIFF
--- a/src/robot/libdoc.py
+++ b/src/robot/libdoc.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 
-#  Copyright 2008-2015 Nokia Networks
-#  Copyright 2016-     Robot Framework Foundation
+#  Copyright 2008-2015 Nokia Solutions and Networks
 #
 #  Licensed under the Apache License, Version 2.0 (the "License");
 #  you may not use this file except in compliance with the License.
@@ -77,6 +76,8 @@ Options
  -n --name newname        Sets the name of the documented library or resource.
  -v --version newversion  Sets the version of the documented library or
                           resource.
+ -S --sourcelines         Search source lines for source code keywords where is possible (Jython,
+                            Python).
  -P --pythonpath path *   Additional locations where to search for libraries
                           and resources.
  -E --escape what:with *  Escapes characters which are problematic in console.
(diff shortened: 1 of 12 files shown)


HelioGuilherme66:
I think you should update the files with the current copyright notes.

pekkaklarck:
It's hard to comment the overall functionality when it's not possible to really test it. As far as I can see, this PR just has code to somehow get the line numbers but doesn't actually add them into the HTML output.

The bigger problem here is that, as I commented in #2350, I fail to see use cases to show line numbers in Libdoc's output. You said request come from testers who already know Python, but I would then consider just using normal Python documentation generators (e.g. Sphinx) in the first place rather than Libdoc.

Another problem is that the code uses lot of Robot's internal APIs related to test libraries to get information it needs. These APIs are subject to change in the future, in fact the whole library import/usage logic is likely to be rewritten at some point. That would mean breaking all this code, and we maintainers then would need to fix it. That's not a chore I want to accept unless benefits are very big.

Then there's also the problem that, as I commented in #2350, this functionality could only be supported with a subset of library/resource types Libdoc in general supports.

Instead of trying to add this functionality into Libdoc, I would recommend you to consider these alternatives:

- Implement a custom tool that extends Libdoc. It wouldn't need to care about library/resource types you don't need.

- Switch to other documentation generators. If we enhance Libdoc's reStructuredText support as planned (see #2507), you could actually generate docs from the same source both using Libdoc and some more programmer oriented tool.

adam-wypych:
Hi Pekka,

the lines can be see in `Console View` and `XML` output if You add switch `-S`. I do not even consider `HTML` since normal lib specs imported by RIDE are generated as `XML` and its enough. 
If You need to check it just create simple 'debugBase.py` file with content like below:
```python
class debugBase:
    def done1(self, name):
        return
```
And execute command: `python -m robot.libdoc -S debugBase.py show` and You will see in output additional line: `Source D:\wyp dump\untitled1\untitled1\debugBase.py:done1:2`.

About RF and Python documentation are big difference since if You will take look regarding _decorators (embedded keywords possibility)_, _dynamic keywords API_ and function names You will see that is not so tricky to say what is really keyword name. About real requests just see: https://github.com/nokia/RED/issues/54 . I think Sphinx will do not know where Java keywords exists or dynamic keywords and the source code upload by me deals with Java and Bytecode files, so is not economical to made in each editor own logic to get an information where RF keywords exists. If You check the logic is very simple get method line.

pekkaklarck:
A decision was made not to add this functionality to Libdoc. Seeing source lines is not a feature normal Libdoc users would need, implementation is complex, and it isn't possible to implement this in generic manner with all supported library types. It is better to implement a custom tool that uses Libdoc, [inspect](https://docs.python.org/2/library/inspect.html) and possibly Javadoc internally. See discussion at #2350 for more details.

Issue #2439, opened 2016-09-01
Builtin: Add case-insensitivity support to comparison keywords


HelioGuilherme66:
For reference see [topic on google group](https://groups.google.com/forum/#!topic/robotframework-users/0ncVqpLu0-4).

> (...) Perhaps it could work like this:

```
Should Be Equal    ${string}    expected    ignore_case=True
Should Contain    ${list}    expected    ignore_case=True
```
- Must be backward compatible (default value as before)
- Must have Unit Tests expanded for new option.

---

Update by @pekkaklarck:

These keywords should get the new `ignore_case` argument:
- [x] Should (Not) Be Equal
- [x] Should (Not) Be Equal As Strings
- [x] Should (Not) Start With
- [x] Should (Not) End With
- [x] Should (Not) Contain
- [x] Should Contain X Times
- [x] Should (Not) Contain Any (added in RF 3.0.1, see #2120)
- [x] Should (Not) Match

pekkaklarck:
Yes, this would be nice. If someone is interested to implement this, I'm willing to review PRs. This shouldn't cause backwards compatibility issues so the enhancement can be included to any release. The first step would be listing affected keywords in BuiltIn, String, and possibly in other libraries.

Few things to take into account:
- I'd prefer acceptance tests over unit tests
- Need to decide what is the name of the new option. `ignore_case` is my current favorite, but could also consider e.g. `case_insensitive` (longer) and `caseless` (shorter but not real word).
- What to do if value is not a string? I think it would be best to leave the value as-is.

chriscallan:
Doesn't look like it would be too difficult to implement this, although it could be a death by 1000 cuts if implemented everywhere that a string "could" be used.  
It would seem that adding a case insensitivity option to generic methods like "Should Be Equal" may be a bit error prone with the assumptions about the values sent in.  I think that adding the case sensitivity option to only "Should Be Equal As Strings" is a more logical, and limited, way to implement this functionality while limiting the assumptions about values being passed in.  
Besides, hasn't the consumer of this keyword already declared that they want to compare as strings by indicating they want to do a case-insensitive comparison? :-)
I'd be happy to take this on, preferably the more limited approach I've suggested.

pekkaklarck:
I agree adding this functionality to each and every keyword possibly operating with strings can be too big a task. Better to add this to the most important keywords only and then add it to new keywords later if needed. That said, I think there are quite many important keywords that ought to get this functionality from the beginning:

[...]

PR #2518, opened 2017-01-11
 "Should (Not) Start With" keywords with case insensitivity option

chriscallan:
Sorry for the delay in getting this one done, trouble finding any free time lately.
I've implemented the ignore_case option for these two keywords, included updated acceptance tests, and took a swing at adding "(case-insensitive)" to the failure message string.


pekkaklarck:
I've already implemented this. I wanted to get it into RF 3.0.1 that was released last friday. See #2439 for details.

PR #2521, opened 2017-01-17
Adds "context" to invoke>=0.13 methods.

HelioGuilherme66:
As mentioned on `invoke` changelog, there was a backward incompatible change since version 0.13 [see changelog](https://github.com/pyinvoke/invoke/blob/master/sites/www/changelog.rst#id51)

This PR fixes invoke tasks on Python 2.7.

### DIFF
--- a/BUILD.rst
+++ b/BUILD.rst
@@ -12,8 +12,8 @@ executed on the command line.
 Using Invoke
 ------------
 
-Some tasks are automated using `Invoke <http://pyinvoke.org>`_. Generating
-releases requires it to be installed::
+Some tasks are automated using `Invoke <http://pyinvoke.org>`_ (version 0.13 or
+ newer). Generating releases requires it to be installed::
 
     pip install invoke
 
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Looks good on a quick look. Did you test different tasks too?

Few things that should still be done:

1. There shouldn't be need to `from invoke import run` anymore.

2. Should mention in BUILD.rst under the Using invoke section that invoke 0.13 or newer is needed.

3. Could also check in tasks.py that invoke is 0.13 or newer. Perhaps something like this would work:

   ```python
   if invoke.__version_info__ < (0, 13):
       sys.exit('invoke 0.13 or newer needed')
   ```

HelioGuilherme66:
@pekkaklarck Done all your suggestions.

(TODO: fix execfile and basestring calls, to make it compatible with Python 3.x)

pekkaklarck:
Thanks @HelioGuilherme66 for the PR!

PR #2526, opened 2017-01-20
formally close the file in get_binary_file()

sdesbure:
in order to avoid `IOError: [Errno 24] Too many open files:` when many files are opened during a test with several `Get Binary File`


pekkaklarck:
Isn't the file closed automatically by the `with` statement??

PR #2530, opened 2017-01-25
Remote Library: Small grammar fixes and rewordings

kennyballou:


pekkaklarck:
Thanks! As a non-native English speaker I highly appreciate grammar fixes. I'm going to change the titles to use the `Only first word capitalized` format, though, because that's what we use also elsewhere in the User Guide.

kennyballou:
@pekkaklarck: I'll keep the header format in mind if I do some more reading/editing. Thanks for the good work!

PR #2533, opened 2017-01-30
issue 2517: Resize listbox for long values

MisterChild:
Added one line to dialogs_py.py to resize the listbox. Added tests and run was ok for me, but I didn't check other versions. (Only on Ubuntu with python).

As this is my first contribution to a github project I'll be thankful for any help to improve my work.


pekkaklarck:
Tested this locally and test and fix look good. The added test can be combined with the existing test for selecting a value to avoid the need for one more manual click. I'll do that after myself.

Thanks for the PR  @MisterChild!

PR #2540, opened 2017-02-07
Validating variables used as parameters during dryrun

### LINKED ISSUE #2322: Variable validation support to dry-ryn mode
If a keyword uses a variable not defined the dryrun mode doesn't report the error but the execution test fails.
It's possible to add this feature?

I'm using the robotframework version 2.8.6 but I didn't see this feature in the later versions.

Expected: Test should fails in dryrun mode
Actual behavior: dryrun PASS, test run FAIL

Simple example:

```
$ cat Suites/MyTestSuite.robot
*** Settings ***
Resource          ../BusinessObjs/MyTest.robot

*** Test Cases ***
MyTestCase
    mykeyword

$ cat BusinessObjs/MyTest.robot
*** Keywords ***
mykeyword
    Run Keyword If    ${VARIABLE_NOT_DEFINED}    otherkeyword

otherkeyword
    log    test

$ pybot --dryrun Suites/MyTestSuite.robot 
MyTestSuite                                                                   
MyTestCase                                                            | PASS |
MyTestSuite                                                           | PASS |
1 critical test, 1 passed, 0 failed
1 test total, 1 passed, 0 failed


$ pybot Suites/MyTestSuite.robot 
MyTestSuite                                                                   
MyTestCase                                                            | FAIL |
Variable '${VARIABLE_NOT_DEFINED}' not found.
MyTestSuite                                                           | FAIL |
1 critical test, 0 passed, 1 failed
1 test total, 0 passed, 1 failed
```

rlucca:
During dryrun, it is possible to verify that a variable
used as argument is not missing from the variables section (*** Variables ***).
But it will not work obviously for keywords like Set Variable.

Fixes #2322

### DIFF
--- a/src/robot/running/librarykeywordrunner.py
+++ b/src/robot/running/librarykeywordrunner.py
@@ -113,7 +113,7 @@ class LibraryKeywordRunner(object):
         if self._handler.longname in self._executed_in_dry_run:
             self._run(context, args)
         else:
-            self._handler.resolve_arguments(args)
+            self._handler.resolve_arguments(args, context.variables)
 
 
 class EmbeddedArgumentsRunner(LibraryKeywordRunner):
--- a/src/robot/running/userkeywordrunner.py
+++ b/src/robot/running/userkeywordrunner.py
@@ -199,7 +199,7 @@ class UserKeywordRunner(object):
             self._dry_run(context, kw.args, result)
 
     def _dry_run(self, context, args, result):
-        self._resolve_arguments(args)
+        self._resolve_arguments(args, context.variables)
         with context.user_keyword:
             timeout = self._get_timeout()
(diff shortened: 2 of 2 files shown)


pekkaklarck:
Unfortunately it is not this easy. If it was, it would have been implemented already ages ago. There are way too many cases how to create variables that aren't handled in dry-run and would cause false negatives if this approach would be taken. You can, for example, see some failures in Robot's acceptance tests if you run `atest/run.py python atest/robot/cli/dryrun/` and I doubt they even cover all cases.

Are you interested to look at how these false positives could be avoided? We could also make variable validation opt-in with a separate command line option, but at least some the most common variable usages should be handled correctly for it to make sense.

Issue #2471, opened 2016-11-09
Support `[...]` syntax with glob patterns


nottyo:
According to the robotframework documentation, there is an option "--test" which I can specify my test cases name to be run. The document also told me that I can use the glob-like pattern to specify my test case name. For an example
My Test Case Name
1. TestCase1
2. TestCase2
3. TestCase3

And I want to run the test only 'TestCase1' and 'TestCase2' I want to specify the command line using a glob-like range pattern.

    pybot -d result --test TestCase[1-2] MyTest.robot

But it seems like the robotframework support only simple pattern with '**\***' (matching any string) and '**?**' (matching any single character)

I want to request for support more glob-like pattern refer to the basic glob-like pattern here https://en.wikipedia.org/wiki/Glob_(programming) including range pattern e.g. [a-z][0-9]

pekkaklarck:
I'd be fine with this addition. Should also support listing possible characters like `[ax1]` in addition to requested ranges, though. The main reason these weren't added initially is that most of the use cases are covered by the currently supported `?` and `*` and nobody needed `[...]` badly enough for implementation to be worth the effort.

The place where this should be implemented is `u

nottyo:
@pekkaklarck Thanks for your comment. I've summited my contribution with the pull requests #2560. So please review and if you have any concern, feel free to let me know.

pekkaklarck:
One a very quick look the PR looks good. The main problem merging it is that this change is not fully backwards compatible, and I thus don't want to include it in RF 3.0.x minor releases. My plan is to start RF 3.1 development pretty soon (hopefully in April), but there's still a chance that there are enough fixes to warrant 3.0.3 release before that. I try to do a full review for the PR still this week so that we get it into 100%-ready-to-be-merged state.

pekkaklarck:
PR #2560 has been merged some cleanup/fixes done afterwards. This issue can be considered done, but I'm slightly worried how much backwards compatibility issues this is going to create. 

The reason the change is backwards incompatible is that pattern matching where the pattern happens to have `[whatever]` will fail afterwards. Earlier that `[whatever]` has matched a literal string, but after this change that will match a single character from `whatever`.

The same `utils.Matcher` that has been enhanced is used in many places that are thus all affected:
- Various command line options like `--test` mentioned in the description.
- BuiltIn keywords `Should Match`, `Should Not Match` and `Run Keyword And Expect Error`.
- Various `Match` keywords in XML and Collections libraries.

[...]

PR #2560, opened 2017-03-12
[enhancement-2471] update matcher to support more glob-like pattern

nottyo:
I added my code to RF for #2471 issue to support more glob like patterns (test filtering) I referenced the glob-like patterns from the Wiki page https://en.wikipedia.org/wiki/Glob_(programming). Unit tests and acceptance tests are also included in this commit. So please review and if you have any concern, feel free to let me know.


pekkaklarck:
Finally working with RF 3.1 and finally having time to really look at this PR. Changes to code as well as the added tests and documentation look very good.

While running all tests with the changes I noticed few problems, though. They are listed below and because most of them aren't exactly related to this change I can take care of fixing them myself.

- [ ] Lot of our acceptance tests use `Should Match` with patterns containing brackets (e.g. `[ ERROR ] Error occurred: *`) and these obviously fail now. Need to fix them by using some other keyword or escaping `[` and `]` with a question mark (e.g. `? ERROR ? Error occurred: *`).

- [ ] Old implementation apparently was buggy and considered `foo` to be match `foo\n` even when matching was space-insensitive. The new implementation fixes that, and as a result even more of our acceptance tests fail. They can be easily fixed, but we also need to submit an issue about this problem and add explicit tests for it.

- [ ] Some unit tests are broken due to new tests affecting the data they use. Easy to fix.

- [ ] New pattern matching syntax should be documented also with `Should Match` and related keywords. Might be best to add separate section about pattern matching into introduction part of the BuiltIn docs. It seems `Should Match` documentation contains also other problems that can be fixed at the same time.

- [ ] `fnmatch.translate` doesn't seem to work with bytes on Python 3. Probably just best not to support matching bytes on Python 3 at all, but this needs to be made explicit.

All and all, the PR is great but it had surprisingly many side effects. It definitely was a good idea not to include this into minor release.

pekkaklarck:
Tests and documentation give an impression that `[[xxx]]` would match `[xxx]`. The documentation has example

```
--test Example[[5]] # Matches tests with name Example[5].
```

and there are tests like matching `'Weird*[[name]]'` to `'Weird my[name]'`.

I don't think this statement is true. For example, the referenced test passes because `Weird*` matches `Weird my[nam`, then `[[name]` matches `e`, and finally `]` matches `]`. If the `*` would be replaces with ` my`, the test would not pass. I'll update docs and tests accordingly.

nottyo:
Thanks for updates and great fixed @pekkaklarck

pekkaklarck:
Thanks for the PR!

PR #2569, opened 2017-03-31
Update a dead link to google code

davidshepherd7:


pekkaklarck:
Hmm, we haven't really migrated the Python tutorial from Google Code. At least at the moment it seems to be too big a task to happen in the foreseeable future. Probably better to remove references to it from the User Guide entirely.

davidshepherd7:
Ah ok, all I  wanted it for was to get to the example library, which seems to work fine.

vairamsvsjdo:
@pekkaklarck  please provide the python tutorial code from google code in a separate repo. Can contribute in identifying what can be salvaged. This will become a "contrib" project. Which examples from google code are not there in github?
What is that we are loosing from those code?

PR #2582, opened 2017-04-21
adds a visual pie chart of success vs. fail to the report summary

olivier-verville:
This will add a visual pie chart to the head summary section of the html report, representing the success vs. fail. This is more of a visual flavor feature. Was originally designed as an external plugin but it was brought to my attention that it could be nice to integrate within the framework itself.

Preview:
preview

### DIFF
--- a/src/robot/htmldata/rebot/report.html
+++ b/src/robot/htmldata/rebot/report.html
@@ -69,9 +69,38 @@ function setBackground(topsuite) {
     $('body').css('background-color', color);
 }
 
-function addSummary(topsuite) {
+function addSummary(topsuite)
+{
     var opts = {logURL: window.settings.logURL};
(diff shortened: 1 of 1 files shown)


boakley:
Would it be possible to move the graph to the right, to use the space more efficiently?

pekkaklarck:
Few quick comments:

1. Isn't this a bit of duplicate information when the report already has all and critical tests graphs which show this information?

2. If we would add the skip status and remove all/critical distinction, showing pass/fail/skip in a pie chart might be a good idea.

3. It might be a good idea to make it possible to inject custom content to reports and logs. The main problem with that would be declaring which APIs in reports/logs could be considered stable.

4. I'm surprised how simple the implementation looks like!

cjnewman:
Any chance you could do a bar chart instead?

https://www.stevefenton.co.uk/2009/04/pie-charts-are-bad/

Issue #2564, opened 2017-03-24
When using `--suite parent.child`, all test case files with name `child.robot` are parsed 


KrzysztofJozefowicz:
We came to this issue while investigating why robot returns errors from suites which are not being executed (RF 3.0.1).
Common point was that each reported file had the same filename but were in different folder.

From RF UserGuide:

> Test suites and test cases can be selected by their names with the command line options --suite (-s) and --test (-t), respectively. Both of these options can be used several times to select several test suites or cases. Arguments to these options are case- and space-insensitive, and there can also be simple patterns matching multiple names. If both the --suite and --test options are used, only test cases in matching suites with matching names are selected.

But more important part:

> Using the --suite option is more or less the same as executing only the appropriate test case file or directory. One major benefit is the possibility to select the suite based on its parent suite. 

As far as I understands, when running following command, Robot will execute selected suite and testcase (if specified by --test):
`python -m robot.run --suite A.B.Suite --test  myTest  <root_robot_test_folder>`

We observed when somewhere inside <root_robot_test_folder> there are multiple files with the same name ( Suite.robot) with fatal errors (for instance duplicated Documentation in Settings part), Robot reports each of those problems for files which in fact were not anywhere executed (different parent path).

> root_robot_test_folder structure:
> A/B/Suite.robot
> C/Suite.robot (with duplicated Docs)

Command output will run A.B.Suite tests but also will report errors inside C/Suite.robot (C.Suite), using --test does not change output.

Is it desired behavior to scan for files outside parent path even though Robot get exact path from command line? Looks like "too eager" for me.

pekkaklarck:
Does this happen also if you don't specify `--test`?

KrzysztofJozefowicz:
Yes,it produces the same output as with --test (suite name is searched regardless of parents). 

In zip you will find reproduction suite. in A/B/main.robot is a test to run, C/D/main.robot is a suite with Robot Error (duplicated docs section in settings), E/F/main.robot - a proper suite like in A/B. output reports for commands:
**python -m robot.run --suite A.B.main d:\robotTest**  => Report_justSuiteAB.zip , output: executed A\B\main,robot with 20170324 20:32:38.868 	ERROR 	Error in file 'd:\robottest\C\D\main.robot': Setting 'Documentation' used multiple times. 
[...]

PR #2583, opened 2017-04-25
Parse only relevant test cases for -suite parent.child option (#2564)

andreeakovacs:
Fix for #2564.


src/robot/parsing/populators.py
```diff
@@ -199,17 +199,27 @@ def _extension_is_accepted(self, ext, incl_extensions):
...
             return base not in self.ignored_dirs or ext
         if not self._extension_is_accepted(ext, incl_extensions):
             return False
-        return self._is_in_included_suites(base, incl_suites)
+        return self._is_in_included_suites(base, incl_suites, self._get_longname(base, datadir, incl_suites))
 
-    def _is_in_included_suites(self, name, incl_suites):
+    def _is_in_included_suites(self, name, incl_suites, longname=None):
+        if longname:
+            return not incl_suites or incl_suites._match_longname(longname)
```
pekkaklarck: Using private method `incl_suites._match_longname` is not OK. It also seem to be unnecessary because `incl_suites.match` would accept `longname` as an optional argument.

src/robot/parsing/populators.py
```diff
@@ -199,17 +199,27 @@ def _extension_is_accepted(self, ext, incl_extensions):
...
         return not incl_suites or incl_suites.match(self._split_prefix(name))
 
     def _split_prefix(self, name):
         return name.split('__', 1)[-1]
+
+    def _get_longname(self, base, datadir, include_suites):
+        if datadir.parent and len(include_suites._matcher) > 1:
```
pekkaklarck: Accessing private attribute `include_suites._matcher` is not OK. Also, I don't understand why it must have more than one entry. If more than zero entries were fine, it would be possible to use just `if datadir.parent and include_suites:`.

src/robot/parsing/populators.py
```diff
@@ -160,10 +160,10 @@ def _directory_is_included(self, path, incl_suites):
         name = os.path.basename(os.path.normpath(path))
         return self._is_in_included_suites(name, incl_suites)
```
pekkaklarck: Should't inclusion be checked against both the normal name and the long name also here?

src/robot/parsing/populators.py
```diff
@@ -199,17 +199,28 @@ def _extension_is_accepted(self, ext, incl_extensions):
...
         if not self._extension_is_accepted(ext, incl_extensions):
             return False
-        return self._is_in_included_suites(base, incl_suites)
+        longname = self._get_longname(base, datadir, incl_suites)
+        return self._is_in_included_suites(base, incl_suites, longname)
 
-    def _is_in_included_suites(self, name, incl_suites):
-        return not incl_suites or incl_suites.match(self._split_prefix(name))
+    def _is_in_included_suites(self, name, incl_suites, longname=None):
+        return not incl_suites or incl_suites.match(longname or self._split_prefix(name), longname)
```
pekkaklarck: What's the motivation to use `longname` in two places? If `longname` is given, this call is `incl_suites.match(longname, longname)` and that looks strange.

src/robot/parsing/populators.py
```diff
@@ -199,17 +199,28 @@ def _extension_is_accepted(self, ext, incl_extensions):
...
 
     def _split_prefix(self, name):
         return name.split('__', 1)[-1]
+
+    def _get_longname(self, base, datadir, include_suites):
+        if datadir.parent and include_suites:
+            if not ['.' in x for x in include_suites][0]:
```
pekkaklarck: I don't understand this code at all. Could you clarify?

atest/testdata/misc/parent_suite/child_suite2/test.robot
```diff
@@ -0,0 +1,7 @@
+*** Settings ***
+Documentation  Should not execute
+Documentation  Should not execute
+
+*** Test Cases ***
+Child Suite Test
+  Log  Should not execute!
```
pekkaklarck: This test suite, as well as the other added suite, is executed by other tests running `atest/testdata/misc`. There are some tests that validate that correct suites/tests are run in that case, and they are going to fail because new suites/tests are added. Also, it's a bit confusing that a suite with documentation `Should not execute` (twice) is actually executed. If these suites are added, the affected tests must naturally be fixed.

An alternative would be just using the existing stuff under `atest/testdata/misc`. It seems that the main reason this suite is added is to validate it is not parsed at all, and that is done by looking there's no error about duplicate documentation in the stderr. Could the actual test be changed to use the existing `warnings_and_errors.robot` somehow? It does't have duplicate documentation, but it has a non-existing setting that causes a similar error.

pekkaklarck:
Added some more comments to the code (and tests) after earlier enhancements.

atest/testdata/misc/parent_suite/child_suite2/test.robot
mihaiparvu: Went on and used the first proposal in the updated PR.
I completly re-done the directory structure, now we will have two suites:
-	Suite Should.Be.Parsed
-	Suite Should.Not Be.Parsed

The second suite contains a test with a non-existing setting that should not be normally encountered when executing with command line option `--suite suite_should.be.parsed`. I also updated the unit tests with the new suites/tests.

Regarding the alternative proposal, I thought about different ways to do this but didn't find one that could rely only on the `warnings_and_errors.robot`. Or to use it elegantly with another test file. If you could give me a suggestion I would glady update the PR.

src/robot/parsing/populators.py
mihaiparvu: Yes, you are right! Fixed it.

src/robot/parsing/populators.py
mihaiparvu: There were some problems with this code, changed it altogether in the new commit.

src/robot/parsing/populators.py
mihaiparvu: I didn't find a case that would benefit from this, so I don't think so.

mihaiparvu:
Thanks for the feedback @pekkaklarck! The new PR contains changes based on your suggestions (and questions). All the acceptance tests are passing now.
The unit test `test_included_suites_with_dot` has been removed, because it conflicts with the change.

src/robot/parsing/populators.py
```diff
@@ -145,7 +145,7 @@ def _populate_children(self, datadir, children, include_extensions,
 
     def _get_include_suites(self, path, incl_suites):
         if not isinstance(incl_suites, SuiteNamePatterns):
-            incl_suites = SuiteNamePatterns(self._create_included_suites(incl_suites))
+            incl_suites = SuiteNamePatterns(incl_suites)
```
pekkaklarck: If `_create_included_suites` isn't used anymore, the helper method itself should also be removed.

atest/robot/core/filter_by_names.robot
```diff
@@ -157,6 +157,8 @@ Filter Using Suite Long Name
     Run Suites    --suite Subsuites.Sub1
     Should Contain Suites    ${SUITE}    Subsuites
     Should Contain Tests   ${SUITE}   SubSuite1 First
+    Run Suites    --suite suite_should.be.parsed  ${DATA DIR}/misc
```
pekkaklarck: This is a bit nitpicky, but we try to be consistent and use four spaces between arguments (unless aligning arguments otherwise makes sense). Here there's two spaces before the last argument, and also the newly added suites have just two space indentation.

atest/robot/core/filter_by_names.robot
```diff
@@ -157,6 +157,8 @@ Filter Using Suite Long Name
...
+    Stderr Should Be Empty
```
pekkaklarck: It would be good to test that the right tests, and only them, have been actually run. Should be pretty easy with `Should Contain Suites` that's used already above.

Related to this, it would be better to have different names in tests in the new `should/be/parsed.robot` and `should/not be/parsed.robot` files to make it explicit which test we expect to be executed. A name like `A test in 'should/be/parsed.robot'` would be extremely explicit.

pekkaklarck:
Thanks for the PR @andreeakovacs and @mihaiparvu!

michalanglart:
Hi guys @mihaiparvu, @pekkaklarck !

We've found out just recently (https://github.com/nokia/RED/issues/262) that RF3.1b1 breaks  `--suite` (and probably `--test`) selector when folders are using dots in their names and this change looks like the one which is responsible. 

Create following structure:
```
A
|__B.C
   |__ D.robot
```
Please have a look at following command line session and outputs:
```
python -m pip install robotframework==3.1a2
>> Successfully installed robotframework-3.1a2

python -m robot.run --version
>> Robot Framework 3.1a2 (Python 3.7.1 on win32)

python -m robot.run --suite A.B.C c:\path\to\directory\A
>> ... PASS ...

python -m pip install robotframework==3.1b1
>> Successfully installed robotframework-3.1b1

python -m robot.run --version
>> Robot Framework 3.1b1 (Python 3.7.1 on win32)

python -m robot.run --suite A.B.C c:\path\to\directory\A
>> [ ERROR ] Suite 'A' contains no tests in suite 'A.B.C'
```

@mihaiparvu Can you confirm?

pekkaklarck:
I can confirm the problem and this change is most likely the culprit. Do you @mihaiparvu have time to look at this? I was planning to create RF 3.1 rc 1 today but this is now a blocker.

If fixing this issue is complicated, it is better to revert the change. The benefits are pretty small anyway.

pekkaklarck:
Regression fixed now. Can you @michalanglart test the fix? That can wait until rc1 if you don't have the project cloned.

michalanglart:
Sure, I'll check that! Just give me couple minutes

michalanglart:
Everything looks fine to me with the dev version. Thank you @pekkaklarck

pekkaklarck:
Awesome! Thanks for testing!

pekkaklarck:
Issue #3029 explains two serious regressions caused by this PR. There doesn't seem to be easy way to fix them without reverting these changes.

Issue #2576, opened 2017-04-11
Possibility to configure number of error message lines in reports


mchfrnc:
Hi. I have a test, where I’m expecting keyword’s fail, which causes another keyword’s fail and I would like to catch a full (long) error message generated by the first one. But I met an limitation of maximum number of error lines visible.
Please see the following issue: https://code.google.com/archive/p/robotframework/issues/340
As far as I know it is impossible to change maximum error lines during execution from CLI. 
The only way is to change it directly in the Robot Framework source code, http://robot-framework.readthedocs.org/en/2.9.1/_modules/robot/utils/text.html, and setting _MAX_ERROR_LINES variable.

My temporary workaround is to use sed in cron (it is executed very often to avoid situation, when after some updates we have MAX_ERROR_LINES set back to 40): 

`sudo sed -i 's/_MAX_ERROR_LINES = 40/_MAX_ERROR_LINES = 999/g' /usr/local/lib/python2.7/dist-packages/robot/utils/text.py`

pekkaklarck:
Do you propose making this configurable somehow? I'm OK with that but it needs someone to provide a PR.

mchfrnc:
In my opinion there is no need to make it configurable, for example, with a new additional Robot's argument. I would say, if user sets --loglevel TRACE it means that he wants to have as much logs as possible, so maybe a good idea is to set _MAX_ERROR_LINES to 999 lines (or whatever, bigger number) when loglevel is set to TRACE and some smaller number when loglevel is set to INFO.

pekkaklarck:
I don't think log level should affect this. You already now see the full error message in the log file regardless the log level you use.

PR #2589, opened 2017-05-04
Added --maxerrorlines CLI option (#2576)

muresanrazvan:
Fix for #2576 
Added CLI option for the number of error lines to be displayed on the console output.


muresanrazvan:
I have tried running different scenarios (even programmatically), the value is set back to the original when RobotFramework.main() method is called. The RobotSettings object contains the default value (if no --maxerrorlines option is passed).

src/robot/utils/text.py
```diff
@@ -27,6 +27,9 @@
 _ERROR_CUT_EXPLN = '    [ Message content over the limit has been removed. ]'
 _TAGS_RE = re.compile(r'\s*tags:(.*)', re.IGNORECASE)
 
+def set_max_error_lines(limit):
+    global _MAX_ERROR_LINES
+    _MAX_ERROR_LINES = limit
```
pekkaklarck: This helper isn't really needed. Better to just rename the constant to `MAX_ERROR_LINES` and let external code to set the module global value directly like

```python
from robot.utils import text

text.MAX_ERROR_LINES
```

src/robot/run.py
```diff
@@ -442,6 +444,7 @@ def main(self, datasources, **options):
             suite.visit(ModelModifier(settings.pre_run_modifiers,
                                       settings.run_empty_suite, LOGGER))
         with pyloggingconf.robot_handler_enabled(settings.log_level):
+            text.set_max_error_lines(settings.max_error_lines)
             result = suite.run(settings)
             LOGGER.info("Tests execution ended. Statistics:\n%s"
                         % result.suite.stat_message)
```
pekkaklarck: As commented below, `set_max_error_lines` isn't really needed because the code here can just set `text.MAX_ERROR_LINES = settings.max_error_lines`. The code here should store the old value and reset if after execution.

src/robot/conf/settings.py
```diff
@@ -157,6 +159,13 @@ def _validate_log_level_and_default(self, log_level, default):
             raise DataError("Default visible log level '%s' is lower than "
                             "log level '%s'" % (default, log_level))
 
+    def _process_max_error_lines(self, name, value):
+        value = self._convert_to_integer(name, value)
+        if value < 0:
+            raise DataError("Option '--%s' expected positive integer value but got '%s'."
+                            % (name.lower(), value))
+        return value
```
pekkaklarck: 1. We could consider making it so that negative values like `-1` means not cutting the message at all. Alternatively could use `NONE` for that purpose because `--maxerrorlines NONE` would be more explicit. Not sure is this really needed, though.

2. Very small values, especially 0, aren't really valid. They'd make the message *really* short and I'm not sure would our splitting algorithm work with such values either. Should we make it so that the value should be at least 10?

doc/userguide/src/CreatingTestData/CreatingTestCases.rst
```diff
@@ -446,9 +446,10 @@ some keywords allow configuring them.
 In some circumstances, for example when continuable failures are used,
 a test case can fail multiple times. In that case the final error message
 is got by combining the individual errors. Very long error messages are
-automatically cut from the middle to keep reports_ easier to read. Full
-error messages are always visible in log_ file as a message of the failed
-keyword.
+automatically cut from the middle to keep reports_ easier to read. The
+default limit for the number of error lines is 40. The limit can be changed
+with `--maxerrorlines` command line option. Full error messages are always
+visible in log_ file as a message of the failed keyword.
```
pekkaklarck: Should mention that this is new in RF 3.1. Either by adding a separate note about it or using something like `Starting from Robot Framework 3.1, the default limit ...`.

mihaiparvu:
@muresanrazvan is no longer part of our team, inside the company, but I have access to his repository and will make the changes. 
Hopefully before the RF 3.1 beta 1.

src/robot/conf/settings.py
mihaiparvu: 1. Opted for using `NONE` for not cutting the message. `-1` seems a bit confusing to me and would require an extra check in `_process_max_error_lines()` from `settings.py`. Documented this in `CreatingTestCases.rst` - not sure if it has to be specified in other places too.
2. A limit of 10 sounds good.

mihaiparvu:
@pekkaklarck thank you for the detailed review. See my responses to your comments below:
1. Done
2. Updated documentation and `--help`. Also included details about the minimum value of 10. Wasn't sure if it's relevant of not, but I can remove them if need be.
3. Done. Created a new variable to store the default value before changing it in the unlikely event the default will be changed from `40`, in the future. Again, if necessary I could modify to use the hardcoded `40.`
4. Used Tidy for that :-) sorry for the reviewing troubles.

Because the pull request was submitted a long time ago, after merging with master had to also update the acceptance test by escaping the square brackets from `[ Message content over the limit has been removed.]`.

pekkaklarck:
This PR looks very good. There are some minor enhancements/fixes that could be done, but I do them myself after merge. Thanks @muresanrazvan and @mihaiparvu!

PR #2605, opened 2017-05-29
Fix demo links in user guide

compact:
The current links lead to an error page on Bitbucket that says "This repository does not have wiki enabled."


pekkaklarck:
Demos have been moved to GitHub and links nowadays point to them.

PR #2607, opened 2017-05-30
add try/catch block to prevent SecurityErrors when accessing localSto…

### LINKED ISSUE #2606: Content Security Policy (CSP) causes problems with report
I noticed a javascript error in report.html when [Content Security Policy](https://content-security-policy.com/) (CSP) is enabled and "sandbox allow-same-origin" is not set. CSP is activated in Jenkins by default since Jenkins 1.641 / Jenkins 1.625.3.

When clicking a tag in the report.html the report output should be filtered by the selected tag and the result shown below "Test Details". However with recent browsers (Edge & Firefox 48+) the result remains empty and the Console shows the error "SecurityError: The operation is insecure.".

As a workaround this can be circumvented by disabling CSP or setting "sandbox allow-same-origin". 

However, the javascript should not fail at this point only because the access to the browser's localStorage is prevented. The report should still work and be more robust in this case.

I have created a fix and will add a pull request.

PeerMueller:
…rage

Fixes  #2606 catching exceptions that may occur when accessing browser's localStorage object, see [explanation](https://mathiasbynens.be/notes/localstorage-pattern#comment-9)

### DIFF
--- a/src/robot/htmldata/common/storage.js
+++ b/src/robot/htmldata/common/storage.js
@@ -8,21 +8,19 @@ storage = function () {
 
     function get(name, defaultValue) {
         try {
-            if (!localStorage)
-                return defaultValue;
             var value = localStorage[prefix + name];
-            if (typeof value === 'undefined')
-                return defaultValue;
-            return value;
         } catch (exception) {
             return defaultValue;
         }
+        if (typeof value === 'undefined')
+            return defaultValue;
+        return value;
+
     }
 
     function set(name, value) {
         try {
-            if (localStorage)
-                localStorage[prefix + name] = value;
+            localStorage[prefix + name] = value;
         } catch (exception) {}
     }
 
--- a/src/robot/htmldata/common/storage.js
+++ b/src/robot/htmldata/common/storage.js
@@ -7,17 +7,23 @@ storage = function () {
     }
 
     function get(name, defaultValue) {
-        if (!localStorage)
+        try {
+            if (!localStorage)
+                return defaultValue;
+            var value = localStorage[prefix + name];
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Catching exception when getting/setting values is probably the best we can do. I'd prefer try/catch to wrap only the actual getting/setting parts, though.

PeerMueller:
Ok, I adapted the PR and now only the get and set calls are wrapped. Also removed the obsolete "if (localStorage)" clauses

pekkaklarck:
1. Why is `if (localStorage)` obsolete?

2. What do you think about storing values in a normal Javascript object that we store on the `storage` namespace if `localStorage` cannot be used? That ought to make it possible to store values at least temporarily in memory. Would that bring some functional benefits?

PeerMueller:
1. When access to localStorage is disabled calling `if (localStorage)` results in an exception. We can also just "use" the localStorage object by setting/getting values and catching exceptions to detect whether its available.
2. I'm not sure whether using a replacement is necessary at all. In my analysis the localStorage has only been used to store whether the user has chosen to hide certain columns (such as "Tags", "Elapsed Time") from the test details table. When the localStorage object is not available the report will "forget" which columns where chosen to be hidden when the output changes (i.e. the user clicks a different tag or changes the filter). This would be acceptable in my opinion unless the localStorage is used for additional purposes.

pekkaklarck:
Decided to implement this a bit differently. Mainly wanted to always have a storage available and now use a normal object if `localStorage` isn't accessible. The change is in commit cb6ec41055bac03ba34ff6129bfdd6f85fcb1845.

Issue #2594, opened 2017-05-15
`Run Keyword And Return` not returning in teardown context if executed test fails


juhamkilpinen:
"Run And Return" keywords do not return when they should in teardown. If executed keyword fails the execution continues. Example:

```
*** Test Cases ***
TEST_001
    [Teardown]    Test Teardown
    Log    Testing

*** Keywords ***
Test Teardown
    Do Some Teardown

Do Some Teardown
    Run Keyword And Return If    True    Fail    Should return from keyword now
    Run Keyword And Return    Fail   This should also return now
    Log    This should never happen    console=yes
```

Robot Framework 3.0.2, Python 2.7.13

pekkaklarck:
Could you please modify your example so that it can be just copied to a file and executed? It would make investigating the problem faster. Even better if you can put it into a highlighted code block to make reading the example easier here too.

ombre42:
The problem is that `Run Keyword And Return If`, `Run Keyword If`, and `Run Keyword And Return` are all failing because the keywords they called failed. In this scenario, the teardown behavior by design is to continue executing the next keyword. In order to get the behavior you desire, you need to use a `Run Keyword` variant that will not fail. I suggest `Run Keyword And Return Status` or `Run Keyword And Ignore Error`. These keywords will not fail, and so you can return early if you desire.

```
*** Test Cases ***
Teardown 1
    [Teardown]    My Teardown 1
    Log    Testing
Teardown 2
    [Teardown]    My Teardown 2
    Log    Testing
*** Keywords ***
My Teardown 1
    Some Keyword 1
Some Keyword 1
    ${passed}    Run Keyword And Return Status    Fail    Should return from keyword now
    Return From Keyword If    not $passed
    Log    Some Keyword 1 still executing    console=yes
My Teardown 2
    Some Keyword 2
Some Keyword 2
    ${status}    ${ret val}    Run Keyword And Ignore Error    Fail    Should return from keyword now
    Return From Keyword If    $status != 'PASSED'
    Log    Some Keyword 2 still executing    console=yes
```

juhamkilpinen:
[...]

PR #2608, opened 2017-05-30
Fix for Run Keyword And Return in teardown #2594

juhamkilpinen:
In teardown context, keywords "Run Keyword And Return" and
"Run Keyword And Return If" did not return from the enclosing
user keyword if the given keyword execution failed.


pekkaklarck:
The underlying bug was fix by a similar change in d0b9d53878d80676ab46d4af4306d84c527835e0.

PR #2613, opened 2017-06-12
Fix line continuation of comments bug

alexpilk:
Say you have a really long comment:
`# very very very long comment`
Before this fix, writer would split lines of comments in the following way:
```
# very very very
... # long comment
```
This adds unneeded line continuations and breaks tests. After the fix it looks like this:
```
# very very very
# ... long comment
```

### DIFF
--- a/src/robot/writer/rowsplitter.py
+++ b/src/robot/writer/rowsplitter.py
@@ -75,7 +75,10 @@ class RowSplitter(object):
             current, row = self._split(row)
             yield self._escape_last_cell_if_empty(current)
             if row:
-                row = self._continue_row(row, indent)
+                if self._comment_mark in current:
+                    row = self._continue_row(row, indent, True)
+                else:
+                    row = self._continue_row(row, indent)
 
     def _split(self, data):
         index = min(self._get_possible_split_indices(data))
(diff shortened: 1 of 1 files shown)


bkhouri:
why don't you have the following multi-line comment?

```
# Line 1
# line 2
```

src/robot/writer/rowsplitter.py
```diff
@@ -75,7 +75,10 @@ def _split_row(self, row, indent):
             current, row = self._split(row)
             yield self._escape_last_cell_if_empty(current)
             if row:
-                row = self._continue_row(row, indent)
+                if self._comment_mark in current:
```
bkhouri: This will cause the following line to fail, won't it?

`Log      The # of items are ${x}`

alexpilk: I checked it and the answer is no, it wouldn't break it. But there were other problems in my code that I fixed in a new commit. Please see the updated pull request.

alexpilk:
I checked my solution. Turns out it fixed only this king of comments:
`#    Really long comment`
But such comments as this one were still ignored:
`#Really long comment`
I updated the code with a more elegant solution that handles those cases as well. And no, if you have a '#' in keyword or test case name - it won't break it. Please review :)

alexpilk:
Could anyone please review the changes?

pekkaklarck:
Unfortunately I've been busy with other RF related tasks (mainly Selenium2Library) and spending holiday with the family. On a quick look changes look good but some tests are needed.

alexpilk:
Thanks, hope to see it approved soon :)

pekkaklarck:
This PR isn't valid anymore after the new parser introduced in RF 3.2. I also believe the underlying problem has been fixed by it. If not, please submit an issue about the problem and then possibly a new PR. I promise to keep better track on PRs in the future.

Issue #2610, opened 2017-05-31
Support embedding Text test cases into Markdown files


vairamsvsjdo:
Many source control systems including bitbucket support markdown natively. Adding markdown syntax will be much helpful.

pekkaklarck:
Could you clarify where and how you'd like to use Markdown?

vairamsvsjdo:
Markdown is supported in Visual Studio Team Services, Team Foundation Server and Bitbucket. There is no support for reStructuredText.
This will be more helpful to add "Acceptance Documentations" available naturally. Without which it adds additional dependencies to make documentations and scripts available as HTML, PDF documents.

Tset-Noitamotua:
@vairamsvsjdo Are talking about saving robot test as MarkDown (`.md`) files instead of `.robot` or `.txt`?

Tset-Noitamotua:
As we don't have unlimited history on our RF Slack channel I've saved the [Slack ](https://robotframework.slack.com/files/tset-noitamotua/F3VF9BCG4/mdreader_py____version_0_0_1_.py)into a [Gist][1].

[1]: https://gist.github.com/Tset-Noitamotua/75d15a2beb9ab6f1931d3871172ebbbf

vairamsvsjdo:
@Tset-Noitamotua that is exactly I was looking for. @pekkaklarck  please add this into robotframework support. It works excellent. I have about 50 test script files embedded with acceptance criteria documentations and all work perfect.

pekkaklarck:
OK, you meant creating test cases in Markdown files similarly as you can create tests in reStructuredText files nowadays. I'm totally fine with that idea and, as @Tset-Noitamotua commented, there's already been discussion about that earlier on different forums. You could have also meant using Markdown when defining library documentation which is a valid proposal too.

What needs to be done to actually get the requested Markdown support into the framework is proving a pull request. The Gist by @Tset-Noitamotua is a good start but needs a little cleanup. A bigger task is creating tests and wring documentation. If there's anyone interested, please first take a look at the CONTRIBUTING.rst file in the project root. I'm willing to help here or on Slack too.

vairamsvsjdo:
@pekkaklarck will do that appropriately.
Doc for libraries 
Doc,
 code for PEP8 standard
Support for HTML test cases 
Support for text , tsv test cases
Are features in that support comes to my mind.
Identify others please.

pekkaklarck:
[...]

PR #2621, opened 2017-06-19
Support embedding Text test cases into Markdown files

vairamsvsjdo:
Support embedding Text test cases into Markdown files
fixes issue:  https://github.com/robotframework/robotframework/issues/2610


vairamsvsjdo:
Navigating through legacy code constructs in robotframework: 
1. robotframework code several of them don't comply with PEP8 conventions: example, populators.py did not meet PEP8 and comment requirement.
2. where is utest for restreader? (I was looking for a sample to use for creating utest for mdreader.py)
3. Is there a contribution recommendations for using a file as test resource or use the complex route of html reader.? I am not sold on htmlreader test method. thoughts?

src/robot/parsing/mdreader.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def MarkDownReader():
+    class MarkDownReader(object):
+        def __init__(self):
+            self.robot_lines = []
+            self.robot_data = ''
+
+        def robotize(self, md_file):
+            f = StringIO(md_file.read().decode('UTF-8'))
```
pekkaklarck: Wrapping the read data into a `StringIO` makes no sense. Much easier to just iterate over the lines of the original file.

src/robot/parsing/mdreader.py
```diff
@@ -0,0 +1,70 @@
...
+                include_line = False
+                for line in f.readlines():
+                    if not include_line:
+                        include_line = (line.strip().lower() ==
+                                        "```robotframework")
+                    elif line.strip() == "```":
+                        include_line = False
+                    else:
+                        self.robot_lines.append(line)
+                self.robot_data = str(''.join(self.robot_lines))
```
pekkaklarck: Using `str` here is totally unnecessary. It also causes `UnicodeDecodeError` with Python 2 if the data happens to contain non-ASCII characters.

src/robot/parsing/populators.py
```diff
@@ -28,11 +28,11 @@
 from .tsvreader import TsvReader
 from .txtreader import TxtReader
 from .restreader import RestReader
-
+from .mdreader import MarkDownReader
```
pekkaklarck: Should have two empty lines after imports.

src/robot/parsing/populators.py
```diff
@@ -28,11 +28,11 @@
...
 
 READERS = {'html': HtmlReader, 'htm': HtmlReader, 'xhtml': HtmlReader,
-           'tsv': TsvReader , 'rst': RestReader, 'rest': RestReader,
```
pekkaklarck: Fixing small PEP-8 errors here in the code that you edit is fine, but in general it would be better to have them in separate PRs. Having lot of unrelated changes in a PR makes reviewing it harder.

src/robot/parsing/populators.py
```diff
@@ -83,7 +83,7 @@ def start_table(self, header):
         self._populator.populate()
         table = self._datafile.start_table(DataRow(header).all)
         self._populator = self._populators[table.type](table) \
-                if table is not None else NullPopulator()
+            if table is not None else NullPopulator()
```
pekkaklarck: As I already commented, PEP-8 fixes in general belong to own PRs. In this particular case I also happen to like the original style better. Using four spaces for indentation when continuing a line is confusing.

pekkaklarck:
Answers to your questions:

1. I'm sure there are PEP-8 violations still in the code and especially in unit tests. Fixing them is fine, but fixes to code that isn't otherwise edited by this PR should rather be moved into their own PR.

2. I'm not sure does RestReader have unit tests. It could be thoroughly tested with acceptance tests.

3. Could you clarify what you mean with the complex route in testing HtmlReader? Also, please read `CONTRIBUTING.rst` and `atest/README.rst` for information about testing in general.

src/robot/parsing/mdreader.py
```diff
@@ -0,0 +1,70 @@
...
+    # Introduction
+    This is an example markdown text
+    ## Script Example
+    ```robotframework
+    *** Test Cases ***
+    My Test
+        Log hello world from markdown    console=yes
+    ```
+    ## Some more documentation
+'''
```
vairamsvsjdo: agreed, will make appropriate change

src/robot/parsing/mdreader.py
vairamsvsjdo: makes sense, will make appropriate change.

src/robot/parsing/populators.py
vairamsvsjdo: agreed

src/robot/parsing/populators.py
vairamsvsjdo: agreed

vairamsvsjdo:
1. agreed
2. Its at least not there in the test parsing folder.
3. Please ignore that.

vairamsvsjdo:
@pekkaklarck I was able to find atest for rest:
However, have these questions:

1. These tests seems to duplicate:
[rest](https://github.com/robotframework/robotframework/blob/master/atest/testdata/parsing/data_formats/rest/sample.rst)
[restdirective](https://github.com/robotframework/robotframework/blob/master/atest/testdata/parsing/data_formats/rest_directives/sample.rst)

2. As I was reviewing the atest test cases, I am realizing the volume of different use cases that markdown file format needs to support as equivalent to Rest. Does this pull request meet all those test cases or can this be as different features.

vairamsvsjdo:
@pekkaklarck  how is [${RESTDIR}](https://github.com/robotframework/robotframework/blob/master/atest/robot/parsing/data_formats/rest.robot#L7) variable populated? Is it a global command line variable?
Where is the autobuild test script that runs when a PR is requested?

vairamsvsjdo:
@pekkaklarck  made changes as requested

pekkaklarck:
Great! Unfortunately I don't have time for a review in immediate future but I hopefully have time for this next week or week after that.

vairamsvsjdo:
Great, thanks

pekkaklarck:
Development hasn't been too active in the past, but we are starting RF 3.1 development and I'd like to get this enhancement in. On a quick look the approach used looks pretty good, but there are some places where code could be cleaned up a little. For example, imports weren't in format recommended by PEP-8. More importantly, some of the tests looked pretty strange with Robot data inside the code blocks formatted as tables like this:

```
=============  =========  ==============  ===============
  Settings       Value         Value           Value
=============  =========  ==============  ===============
 Default Tags   default1
￼=============  =========  ==============  ===============
```

My guess is that these are copied from the current reStructuredText tests, where this format is used when the idea is to convert reST files to HTML and parse them then. We are going to deprecate and remove the support for using HTML data (as well as reST what is converted to HTML), so supporting anything like that with Markdown doesn't make sense either. The current code doesn't seem to support this either, so I would assume it's enough to just remove these tests.

vairamsvsjdo:
@pekkaklarck  This was assumed at the time of making it work in markdown, we decided in our implementation to use reST format such that it will enable us to see these in markdown with syntax highlighting and work as a document for Acceptance Testing.

Please provide how you envision and will make changes accordingly.
What formats are we looking to support?
How do you see a test script must look embedded with other documentation.

pekkaklarck:
We got new parser in RF 3.2 (#3076) and this PR is incompatible with it. Support for Markdown would be a nice addition, though, and adding it on top of the new parser ought to be easier than it was earlier.

PR #2622, opened 2017-06-20
Fix keyword remover usage error

fvarose:
The `--removekeywords` option supports tags, but this isn't reflected in the error message if used incorrectly:

> [ERROR ] Invalid value for option '--removekeywords'. Expected 'ALL', 'PASSED', 'NAME:<pattern>', 'FOR', or 'WUKS' but got 'TAGS:not_ready'.

This PR simply adds `'TAG:<pattern>'` to the error message.


pekkaklarck:
Good catch and thanks for the fix!

PR #2625, opened 2017-06-26
Fix when using typing check, keyword is not found

kbfu:
For Python 3, when I use typing check, Robot Framework always shows me "no keyword with name" error. I found it should be the deprecated method "inspect.getargspec" used in argumentparser.py.

So I put some new code with "inspect.getfullargspec", everything works fine for me now.

This change should be compatible between Python 2 and 3.
Typing hint refer: https://docs.python.org/3.6/library/typing.html
Inspect refer: https://docs.python.org/3.6/library/inspect.html#inspect.getfullargspec

### DIFF
--- a/src/robot/running/arguments/argumentparser.py
+++ b/src/robot/running/arguments/argumentparser.py
@@ -40,13 +40,14 @@ class _ArgumentParser(object):
 class PythonArgumentParser(_ArgumentParser):
 
     def _get_arg_spec(self, handler):
-        args_tuple = inspect.getfullargspec(handler)
+        if sys.version_info[0] == 2:
+            args, varargs, kwargs, defaults = inspect.getargspec(handler)
+        elif sys.version_info[0] == 3:
+            args, varargs, varkw, defaults, _, _, _ = inspect.getfullargspec(handler)
         if inspect.ismethod(handler) or handler.__name__ == '__init__':
-            args = args_tuple.args[1:]  # drop 'self'
-        else:
-            args = args_tuple.args
-        defaults = list(args_tuple.defaults) if args_tuple.defaults else []
-        return args, defaults, args_tuple.varargs, args_tuple.varkw
+            args = args[1:]  # drop 'self'
+        defaults = list(defaults) if defaults else []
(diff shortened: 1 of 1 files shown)


src/robot/running/arguments/argumentparser.py
```diff
@@ -40,11 +40,15 @@ def _get_arg_spec(self, source):
 class PythonArgumentParser(_ArgumentParser):
 
     def _get_arg_spec(self, handler):
-        args, varargs, kwargs, defaults = inspect.getargspec(handler)
+        if sys.version_info[0] == 2:
+            args, varargs, kwargs, defaults = inspect.getargspec(handler)
+            varkw = kwargs
```
pekkaklarck: Use either `varkw` or `kwargs`, not both.

src/robot/running/arguments/argumentparser.py
```diff
@@ -40,11 +40,15 @@ def _get_arg_spec(self, source):
 class PythonArgumentParser(_ArgumentParser):
 
     def _get_arg_spec(self, handler):
-        args, varargs, kwargs, defaults = inspect.getargspec(handler)
+        if sys.version_info[0] == 2:
```
pekkaklarck: Could also add `from robot.utils import PY2` and use `PY2` here.

kbfu:
@pekkaklarck Thanks for the comments, my bad, I should not use two different variable names for kwargs.

And I have run the atest, seems ok for me. I ran the test on OS X 10.12.5 with Python 2.7.13 installed.
Here is test results:
[Python-2.7-OSX.zip](https://github.com/robotframework/robotframework/files/1107430/Python-2.7-OSX.zip)

pekkaklarck:
This fix would need new tests too. Could probably add them/it into some existing suite.

kbfu:
The acceptance test is currently running under Python 2, since the typing hint is newly added in 3.5, I don't think it will work in the acceptance test scripts if I add some new tests.

pekkaklarck:
This work was continued by @Kompakti in PR #2751 that's been merged.

Issue #2443, opened 2016-09-15
Global variables not accessible in listener `close` method


stasiek:
Opposed to Robot Framework 2.9.1, in Robot Framework 3.0 in the _close() method of a listener, the Robot "automatic variables" (e. g. ${OUTPUTDIR}) are not accessible anymore.

This is a problem, because you might want to use them during the cleanup actions in _close(). Looking at the "automatic variables" they seem to fall into different "scopes", i. e. some are only useful inside keywords, others in test cases, test suites or globally. In my opinion at least those of them which are useful globally should be available in _close().

an example
`tt.robot`

``` robotframework
*Settings*
Library  tt

*Test Case*
tc1
  Log  hello
```

`tt.py`

``` python
from robot.libraries.BuiltIn import BuiltIn

class MyListener(object):
    ROBOT_LISTENER_API_VERSION = 2

    def _close(self):
        print(BuiltIn().get_variable_value("${OUTPUT_DIR}"))

ROBOT_LIBRARY_LISTENER = MyListener()
```

Robot Framework 2.9.1:

```
% PYTHONPATH=. pybot tt.robot
==============================================================================
Tt                                                                           
==============================================================================
tc1                                                                   | PASS |
------------------------------------------------------------------------------
Tt                                                                    | PASS |
1 critical test, 1 passed, 0 failed
1 test total, 1 passed, 0 failed
==============================================================================
/tmp
Output:  /tmp/output.xml
Log:     /tmp/log.html
Report:  /tmp/report.html
```

Robot Framework 3.0:

```
% PYTHONPATH=. pybot tt.robot
==============================================================================
Tt                                                                           
==============================================================================
tc1                                                                   | PASS |
------------------------------------------------------------------------------
Tt                                                                    | PASS |
1 critical test, 1 passed, 0 failed
1 test total, 1 passed, 0 failed
==============================================================================
[ ERROR ] Unexpected error: RobotNotRunningError: Cannot access execution context
Traceback (most recent call last):
  File ".../local/lib/python2.7/site-packages/robot/utils/application.py", line 80, in _execute
    rc = self.main(arguments, **options)
  File ".../local/lib/python2.7/site-packages/robot/run.py", line 434, in main
    result = suite.run(settings)
  File ".../local/lib/python2.7/site-packages/robot/running/model.py", line 247, in run
    self.visit(runner)
  File ".../local/lib/python2.7/site-packages/robot/model/testsuite.py", line 160, in visit
[...]

PR #2626, opened 2017-06-27
Fix for the access to the global variables in listener 'close' method (#2443)

andreeakovacs:
Moved the close of global listeners before the end of running context, which prevents the access to the global variables in 'close' method.


pekkaklarck:
The fix looks pretty simple which is great. Some tests demonstrating that the fix works would be needed, though. If you cannot find good place where to add tests, I can take a look. For general information about Robot's acceptance tests see `atest/README.rst` file.

andreeakovacs:
To test this solution, I ran succesfully acceptance tests with Python2  on Windows and Linux and also the failing example from the issue description. 
I will try to add some new acceptance tests.

pekkaklarck:
Good to hear that the fix doesn't break any existing tests. New tests are definitely needed to verify the fix and to make sure we don't break this again.

andreeakovacs:
The solution was not complete, the global variables concerning suites were still not available, the close of global listeners needed to be done before end of variables scope. 
  
  I've added an acceptance test that checks the access to a part of the global variables in 'close' method, not for the ones that return an absolute path, because they are os and user configuration dependent.

atest/robot/test_libraries/as_listener_global_vars.robot
```diff
@@ -0,0 +1,22 @@
...
+        ...  Suite name:Global Vars Listener
+        ...  Suite documentation:Test global variables in 'close' listener method
+        ...  Previous test name:Global variables test
+        ...  Previous test status:PASS
+        ...  Log level:INFO
+        ...  Suite name:Global Vars Listener
+        ...  Suite documentation:Test global variables in 'close' listener method
+        ...  Previous test name:Global variables final test
+        ...  Previous test status:PASS
+        ...  Log level:INFO\n
```
pekkaklarck: This test verifies that global variables are logged three times. I assume that's correct because the executed suite has two tests, and the library that has test case scope goes out of the scope (i.e. is closed) once per test and once also at the end of the test suite. This could still be made more explicit, and test coverage enhanced in general, like this:

1. Rename this test to something like `Global variables in listener 'close' method when library has test case scope`.

2. Add two more tests for libraries using the test suite and global scope. This requires also new libraries, but it's easy to create them so that they just extend the existing and set `ROBOT_LIBRARY_SCOPE` accordingly. It's possible to have all libraries in a single module, but that requires imports like `Library    module.Library` and that requires putting the directory containing them into `PYTHONPATH`. Might be easiest just to have separate files for each library.

3. Specify a set of variables that are logged as a list variable and then use it in each test as many time as needed.

atest/testdata/test_libraries/as_listener/global_vars_listenerlibrary.py
```diff
@@ -0,0 +1,19 @@
+import sys
+from robot.libraries.BuiltIn import BuiltIn
+from collections import OrderedDict
```
pekkaklarck: This is test code, but imports could still be ordered as recommended by PEP-8.

atest/testdata/test_libraries/as_listener/global_vars_listenerlibrary.py
```diff
@@ -0,0 +1,19 @@
...
+                               ("Previous test name:", "${PREV_TEST_NAME}"),
+                               ("Previous test status:", "${PREV_TEST_STATUS}"),
+                               ("Log level:", "${LOG_LEVEL}")])
+
+    def __init__(self):
+        self.ROBOT_LIBRARY_LISTENER = self
+
+    def _close(self):
+        for var, value in self.global_vars.items():
+            sys.__stderr__.write(var + (BuiltIn().get_variable_value(value)) +"\n")
```
pekkaklarck: Although this is test code, it would be cleaner to use string formatting like `'%s: %s\n' % (var, BuiltIn().get_variable_value(value))` here. Or perhaps  that `global` vars could be just a list of variables to check and this line could be `'%s: %s\n' % (var, BuiltIn().get_variable_value(var))`.

pekkaklarck:
The fix looks pretty simple and it's great that there's a test now as well. I'd like this to be tested a bit more thoroughly, though. If you don't have time or interest, I can create the needed tests myself.

atest/testdata/test_libraries/as_listener/global_vars_listenerlibrary.py
mihaiparvu: Done. `global_vars` is now a list of variables instead of an OrderedDict.

atest/robot/test_libraries/as_listener_global_vars.robot
mihaiparvu: Created two more tests for the other scopes, however I'm not very happy with the naming of the new files/tests. If you @pekkaklarck find better suggestions I will update the PR.

mihaiparvu:
I took over @andreeakovacs' PR, added the missing tests and made the requested changes.

pekkaklarck:
Thanks @mihaiparvu for finishing this PR and @andreeakovacs for the original work!

PR #2664, opened 2017-08-29
default_tags and force_tags attributes in TestSuite object

Kukkerem:
I want to filter and analyze a suite, and I use TestSuiteBuilder.
After I build it, I have to print the suites with tag informations, but the TestSuite object has no such attributes.
The new attributes filled when TestSuiteBuilder build called.


pekkaklarck:
Executable test suites don't anymore need this information. Individual tests in suites have tags set based on these settings.

Issue #2675, opened 2017-09-11
Acceptance tests don't work in py3


Brian-Williams:
# Summary
The atest structures were never converted to be py3 compatible.

## Immediate fix
Replace mix of bytes and string comparisons with all strings. Replace all print keywords with print function calls. Fix unichr calls.

## Long term fix
Setup CI against py3.x, so that these issues can be caught.

Brian-Williams:
I have some of the fixes above in [this](https://github.com/Brian-Williams/robotframework/tree/testing) branch.

pekkaklarck:
Acceptance tests runner side working only with Python 2 is actually by design. There are no real benefits in supporting both Python 2 and Python 3 when you anyway must have both available to be able to test the framework itself on both of them. More importantly, there are absolutely no benefits in spending time testing that such code works with both Python 2 and Python 3.

Instead of changing the code to work both with Python 2 and Python 3, we should make it Python 3 only. This avoids all the annoyances required when writing Python 2 and 3 compatible code, removes the need to test on both variants, and makes it possible for us to use all the goodies Python 3 has. I'd actually go as far as saying that we can make the runner side Python 3.6+ only to be able to use the great `f'string'` syntax. That's what I already did with https://github.com/robotframework/rellu/.

The CI issue is important. We used to have comprehensive CI but it was maintained by a company not involved with Robot Framework anymore. Back then we executed tests with Python (2.6-3.5), Jython and IronPython on Linux and Windows, and AFAIK this is not something Travis can provide. I'm fairly certain we get CI organized somehow via [Robot Framework Foundation](http://robotframework.org/foundation/), but I've got other more important tasks and haven't spent time with that. I hopefully have time to start RF 3.1 development this autumn and then working CI is a must.

Brian-Williams:
Ah, that makes a lot of sense.

I wasn't aware robot was planning on dropping py2 support with more than a year left of py2 support. Glad to hear it though!

Travis only supports linux and osx. I don't know of a free to open-source CI tool that currently includes windows as an os option. I've only used gitlab ci, travis and shippable and know that shippable has windows support on their roadmap.

```
os:
  - linux
  - osx
```

I will close this issue as it is invalid. Would you like me to leave the [pull](https://github.com/robotframework/robotframework/pull/2676) open as a half-measure until a more comprehensive Robot Framework Foundation guided implementation can be added?

pekkaklarck:
[...]

PR #2676, opened 2017-09-11
Add .travis.yml

Brian-Williams:
# Summary
Add travis unit testing for python versions and acceptance testing for py2.7. [Successful](https://travis-ci.org/Brian-Williams/robotframework/builds/274342159) build on fork.

### pypi and jython addition problems
Also additional unit tests could be added like shown in [this branch](https://github.com/Brian-Williams/robotframework/blob/add-more-ci/.travis.yml), however there are currently pypy and jython unit test [errors](https://travis-ci.org/Brian-Williams/robotframework/builds/274468773) the pypy appears to be an actual failure with OrderedDicts, however I don't know much about jython.

We need a before_install script to correctly install jython.

### Configuration suggestion
Because a github pages branch exists. 'Build only if .travis.yml is present' should be turned on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,26 @@
+language: python
+python:
+  - "2.7"
+  - "3.5"
+  - "3.6"
+  - "pypy"
+  - "pypy3"
+env:
+  - PYTHON_EXE=python
+matrix:
+  include:
+  - python: "2.7"
+    sudo: required
+    env: PYTHON_EXE=jython
+    addons:
+      apt:
+        packages:
+          - jython
+install:
+  - pip install .
+  - pip install invoke -r atest/requirements.txt
+script:
+  - $PYTHON_EXE utest/run.py
+  - if [ "$TRAVIS_PYTHON_VERSION" == "2.7" ];
+    then atest/run.py "$PYTHON_EXE" --exclude no-ciORrequire-screenshot atest/robot;
+    else echo "We only run atest from the py2.7 part of the build matrix."; fi


pekkaklarck:
As I already mentioned in comments of #2675, setting up CI is important but it requires a bit more discussion. Few somewhat random notes about that:

- CI running tests only on Python (+ PyPy) on Linux (or OSX) isn't really enough for us:
  - They can be run locally so fast (< 10min) that benefits on automatic run elsewhere aren't big.
  - PyPy is so close to Python in compatibility that there seldom are problems (and tests run fast on it locally too).
  - Tests take *a lot* more time on Jython and IronPython is pretty slow too. CI would help with them much more.
  - There are often enough Jython/IronPython specific bugs that CI would be *really* useful.
  - There are often enough Windows problems that running tests there would be needed too.
  - Windows is likely needed to run IronPython tests anyway (we haven't tested it on Mono at all)
 
- We used to have CI running tests on Jenkins (using our Jenkins plugin, of course) on Linux and Windows using Python 2.6-3.5, Jython and IronPython. You can read more about it in CONTRIBUTING.rst (which obviously is outdated in that regard nowadays).
  - The company that provided hosting for the main CI machine isn't anymore involved with Robot Framework anymore.
  - I'd like to have similar setup again. Jenkins isn't a must but it did work pretty well.
  - Such a setup is likely to cost something, but [Robot Framework Foundation](http://robotframework.org/foundation/) ought to be able to pay it. Either the foundation will pay directly or some member company can set the CI up.
  - It would be especially awesome if the CI system could then also be used by other projects in the ecosystem.

- If you @Brian-Williams or someone else is interested to help setting up CI, that would be awesome!
  - I don't particularly care about the underlying service but it should OS and Python variants I described above.
  - As I already commented, the Foundation can pay for (reasonably priced) hosting.

- CI isn't urgent for Robot Framework project at the moment, but when RF 3.1 works starts with bigger changes, it's going to be really important.

Brian-Williams:
Additional discussion at `CI` channel on robotframework.slack.com.

hugovk:
I can't access robotframework.slack.com but I think the benefits of turning on Travis CI is better than the current state (no CI or private CI), and it's a very quick win: 

* some test coverage is much better than none
* it's free and ready to go _right now_
* fail fast: quick feedback to the contributor about any failures (eg. simple Py2 v. Py3 problems)
* Travis CI is well-known to open-source contributors on GitHub

None of this would prevent setting up a more thorough testing CI to run as well, or from extending the Travis setup.

(See also AppVeyor CI which may be an option for Windows.)

bkhouri:
@hugovk to access robotframework.slack.com, you need to request access via https://robotframework-slack-invite.herokuapp.com/

MaciejWiczk:
@pekkaklarck, GitHub introduced GitHub Actions CI/CD lately which covers what You mentioned in post above. Currently it's beta and there is a signup need, but after November 13 it will be available to everyone. This solution is free for opensource repositories. More here:
https://github.blog/2019-08-08-github-actions-now-supports-ci-cd/
What do You think?

Brian-Williams:
CI exists now

PR #2684, opened 2017-10-02
fix missing custom message in multi line string diff

DirkRichter:


pekkaklarck:
What's the motivation for this PR?

DirkRichter:
Multiline string diff did not allow custom error message. Support was added.

pekkaklarck:
There are no changes visible anymore. I guess this has been implemented already.

PR #2688, opened 2017-10-14
Update classifiers

hugovk:


pekkaklarck:
We haven't added version specific classifiers to avoid updating them when new Python versions are released. If we say, for example, that we support versions 3.5 and 3.6, it would look like we wouldn't support 3.7 when it is out. This is especially annoying because you cannot update the project metadata in PyPI without creating a new release. The best solution would be having classifiers like `3.5+` or to indicate that also newer versions are OK.

I guess being explicit about versions we supports has enough benefits to counter the above problems, though. Perhaps we could also start listing new Python releases as supported as soon as they hit alpha or beta stage.

The reason I cannot merge this PR at the moment is that it doesn't include Python 2.6 and 3.3 that we currently still support. There's going to be one RF 3.0.x release still with these Python versions supported, and it should have correct metadata.

hugovk:
I've added 2.6 and 3.3.

> We haven't added version specific classifiers to avoid updating them when new Python versions are released. If we say, for example, that we support versions 3.5 and 3.6, it would look like we wouldn't support 3.7 when it is out. This is especially annoying because you cannot update the project metadata in PyPI without creating a new release. The best solution would be having classifiers like 3.5+ or to indicate that also newer versions are OK.

Then make a release :) "Release early, release often", they say. 

I think if it says 3.5 and 3.6, but not 3.7, and 3.7 is just out, most will think it's just that the metadata's not been updated yet. The majority of top PyPI packages explicitly mention versions so it's not at all uncommon. They can always pop over and check the homepage / latest source code for confirmation.

> I guess being explicit about versions we supports has enough benefits to counter the above problems, though. Perhaps we could also start listing new Python releases as supported as soon as they hit alpha or beta stage.

👍 

One approach is to claim support once it's been tested on the new version, or there's enough confidence to believe it works fine.

(It's possible to test against `3.7-dev` on Travis CI (and I expect others) right now, and there's also a 3.7 classifier. Re: https://github.com/robotframework/robotframework/pull/2676)

pekkaklarck:
Perhaps we can think this so that versions that are listed explicitly are tested against. We naturally cannot say anything about Python 3.7 until it is released or at least in beta/rc phase and has all the planned changes.

PR #2691, opened 2017-10-16
Create a xunit v2 reporting writer

RussellBall:
**Provides an xUnit v2 compatible result file.**

This allows the use of VSTS "Publish Test Results" v2.0 with xunit

Attempts to adhere to the schema supplied by http://xunit.github.io/docs/format-xml-v2.html

### DIFF
--- a/xunitv2writer
+++ b/xunitv2writer
@@ -1,0 +1,128 @@
+from robot.result import ResultVisitor
+from robot.utils import roundup, XmlWriter
+
+class XUnitWriter(object):
+    def __init__(self, execution_result, skip_noncritical):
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Could you clarify which tools use this format?

pekkaklarck:
Apparently this xUnit.net v2 XML format is used by various Microsoft tools but it isn't directly related to  the xUnit output we support that's used by jUnit and various other tools. If the xUnit.net v2 XML format is widely used, I'm fine us adding a direct support for it. That would require also tests and documentation, not only code.

Issue #2633, opened 2017-07-14
New keyword in buildin lib: Repeat Keyword And Ignore Errors


PWojtal:
Hello,
We need a keyword which allows us to repeat another keyword in the loop multiple times, let's say 500, and then create small statistics in Robot log (how many failures occurred, how many iterations were made). 
It should pass only if all iterations passed

Sample usage: 

`Repeat Keyword And Ignore Errors    Another Keyword   x500`

I think it may be useful also for other test teams, so I don't want to make it only for our own - it's relatively easy to do task.
Do you think that it is possible to provide us such kw?

Regards,
Paweł

chriscallan:
Sounds like a conglomeration of the Repeat Keyword and Run Keyword And Ignore Error keywords.  I've never had a need for something like that, but obviously you've found a situation where it's useful.  
I'd suggest you create a user-defined keyword that gathers your statistics and loops the keyword, likely easier if you use a 'for' loop internally along with the "Ignore Errors" keyword.  Just a thought.
If other folks would find that useful, they'll weigh in here and a decision can be made about incorporating that via Pull Request.  
Good luck with that.

pekkaklarck:
Sorry for not commenting earlier. I've been mainly working with SeleniumLibrary lately and haven't paid too close attention to Robot's tracker. Anyway, I'm not certain is this a generic enough functionality that should be added to the core framework, especially when there are already related features that ought to solve more or less the same use case.

Could you clarify why you couldn't use test templates instead? Or if they won't work for some reason, why couldn't you then use something like this:

    Repeat Keyword    500x    Run Keyword And Continue On Failure
    ...    My Keyword    arg1    arg2

pekkaklarck:
Now new info lately. Closing.

PR #2700, opened 2017-10-27
New keyword in buildin lib: Repeat keyword and ignore errors

mihaiparvu:
Provided pull request for issue #2633 . @PWojtal let me know if this is the expected behavior.


pekkaklarck:
On a quick look the PR looks pretty good, but I'm not entirely sure is this so useful feature we'd like to add it. As I just commented issue #2633, test templates could typically used to solve this kind of use cases. If they aren't adequate, then it's possible to use the existing keywords like this:

    Repeat Keyword    500x    Run Keyword And Continue On Failure
    ...    My Keyword    arg1    arg2

pekkaklarck:
Issue #2633 has been closed. I guess this can be closed as well.

Issue #2715, opened 2017-11-15
Arguments column is empty in generated HTML documents


LegNBass:
When generating HTML documentation for python-defined robot keywords (using libdoc), the resulting "Arguments" column is empty even when keyword strings contain required args.

![image](https://user-images.githubusercontent.com/6739247/32855638-4cbe9438-ca10-11e7-93ae-1bab5256072a.png)

pekkaklarck:
This is by design, the keyword you have implemented accepts no positional arguments and thus the column is empty. There's even comment "Show empty argument spec for Libdoc" on the line you modified in your PR.

I can see that showing "Arguments" column with keywords containing embedded arguments can be a bit confusing. In my opinion showing the embedded arguments as positional ones would just make the situation more confusing, though, but there are other possibilities we can consider:

- Change the column title to "Positional Arguments".
- Add something like "N/A" to the column with these keywords and/or make the cell background light gray as if was disabled.
- Remove the Arguments column altogether if all keywords have only embedded arguments.

gageorsburn:
May I ask why embedded arguments are not taken into account when generating documentation?

pekkaklarck:
They are. You see them in the keyword name. The arguments column is for *positional* arguments which these keywords don't have.

pekkaklarck:
As I wrote earlier, I don't think embedded arguments should be shown in the arguments column. In addition to reasons listed above, there's a change we allow keywords to have both embedded and normal arguments in the future. I'd be fine changing the column name or otherwise making the situation more explicit, though. Comment here or open a new issue if you think that would be useful.

PR #2716, opened 2017-11-15
Allow arguments to display in libdoc HTML documentation

LegNBass:
Solves issue #2715 

screen shot 2017-11-15 at 2 28 45 pm


pekkaklarck:
As I commented #2715, the current behavior is by design.

PR #2719, opened 2017-11-20
New keyword added for 2706 Convert To Titlecase

bbpatel2001:
Provided pull request for issue # 2706
| ${str1} = | Convert To Titlecase | abc |
| ${str2} = | Convert To Titlecase | i love robot Framework |
| ${str3} = | Convert To Titlecase | i love ROBOT Framework |
| Should Be Equal | ${str1} | Abc |
| Should Be Equal | ${str2} | I Love Robot Framework |
| Should Be Equal | ${str3} | I Love ROBOT Framework |
| ${str4} = | Convert To Titlecase | i love ROBOT    Framework |
| Should Be Equal | ${str4} | I Love ROBOT    Framework |

Preserve spaces as well.


pekkaklarck:
1. Code seems to be totally broken. `''.join()` or something missing around the expression?
2. Tests are needed. See CONTRIBUTING.rst and atest/README.rst for details.
3. The algorithm doesn't seem to handle cases like `jUnit` correctly.

bbpatel2001:
Modified code and added tests as well. 
 #2706_titlecase

kshrestha99:
suggestion: include a separator as an optional parameter. By default it will take `space` as the separator. Like this maybe:
	def title_case(self, string, separator=' '):
		return ''.join( (c.upper() if i == 0 or string[i-1] == separator else c) for i, c in enumerate(string))

pekkaklarck:
Thanks for updates! Few more things:

1. You need to add a test to the "execution side" (i.e. atest/robot) and not only to the data side (i.e. atest/testdata). See atest/README.rst for details.

2. The algorithm still doesn't seem to handle `jUnit` or `iPhone` correctly. I'd expect these *not* to be converted to `JUnit` or `IPhone`. In my opinion words should be title cased only if they are all lower case. Perhaps this could be fixed with regexps:

pekkaklarck:
Do you @kshrestha99 have use cases for using something else than whitespace as a separator?

bbpatel2001:
Hi @pekkaklarck , Should I work on this ticket/code further or your implementation will go as new keyword ?

pekkaklarck:
@bbpatel2001 Please continue working! The regexp based solution just occurred to me while thinking how this could be implemented and wanted to share it. You can use it or some other solution as long as it works correctly. =)

kshrestha99:
@pekkaklarck I was only thinking for scenarios where you would want title case for situations like these:
father-in-law => Father-In-Law
mother_in_law => Mother_In_Law
and the likes where the separator is not whitespace but some other character that the user can input if needed. And if not specified, would default to whitespace as it is now.

pekkaklarck:
That "father-in-law" is a good example! Instead of making the separator configurable, we could enhance my earlier regular expression based implementation proposal to handle it out-of-the-box (another good example!) by using `r'\w+'` instead of `r'\S+'` as the pattern:

kshrestha99:
@pekkaklarck Your regex solution is elegant, I like it a lot, but it will not work when you want to title case a word that already has a capital letter, like "hubSpot" to "HubSpot". Maybe it's not an issue for most users but it could be for some. This issue will not be an issue if we allow users to specify the separator and not use `return string.title() if string.islower() else string` clause.

pekkaklarck:
It's a good question how to handle words like "hubSpot". My algorithm above titlecases words that are all lowercase but leaves others as-is. Another solution would be turning the first letter uppercase always, but that would then also change "iPhone" to "IPhone". The third approach is always using `str.title()`, but it turns "XML" to "Xml" and isn't a good default behavior.

If we want to make this really generic, we can make the algorithm configurable. I'm not sure is it worth the effort, though. Most important is documenting the logic so that users know what they get.

bbpatel2001:
```
def convert_to_titlecase(s,ignore_caps=False,small_second_to_last=False):
  def to_title(match):
    string = match.group()
    if small_second_to_last == True:
      return string.title() 
    if ignore_caps == True:      
        return string[0].upper() + string[1:]   
    else:
        if string.islower() :
          return string.title() 
        else:
          return string 
  return (re.sub(r'\w+', to_title, s))
```
===================== Tests ===================
convert_to_titlecase("i love rObot Framework")
Case 1: Do not Ignore Cap in word/Sentence: i love rObot Framework ==> I Love rObot Framework

convert_to_titlecase(word6,ignore_cap=True)
Case 2: Ignore Cap in word/Sentence: i love rObot Framework ==> I Love RObot Framework

Case 3: Do not Ignore Cap in word/Sentence: jUnit ==> jUnit

Case 4: Ignore Cap in word/Sentence: jUnit ==> JUnit

Case 5: Ignore Cap in word/Sentence: i love rObot      Framework ==> I Love RObot      Framework

Case 6: Ignore Cap in word/Sentence: i\nlove\nrObot\nFramework ==> I
Love
RObot
Framework
Case 7: Do not Ignore Cap in word/Sentence: i\nlove\nrObot\nFramework ==> I
Love
rObot
Framework

Case 8: Do not Ignore Cap in word/Sentence: father-in-law ==> Father-In-Law

Case 9: Ignore Cap in word/Sentence: father#in#Law ==> Father#In#Law

Case 10: Ignore Cap in word/Sentence: hello XML iPhone ==> Hello XML IPhone

Case 11: Do not Ignore Cap in word/Sentence: hello XML iPhone ==> Hello XML iPhone

Case 12: Do not Ignore Cap in word/Sentence: Hyphen Delimited Sub-Phrase With a Small Word in Quotes - 'a Trick, Perhaps?' ==> Hyphen Delimited Sub-Phrase With A Small Word In Quotes - 'A Trick, Perhaps?'

Case 13: Ignore Cap in word/Sentence: Hyphen Delimited Sub-Phrase With a Small Word in Quotes - 'a Trick, Perhaps?' ==> Hyphen Delimited Sub-Phrase With A Small Word In Quotes - 'A Trick, Perhaps?'


word6 = "hello XML iPhone'"

convert_to_titlecase(word6,ignore_cap=True,small_second_to_last=True)
Case 14: small_second_to_last=True : Only First charecter of word will be Capitalhello XML iPhone' ==> Hello Xml Iphone'

convert_to_titlecase(word6,ignore_cap=False,small_second_to_last=False)
Case 15: small_second_to_last=False : Only First charecter of word will be Capitalhello XML iPhone' ==> Hello XML iPhone'

convert_to_titlecase(word6,ignore_cap=True,small_second_to_last=False)
Case 16: small_second_to_last=False : Only First charecter of word will be Capitalhello XML iPhone' ==> Hello XML IPhone'

pekkaklarck:
Instead of having separate arguments for controlling the algorithm, perhaps it would be better to have just one, named `algorithm` or something similar, that takes the algorithm name as an argument. Just needed to come up with good names.

bbpatel2001:
@pekkaklarck , can you please guide me on naming

bbpatel2001:
Conflicts resolved.

bbpatel2001:
@pekkaklarck , please take next action on the PR

bbpatel2001:
@cepter888 can you please review this PR

src/robot/libraries/String.py
```diff
@@ -74,6 +74,21 @@ def convert_to_uppercase(self, string):
...
+        | ${str1} = | Convert To Titlecase | abc |
+        | ${str2} = | Convert To Titlecase | i love robot Framework |
+        | ${str3} = | Convert To Titlecase | i love ROBOT Framework |
+        | Should Be Equal | ${str1} | Abc |
+        | Should Be Equal | ${str2} | I Love Robot Framework |
+        | Should Be Equal | ${str3} | I Love ROBOT Framework |
+
+        New in Robot Framework 3.0.2.
+        """
+        return  ''.join( (c.upper() if i == 0 or string[i-1] == ' ' else c) for i, c in enumerate(string))
```
pekkaklarck: This implementation is pretty cleaver, but it doesn't take into account that there are other characters than the ASCII space that should be taken into account when deciding which characters to capitalize. See how Python's `.title()` handles, for example, the tab character (`\t`) and NO-BREAK-SPACE (`u`xa0`):

```python
>>> u'foo\tbar'.title()
u'Foo\tBar'
>>> u'foo\xa0bar'.title()
u'Foo\xa0Bar'
```

We can discuss about better algorithms in the comments of this PR separately.

src/robot/libraries/Collections.py
```diff
@@ -386,7 +389,7 @@ def lists_should_be_equal(self, list1, list2, msg=None, values=True,
         default = 'Lengths are different: %d != %d' % (len1, len2)
         _verify_condition(len1 == len2, default, msg, values)
         names = self._get_list_index_name_mapping(names, len1)
-        diffs = list(self._yield_list_diffs(list1, list2, names))
+        diffs = list(self._yield_list_diffs(list1, list2, names,ignore_order))
```
pekkaklarck: This change, nor related tests, have nothing to do with title casing. Are you possibly using the same branch when implementing two different features? That won't work with pull requests, you need to have a separate branch for each PR.

pekkaklarck:
Coming up with an algorithm that handles this seemingly simple problem well is surprisingly complicated. One thing that helps is creating a test that goes through all special cases to make sure the algorithm really works and I'd propose doing that first. The existing test is already pretty good, but it needs more variations. At least these should be added:

```
foo\tbar            -> Foo\tBar
foo\nbar            -> Foo\nBar
foo\xa0bar          -> Foo\xa0Bar
äiti åse            -> Äiti Åse
they're from UK     -> They're From UK
multiple    spaces  -> Multiple    Spaces
```

There are various solutions for this on Stackoverflow both [here](https://stackoverflow.com/questions/42228427/python-capitalize-first-letter-of-string-without-using-title) and [here](https://stackoverflow.com/questions/1549641/how-to-capitalize-the-first-letter-of-each-word-in-a-string-python) but most of them fail some of the above examples. I got a feeling that it's best to handle this using regular expressions wither by using `re.sub` or by splitting from whitespace, but there may be still better solutions than those.

src/robot/libraries/String.py
joaonc: Why not simply `return string.title()`?
That way Robot Framework keeps in line with Python's implementation and makes this code easier.

pekkaklarck: Because `string.title()` doesn't work well in many cases. For example, `'XML and HTML'.title()` returns `'Xml And Html'` which most people consider wrong.

As I've commented below, this is surprisingly complicated problem and it may also depend on the context how title-casing should work. One possibility would be making the algorithm to use configurable. Or then we can decide not to implement this keyword in the first place and let everyone implement their own context specific keywords.

pekkaklarck:
Cannot merge in the current state and no activity in close to 10 months. Closing.

pekkaklarck:
#3308 is a new attempt to implement this.

Issue #2703, opened 2017-11-01
`Lists Should Be Equal` keywords in Collections should have the option to ignore order


bkhouri:
when comparing lists, we sometimes want to compare only the content irregardless of the order.

Currently, the `Lists should be equal` ensures each index in the list is the same.

So the following test case fails

```
*** Settings ***
Library     Collections

*** Test Cases ***
Compare lists
    ${l1}=  create list     a   b    c
    ${l2}=  create list     c   a    b
    lists should be equal   ${l1}   ${l2}
```

If I wanted to get this test to pass, I would need to use a combination of keywords found in `Collections` and `BuiltIn`.

```
*** Settings ***
Library     Collections

*** Test Cases ***
Compare lists
*** Settings ***
    ${l1}=  create list     a   b    c
    ${l2}=  create list     c   a    b
    list should contain sub list  ${l1}     ${l2}
    ${l1_length}=       get length  ${l1}
    length should be    ${l2}       ${l1_length}

```

I would be nice to be able to do something as follows:

```
*** Settings ***
Library     Collections

*** Test Cases ***
Compare lists
    ${l1}=  create list     a   b    c
    ${l2}=  create list     c   a    b
    lists should be equal   ${l1}   ${l2}      ignore_order=True
```

This ticket has been opened to improve the `Lists should be equal` keywords to include a way to ignore the orders of the list.

pekkaklarck:
Sounds reasonable. What about duplicates?

bkhouri:
I haven't given some thoughts about duplicates, but I guess we can have an "ignore_duplicates" argument as well.

bbpatel2001:
This `Lists should be equal` keyword first compares  length of two list and if it is not same it fails. 
if we ignore_order=True as argument, we can check first list's element in 2nd list  and if it is available at any position, test should pass.  

**Pseudo code** 
If  ignore_order=True
    remove duplicates from lists
    Fail if  not same length
    if  element is not available in 2nd list, assert fail.
else  ##  order should be checked 
    compare item1 and item2

```
if ignore_order == True: 
    assert_true(item1 in list2, msg='Index %d%s' % (index,  " Not in second list"))
else:
    assert_equal(item1, item2, msg='Index %d%s' % (index, name))
```

As length of both lists are, same we need to check first list's element in 2nd list in case of ignore_order.

bbpatel2001:
Hi @pekkaklarck, 
Can you please review following `_yield_list_diffs` function. 
[...]

PR #2720, opened 2017-11-23
#2703 "Lists should be equal" keywords in the Collection library should have the option to ignore order 

bbpatel2001:


bbpatel2001:
Lists Should Be Equal With Different Values Ignore_order :: FAIL L... | PASS |
Lists Should Be Equal With Different Values And Own Error Message ... | PASS |
Lists Should Be Equal With Different Values And Own And Default Er... | PASS |
Lists Should Be Equal With Named Indices As List Ignore_order :: F... | FAIL |
Lists are different:
Index 0 (a): 11 != 13
Index 2 (c): 13 != 11
Lists Should Be Equal With Named Indices As List With Too Few Valu... | PASS |
Lists Should Be Equal With Duplicate Value Ignore_order :: PASS li... | PASS |
Lists Should Be Equal With Duplicate Value Second Small List :: FA... | FAIL |
Lists  are  different
Lists Should Be Equal With Two Duplicate Value Ignore_order :: FAI... | PASS |
Lists Should Be Equal With Duplicate And Different Value Ignore_or... | FAIL |
Lists are different:
Index 2: A != E
Index 3: D != A

bbpatel2001:
@pekkaklarck , any suggestions   on this ?

bbpatel2001:
@pekkaklarck can you please check this is acceptable PR?

kblomqvist:
Would `List Should Contain Sub List    ${list1}    ${list2}` work for you? It fails if not all of the elements in `list2` are found in `list1`. The order of values and the number of values are not taken into account.

PR #2730, opened 2017-12-01
Documentation fix: add forgotten extension .robot

aksenov007:
There is the list in the section "Test suite directories" has the list of ignored extensions. Although there is a link in the sentence with the actual list of supported extensions, I think it makes sense to keep the list up-to-date anywhere in the documentation.


pekkaklarck:
Thanks!

PR #2746, opened 2018-01-09
Fixes Selenium2Library to SeleniumLibrary in External libraries chapter

aaltat:


pekkaklarck:
- The related link should also be updated.
- Did you check are there other references to S2L in Robot's code base?

aaltat:
1) Will fix the link too
2) I didn't check in any other places in the user guide. I did run into this by accident. But will check out.

aaltat:
Did use `pss` to find occurrences of Selenium2Library from the user guide and fixed them to SeleniumLibrary.

pekkaklarck:
Thanks!

PR #2751, opened 2018-01-19
Rf 2627

Kompakti:


pekkaklarck:
Looks great! `Calling Using` isn't really needed in test case names but that's not worth changing.

PR #2752, opened 2018-01-22
fix for Set Test Message with the combined INFO and HTML

### LINKED ISSUE #2732: Appending to test message using HTML doesn't work if old message is not HTML
I would like insert a line between test messages :
`Set Test Message | *HTML* Insert line here<hr> | append=True`
But when i use optional argument 'append=True', message appear in text format. 

(Test with robotframework 3.0.2)

PavelBbb:
Fixes #2732

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ doc/api/_build
 doc/libraries/*.html
 doc/userguide/RobotFrameworkUserGuide.html
 src/robotframework.egg-info
+log.html
+output.xml
+report.html
--- a/atest/testdata/standard_libraries/builtin/set_test_message.robot
+++ b/atest/testdata/standard_libraries/builtin/set_test_message.robot
@@ -21,6 +21,23 @@ Set Multiline Message
     Set Test Message    1\n2\n3
     [Teardown]    Should Be Equal    ${TEST MESSAGE}    1\n2\n3
 
+Use the HTML
+    [Documentation]    PASS *HTML* custom message and another custom message in <HR> with a line
+    Set Test Message  custom message
+	Set Test Message  *HTML* and another custom message in <HR> with a line  append=True
+
+Use the HTML first
+    [Documentation]    PASS *HTML* and another custom message in <HR> with a line custom message
+    Set Test Message    *HTML* and another custom message in <HR> with a line
+	Set Test Message    custom message  append=True
+
+Use the HTML combination
+    [Documentation]    PASS *HTML* and another custom message in <HR> with a line custom message custom &lt;message&gt; and another custom message in <HR> with a line
+    Set Test Message    *HTML* and another custom message in <HR> with a line
+	Set Test Message    custom message  append=True
+	Set Test Message    custom <message>  append=True
+	Set Test Message    *HTML* and another custom message in <HR> with a line  append=True
+
 Set Non-String Message
     Set Test Message    ${42}
     [Teardown]    Should Be Equal    ${TEST MESSAGE}    42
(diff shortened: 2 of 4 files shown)


PavelBbb:
Please use the last commit ) There was a little typo )

pekkaklarck:
Sorry for not reviewing this yet. Been rather busy lately and soon leaving for a holiday. Should have time to really concentrate on RF development after that.

pekkaklarck:
I finally have time for RF development. I looked that changes and noticed that the change was in a helper method that's used also by other keywords that don't support `*HTML*` at all. Either the helper should be made configurable so that it handles possible `*HTML*` when needed or the fix should be done somewhere else. Also, there seem to be logic bugs in some cases.

Are you @PavelBbb still around and interested to work with this PR? If yes, I can do a more detailed review. If not, I can look at fixing this myself later this week. I try to get RF 3.0.3 ready latest next week.

PavelBbb:
Hi , 
I’m interested ) .
Will appreciate to continue with your feedback and review ) 

Paul

.gitignore
```diff
@@ -17,3 +17,6 @@ doc/api/_build
 doc/libraries/*.html
 doc/userguide/RobotFrameworkUserGuide.html
 src/robotframework.egg-info
+log.html
+output.xml
+report.html
```
pekkaklarck: I think we had some reason not to add these files to .gitignore but I don't remember what was it. Probably they could be added, but that should be done as a separate commit/PR outside of this one.

atest/testdata/standard_libraries/builtin/set_test_message.robot
```diff
@@ -21,6 +21,23 @@ Set Multiline Message
     Set Test Message    1\n2\n3
     [Teardown]    Should Be Equal    ${TEST MESSAGE}    1\n2\n3
 
+Use the HTML
+    [Documentation]    PASS *HTML* custom message and another custom message in <HR> with a line
+    Set Test Message  custom message
+	Set Test Message  *HTML* and another custom message in <HR> with a line  append=True
```
pekkaklarck: There seems to be a tab on the line above. More tabs on some of the lines below. They should all be converted to spaces.

src/robot/libraries/BuiltIn.py
```diff
@@ -3088,7 +3088,17 @@ def _get_possibly_appended_value(self, initial, new, append):
...
+
+            if new.startswith('*HTML*'):
+                new = new[6:].lstrip()
+                if initial.startswith('*HTML*'):
+                    return '%s %s' % (initial, new)
+                else:
+                    initial = '*HTML* ' + initial
+            else:
+                new = html_escape(new)
             return '%s %s' % (initial, new)
```
pekkaklarck: Various bigger and smaller issues in this code:

- This helper method is used also by other methods than `set_test_message` and in those usages nothing should be done if message happens to start with `*HTML*`. Should either implement separate helper that handles `*HTML*` or make the current one configurable with something like `handle_html=False`.

- `html_escape(new)` should be one *only* when `initial` is HTML (and when appending). Existing tests for non-HTML appending should be enhanced to contain something like `<message>` to see that they are not altered.

- It's bit hard to follow the logic when some branches return the value and others just modify values that are then returned later. Should handle this consistently. Also, it probably would be best to have

  ```python
  if not (is_truthy(append) and initial):
      return new
  ```
  before handling appending in general. That would avoid one level of nested if/else structures.

- We try to avoid empty rows inside methods. If methods grow so big/complicated that code should be grouped, we rather create helper methods. Helper may not be necessary here, but please remove the empty rows.

pekkaklarck:
Do you have time to look at this @PavelBbb? I'm planning to create RF 3.0.3 final or at least RC this week and would like to get this one in. I can take care of it myself as well if you don't have time.

PavelBbb:
Hi this week not sure if I'll have time. Maybe on a weekends.

Paul

pekkaklarck:
OK. I totally understand as this is somewhat tight schedule. As I wrote earlier, I try to get at least the release candidate of RF 3.0.3 out still this week. I'll take a look at needed changed myself.

pekkaklarck:
Merged and then took care of the issues I commented above. Thanks @PavelBbb for the PR!

PR #2759, opened 2018-01-31
Fix a warning message of keyword  'Take Screenshot' when using wxPython 4

Teddy12090:
Use :class:`wx.Bitmap` instead of `wx.EmptyBitmap` due to a warning message below in execution:

wxPyDeprecationWarning: Call to deprecated item EmptyBitmap. Use :class:`wx.Bitmap` instead
  bitmap = wx.EmptyBitmap(width, height, -1)

The version of wxPython 4.0.0b2 (the latest one).

### DIFF
--- a/src/robot/libraries/Screenshot.py
+++ b/src/robot/libraries/Screenshot.py
@@ -346,7 +346,10 @@ class ScreenshotTaker(object):
             self._wx_app_reference = wx.App(False)
         context = wx.ScreenDC()
         width, height = context.GetSize()
-        bitmap = wx.Bitmap(width, height, -1)
+        if wx.__version__ >= '4':
+            bitmap = wx.Bitmap(width, height, -1)
+        else:
+            bitmap = wx.EmptyBitmap(width, height, -1)
         memory = wx.MemoryDC()
         memory.SelectObject(bitmap)
         memory.Blit(0, 0, width, height, context, -1, -1)
--- a/src/robot/libraries/Screenshot.py
+++ b/src/robot/libraries/Screenshot.py
@@ -346,7 +346,7 @@ class ScreenshotTaker(object):
             self._wx_app_reference = wx.App(False)
         context = wx.ScreenDC()
         width, height = context.GetSize()
-        bitmap = wx.EmptyBitmap(width, height, -1)
+        bitmap = wx.Bitmap(width, height, -1)
         memory = wx.MemoryDC()
         memory.SelectObject(bitmap)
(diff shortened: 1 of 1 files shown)


HelioGuilherme66:
@Teddy12090 This is the wrong project. You are probably using RIDE from my [fork](https://github.com/HelioGuilherme66/RIDE). Please do the PR to my branch `wxPhoenix`.

pekkaklarck:
No, this is the right project. The fix is for the Screenshot library that uses wxPython as one of its backends.

Have you @Teddy12090 tested does the changed code also work with older wxPython versions?

pekkaklarck:
According to the [wxPython 2.8 docs](http://xoomer.virgilio.it/infinity77/wxPython/Widgets/wx.Bitmap.html#__init__), the old `wx.Bitmap` is not compatible. Testing on the console reveals the same:

```
$ python
Python 2.7.6 (default, Nov 23 2017, 15:49:48) 
[GCC 4.8.4] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import wx
>>> wx.Bitmap(100, 100, -1)
  File "/usr/lib/python2.7/dist-packages/wx-2.8-gtk2-unicode/wx/_gdi.py", line 552, in __init__
    _gdi_.Bitmap_swiginit(self,_gdi_.new_Bitmap(*args, **kwargs))
TypeError: new_Bitmap() takes at most 2 arguments (3 given)
>>> wx.__version__
'2.8.12.1'
```

An easy solution is using `wx.BitMap` if `wx.__version__ >= '4'` and using `wx.EmptyBitmap` otherwise. Can you @Teddy12090 update the PR? Could you please also submit an issue about this so that we get it listed in the release notes automatically.

Teddy12090:
@pekkaklarck I have updated the PR and submitted an issue.

pekkaklarck:
Great! The PR looks good. I want to test it locally as well, though, and don't have time for that now. Due to my forthcoming holiday merging may take a while but I'll add the issue to 3.0.3 scope so it won't be forgotten.

HelioGuilherme66:
Sorry about my comment.
@pekkaklarck Does it use wx exclusively? Or Tcl/Tk if available?
Thanks.

pekkaklarck:
There are various backends but I don't think Tk is supported. Not sure can it even take screenshots. See the library documentation (or source code) for more details.

pekkaklarck:
Finally merged. Thanks @Teddy12090 for the PR!

Issue #2376, opened 2016-05-25
"Run Keyword And Ignore Error" does not ignore timeout errors


dimanhsb:
Hello,

Consider the following code:

```
*** Test Cases ***
TC
    Run Keyword And Ignore Error  kw
    Log to console  b

*** Keywords ***
kw
    [timeout]  1s
    Log To Console  a
    Sleep  2s

```

The keyword `Log to console  b` is not executed, and the TC is failed. Is there a way around this limitation?

dimanhsb:
I did edit the errors.py in my local installation, the dont_continue function, in order to not fail on timeouts, but I was wondering if a "soft timeout" feature or a command line switch should be added.

pekkaklarck:
Errors caused by timeouts not being caught by this keyword is a [documented feature](http://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Run%20Keyword%20And%20Ignore%20Error) that cannot be changed. If someone wants to implement a separate keyword `Run Keyword And Ignore Timeout` it can be added.

dimanhsb:
I was mostly considering a `Run Keyword And Continue On Timeout` feature, as I want to capture the failure (or maybe both)

pekkaklarck:
What would be the semantics of this keyword? Continue execution but in the end fail anyway?

dimanhsb:
Yes, I was thinking of a way to revert the timeout failures to the standard handling of KW failures, and let the user control those via keywords like `Run Keyword And Ignore Error` , `Wait Until Keyword Succeeds`, `Run Keyword And Continue On Failure`, `Run Keyword And Expect Error` etc.

dimanhsb:
By the way, let me give my specific example:

There is either a bug, or a misconfiguration from my part in Selenium2Library:
The KW `Open Browser` sometimes hangs, without returning any failure (for some reason, maybe the driver for the specific browser has a bug), I would say 1% of the time (or less). Either way, if the browser doesn't start in 20sec or so, I want to "kill" this keyword and retry opening the browser (with the dont_continue hack I've done, this now works succesfully).

sameer49:
@dimanhsb can you tell how handled above problem?

sameer49:
@dimanhsb
since I am using python binding,  I handled open browser hang issues by writing custom python function which create own browser instance using core selenium library and assign that instance session to robot selenium library.

```
se2lib = BuiltIn().get_library_instance('Selenium2Library')
[...]

PR #2764, opened 2018-02-02
Added Run Keyword With Maximum Time keyword in BuiltIn library

mihaiparvu:
PR for #2376 
We had requests for something like this and implemented it as @pekkaklarck suggested: a new keyword that uses Robot's timeout logic.
If it is decided to be included in the next version of RF, I would update the User Guide also.

### DIFF
--- a/src/robot/libraries/BuiltIn.py
+++ b/src/robot/libraries/BuiltIn.py
@@ -25,7 +25,9 @@ from robot.errors import (ContinueForLoop, DataError, ExecutionFailed,
                           PassExecution, ReturnFromKeyword)
 from robot.running import Keyword, RUN_KW_REGISTER
 from robot.running.context import EXECUTION_CONTEXTS
+from robot.running.timeouts import KeywordTimeout
 from robot.running.usererrorhandler import UserErrorHandler
+
 from robot.utils import (DotDict, escape, format_assign_message,
                          get_error_message, get_time, is_falsy, is_integer,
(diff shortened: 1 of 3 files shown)


src/robot/libraries/BuiltIn.py
```diff
@@ -1872,6 +1874,31 @@ def run_keyword_and_expect_error(self, expected_error, name, *args):
...
+        """
+        if not is_string(name):
+            raise RuntimeError('Keyword name must be a string.')
+        timeout = KeywordTimeout(timeout_value, variables=self._variables)
+        kw = Keyword(name, timeout=timeout, args=args)
+        with self._context.timeout(kw.timeout):
+            try:
+                return kw.run(self._context)
+            except ExecutionFailed:
+                raise
```
miktuy: maybe add some message about timeout error?

VenkiDeveloper (approved):

pekkaklarck:
It seems I had missed this old PR. It looks pretty good, but APIs have changed so that `robot.running.Keyword` doesn't anymore accept `timeout` so the code wouldn't work as-is. I don't know would resolving that issue be too big a task, but there doesn't seem to be too much interest related to this otherwise. The referenced issue #2376 was closed already long time ago, and nobody there showed any interest towards a keyword like this. Finally, it is pretty easy to wrap existing keywords with a user keyword having a timeout if a behavior like this is decided. For all these reasons I think it's best to just close this.

PR #2771, opened 2018-02-10
Introduce an option for Tidy to set the max line length (Fix #1891)

### LINKED ISSUE #1891: Tidy should split lines longer than n characters
It's pretty common in most languages to have a style convention limiting line length to something like 80 or 100 characters. Currently Tidy will split rows that have more than 8 cells, but if individual cells are long, then rows can get very long and hard to read. It would be nice if Tidy could split lines longer than some configurable value (with a reasonable default).

richardsavio:
PR for #1891.

Tidy currently reformats suites with eight cells per line regardless of
line length. This enhancement adds a `--maxlinelength` argument to tidy
that splits the line at the specified line length when possible.

A summary of changes introduced in this commit:

* Introduced a `--maxlinelength` option for Tidy
* `--maxlinelength` defaults to 80 characters by default
* It is only applicable to space separated and pipe separated robot
files. Tabs can have a configurable length, and many 3rd party HTML
formatters exist already which can do a very good job.
* Updated existing test data to have the correct default line length
(80 characters)
* Added new tests and new test data to test custom line length
* Updated usage text for Tidy with information about the option
* Update the userguide
* This PR however introduces a tighter coupling between rowsplitter.py,
formatters.py and filewriters.py. It may not be desirable.

Changes to the rowsplitter logic:
Originally, the row splitter would either split the row at 8 cells or split
depending on the presence of IF, ELSE, etc. so that they begin on a new
line.

Now, it calculates the line length based on how the row and the table
would be formatted, and calculates split indices based on that. It would
always split at 80 characters by default. And it will only use the new logic for
space and pipe separated txt files.

### DIFF
--- a/atest/robot/tidy/tidy_maxlinelength.robot
+++ b/atest/robot/tidy/tidy_maxlinelength.robot
@@ -1,0 +1,15 @@
+*** Settings ***
+Resource    tidy_resource.robot
+Test Setup        Create Directory     ${TEMP}
+Test Teardown     Remove Directory     ${TEMP}    recursive=True
+
+*** Test Cases ***
+Custom maxlinelength for test suite
+    [Template]    Run tidy with golden file and check result
+    -m 120 -f txt               golden_120_maxlinelength.robot
+    -m 120 -f txt --usepipes    golden_pipes_120_maxlinelength.robot
+
+Custom maxlinelength for resource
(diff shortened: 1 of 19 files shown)


fabo-at-noona:
@pekkaklarck ping

PR #2796, opened 2018-03-28
updated docs to tell how to install from github using pip

rubygeek:
formatting looks different? but code looks same as other commands 🤔


rubygeek:
takes a short amount of time, and if a user runs it they can see it puts it in a temp place, so I dont think we need to explain.

```
time  pip install git+https://github.com/robotframework/RobotFramework.git
Collecting git+https://github.com/robotframework/RobotFramework.git
  Cloning https://github.com/robotframework/RobotFramework.git to /private/var/folders/dg/wrdwgxns5wj2lsyh9s453qn1nzl1qj/T/pip-_f78ggxz-build
Installing collected packages: robotframework
  Found existing installation: robotframework 3.0.2
    Uninstalling robotframework-3.0.2:
      Successfully uninstalled robotframework-3.0.2
  Running setup.py install for robotframework ... done
Successfully installed robotframework-3.0.3.dev20180315
...
pip install git+https://github.com/robotframework/RobotFramework.git  9.14s user 6.58s system 45% cpu 34.744 total
```

I looked in pip change log to find when installing with a git url was added, best i could determine was 0.3 :)

pekkaklarck:
pip 0.3 is so ancient it probably won't even work anymore. Perhaps it should just say something like "If you want to install the latest and possibly unreleased code directly from GitHub, you can use..."?

pekkaklarck:
Thanks @rubygeek!

PR #2802, opened 2018-04-08
Added note on calling pdb from a test case using evaluate

MarkJB:
Added a note to the User Guide on calling pdb directly from within a test case using the builtin evaluate keyword.

### DIFF
--- a/doc/userguide/src/ExecutingTestCases/BasicUsage.rst
+++ b/doc/userguide/src/ExecutingTestCases/BasicUsage.rst
@@ -668,18 +668,26 @@ Using the Python debugger (pdb)
 
 It is also possible to use the pdb__ module from the Python standard
 library to set a break point and interactively debug a running test.
-The typical way of invoking pdb by inserting
+The typical way of invoking pdb by inserting:
 
 .. sourcecode:: python
 
    import pdb; pdb.set_trace()
 
 at the location you want to break into debugger will not work correctly
-with Robot Framework, though, as the standard output stream is
+with Robot Framework, as the standard output stream is
 redirected during keyword execution. Instead, you can use the following:
 
 .. sourcecode:: python
 
    import sys, pdb; pdb.Pdb(stdout=sys.__stdout__).set_trace()
+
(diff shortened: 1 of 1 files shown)


MarkJB:
I missed a comma from the library imports, will update and resubmit the PR.

pekkaklarck:
You could have updated this one as well.

MarkJB:
Ok, I'll reopen and update.

MarkJB:
Ok, I didn't have to do anything other than commit and push to my fork and this PR automagically updated.

Is having two commits rather than one ok? (I don't know how to merge them into one).

Also what is the convention for the commit comment? Should I just have added a note that says 'Fixed syntax error' rather than using the same commit note again?

pekkaklarck:
Yeah, the PR is updated when you push something to the branch. A good reason to have dedicated topic branches if you are working with something bigger and, especially, if you want to work with multiple issues. Earlier it was advised to use Git magic to squash commits if the latter were just fixes for the original, but nowadays it's possible to squash while merging the PR.

doc/userguide/src/ExecutingTestCases/BasicUsage.rst
```diff
@@ -668,18 +668,26 @@ Using the Python debugger (pdb)
...
 
 .. sourcecode:: python
 
    import sys, pdb; pdb.Pdb(stdout=sys.__stdout__).set_trace()
 
+from within a python library or alternativley:
+
+.. sourcecode:: robotframework
+
+  evaluate  pdb.Pdb(stdout=sys.__stdout__).set_trace()  sys, pdb
```
pekkaklarck: We've used title case with keyword names and four spaces between arguments in examples elsewhere. Consistency with that style would be good. I'd also say `modules=sys, pdb` as the second argument would be a little clearer.

MarkJB: I've added the suggested changes.

pekkaklarck:
Thanks, merged!

PR #2831, opened 2018-05-07
Add Format String Keyword

### LINKED ISSUE #2822: Add a generic string format keyword
I would like to propose a generic string format keyword giving a more clean and generic way to format strings.
As suggested by @pekkaklarck in Slack this keyword could be something like this example:
```
${result} =    Format String    Template with {x} and {y}    x=foo    y=bar
```

This enhancement should solve some substituitions issues like this [gist](https://gist.github.com/mayribeirofernandes/104595f2a92fe2b6939b0b7ba6ca3a61) from @mayribeirofernandes.

elyssonmr:
Adds the **Format String** keyword.

Closes #2822


src/robot/libraries/String.py
```diff
@@ -125,6 +126,48 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+        | ${formatted_string} = | Format String | Document {} is missing on folder {folder} | tests.robot | folder=/home |
+        | Should Be Equal | ${formatted_string} | Document tests.robot is missing on folder /home |
+        | ${formatted_string} = | Format String | Uploaded file: {} should not be bigger than {}. | photo.jpg | 5MB |
+        | Should Be Equal | ${formatted_string} | Uploaded file: photo.jpg should not be bigger than 5MB. |
+
+        New in Robot Framework 3.1
+        """
+        if isfile(template_or_str):
+            logger.info('Opening file %s to be formated' % (template_or_str,))
+            format_file = open(template_or_str, 'r')
```
bkhouri: I would use the `with` statement to allow python to handle closing the file objects.  so this would be change to

```
with open(template_or_str, 'r') as format_file:
       format_string = format_file.read()
```

This way, if an error occurs during the read operation, the file handler will automatically be closed.

src/robot/libraries/String.py
```diff
@@ -125,6 +126,48 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
             raise TypeError('Can not decode strings on Python 3.')
         return bytes.decode(encoding, errors)
 
+    def format_string(self, template_or_str, *posicional_search_replace,
```
bkhouri: can we add unit tests?

elyssonmr: Yes we can, but I don't know how to do that.

pekkaklarck: Better actually to add acceptance tests. See `atest/README.rst` for details.

elyssonmr:
add acceptance tests.
@pekkaklarck  or @bkhouri can you review it?

tminakov:
I stumbled on this PR looking for something similar before implementing it myself, glad you did it against the standard library.  
One major comment here - the current implementation accepts as positional argument either a source string to be formatted, or a file which content will be used for that. It distinguishes between the two by a call to `os.path.isfile()`, but this approach has two major drawbacks:

- mostly, performance hit - every keyword call will start off with file system access ( `os.path.isfile()` is a wrapper over `os.stat()`), which is order of magnitudes slower than the actual work of string formatting (the data for it is already in-memory),
- in the corner case where the user wants to pass a string, but there's a file with the same name in the cwd,  this will lead to an unexpected result.

Thus I'd suggest to split the keyword in two, `Format String` and `Format String From File` (or similar name).

src/robot/libraries/String.py
```diff
@@ -121,6 +122,47 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
```
tminakov: Typo, should be "positional"

src/robot/libraries/String.py
```diff
@@ -121,6 +122,47 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+
+        New in Robot Framework 3.1
+        """
+        if isfile(template_or_str):
+            logger.info('Opening file %s to be formated' % (template_or_str,))
+            with open(template_or_str, 'r') as format_file:
+                format_string = format_file.read().rstrip()
+        else:
+            format_string = template_or_str
+        logger.info('Formating string')
```
tminakov: If there's a logging event, perhaps it would be more helpful if it includes some actual info, like the source string? For example:. 
```suggestion
        logger.info('Formatting the source string "{}"{}'.format(format_string[:80]), 
                    ' (log trimmed)' if len(format_string)>80 else '')
```
Trimmed to the first 80 chars not to fill up the log for lengthy strings.

There's a small typo in "Formatting" btw.

src/robot/libraries/String.py
```diff
@@ -121,6 +122,47 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+
+        If the given ``string`` is a valid file path (absolute or relative), opens the file in read
+        mode and then format its content using the given
+        ``posicional_search_replace`` and ``named_search_replace``.
+
+        This keyword uses python's string format. For more information see:
+        [https://docs.python.org/2.7/library/string.html#formatstrings]|Python2 format syntax]
+        or
+        [https://docs.python.org/3.6/library/string.html#formatstrings]|Python3 format syntax]
+
```
tminakov: There's a very good 3rd party formatting resource, which you might consider adding here - https://pyformat.info.  
It's quite in-depth though, I'm not sure won't it be a bit too intimidating for RF users w/o python background.

pekkaklarck:
It was probably me who proposed adding the ability to read the template from a file. XML library has similar functionality and it has worked great.

I doubt the performance hit from checking does the file exist is big and it is also rather unlikely that a normal template string would match a path to an existing file. Both of these worries can be mostly mitigated by mandating that the path must be absolute and checking that with `os.path.isabs` before using `isfile`.

tminakov:
@pekkaklarck there is an existing issue for this addition - #2822, but I don't think it's currently targeted for 3.1; it would be good if it gets there.

pekkaklarck:
Added #2822 to 3.1 scope.

pekkaklarck:
If the possible problem with a template string matching an exist file with an absolute path turns out to actually cause problems, we can consider adding an optional argument to control is the template a string or path. I considered that so unlikely that I wouldn't do it now. The `isabs` check I mentioned earlier would probably be a good idea. The requirement for the path to be absolute would then need to be mentioned also in the docs.

tminakov:
> If the possible problem with a template string matching an exist file with an absolute path turns out to actually cause problems, we can consider adding an optional argument to control is the template a string or path.

I also think we're spending too many cycles on a small issue - hitting a file name collision with a templated string is really a corner case. I fully agree `os.path.abs` will weed out the majority of the values usages, with no perf hit, leaving only a very small percentage getting down to the `isfile()` check; let's leave it at that.

Thanks for adding it to the scope 3.1 👍

elyssonmr:
@tminakov thanks for your review, I'm changing the code as you suggest tonight.

I think that absolute path is not easy to archive in some projects. For example, in my team I use Linux but the people who make our ATDD use Windows.

Another point is that relative path could be trick sometimes due the CWD changes as the tests are executed from another folder.

atest/testdata/standard_libraries/string/format_string_template.txt
```diff
@@ -0,0 +1 @@
+Format String also {condition} files templates!
```
pekkaklarck: Some tests for the format specification mini-language would be nice. At least some tests for the format spec, but possibly also for alignment and attribute/item access. Obviously no need to test the whole mini-language, but making sure the common and for us relevant cases actually work would be nice.

The attribute and item access test could use something like this:

```robotframework
*** Variables ***
&{USER}                 name=John Doe    email=john@example.com

*** Test Cases ***
Attribute access
    ${result} =    Format String    {user.name} <{user.email}>    user=${USER}
    Should Be Equal    ${result}    John Doe <john@example.com>

Item access
    ${result} =    Format String    {user[name]} <{user[email]}>    user=${USER}
    Should Be Equal    ${result}    John Doe <john@example.com>
```

Notice that I don't know do the above work but I expect they do!

pekkaklarck:
As I commented in my review, I think we should require the template path to be absolute. In addition to the benefits discussed earlier, absolute paths aren't affected by changes to the current working directory. It's easy to create such paths by using, for example, `${CURDIR}`.

Anyway, the changes look good in general and I'm sure we can get this into RF 3.1. I'm trying to get beta 1 out this week before leaving for a one week holiday, but there's going to be beta 2 (or rc 1) if that deadline is missed.

src/robot/libraries/String.py
```diff
@@ -121,6 +122,48 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+        | ${formatted_string} = | Format String | C:\\template.txt | | test=awesome |
+        | Should Be Equal | ${formatted_string} | My awesome String |
+        | ${formatted_string} = | Format String | User {} is not a admin user. | non-admin | |
+        | Should Be Equal | ${formatted_string} | User non-admin is not a admin user. |
+        | ${formatted_string} = | Format String | Username: {username} - Password: {password} | | username=Robot | password=Framework |
+        | Should Be Equal | ${formatted_string} | Username: Robot - Password: Framework |
+        | ${formatted_string} = | Format String | Document {} is missing on folder {folder} | tests.robot | folder=/home |
+        | Should Be Equal | ${formatted_string} | Document tests.robot is missing on folder /home |
+        | ${formatted_string} = | Format String | Uploaded file: {} should not be bigger than {}. | photo.jpg | 5MB |
+        | Should Be Equal | ${formatted_string} | Uploaded file: photo.jpg should not be bigger than 5MB. |
```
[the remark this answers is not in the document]
elyssonmr: I changed to result

src/robot/libraries/String.py
```diff
@@ -121,6 +122,48 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
             raise TypeError('Can not decode strings on Python 3.')
         return bytes.decode(encoding, errors)
 
+    def format_string(self, template_or_str, *positional_search_replace,
+                      **named_search_replace):
+        """Formats a ``string`` using the given ``positional_search_replace``
+        and ``named_search_replace``.
+
+        If the given ``string`` is a valid file path (absolute or relative), opens the file in read
```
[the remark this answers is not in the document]
elyssonmr: Changed to "absolute file path"

atest/testdata/standard_libraries/string/format_string.robot
```diff
@@ -0,0 +1,33 @@
...
+Format String With Named Search Replace Arguments
+    ${result} =    Format String    Username: {username} - Password: {password}    username=robotframework    password=S3¢r3t
+    Should be equal    ${result}    Username: robotframework - Password: S3¢r3t
+
+Format String With Named And Search Replace Arguments
+    ${result} =    Format String    Document {} is missing on folder {folder}    tests.robot    folder=/home
+    Should be equal    ${result}    Document tests.robot is missing on folder /home
+
+Format String From Template File
+    ${result} =    Format String    atest/testdata/standard_libraries/string/format_string_template.txt    condition=supports
```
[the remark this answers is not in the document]
elyssonmr: Changed to use this keyword.

src/robot/libraries/String.py
```diff
@@ -121,6 +122,48 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+        mode and then format its content using the given
+        ``positional_search_replace`` and ``named_search_replace``.
+
+        This keyword uses python's string format. For more information see:
+        [https://docs.python.org/2.7/library/string.html#formatstrings]|Python2 format syntax]
+        or
+        [https://docs.python.org/3.6/library/string.html#formatstrings]|Python3 format syntax]
```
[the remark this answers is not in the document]
elyssonmr: I changed the link.

src/robot/libraries/String.py
```diff
@@ -121,6 +122,48 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+        | ${formatted_string} = | Format String | User {} is not a admin user. | non-admin | |
+        | Should Be Equal | ${formatted_string} | User non-admin is not a admin user. |
+        | ${formatted_string} = | Format String | Username: {username} - Password: {password} | | username=Robot | password=Framework |
+        | Should Be Equal | ${formatted_string} | Username: Robot - Password: Framework |
+        | ${formatted_string} = | Format String | Document {} is missing on folder {folder} | tests.robot | folder=/home |
+        | Should Be Equal | ${formatted_string} | Document tests.robot is missing on folder /home |
+        | ${formatted_string} = | Format String | Uploaded file: {} should not be bigger than {}. | photo.jpg | 5MB |
+        | Should Be Equal | ${formatted_string} | Uploaded file: photo.jpg should not be bigger than 5MB. |
+
+        New in Robot Framework 3.1
```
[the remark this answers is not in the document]
elyssonmr: Added

src/robot/libraries/String.py
```diff
@@ -121,6 +122,48 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+        """
+        if isfile(template_or_str):
+            logger.info('Opening file %s to be formated' % (template_or_str,))
```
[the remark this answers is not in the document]
elyssonmr: I am not understanding the log part. I've done th modifications but when I use the Check Log Message keyword my test fails telling me that this keyword doesn't exist.

I coded that the log comes before the replace, is that correct? or I need to change it first and then log?

src/robot/libraries/String.py
```diff
@@ -121,6 +122,48 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+            with open(template_or_str, 'r') as format_file:
+                format_string = format_file.read().rstrip()
+        else:
+            format_string = template_or_str
+        logger.info('Formatting the source string "{}"{}'.format(format_string[:80], 
+                                                          ' (log trimmed)' if len(format_string)>80 else ''))
```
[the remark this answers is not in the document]
elyssonmr: Removed this log.

atest/testdata/standard_libraries/string/format_string_template.txt
elyssonmr: Added those tests. And they passed :)

src/robot/libraries/String.py
```diff
@@ -19,6 +19,8 @@
 from fnmatch import fnmatchcase
 from random import randint
 from string import ascii_lowercase, ascii_uppercase, digits
+from os.path import isfile, isabs, sep
```
pekkaklarck: - `sep` is officially exposed directly via `os`, not via `os.path`. Should also just `import os` and then use 'os.sep`.
- Could also consider just importing `os.path` and using `os.path.isfile`, but also this style is fine as well.

src/robot/libraries/String.py
```diff
@@ -121,6 +123,47 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+
+        New in Robot Framework 3.1.
+        """
+        if isabs(template_or_str) and isfile(template_or_str):
+            logger.info(
+                'Reading template from file <a href="%s">%s</a>' % (template_or_str, template_or_str),
+                html=True)
+            template_or_str.replace("/", sep)
+            with open(template_or_str, 'r') as format_file:
+                format_string = format_file.read().rstrip()
```
pekkaklarck: 1. I agree it's a good idea to use `rstrip()`, but stripping trailing whitespace should be mentioned in the documentation as well. Obviously stripping is annoying if you want to have whitespace at the end, but leaving the newline that many editors automatically add to the end of a file would be annoying as well. One possibility would be stripping only possible newline. Do you thin that would be worth the (small) effor?

2. The current code won't work with Python 2 if the file contains non-ASCII characters and with Python 3 encoding would be operating system specific. We should decode the contents using UTF-8 like we do with all other data files. The easiest way to do that in Python 2 and 3 compatible manner is using our `robot.utils.UtfReader` like

```python
from robot.utils import Utf8Reader

with Utf8Reader(template) as reader:
    content = reader.read().rstrip()
```

3. Above reminds some new tests are needed:
   - Stripping whitespace from template files
   - Non-ASCII content in template files
   - Non-ASCII template strings in general

src/robot/libraries/String.py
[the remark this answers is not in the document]
elyssonmr: I changed to only import OS. Then I use `os.sep`, `os.path.isabs` and `os.path.isfile`

src/robot/libraries/String.py
```diff
@@ -121,6 +123,47 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+        | ${result} = | Format String | Uploaded file: {} should not be bigger than {}. | photo.jpg | 5MB |
+        | Should Be Equal | ${result} | Uploaded file: photo.jpg should not be bigger than 5MB. |
+
+        New in Robot Framework 3.1.
+        """
+        if isabs(template_or_str) and isfile(template_or_str):
+            logger.info(
+                'Reading template from file <a href="%s">%s</a>' % (template_or_str, template_or_str),
+                html=True)
+            template_or_str.replace("/", sep)
```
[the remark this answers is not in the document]
elyssonmr: By Bad... Fixed

src/robot/libraries/String.py
```diff
@@ -121,6 +123,47 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
             raise TypeError('Can not decode strings on Python 3.')
         return bytes.decode(encoding, errors)
 
+    def format_string(self, template_or_str, *positional_search_replace,
```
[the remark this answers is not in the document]
elyssonmr: I'm concerned about people who do not have technical expertise to understand what `args` and `kwargs` are. I changed to `positional` and `named`.

src/robot/libraries/String.py
elyssonmr: 1. We can acomplhis it with rstrip, I just need to pass the arg **'\r\n'** and it would only remove those characters.
2. Changed to use this reader
3. Added the new suggested tests

tminakov: > Obviously stripping is annoying if you want to have whitespace at the end, but leaving the newline that many editors automatically add to the end of a file would be annoying as well. 

What's the use case of stripping the end newlines? Having `\n` as the last byte of a file is a good practice - pep8 requires it through W292, diff complains if there isn't one, it's required in c/c++ header files (due to the `#include` directive), and running such file through `cat` is unpleasant as 
it leaves the prompt at the line's end, to name a few examples.

If we include it here, this will turn the keyword from a formatter to a mutator; for some users the trailing newlines may be significant. Also this adds behavior duality - the chars at stripped in file content, but not when a string is passed.

src/robot/libraries/String.py
```diff
@@ -121,6 +123,45 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
             raise TypeError('Can not decode strings on Python 3.')
         return bytes.decode(encoding, errors)
 
+    def format_string(self, template, *positional, **named):
+        """Formats a ``string`` using the given ``positional`` and ``named``
```
tminakov: The parameter name is "template", but it is referenced as "string" in the docstring. I'm not sure how will this turn up in the generated documentation (haven't ever done it myself), but won't it lead to a difference - "template" in the keyword signature (the left column), and "string" in the documentation body (the right column)?

If so, that can be a bit confusing; it'd be better if the to names are the same, e.g.
"Formats a \`\`template\`\` (a string, or a file's content) using the ..."

src/robot/libraries/String.py
```diff
@@ -121,6 +123,45 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+        | Should Be Equal | ${result} | Username: Robot - Password: Framework |
+        | ${result} = | Format String | Document {} is missing on folder {folder} | tests.robot | folder=/home |
+        | Should Be Equal | ${result} | Document tests.robot is missing on folder /home |
+        | ${result} = | Format String | Uploaded file: {} should not be bigger than {}. | photo.jpg | 5MB |
+        | Should Be Equal | ${result} | Uploaded file: photo.jpg should not be bigger than 5MB. |
+
+        New in Robot Framework 3.1.
+        """
+        if os.path.isabs(template) and os.path.isfile(template):
+            template = template.replace("/", os.sep)
```
tminakov: Why is this replacement needed? The check in the previous like has already deemed the value as an existing file for the current OS,  replacing its forwarded slashes now is a bit late (and changes the filename).  
By the replacement I guess this is for Windows clients to be able to use *nix path values, but Windows does work with both path separators (`\` and `/`).

src/robot/libraries/String.py
pekkaklarck: Yeah, should use `string` in the doc. Don't think it's necessarily to explain already here hat the template can be a string or a path to a file, though.

src/robot/libraries/String.py
pekkaklarck: The main reason is that logging is consistent. I doubt that's a big issue for users, but it makes it easier to create tests. Not sure is the log message actually tested, though.

src/robot/libraries/String.py
```diff
@@ -121,6 +123,45 @@ def decode_bytes_to_string(self, bytes, encoding, errors='strict'):
...
+            logger.info(
+                'Reading template from file <a href="%s">%s</a>' % (template, template),
+                html=True)
+            with Utf8Reader(template) as reader:
+                format_string = reader.read().rstrip('\r\n')
```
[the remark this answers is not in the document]
pekkaklarck: Agreed. It's not good to use the same name as the function has. Just re-using `template`, as suggested, would work nice and avoid the need for the `else` branch below.

src/robot/libraries/String.py
tminakov: Sorry @pekkaklarck did you actually meant to say "should use `template` in the doc"?

src/robot/libraries/String.py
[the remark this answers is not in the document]
elyssonmr: Updated to template.

src/robot/libraries/String.py
elyssonmr: All right, I'm removing it. But I really think people are going to have issues with new line from editors. I'm going to also add it to the documentation.

src/robot/libraries/String.py
pekkaklarck: Yes, `template` is better as an argument name and should then obviously be used in the documentation as well.

pekkaklarck:
Looks good to me now. There are few things that can still be cleaned up, but I'll merge this now and do that myself.

PR #2845, opened 2018-05-28
Correct example of "Should Contain"

jiranat:
The example of "Should Contain" keyword (documentation) should use ignore_case instead of case_insensitive.


pekkaklarck:
Good catch and thanks for PR!

Issue #2835, opened 2018-05-10
Screenshot library enhancements


piotrpruski:
Hello,

On behalf of my team I would like to ask you to implement new features/enhancements in Screenshot library. When using Take Screenshot we would like to have possibility to change format of saved file for example for *.png and have a possibility to change compression ratio - this would help to reduce size of files generated during bigger suites where a lot of Take Screenshot is triggered.

ksokolowski:
Hello.
Yes, the possibility to change file format and compression level would be really helpful. Even with existing default JPG file format it would be great to be able to change compression level. In my team, any larger RF test round is currently generating now about 100 MB of data and most of it are screenshots. Cutting this in half would be great improvement, 
https://blog.codinghorror.com/a-comparison-of-jpeg-compression-levels-and-recompression/
We could safe a lot of disk space on our Jenkins servers, if we could already choose specific compression/quality ratio for JPG files. This could be an option when importing Screenshot library.
When considering new file formats IMO it would be great to also to add support for new WebP files. 
https://developers.google.com/speed/webp/

piotrpruski:
We are using pygtk module.
Seems that in Screenshot library there is used "gdk_pixbuf_save ()" which can be parametrized for format and quality like:
gdk_pixbuf_save (pixbuf, handle, "**jpeg**", &error, "**quality**", "100", NULL);
( https://developer.gnome.org/gdk-pixbuf/2.36/gdk-pixbuf-File-saving.html ).
But for jpeg and png there is different approach/numbers for quality factor. This could be perhaps made somehow like this:
When importing Screenshot library with PyGTK there would be default jpeg format with default compression, but this could be optionally parametrized. When parametrizing user would choose format and compression ratio, for example png and compression ratio between 0-9. Then inside in case of jpeg format compression ratio would be scaled for example multiplied by 11.

pekkaklarck:
Can we come up with config that has a good balance between quality and file size? If yes, I'd be fine just using that without making it possible for users to configure it further.

pekkaklarck:
The problem with making quality configurable is that we have many different backends (Java and .NET native APIs, wxPython, PyGTK, PIL, Scrot, OS screencapture) and supporting them all the same is lot of work and may not even be possible. Creating images that are good enough for normal usages (no visible compression artifacts) but not overly big is a lot simpler solution. If more features are needed, it would probably be better to create a new external library that only supports a single backend like PyGTK.

pekkaklarck:
[...]

PR #2868, opened 2018-06-13
Screenshot enhancements

mihaiparvu:
PR for https://github.com/robotframework/robotframework/issues/2835.

Added two arguments to `Take Screenshot` and `Take Screenshot Without Embedding` keywords: `image_format` [jpg, jpeg, png] and `quality` [0-100]. The different approach between PNG and JPEG compression/quality is handled internally and the end user would not have to care about that.

The quality can only be customized when using gtk, wx, pil and scrot. Unfortunately it does not work 'out of the box' with Jython, IronPython and OSX.

### DIFF
--- a/atest/robot/standard_libraries/screenshot/take_screenshot.robot
+++ b/atest/robot/standard_libraries/screenshot/take_screenshot.robot
@@ -15,18 +15,6 @@ Basename May Be Defined
     ${tc}=  Check Test Case  ${TESTNAME}
     Check Embedding In Log  ${tc.kws[0].kws[0].msgs[1]}  foo_1.jpg
 
-Screenshot With Png Extension
-    ${tc}=  Check Test Case  ${TESTNAME}
-    Check Embedding In Log  ${tc.kws[0].kws[0].msg[1]}  screenshot_1.png
-
-Png Screenshot Quality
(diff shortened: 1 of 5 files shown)


pekkaklarck:
As I've commented issue #2835, I'm still not sure is making quality configurable worth the effort. It complicates the code, makes testing harder, isn't supported by all backends, and requires user configuration to actually get smaller size images. I've asked few times wouldn't it be possible to just adjust the quality so that it is good enough for all practical purposes but doesn't create overly large images, but nobody has replied.

pekkaklarck:
Also, wouldn't it be more convenient to configure image format when taking the library into use? Using `format=png` always when using `Take Screenshot` would be a bit annoying. Having separate keyword to change the format (like Telnet and SSHLibrary has for altering their default settings) probably wouldn't be needed.

pekkaklarck:
My proposal for this PR:

1. Implement only PNG support. Having one feature per one PR is generally a good idea, and we can continue discussion how to handle compression separately.

2. Make it possible to give the format when importing the library to avoid the need to repeat it everywhere. Keywords can also accept it as an optional argument overriding the default value set when importing the library.

mihaiparvu:
I've made the proposed changes. Removed quality altoghether for the time being and added possibility to set the format at library import.

PR #2887, opened 2018-07-02
xUnit: represent times as floats not ints

### LINKED ISSUE #2397: Times in xUnit outputs should be floats not integers
I noticed that the current xunit xml report file only shows time in seconds and does not have milliseconds component to it. So for some of the testcase which just finishes in say 330ms it shows 0; 660ms shows 1 (not sure about the rounding off), but it just shows seconds.
example: 

<testcase classname="Tests.User" name="Read Single User" time="1"> </testcase>

It would be useful if we shows the time in milliseconds or in seconds with fractions ms, like time="1.182"
See attached report.xml file generated using --xunit flag (i have to change the file to .txt for attaching)
[report_xml.txt](https://github.com/robotframework/robotframework/files/314919/report_xml.txt)

juusoi:
Fixes #2397.

### DIFF
--- a/atest/testdata/misc/non_ascii.robot
+++ b/atest/testdata/misc/non_ascii.robot
@@ -5,6 +5,7 @@ Library         NonAsciiLibrary
 Non-ASCII Log Messages
     Print NonASCII Strings
     Log    Français
+    Sleep    0.001
 
 Non-ASCII Return Value
     ${msg} =    Evaluate    u'Fran\\xe7ais'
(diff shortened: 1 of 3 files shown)


atest/testdata/misc/non_ascii.robot
```diff
@@ -5,6 +5,7 @@ Library         NonAsciiLibrary
 Non-ASCII Log Messages
     Print NonASCII Strings
     Log    Français
+    Sleep    0.001
```
pekkaklarck: Example note.

PR #2889, opened 2018-07-03
Issue 2580

### LINKED ISSUE #2580: `Collections` keywords fail with bad error message when used with invalid input
When I have this: 
```
*** Settings ***
Library    Collections
*** Variables ***
${foo}    a=1   b=2

*** Text Cases ***
Example
    ${value} = Collections.Get From Dictionary    ${foo}    a
```
Then I get error: `TypeError: string indices must be integer`, which is little bit confusing. I understand that error comes directly from Python side, but spotting the error is quite difficult. Because the dictionary is declared in wrong way, in variables section, but it is used correctly in the keyword.

juusoi:
Fixes #2580.


atest/testdata/standard_libraries/collections/dictionary.robot
```diff
@@ -3,6 +3,10 @@ Test Setup        Create Dictionaries For Testing
 Resource          collections_resources.robot
 Library           CollectionsHelperLibrary.py
 
+*** Variables ***
+${dict_string}    This is a string not a dictionary.
+${dict_error}     TypeError: Expected first argument to be a dictionary, got 'string' instead.
```
pekkaklarck: These don't seem to be used anywhere.

src/robot/libraries/Collections.py
```diff
@@ -396,6 +415,8 @@ def list_should_contain_sub_list(self, list1, list2, msg=None, values=True):
         See `Lists Should Be Equal` for more information about configuring
         the error message with ``msg`` and ``values`` arguments.
         """
+        self._validate_list(list1)
+        self._validate_list(list2, 2)
```
pekkaklarck: Couldn't this be replaces with `self._validate_lists(list1, list2)`?

src/robot/libraries/Collections.py
```diff
@@ -363,6 +380,8 @@ def lists_should_be_equal(self, list1, list2, msg=None, values=True,
         message would contain a row like ``Index 2 (email): name@foo.com !=
         name@bar.com``.
         """
+        self._validate_list(list1)
+        self._validate_list(list2, 2)
```
pekkaklarck: Couldn't this be replaces with `self._validate_lists(list1, list2)`?

src/robot/libraries/Collections.py
```diff
@@ -432,6 +454,13 @@ def _index_error(self, list_, index):
         raise IndexError('Given index %s is out of the range 0-%d.'
                          % (index, len(list_)-1))
 
+    def _validate_list(self, list, position=1):
+        if not is_list_like(list):
+            raise TypeError("Expected %i. argument to be a list, got '%s' instead." % (position, type_name(list)))
+
+    def _validate_lists(self, *lists):
+        for item in lists:
+            self._validate_list(item, lists.index(item)+1)
```
pekkaklarck: This is a clever utility that, as I commented above, can be used more. The code should, however, be changed to:

```python
for index, item in enumerate(lists, start=1):
    self._validate_list(item, index)
```

src/robot/libraries/Collections.py
```diff
@@ -432,6 +454,13 @@ def _index_error(self, list_, index):
         raise IndexError('Given index %s is out of the range 0-%d.'
                          % (index, len(list_)-1))
 
+    def _validate_list(self, list, position=1):
+        if not is_list_like(list):
+            raise TypeError("Expected %i. argument to be a list, got '%s' instead." % (position, type_name(list)))
```
pekkaklarck: Nitpicky style comments:
- It would be better to use `list_`, not `list` as the argument name.
- `%d` is used a lot more commonly than `%i` in Python. Didn't myself even know what `%i` does without googling.
- I don't think `1.` style is used in English and it should be `Expected 1st argument to be...` instead. A lot easier to change it to `Expected argument 1 to be...`, though.
- We don't expect the argument to be a list, only list-like, so the message should be something like `Expected argument %d to be list or list-like, ...`.
- Could actually consider renaming this helper to `_validate_list_like` as well.
- Quotes aren't needed around the name returned by `type_name`. The message should thus end just `..., got %s instead.`
- We try to follow PEP-8 and keep lines below 80 characters.

This implementation ought to fix all the (minor) problems listed above:

```python
    def _validate_list(self, list_, position=1):
        if not is_list_like(list_):
            raise TypeError("Expected argument %d to be list or list-like, "
                            "got %s instead." % (position, type_name(list_)))
```

atest/testdata/standard_libraries/collections/dictionary.robot
```diff
@@ -232,7 +232,34 @@ Pop From Dictionary With Default
...
+    Get from dictionary                        string    1    I'm not a dict, I'm string.    a
+    Keep in dictionary                         string    1    I'm not a dict, I'm string.    a
+    Log Dictionary
+    Pop From Dictionary                        string    1    I'm not a dict, I'm string.    a
+    Remove From Dictionary                     string    1    I'm not a dict, I'm string.    a
+    Set To Dictionary                          string    1    I'm not a dict, I'm string.    a    b
+
 *** Keywords ***
+Validate invalid argument error
+    [Arguments]  ${keyword}    ${type}=string    ${position}=1    ${argument}=I'm not a dict, I'm string.    @{kwargs}
```
pekkaklarck: Should be `@{args}`, not `@{kwargs}`. The latter has a different meaning. Possibly could add additional `&{kwargs}` here and pass those forward to the called keyword as well, but I don't think that's needed.

pekkaklarck: Also, this would be an *excellent* place where to use the brand new named-only arguments syntax (#2896). Would require rebasing the branch and can be done also afterwards.

atest/testdata/standard_libraries/collections/list.robot
```diff
@@ -551,7 +551,43 @@ List Should Not Contain Value, Value Found And Own Error Message Glob
...
+    Set List Value                        string    1    I am a string. Not a list.    0    a
+    Should Contain Match                  string    1    I am a string. Not a list.    a
+    Should Not Contain Match              string    1    I am a string. Not a list.    xyz
+    Sort List
+
 *** Keywords ***
+Validate invalid argument error
+    [Arguments]    ${keyword}    ${type}=string    ${position}=1    ${argument}=I'm not a dict, I'm a string.    @{kwargs}
+    ${error} =    Set variable    TypeError: Expected ${position}. argument to be a list, got '${type}' instead.
+    Run keyword and expect error    ${error}    ${keyword}    ${argument}    @{kwargs}
```
pekkaklarck: Same comments here than to the above keyword validating dicts. An extra comment for changing the default value of `${argument}` from `I'm not a dict, I'm a string.` into `I'm not a list, I'm a string.` or just `I'm a string!` or something similar.

src/robot/libraries/Collections.py
```diff
@@ -727,6 +773,9 @@ def _yield_dict_diffs(self, keys, dict1, dict2):
             except AssertionError as err:
                 yield unic(err)
 
+    def _validate_dictionary(self, dictionary, position=1):
+        if is_string(dictionary) or is_number(dictionary):
+            raise TypeError("Expected %i. argument to be a dictionary, got '%s' instead." % (position, type_name(dictionary)))
```
pekkaklarck: Same comments here than to `_validate_list` above. An extra comment to have two empty rows between classes as PEP-8 mandates.

pekkaklarck:
Thanks for a great PR @juusoi!

PR #2892, opened 2018-07-05
improve clarity, at first I was confused... maybe its not just me :)

rubygeek:


rubygeek:
perhaps not needed since you have a PR that makes it work in python3 :)

PR #2893, opened 2018-07-05
Adapted u2html.py to run under Python 3.

HelioGuilherme66:
Tested only in Linux with ~~both python 2.7.15 and~~ 3.6.5.

It was modified to run only with python 3.

### DIFF
--- a/doc/userguide/ug2html.py
+++ b/doc/userguide/ug2html.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 
 """ug2html.py -- Creates HTML version of Robot Framework User Guide
 
@@ -11,6 +11,8 @@ dist .... Creates the user guide under 'robotframework-userguide-<version>'
           directory and also copies all needed images and other link targets
           there. The created output directory can thus be distributed
           independently.
+          Note: Before running this command, you must generate documents at
+          project root, for example, with: invoke library-docs all
 
 zip ..... Uses 'dist' to create a stand-alone distribution and then packages
           it into 'robotframework-userguide-<version>.zip'
@@ -103,8 +105,7 @@ def pygments_directive(name, arguments, options, content, lineno,
     try:
         lexer = get_lexer_by_name(arguments[0])
     except ValueError as err:
-        raise ValueError('Invalid syntax highlighting language "%s".'
-                         % arguments[0])
+        raise ValueError(f'Invalid syntax highlighting language "{arguments[0]}".')
     # take an arbitrary option if more than one is given
     formatter = options and VARIANTS[options.keys()[0]] or DEFAULT
     # possibility to read the content from an external file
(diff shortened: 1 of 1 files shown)


rubygeek:
tested this branch and verified it works  on both python2 and python3 👍

pekkaklarck:
Thanks for the PR. There is, however, no real need to this script to work both with Python 2 and 3 and supporting both just adds extra complexity. All other development tools actually require Python 3.6, which allows using f-strings, so it would be great to port this to Python 3(.6) as well. Would you be interested to do that?

In the longer term this whole script should be merged with the other dev tooling that is generally used via Invoke and its `tasks.py` file (see BUILD.rst). I haven't yet thought what kind of interface it should have there, but if you are interested to look at it we can discuss about it on #devel. Probably best to get this script first ported to Python 3 separately anyway.

doc/userguide/ug2html.py
```diff
@@ -193,24 +197,27 @@ def _copy_installation_instructions():
 #
 def create_distribution():
     import re
-    from urlparse import urlparse
+    try:
+        from urlparse import urlparse
+    except ImportError:
+        from urllib.parse import urlparse
```
pekkaklarck: No need for Py 2/3 compatibility.

doc/userguide/ug2html.py
```diff
@@ -248,20 +255,20 @@ def copy(source, dest):
     with open(ugpath) as infile:
         content = link_regexp.sub(replace_links, infile.read())
     with open(os.path.join(outdir, os.path.basename(ugpath)), 'wb') as outfile:
-        outfile.write(content)
-    print os.path.abspath(outfile.name)
+        outfile.write(content.encode(sys.getdefaultencoding()))
```
pekkaklarck: When working with various operating systems, we definitely don't want outputs to be written using OS default encoding. Should be using UTF-8 and it should be given as an argument to both of the `open()` calls above. Also, no need to open text files in binary mode.

doc/userguide/ug2html.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
```
pekkaklarck: This should be changed to `python3`. Otherwise running the script like `doc/userguide/ug2html.py` will typically run it on Python 2.

doc/userguide/ug2html.py
```diff
@@ -155,21 +157,23 @@ def create_userguide():
...
     ugpath = os.path.abspath(arguments[-1])
-    print ugpath
+    print(ugpath)
     return ugpath, version
 
 
 def _update_version():
     version = _get_version()
-    print 'Version:', version
+    print('Version: %s' % version)
```
pekkaklarck: Using `print('Version', version)` would work fine on Python 3 as well. And we don't need to care about Python 2. Alternatively this script could be updated to Python 3.6 making it possible to use the awesome f-strings.

doc/userguide/ug2html.py
HelioGuilherme66: I have forgotten this one :)

HelioGuilherme66:
@pekkaklarck Done all requested changes.

pekkaklarck:
Thanks! Still need to change the shebang line to 3.6 because the code now uses f-strings. I'll do it myself.

Issue #2903, opened 2018-07-18
Add note in documentation about `${SPACE}` to clarify that it means the ASCII space


laurentbristiel:
From @pekkaklarck : "The above said, the documentation of `${SPACE}` could probably be enhanced to clarify that it means the ASCII space (`\x20`) and other spaces should be specified using the aforementioned escape sequences"

laurentbristiel:
https://github.com/robotframework/robotframework/pull/2904

pekkaklarck:
Thanks for the PR #2904! Not sure would this have required an issue, but since you added it let's include it into RF 3.1 scope as well.

pekkaklarck:
Nice to get a reference to "Yucca's" pages by the way. He was working for Helsinki University of Technology back in the day when I studied there.

Also, didn't know there are *that* many spaces: http://jkorpela.fi/chars/spaces.html

PR #2904, opened 2018-07-18
note in documentation about  to clarify that it means the ASCII space #2903

laurentbristiel:


pekkaklarck:
Thanks! There could be some examples but I can add them myself.

PR #2905, opened 2018-07-19
Named-only arguments in dictionary error tests

juusoi:
Changing list and dictionary tests for argument error to use named-only arguments.


pekkaklarck:
Thanks! Great to see named-only args in such good use. =)

PR #2907, opened 2018-07-22
Add normalize path to USER env. var in path_expansion test case.

HelioGuilherme66:
Without this change, there were two test failures on Fedora Linux 28.


pekkaklarck:
Thanks @HelioGuilherme66!

Issue #2276, opened 2015-12-22
Remove support for Python 2.6 and 3.3


pekkaklarck:
It's time to drop Python 2.6 support in Robot Framework 3.1.

The main motivation:
- Python 2.6 hasn't been officially supported, even with security updates, since [October 2013](https://www.python.org/download/releases/2.6.9/).
- Supporting Python 2.6 requires some time. That time is better spent fixing bugs and developing new features.
- We get to use set literals, dict comprehensions, and other such syntax and modules not supported by Python 2.6.
- All test libraries and other code running on Python 2.6 ought to run unchanged on Python 2.7. Upgrading is thus only matter of upgrading the underlying Python version. If that is impossible, it is unlikely these users can upgrade to newer RF releases either.
- Users of older RedHat based Linux distributions that only contain Python 2.6 by default can easily install Python 2.7 from the official [Software Collections repository](https://www.softwarecollections.org/en/scls/rhscl/python27/).

For some more reasoning see http://www.snarky.ca/stop-using-python-2-6. Notice that we do not plan to drop Python 2.7 support in the foreseeable future.

pekkaklarck:
Should we also drop official support from Python 3.3 at the same time? It [reached EOL at 2017-09-29](
https://www.python.org/downloads/release/python-337/). I doubt this would make any difference in code, but having one less Python versions to test against would be nice. Are there Linux distributions in wide use that have Python 3.3 by default and where installing newer version is inconvenient?

HelioGuilherme66:
I agree that support for Python 3.3 should be dropped. I think that Python 3.4 came really fast to correct problems on 3.3. Don't know of any Linux distributions with Python 3.3.

hugovk:
I agree to drop support for Python 3.3.

None of the currently-supported enterprise Linux distros ship 3.3:
![image](https://user-images.githubusercontent.com/1324225/31711628-73d761f0-b401-11e7-86c0-fa0c5442d098.png)
https://en.wikipedia.org/wiki/CPython#Enterprise_Linux

By the way, here's pip installs for robotframework from PyPI for the last month:
```
$ pypinfo --percent --pip robotframework pyversion
python_version percent download_count
-------------- ------- --------------
2.7              90.1%         74,397
3.6               5.1%          4,181
3.5               2.3%          1,879
3.4               1.5%          1,225
2.6               0.7%            538
3.3               0.3%            236
None              0.1%             57
3.7               0.0%             21
3.2               0.0%              1
```

[...]

PR #2908, opened 2018-07-23
Adds enum support to the typeconverter.

jeroenroosen:
This PR adds Enum support to the typeconverter and relevant tests.

Because the `enum` I added for the test starts with an upper case character I wasn't able to use the existing template for the expected failure.

### DIFF
--- a/atest/testdata/keywords/type_conversion/python_annotations.robot
+++ b/atest/testdata/keywords/type_conversion/python_annotations.robot
@@ -99,6 +99,14 @@ String None is converted to None object
     Dictionary
     Set
 
+Enum
+    Enum          BAR                 Foo.BAR
+
+Invalid Enum
+    Run Keyword And Expect Error
+    ...    ValueError: Argument 'argument' cannot be converted to Foo, got 'foobar'.
+    ...    Enum    foobar
+
 *** Keywords ***
 Conversion Should Fail
     [Arguments]    ${kw}    ${arg}
--- a/atest/testdata/keywords/type_conversion/python_annotations.robot
+++ b/atest/testdata/keywords/type_conversion/python_annotations.robot
@@ -103,16 +103,15 @@ Enum
     Enum          BAR                 Foo.BAR
 
 Invalid Enum
-    Run Keyword And Expect Error
-    ...    ValueError: Argument 'argument' cannot be converted to Foo, got 'foobar'.
-    ...    Enum    foobar
+    [Template]    Conversion Should Fail
+    Enum          foobar    type=Foo
 
 *** Keywords ***
 Conversion Should Fail
-    [Arguments]    ${kw}    ${arg}
+    [Arguments]    ${kw}    ${arg}    ${type}=${kw.lower()}
     ${error} =    Run Keyword And Expect Error    *    ${kw}    ${arg}
     Should Be Equal    ${error}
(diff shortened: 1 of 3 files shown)


atest/testdata/keywords/type_conversion/python_annotations.robot
```diff
@@ -99,6 +99,14 @@ String None is converted to None object
...
+    Enum          BAR                 Foo.BAR
+
+Invalid Enum
+    Run Keyword And Expect Error
+    ...    ValueError: Argument 'argument' cannot be converted to Foo, got 'foobar'.
+    ...    Enum    foobar
+
 *** Keywords ***
 Conversion Should Fail
     [Arguments]    ${kw}    ${arg}
```
pekkaklarck: You ought to be able to add new argument `${type}=${kw.lower()}` here and use this keyword also with the new `Invalid Enum` test like

    Conversin Should Fail    Enum    foobar    type=Foo

or with `[Template]` if you want to test multiple values like existing tests do.

src/robot/running/arguments/typeconverter.py
```diff
@@ -14,6 +14,7 @@
 #  limitations under the License.
 
 from ast import literal_eval
+from enum import EnumMeta
```
pekkaklarck: Enums being new in Python 3.4, this is probably the first change not compatible with Python 3.3. Good that we decided to drop its support along with Python 2.6 in #2276.

src/robot/running/arguments/typeconverter.py
```diff
@@ -39,12 +40,17 @@ def convert(self, positional, named):
     def _convert(self, name, value):
         if name not in self._argspec.types or not is_string(value):
             return value
+        if value.upper() == 'NONE':
+            return None
```
pekkaklarck: Is moving these two lines related to enums? For me it feels a bit wrong to convert `NONE` string to `None` always if an argument has annotation. For example, if we have `def example(arg: str)`, we should pass all strings (and all other arguments) without changes. This conversions is ought to be safe if the annotation is a type we are going to convert and know wouldn't work with strings in the first place.

src/robot/running/arguments/typeconverter.py
```diff
@@ -39,12 +40,17 @@ def convert(self, positional, named):
...
         type_ = self._argspec.types[name]
+        if isinstance(type_, EnumMeta):
+            try:
+                return type_[value]
+            except KeyError:
+                self._raise_convert_failed(name, type_.__name__, value)
```
pekkaklarck: Although enums need to be recognized differently than other types, it would be better to have a separate `_conver_enum` helper for conversion. Code here could hen be change to something like:

```python
if isinstance(type_, EnumMeta):
    converter = self._convert_enum
else:
    converter = self._converters.get(type_)
```

jeroenroosen:
I incorporated your feedback, is [this enum converter function](https://github.com/robotframework/robotframework/pull/2908/files#diff-19c1073eff418b20b5a45af0d0df17b5R86) what you had in mind?

pekkaklarck:
Didn't realize `_convert_enum` needs a different signature than other convert functions. How you handled that with `_enum_converter_for` works fine.

pekkaklarck:
I'll add the missing tests myself. I got some extra time for coding and want to merge this PR to avoid conflicts.

PR #2911, opened 2018-07-25
Allow Remote Library connection over HTTPS

LSumbler:
Previously remote library connections only worked with HTTP, now works with both HTTP and HTTPS.

### DIFF
--- a/src/robot/libraries/Remote.py
+++ b/src/robot/libraries/Remote.py
@@ -197,7 +197,7 @@ class RemoteResult(object):
 class XmlRpcRemoteClient(object):
 
     def __init__(self, uri, timeout=None):
-        if 'https' in uri:
+        if uri.startswith('https://'):
             transport = TimeoutHTTPSTransport(timeout=timeout)
         else:
             transport = TimeoutHTTPTransport(timeout=timeout)
@@ -249,6 +249,7 @@ class XmlRpcRemoteClient(object):
 
 
 class TimeoutHTTPTransport(xmlrpclib.Transport):
+    _connection_class = httplib.HTTPConnection
 
     def __init__(self, use_datetime=0, timeout=None):
         xmlrpclib.Transport.__init__(self, use_datetime)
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Looks good on a quick look on mobile. Will look more closely and merge later. Could you please submit an issue about this for tracking purposes as well?

LSumbler:
Issue #2912 raised to track this issue

src/robot/libraries/Remote.py
```diff
@@ -197,7 +197,10 @@ def _convert(self, value):
 class XmlRpcRemoteClient(object):
 
     def __init__(self, uri, timeout=None):
-        transport = TimeoutTransport(timeout=timeout)
+        if 'https' in uri:
```
pekkaklarck: It would be safer to use `if uri.startswith('https://'):` here. Could even use `if urlparse(uri).scheme == 'https':` but that's perhaps an overkill and importing `urlparse` is a bit annoying because it is different modules depending on Python versions.

src/robot/libraries/Remote.py
```diff
@@ -260,6 +263,21 @@ def make_connection(self, host):
...
+        xmlrpclib.Transport.__init__(self, use_datetime)
+        if not timeout:
+            timeout = socket._GLOBAL_DEFAULT_TIMEOUT
+        self.timeout = timeout
+
+    def make_connection(self, host):
+        if self._connection and host == self._connection[0]:
+            return self._connection[1]
+        chost, self._extra_headers, x509 = self.get_host_info(host)
+        self._connection = host, httplib.HTTPSConnection(chost, timeout=self.timeout)
```
pekkaklarck: Do I get it right that the only difference between `TimeoutHTTPTransport` and `TimeoutHTTPSTransport` is that on this line the former uses `httplib.HTTPConnection` while the latter uses `httplib.HTTPSConnection` ? In that case there's a lot of unnecessary duplication.

One simple way to avoid duplication would be adding a class variable like `_connection_class =  httplib.HTTPConnection` to `TimeoutHTTPTransport` and then changing this line in that class to:

```python
self._connection = host, self._connection_class(chost, timeout=self.timeout)
```

That would allow changing `TimeoutHTTPSTransport` implementation to this:

```python
class TimeoutHTTPSTransport(TimeoutHTTPTransport):
    _connection_class = httplib.HTTPSConnection
```

pekkaklarck:
Can you @LSumbler take a look at the comments I added to the code?

I'd generally like all changes to be backed with automated tests, but in this case setting up a remote server using HTTPS is probably not worth the effort. Do you @LSumbler possibly have such a remote server publicly available so that I could manually test the changes?

LSumbler:
@pekkaklarck Thanks for your review notes, I've now updated the code to reflect the changes you suggested.
With regards to testing with a remote server, the one I am using is unfortunately not publicly available.

pekkaklarck:
Changes look great. It's a shame I cannot test this myself easily, but it's certain this won't break anything and I believe you've tested this at your end. Do you still want to add yourself to AUTHORS.txt file?

LSumbler:
Yes, tested it and using the changes already, hopefully shouldn't affect anyone else if there is something I've overlooked as it is new functionality. I'm happy not to be added to the AUTHORS.txt, after all this is only a small change.

pekkaklarck:
Thanks for the PR @LSumbler!

Issue #2937, opened 2018-08-21
Possibility to reset remote connection


juhoarvid:
There would be need to do reset for remote connection at least in following conditions:
- remote server is requested to do restart for testing purpose
- remote server is requested to do something which is expected to broke remote connection for a while

juhoarvid:
Any comments or review for:
https://github.com/robotframework/robotframework/pull/2938

pekkaklarck:
I guess this could be useful. There are some problems with the code in the PR that ought to be pretty easy to fix. More importantly it would need tests and documentation.

juhoarvid:
> 
> 
> I guess this could be useful. There are some problems with the code in the PR that ought to be pretty easy to fix. More importantly it would need tests and documentation.

Can I do something? Is there documentation rule follow, test procedure to execute and descritption what should be fixed?

juhoarvid:
Short check with "Reload Library" shows it is working a bit differently, but may lead to same situation. Will need a bit more time that I can verify it fulfills my need

pekkaklarck:
What's the situation with this issue/PR?

PR #2938, opened 2018-08-21
[#2937] ability to do remote server reset

juhoarvid:
- remote library instace connection can be resetted
- needs remote library used in own library, no possibility to straight
  call "Reset Connection" kw from robot scripts


pekkaklarck:
Could you clarify the use case and most importantly the usage?

PR #2960, opened 2018-09-17
libraries/Telnet: fix window size by unifying order of rows and columns

pietrushnic:
This patch fix inconsistency in row and column parameter across Telnet
library. Without this patch users following documentation cannot
correctly set terminal emulator, because window_size is interpreted in
incorrect order.

Signed-off-by: Piotr Król <piotr.krol@3mdeb.com>


pekkaklarck:

PR #2961, opened 2018-09-17
libraries/Telnet: add write_bare_slow functions

pietrushnic:
Some environments cannot consume characters with speed that Telnet
library provide, because of that we added function that can write
character by character with given delay.

Signed-off-by: Piotr Król <piotr.krol@3mdeb.com>

### DIFF
--- a/src/robot/libraries/Telnet.py
+++ b/src/robot/libraries/Telnet.py
@@ -795,14 +795,22 @@ class TelnetConnection(telnetlib.Telnet):
             return self._encode(self._newline)
         return self._newline
 
-    def write_bare(self, text):
+    def write_bare(self, text, char_delay=None):
         """Writes the given text, and nothing else, into the connection.
+
+        If char_delay parameter specified function sends characters one by one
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Why not to add an optional delay to the existing `Write Bare` keyword?

pietrushnic:
@pekkaklarck please let me know if it looks better now.

pekkaklarck:
Looks OK but there are various things to do still before this can be merged:

- Tests are needed.
- Documentation should get a note that this is a new feature in RF 3.1 (assuming this makes into that release).
- Time should be interpreted as Robot's "time string". See `timestr_to_secs` usage in the code and an appendix of the User Guide for more details about time formats.
- There's no need to have `list()` in `for ch in list(text)`.
- Need to decided to you want to have delay between Unicode characters or between bytes. If the latter, text should be encoded first and only then bytes iterated over (e.g. `for byte in self._encode(text)`).

pekkaklarck:

Issue #2952, opened 2018-09-05
Signal handler registered outside Python causes error


JonKoser:
I am using robot.run in a script of mine, which I execute from another open application, to run some robot tests. Trouble is, I keep getting the following TypeError when I try and execute my script.

![image](https://user-images.githubusercontent.com/9324338/45095799-94c87f80-b0e4-11e8-931e-4db70e86aac5.png)

This is preventing my tests from finishing and tearing down properly.

Upon some inspection, it looks like self._orig_sigint and self._orig_sigterm are getting set to "None" from signal.getsignal()
![image](https://user-images.githubusercontent.com/9324338/45096004-1ddfb680-b0e5-11e8-8aa8-93c28894bf34.png)

Oddly enough, this only happens when I first run my script. If I re-run the script again, it works fine.

I can edit some things myself in signalhandler.py to get it to work, but I'd rather any changes like that come from the official robotframework folks :)

pekkaklarck:
According to the docs of [signal.getsignal](https://docs.python.org/2/library/signal.html#signal.getsignal), it returns `None` if the signal handler has been defined outside Python. In that case we shouldn't try to restore the old value because that won't work. We don't want to leave our signal handler registered either, so I guess we need to either register `SIG_IGN` or `SIG_DFL` when we exit. Probably the latter is better.

It's pretty easy to fix this, but It would be great to be able to create a test for it also. Does anyone know how can we register a signal handler so that `signal.getsignal()` returns `None`? If that's not easy, we can obviously mock it.

PR #2964, opened 2018-09-20
Fixed issue #2952 - signal handler registered outside Python causes e…

JonKoser:
Fixing issue #2952

For the solution, I also went with restoring signal.SIG_DFL, not signal.SIG_IGN when self._orig_sigint/sigterm are 'None'.

I couldn't figure out a nice way to register a signal outside of Python during test execution so I went with mocking for the test. I created an instance of the _StopSignalMonitor then set self._orig_sigint/sigterm to 'None' as if they had been set to 'None' in __enter__. On __exit__, we should no longer get the TypeError from signal.signal().

 If there is a better and/or cleaner way to do this, I'd love to know!

### DIFF
--- a/src/robot/running/signalhandler.py
+++ b/src/robot/running/signalhandler.py
@@ -56,8 +56,8 @@ class _StopSignalMonitor(object):
 
     def __exit__(self, *exc_info):
         if self._can_register_signal:
-            signal.signal(signal.SIGINT, self._orig_sigint)
-            signal.signal(signal.SIGTERM, self._orig_sigterm)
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Looks great! But is try/except/else really needed in test?

pekkaklarck:
In my opinion these assertions in the test are unnecessary and just makes the test more complex:

```python
            assert_equal(None, monitor._orig_sigint)
            assert_equal(None, monitor._orig_sigterm)
```

Otherwise everything looks so good that I'll just merge this and may remove those lines myself later.

PR #2972, opened 2018-10-07
Fix collections documentation

adongy:
Libdoc parser has difficulty parsing the attached ``key`` / ``value``.


pekkaklarck:
Thanks!

PR #2984, opened 2018-10-13
add time attribute into the testsuite element

kanchi240:
Add `time` attribute into the `<testsuite>` element of the Xunit output file.

Issue #2979, opened 2018-10-11
xUnit outputs: Add `time` attribute to `<testsuite>` element


kanchi240:
Hi, 

I'm using [unitth](http://junitth.sourceforge.net/) to generate test execution history report, which need `time` attribute at the `testsuite` node of the xunit output file as follow:

![15392298978774](https://user-images.githubusercontent.com/4908696/46780102-8c034300-cd4d-11e8-8dc8-fc2215c14c2f.png)

and it is better to add `testsuite` node under the `testsuites` top-level node, see: http://llg.cubic.org/docs/junit/
#2292 

Thank you for your support!

----

UPDATE by @pekkaklarck: This issue covers only adding the `time` attribute and #2982 is possibly changing the file structure.

pekkaklarck:
This is easy to do, but requires testing that changes are backwards compatible and work with various tools consuming xUnit outputs. I'm fairly certain adding `time` is safe, but adding separate `testsuite` elements for each nested suite is a bigger change. Can you @kanchi240 test this? Or could @rasjani possibly help?

kanchi240:
Not sure have time, please make this issue to lower priority, I have written a python script to make it work.
Or just add the time attribute into the top-level testsuite node firstly?

pekkaklarck:
Adding the `time` attribute to the current `<testsuite>` element and representing each suite with a separate `<testsuite>` element are separate enhancements. Let's concentrate only on the former in this issue. I already submited #2982 about the latter.

pekkaklarck:
Adding the `time` attribute is a very simple change. Anyone interested to do it and get some experience contributing to Robot Framework?

It's also pretty safe, and can only cause backwards compatibility problems if someone validates our xUnit outputs and expect that such attribute doesn't exist. I don't see any valid reason to do such validation, though, and making the attribute optional in a schema would nevertheless be easy.

Because this is simple to implement and safe enhancement, I've added this to RF 3.1 scope.

PR #2986, opened 2018-10-16
add time attribute into the testsuite element of the Xunit out put file

kanchi240:
please close the previous commit #2984
and review this commit according to #2979 

Thanks for your help!


pekkaklarck:
Thanks for the PR @kanchi240!

Issue #2906, opened 2018-07-20
Add a new keyword like Get Selections From User


rfabbris:
In `Dialogs` library, nowadays there is only the possibility to choose one selection through a dropdown list (see `Get Selection From User`).
It would be really helpful to have a keyword to display tick boxes and to give the possibility to choose multiple selections.
The input could be a list with the possible selections (e.g. `${fruits} =  Create List  apple  banana`)
The output could be either a list with the chosen options or a dictionary with all the possibilities and the information of which one of them has been chosen (e.g. `${fruits_chosen} =  Get Selections From User  ${fruits}`)

pekkaklarck:
Sounds good to me. I think inputs should be same as to `Get Selection From User`. Return value probably should be a list of selected values. Anyone interested to implement this?

marq35:
You can assign me to this issue :)

pekkaklarck:
We got a PR so let's try to get this into 3.1.

PR #2987, opened 2018-10-16
Added keyword Get Selections From User

cristii006:
PR for issue #2906

### DIFF
--- a/atest/robot/standard_libraries/dialogs/dialogs.robot
+++ b/atest/robot/standard_libraries/dialogs/dialogs.robot
@@ -46,6 +46,9 @@ Get Selection From User Cancelled
 Get Selection From User Exited
     Check Test Case    ${TESTNAME}
 
+Get Selections From User
+    Check Test Case    ${TESTNAME}
+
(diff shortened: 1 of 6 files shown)


pekkaklarck:
Did a little cleanup and enhanced tests a bit as well in the commit referenced above. Noticed that with Java the dialog shows only four items. This combined with the issue that items need to be selected while pressing control decrease usability a bit. I looked at these problems and learned the following:

- Couldn't find anything in API docs about controlling how to select values. Needing to use Ctrl isn't that bad so perhaps we can forget about this.

- Only four items are shown because the new code explicitly calls `setVisibleRowCount(4)`. The default would be 8, but that causes the layout to look *really* ugly. The uglyness is due to using `GridLayout` which stores each item inside it (the message and the option list in our case) in separate cells and *makes sure each cell has same size*. I guess this should be fixed by using some other layout manager (e.g. `GridBagLayout` or `BoxLayout`) but my Swing skills are very limited. Do you @cristii006 have any more experience with Swing layouts?

PR #2989, opened 2018-10-16
Alternate name for "Run Keyword And Expect Error": "Expect Error From Keyword"

DanielPBak:
Keywords should mention their arguments in the same order that they are expected: Run Keyword And Expect Error expects an error first and then a keyword, which is confusing for somebody not familiar with the BuiltIn library.

This is a structure that is common among all of the standard libraries, and this is the only exception that I am aware of.

For example: Wait Until Element Contains (locator, text) and Send Signal To Process (signal, handle).

I've added an alternative name for this keyword where the name mirrors the expected arguments.


HelioGuilherme66:
I don't think this needs to be BuiltIn in RoboFramework. After all it is just an alias for the keyword, so you should keep it in your own library extension, or even an user keyword.
I agree that, by looking isolated to the keyword it may standout from the others, but the keyword belongs to the group of "Run Keywords ..." and it is well documented (if not, then documentation improvements would be welcomed).

pekkaklarck:
I totally agree with @HelioGuilherme66. The name is consistent with other similar keywords and documentation can be enhanced if the usage is not clear. Notice also that in the future way may add some dedicated control structure like `TRY/EXCEPT` to Robot's syntax to handle this case.

Issue #2285, opened 2016-01-02
Support controlling continue-on-failure mode using test and keyword tags


jaloren:
The [continue on failure](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#continue-on-failure)  mode allows execution to continue after a keyword has failed. This functionality is very useful but has a couple short comings.
## Problem
- Outside of custom keywords defined in a python library, a user must leverage additional keywords on top of the existing keyword which can significantly increase verbosity and complexity of custom keywords (e.g Run Keyword And Continue On Failure, Run Keyword and Return Status, and Run Keyword If). 
- there is no easy way to turn continue on failure mode on for setup like one has it for teardown
- there is no easy way to run continue on failure mode off for teardown like it works for set up.
## proposed solution

after discussing this with Pekka, current proposal is to implement reserved tags.
- robot-continue-on-failure
- robot-stop-on-failure
- robot-exit-on-failure.

---

**UPDATE:** Tags that int the end were implemented are `robot:continue-on-failure` and `robot:recursive-continue-on-failure`. See the discussion below for details.

pekkaklarck:
I agree on the problem. Instead of having separate keywords to turn the mode on or off, I would prefer user keywords having a way to explicitly set the mode. It could be done with a new setting like `[Continue Mode]` or with a special keyword tag like `robot:continue-on-failure`. Using tags would be a lot easier to implement and would not require syntax changes that editors and other tools would need to support.

jaloren:
@pekkaklarck: I forgot that newer versions of RF allow setting tags in a keyword and I hadn't realized there was such a thing as a reserved tag . Given that, the tag approach sounds nice, especially since that means the RF report and stats will automatically identify and flag any keyword or test cases that use it.

According to the documentation, a reserved tag is denoted with robot- so does that mean it would be something like:

robot-continue-on-failure

pekkaklarck:
Yeah, something like that. Could then also add `robot-exit-on-failure`.

pekkaklarck:
No activity here lately, but I consider this so important that hopefully we get it implemented at some point. 

We can still consider using `robot:` prefix instead of `robot-` prefix with reserved tags. For some reason I know think the former feels better. I think they only have one usage at the moment so changing is possible.

jaloren:
I tend to agree that the colon looks cleaner -- mostly aesthetic from my point of view though. Could see someone arguing it either way.

ISY-Thomas:
[...]

PR #3000, opened 2018-11-02
Add robot:exit-on-failure tag for keyword

hellowangsai:
How and why do it, see:
https://github.com/robotframework/robotframework/issues/2285

After this PR, keywords with tag `robot:exit-on-failure` exit immediately if failure happens. Even it run in tear down phase. By the way, for loops inherit this tag from loop's parent keyword.


pekkaklarck:
This PR has some serious problems:

- There are changes to two unrelated features (exit-on-failure mode and new Set Test Appendix keyword).
- The implemented exit-on-failure feature seems to only affect how teardowns are handled. In Robot's terminology exit-on-failure means stopping execution if there's a failure. I don't know is the implementation wrong or is it just named wrong.
- No tests (nor documentation).

Could you clarify what is the idea behind this PR?

pekkaklarck:
No answers to my questions related to the problems in this PR. Closing.

PR #3001, opened 2018-11-02
Iterate through list of connections to return right one

jluhrsen:
This is an attempt to solve a new bug seen in SSHLibrary
3.2.0. Issue filed here:

  https://github.com/robotframework/SSHLibrary/issues/279

Signed-off-by: Jamo Luhrsen <jluhrsen@redhat.com>


pekkaklarck:
This fix seems to be specific to SSHLibrary and would break all other usages of ConnectionCache. There is not guarantee that a connection would have `config`.

PR #3018, opened 2018-12-01
Add skip console highlighting.

windelbouwman:
Regarding the long open issue about skipping tests, maybe this might be an approach to break up the changes in easily reviewable pieces.

This work contains highlighting on the console for skipped tests only.

### DIFF
--- a/atest/robot/cli/console/console_resource.robot
+++ b/atest/robot/cli/console/console_resource.robot
@@ -7,8 +7,8 @@ ${SEP_CHAR1}      =
 ${SEP_CHAR2}      -
 ${SEP_LINE1}      ${SEP_CHAR1 * 78}
 ${SEP_LINE2}      ${SEP_CHAR2 * 78}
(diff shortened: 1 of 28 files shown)


windelbouwman:
@pekkaklarck do you think this change is a good idea? Or do you prefer the work in one big chunk?

pekkaklarck:
Does this change have any effect?

pekkaklarck:
I definitely don't want to add unused code to the master.

windelbouwman:
Okay, let me re-phrase. At the moment, this code is unused. In future, this will be used when the SKIP status is added.

windelbouwman:
I added some more changes that include adding the SKIP status.

pekkaklarck:
I don't want to add code to master that's not used regardless would be used in the future or not. Especially not when it's unclear if/when it is actually used and how would it be used. Such code could be added to a feature branch, but absolutely not to master.

windelbouwman:
Okay, I understand.

Issue #2512, opened 2016-12-30
BuiltIn: Enhance error reporting of comparison keywords especially when `ignore_case=True`


pekkaklarck:
Case-insensitivity support added in RF 3.0.1 (#2439) works fine otherwise, but error reporting isn't always consistent. Sometimes error messages show the compared values turned to lower case, and in others they show the original values. It would probably be best to to always show the original values but add a note like `(case-insensitive)` to the message automatically.

There are also other inconsistencies in error reporting (e.g. do messages end with `.` or not) that can be fixed when refactoring error reporting in general. Having a better way to report errors involving large strings or containers with many items would be great too, but that is probably so big task that it would deserve a separate issue.

pekkaklarck:
I don't have time to implement this before RF 3.1. If someone else is interested, I can help. Shouldn't be too complicated even for new contributors, but decided not to add the `good first issue` label because there can be quite a bit of work to go through all related keywords.

PR #3056, opened 2019-01-17
Enhancing error report for keywords containing 'case insensitive'

cristii006:
This PR implements the fixes requested by #2512 .


pekkaklarck:
Don't have time to look at this PR in detail but two quick comments:

- The information about case-insensitivity should be in the error message, not only logged. Otherwise it's not easily visible in the log file and not at all visible in report. A problem is that error messages are partly created by the functions in the `robot.utils.asserts` module and partly by BuiltIn itself.

- I don't think new functions are needed in the `asserts` module. Instead existing ones could get `ignore_case=False` option that they could use to format the error message correctly.

PR #3117, opened 2019-03-12
Update Easter.py

MisterChild:
None Shall Pass !

### DIFF
--- a/src/robot/libraries/Easter.py
+++ b/src/robot/libraries/Easter.py
@@ -19,9 +19,9 @@ from robot.api import logger
 def none_shall_pass(who):
     if who is not None:
         raise AssertionError('None shall pass!')
-    logger.info('<object width="480" height="385">'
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Is the old way to embed the video outdated or why should it be changed? If there are problems, fixing them before Easter is obviously very high priority. =)

MisterChild:
The linked video was not available anymore and youtube suggests this way of implementation. So I updated video and way of embedding. I'm glad to hear, that development priorities are fully grounded and long-sighted the same time. :-)

pekkaklarck:
Thanks for the PR and happy Easter!

pekkaklarck:
This change broke a test we had but I fixed it. At the same time noticed a small problem with the code and fixed that as well. The code problem was related to this:

```python
>>> len('Beginning \
...      end')
18
>>> len('Beginning'
...     'end')
12
```

Issue #2698, opened 2017-10-26
Possibility to automatically expand certain keywords in log file


DirkRichter:
When opening a *log.html* file, all failed suites and keywords are shown directly (auto expanded).

I suggest a new **rebot** cli argument **--autoexpand keyword**
in order to auto expand all matching keywords when opening a **log.html** file (e.g. **Take Screenshot** or SelemiumLibrary.**Capture Page Screenshot**).

pekkaklarck:
Sounds useful but would probably be somewhat complicated to implement. The main complexity is due to the fact that logs are dynamically created while you open keywords, and automatically opening some of the keywords would likely take a lot of time with larger log files. Most likely the list of keywords to open should be collected already when generating the log file so that unnecessary keywords wouldn't need to be opened. If this would be implemented, I'd like to support keyword tags as well as names similarly as e.g. `--removekeywords`.

Are you interested to implement this yourself?

DirkRichter:
I tried, but run into trouble.

The problem is: The expandDecider do not know anything about child's and keywords inside. Scanning all child's and subchild's is not an option. But: expanding nodes with status=FAIL works, because the status is propagated to mother nodes. A similar **expandStatus* is needed in stored data, which is a complex change.

Currently we have a workaround: we make screenshots in teardown when tests fails and apply a postprocessing to replace the default expandDecider with a more powerfull custom version, which is a kind of hack and not suited for a pull request. Thus: this topic is not urgent :-)

pekkaklarck:
Are you @DirkRichter still interested in this enhancement? As I commented above, this could be implemented by collecting ids of keywords to be opened, writing that to the log file, and log file then opening the listed keywords.

DirkRichter:
Hi. Thanks for your hints. Yes we are still interested in this (on my todo list). But currently the topic is not top priority due to the workaround.

pekkaklarck:
PR #3118 looks so promising (and safe) that I'll add this to RF 3.1.2 scope.

pekkaklarck:
Talked with @DirkRichter on Slack and he doesn't have time to finalize PR #3118 in time for RF 3.1.2. Moving to RF 3.2 scope.

pekkaklarck:
Are you @DirkRichter still interested in this issue? RF 3.2 is getting ready and if we want this in, PR #3118 should somewhat soon be finalized.

jura84:
I would like to know the reason behind this.

[...]

PR #3118, opened 2019-03-12
new rebot cli arg --expandkeywords for #2698

DirkRichter:
As requested in #2698 the new cli arg **--autoexpandkeywords** was implemented, which automatically expands all keywords in log.html.

Cite from --help:
>    --autoexpandkeywords kw1,kw2,..
>                          After opening log.html these types of keywords 
>                          separated by comma will be expanded per default.
>                          Example: --autoexpandkeywords "Take Screenshot,Capture Page Screenshot"


pekkaklarck:
Looks great on a quick look. Some comments based on that:

- It would be good to be able to expand keywords based on tags as well. Somewhat similar '--removekeywords` and `--flattenkeywords` support that.
- Instead of separating keywords with a comma, I'd consider making it so that you need to use this option multiple times to match multiple keywords. That's how the aforementioned `--remove/flattenkeywords` and other similar options work. Would also be pretty much a requirement to allow expanding both using a name and using tags.
- New code didn't always follow our coding conventions. They are documented in CONTRIBUTING.rst but basically we follow PEP-8.
- Documentation in the User Guide would be needed. That's also explained in CONTRIBUTING.rst.
- It would be good to have tests for this functionality but I'm not sure how easy that is. We don't have any tests for rendering log file, and the new JS and HTML code is just using existing functionality anyway. Could check that correct info is written to log.html file (we got some tests like that) but not sure is it worth the effort. Would be good to check that anyway.

DirkRichter:
i improved the PR based on your comments and:
- added support for matching by tags
- refactoring of current matching to name:/tag: prefix
- changed parameter type to same as flattenkeywords
- follow PEP-8 rules + rules from CONTRIBUTING.rst
- added documentation in User Guide

pekkaklarck:
On a very quick look latest change look good. I try to do a full review soonish and also test this feature myself. The issue is targeted to RF 3.1.2, but you can also ping me here or on Slack if it takes time.

pekkaklarck:
Finally looking this for real. Initial impression is very positive, but I noticed one somewhat big problem we need to discuss and probably fix.

It seems that when matching with name, the given name is matched against keywords "short name" (e.g. `Log`), not against its "full name" (e.g. `BuiltIn.Log`) like, for example, `--removekeywords` behaves. In my opinion matching the full name would be better because it's more explicit, avoids problems with different keywords having keywords with same names (or matching name when using patterns), and is consistent with other similar usages. The consistency argument is especially strong when the `--help` text tells to look at `--removekeywords` for details.

We could also consider implementing this so that both short and long names could be used, but then we needed to change also `--removekeywords` and others to keep functionality consistent. I doubt that's worth the effort, though.

doc/userguide/src/Appendices/CommandLineOptions.rst
```diff
@@ -56,6 +56,8 @@ Command line options for test execution
   --tagstatcombine <tags:title>  Creates `combined statistics based on tags`_.
   --tagdoc <pattern:doc>  Adds `documentation to the specified tags`_.
   --tagstatlink <pattern:link:title>  Adds `external links`_ to the *Statistics by Tag* table.
+  --autoexpandkeywords <name:pattern|tag:pattern>  initially `expand keywords`_
```
pekkaklarck: Not sure is "initially" that good word here. Perhaps "automatically" would be better?

src/robot/htmldata/rebot/log.js
```diff
@@ -62,6 +62,16 @@ function expandElementWithId(elementid) {
     expandElement(window.testdata.findLoaded(elementid));
 }
 
+function expandElementsWithIds(ids) {
+    util.map(ids,expandElementWithId)
```
pekkaklarck: Space needed after comma and semicolon at the end. Semicolon missing also from the next function. #nitpick

src/robot/htmldata/rebot/log.html
```diff
@@ -52,6 +52,7 @@ <h1>Opening Robot Framework log failed</h1>
     addErrors();
     addExecutionLog(topsuite);
     addLogLevelSelector(window.settings['minLevel'], window.settings['defaultLevel']);
+    loadAndExpandElementIds(window.output['auto_expand_ids'])
```
pekkaklarck: Semicolon missing at the end.

doc/userguide/src/ExecutingTestCases/OutputFiles.rst
```diff
@@ -552,6 +552,26 @@ deeply nested keyword structures.
...
+
+`TAG:<pattern>`
+   Expand keywords with tags matching the given pattern. Pattern matching
+   rules are same as when `removing keywords`_ using `TAG:<pattern>` mode.
+
+Examples::
+
+   rebot --autoexpandkeywords name:CapturePageScreenshot output.xml
+   rebot --autoexpandkeywords tag:ShowInLog output.xml
+
```
pekkaklarck: Need to add a note somewhere that this is a new feature in RF 3.1. There ought to be other features that have similar notes that can be used as examples.

src/robot/rebot.py
```diff
@@ -200,6 +200,10 @@
                           match (starting from 1).
                           Examples: --tagstatlink mytag:http://my.domain:Title
                           --tagstatlink "bug-*:http://url/id=%1:Issue Tracker"
+    --autoexpandkeywords name:<pattern>|tag:<pattern> *
+                          Matching keywords will be initially expanded in log.html.
+                          For pattern matching rules see --removekeywords option.
+                          Example: --autoexpandkeywords name:CapturePageScreenshot
```
pekkaklarck: `New in RF 3.1.2.` should be added somewhere.

src/robot/reporting/jsbuildingcontext.py
```diff
@@ -20,13 +20,16 @@
...
 
 from .stringcache import StringCache
 
 
+
 class JsBuildingContext(object):
 
-    def __init__(self, log_path=None, split_log=False, prune_input=False):
+    def __init__(self, log_path=None, split_log=False, prune_input=False,
+                 auto_expand_args=[]):
```
pekkaklarck: Mutable default values are always a bit questionable. Would be better to use `None` or possibly `()` here.

src/robot/reporting/jsexecutionresult.py
```diff
@@ -28,19 +28,21 @@
 class JsExecutionResult(object):
 
     def __init__(self, suite, statistics, errors, strings, basemillis=None,
-                 split_results=None, min_level=None):
+                 split_results=None, min_level=None, auto_expand_ids=[]):
```
pekkaklarck: Mutable default values are always a bit questionable. Would be better to use None or possibly () here.

src/robot/reporting/jsmodelbuilders.py
```diff
@@ -22,9 +22,10 @@
 class JsModelBuilder(object):
 
     def __init__(self, log_path=None, split_log=False,
-                 prune_input_to_save_memory=False):
+                 prune_input_to_save_memory=False, auto_expand_keywords=[]):
```
pekkaklarck: Mutable default values are always a bit questionable. Would be better to use None or possibly () here.

src/robot/result/autoexpandkeywordmatcher.py
```diff
@@ -0,0 +1,53 @@
...
+from robot.errors import DataError
+from robot.model import TagPatterns
+from robot.utils import MultiMatcher, is_list_like, py2to3
+
+
+def validate_autoexpandkeywords(options):
+    for opt in options:
+        low = opt.lower()
+        if not (low.startswith('name:') or
+                low.startswith('tag:')):
```
pekkaklarck: `str.startswith()` accepts a tuple of strings to match. This could be shortened to `opt.lower().startswith(('name:', 'tag:'))`. 

Validation is so simple that having it in `settings.py` could be considered as well.

src/robot/run.py
```diff
@@ -237,6 +237,10 @@
                           match (starting from 1).
                           Examples: --tagstatlink mytag:http://my.domain:Title
                           --tagstatlink "bug-*:http://url/id=%1:Issue Tracker"
+    --autoexpandkeywords name:<pattern>|tag:<pattern> *
+                          Matching keywords will be initially expanded in log.html.
+                          For pattern matching rules see --removekeywords option.
+                          Example: --autoexpandkeywords name:CapturePageScreenshot
```
pekkaklarck: `New in RF 3.1.2.` needed.

src/robot/result/autoexpandkeywordmatcher.py
```diff
@@ -0,0 +1,53 @@
...
+        self._matched_ids = []
+        if not is_list_like(auto_expand_list):
+            auto_expand_list = [auto_expand_list]
+        names = [n[5:] for n in auto_expand_list if n[:5].lower() == 'name:']
+        tags  = [p[4:] for p in auto_expand_list if p[:4].lower() == 'tag:']
+        self._namematcher = MultiMatcher(names)
+        self._tagmatcher = MultiMatcher(tags)
+
+    def check(self, kw):
+        name = kw.kwname
```
pekkaklarck: Changing this to `name = kw.name` ought to change the functionality so that matching name is done against keywords "long name".

Could also inline this variable assignment similarly as the one on the next line.

doc/userguide/src/ExecutingTestCases/OutputFiles.rst
DirkRichter: added a note to be new in Robot Framework 3.2

src/robot/rebot.py
DirkRichter: RF 3.2 note added

src/robot/reporting/jsbuildingcontext.py
DirkRichter: changed default to None + added support of None in AutoExpandKeywordMatcher

src/robot/reporting/jsexecutionresult.py
DirkRichter: refactoring to None (deserialized to null in JS). I checked ECMAScript specification and also do some tests, which showed JavaScript handles [] exactly like null in for loops (None from Python).

src/robot/reporting/jsmodelbuilders.py
DirkRichter: refactoring to None. It's used in JsBuildingContext, which now also supports None.

src/robot/result/autoexpandkeywordmatcher.py
DirkRichter: changed + checked required python version = 2.5, thus it's safe to use.

src/robot/run.py
DirkRichter: updated with RF 3.2.

src/robot/result/autoexpandkeywordmatcher.py
DirkRichter: reduced confusion by inlining

doc/userguide/src/Appendices/CommandLineOptions.rst
DirkRichter: agree. done.

doc/userguide/src/ExecutingTestCases/OutputFiles.rst
DirkRichter: updated with note new in RF 3.2

src/robot/rebot.py
DirkRichter: updated with note new in RF 3.2

src/robot/reporting/jsbuildingcontext.py
DirkRichter: refactoring to None + implemented handling of None

src/robot/reporting/jsexecutionresult.py
DirkRichter: refactoring to None + implemented handling of None

src/robot/reporting/jsmodelbuilders.py
DirkRichter: refactoring to None + implemented handling of None

src/robot/result/autoexpandkeywordmatcher.py
DirkRichter: applied improvement, but i prefer to stay validation in that file beacuse of SOLID design principle

src/robot/run.py
DirkRichter: updated with note to RF 3.2

src/robot/result/autoexpandkeywordmatcher.py
DirkRichter: inlined as suggested to reduce confusion

src/robot/rebot.py
```diff
@@ -201,6 +201,11 @@
                           match (starting from 1).
                           Examples: --tagstatlink mytag:http://my.domain:Title
                           --tagstatlink "bug-*:http://url/id=%1:Issue Tracker"
+    --autoexpandkeywords name:<pattern>|tag:<pattern> *
+                          Matching keywords will be initially expanded in log.html.
```
pekkaklarck: "log.html" -> "the log file"

src/robot/rebot.py
```diff
@@ -201,6 +201,11 @@
...
+                          For pattern matching rules see --removekeywords option.
+                          Example: --autoexpandkeywords name:CapturePageScreenshot.
```
pekkaklarck: The `.` at the ends should probably be removed.

src/robot/reporting/jsbuildingcontext.py
```diff
@@ -20,13 +20,16 @@
 from robot.utils import (attribute_escape, get_link_path, html_escape,
                          html_format, is_string, is_unicode, timestamp_to_secs,
                          unic)
+from robot.result.autoexpandkeywordmatcher import AutoExpandKeywordMatcher
 
 from .stringcache import StringCache
 
 
+
```
pekkaklarck: Unnecessary empty line.

src/robot/reporting/jsbuildingcontext.py
```diff
@@ -20,13 +20,16 @@
 from robot.utils import (attribute_escape, get_link_path, html_escape,
                          html_format, is_string, is_unicode, timestamp_to_secs,
                          unic)
+from robot.result.autoexpandkeywordmatcher import AutoExpandKeywordMatcher
```
pekkaklarck: We try to avoid imports from other modules so that they need to go to submodules. Instead of `from robot.module.submodule import something` we try to do `from robot.module import something` i.e. expose `something` on the module root.

In this particular case it seems that the `autoexpandkeywordmatcher` module is in wrong place. It is under `robot.result` but seems to be used only under `robot.reporting` and that feels like a better "home" for it in general. Then the import could be like `from .autoexpandkeywordmatcher import xxx`.

src/robot/reporting/resultwriter.py
```diff
@@ -132,7 +132,9 @@ def js_result(self):
         if self._js_result is None:
             builder = JsModelBuilder(log_path=self._settings.log,
                                      split_log=self._settings.split_log,
-                                     prune_input_to_save_memory=self._prune)
+                                     prune_input_to_save_memory=self._prune,
+                                     auto_expand_keywords
+                                     =self._settings.auto_expand_keywords)
```
pekkaklarck: 1. This formatting is a bit odd. Probably would be better to do something like
```python
builder = JsModelBuilder(
    options=here,
    with_suitable=indentation
)
```

2. The `prune_input_to_save_memory` option is somewhat special and I'd keep it as the last one.

src/robot/reporting/jsmodelbuilders.py
```diff
@@ -22,9 +22,10 @@
 class JsModelBuilder(object):
 
     def __init__(self, log_path=None, split_log=False,
-                 prune_input_to_save_memory=False):
+                 prune_input_to_save_memory=False, auto_expand_keywords=None):
```
pekkaklarck: The `prune_input_to_save_memory` is somewhat special argument. I'd keep it last even though it would possibly break backwards compatibility. It's very unlikely anyone calls this outside Robot's code and if they do, using `prune_input_to_save_memory=xxx` is a good idea anyway.

src/robot/result/autoexpandkeywordmatcher.py
```diff
@@ -0,0 +1,51 @@
...
+        self._namematcher = MultiMatcher(names)
+        self._tagmatcher = MultiMatcher(tags)
+
+    def check(self, kw):
+        if self._namematcher.match(kw.kwname) or self._tagmatcher.match_any(kw.tags):
+            self._matched_ids.append(kw.id)
+
+    @property
+    def all_matched_ids(self):
+        return self._matched_ids
```
pekkaklarck: Why property? Who not just have `matched_ids` or `ids` as public attribute?

src/robot/result/autoexpandkeywordmatcher.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __nonzero__(self):
+        return bool(self._matcher)
```
pekkaklarck: This is not correct. There's no `self._matcher` attribute. If nothing breaks, it seems this isn't used and could possibly be removed.

src/robot/reporting/jsbuildingcontext.py
```diff
@@ -68,6 +72,13 @@ def create_link_target(self, msg):
         id = self._top_level_strings.add(msg.parent.id)
         self._msg_links[self._link_key(msg)] = id
 
+    def check_auto_expansion(self, kw):
+        self._auto_expand_matcher.check(kw)
```
pekkaklarck: This might be premature optimization, but I'm slightly worried about the performance if that `.check(kw)` is called with all keywords. There can be a huge amount of keywords and calling matchers internally in that `check()` method is not free. It might be better to set `self._auto_expand_matcher to `None` if this functionality isn't enabled and then here have:
```python
if self._auto_expand_matcher is not None:
    # ...
```

src/robot/result/autoexpandkeywordmatcher.py
```diff
@@ -0,0 +1,51 @@
...
+from robot.errors import DataError
+from robot.model import TagPatterns
+from robot.utils import MultiMatcher, is_list_like, py2to3
+
+
+def validate_autoexpandkeywords(options):
+    for opt in options:
+        if not (opt.lower().startswith(('name:', 'tag:')):
+            raise DataError("Expected 'TAG:<pattern>', or "
+                            "'NAME:<pattern>' but got '%s'." % opt)
```
pekkaklarck: I'm not sure is this a good way to handle command line option validation, but I know same approach is used with other similar options. I guess we can leave it like this and then later change all if needed.

src/robot/result/autoexpandkeywordmatcher.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@py2to3
+class AutoExpandKeywordMatcher(object):
+
+    def __init__(self, auto_expand_list):
+        self._matched_ids = []
+        if auto_expand_list==None: 
+            auto_expand_list=[]
```
pekkaklarck: 1. Should use `is None`, not `== None`.
2. Just checking truth value (i.e. `if not auto_expand_list`) is probably better here.
3. Spaces missing around `==` and `=`.
4. `_list` postfix feels unnecessary for me.

doc/userguide/src/Appendices/CommandLineOptions.rst
```diff
@@ -56,6 +56,8 @@ Command line options for test execution
   --tagstatcombine <tags:title>  Creates `combined statistics based on tags`_.
   --tagdoc <pattern:doc>  Adds `documentation to the specified tags`_.
   --tagstatlink <pattern:link:title>  Adds `external links`_ to the *Statistics by Tag* table.
+  --autoexpandkeywords <name:pattern|tag:pattern>  automatically `expand keywords`_
+                          in the generated log file.
```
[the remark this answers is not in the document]
DirkRichter: agree. auto-prefix dropped

doc/userguide/src/ExecutingTestCases/OutputFiles.rst
```diff
@@ -552,6 +552,28 @@ deeply nested keyword structures.
 
 .. note:: `TAG:<pattern>` mode was added in Robot Framework 2.9.
 
+Automatically expanded keywords in log.html
```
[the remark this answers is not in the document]
DirkRichter: adjusted

doc/userguide/src/ExecutingTestCases/OutputFiles.rst
```diff
@@ -552,6 +552,28 @@ deeply nested keyword structures.
...
+-------------------------------------------
+
+The :option:`--autoexpandkeywords` option set keywords initially to visible in log.html.
```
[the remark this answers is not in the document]
DirkRichter: docu improved

doc/userguide/src/ExecutingTestCases/OutputFiles.rst
```diff
@@ -552,6 +552,28 @@ deeply nested keyword structures.
...
+
+Autoexpandkeywords supports the following modes:
```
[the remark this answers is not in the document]
DirkRichter: agree + improved

src/robot/htmldata/rebot/log.js
```diff
@@ -62,6 +62,16 @@ function expandElementWithId(elementid) {
     expandElement(window.testdata.findLoaded(elementid));
 }
 
+function expandElementsWithIds(ids) {
+    util.map(ids, expandElementWithId);
+}
+
+function loadAndExpandElementIds(ids) {
+    for (var i in ids) {
+      window.testdata.ensureLoaded(ids[i], expandElementsWithIds);
```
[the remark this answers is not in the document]
DirkRichter: fixed

src/robot/rebot.py
DirkRichter: fixed

src/robot/rebot.py
DirkRichter: fixed

src/robot/reporting/jsbuildingcontext.py
DirkRichter: fixed

src/robot/reporting/jsbuildingcontext.py
DirkRichter: refactoring done

src/robot/reporting/resultwriter.py
DirkRichter: parameter order was updated. but implementing a new class storing options would effect more JS-modules/classes. e.g. JsBuildingContext beside JsModelBuilder. This increases code complexity where avoiding 1 extra parameter should improve code readability. Thus i prefer to reject this JS-options-refactoring...

src/robot/reporting/jsmodelbuilders.py
DirkRichter: done

src/robot/result/autoexpandkeywordmatcher.py
DirkRichter: refactoring to public attribute done.

src/robot/result/autoexpandkeywordmatcher.py
DirkRichter: fixed

src/robot/reporting/jsbuildingcontext.py
DirkRichter: refactoring done: now _expand_matcher is set to None if functionality is not needed avoiding calls to .check methods

src/robot/result/autoexpandkeywordmatcher.py
DirkRichter: fixed: double racing of DataError reduced to one by moving validation_code inside of conf/settings.py

src/robot/result/autoexpandkeywordmatcher.py
DirkRichter: 1,2,3,4 fixed

doc/userguide/src/ExecutingTestCases/OutputFiles.rst
```diff
@@ -552,6 +552,32 @@ deeply nested keyword structures.
...
+   same as when `removing keywords`_ using `NAME:<pattern>` mode.
+
+`TAG:<pattern>`
+   Expand keywords with tags matching the given pattern. Pattern matching
+   rules are same as when `removing keywords`_ using `TAG:<pattern>` mode.
+
+Examples::
+
+   rebot --expandkeywords name:CapturePageScreenshot output.xml
+   rebot --expandkeywords tag:ShowInLog output.xml
```
pekkaklarck: This works both with Robot and Rebot, right? Changing the first example to `robot ... tests.robot` would probably be a good idea. Not sure should that fact be explicitly mentioned as well, probably not.

src/robot/conf/settings.py
```diff
@@ -295,6 +298,13 @@ def _validate_flatten_keywords(self, values):
         except DataError as err:
             raise DataError("Invalid value for option '--flattenkeywords'. %s" % err)
 
+    def _validate_expandkeywords(self, values):
+        for opt in values:
+            if not (opt.lower().startswith(('name:', 'tag:'))):
+                raise DataError("Invalid value for option '--expandkeywords'."
+                                " Expected 'TAG:<pattern>', or "
+                                "'NAME:<pattern>' but got '%s'." % opt)
```
pekkaklarck: Good idea to have validation here.

Nitpicking:
- There are unnecessary parentheses in the 'if not` line.
- When using implicit string literal concatenation, it would be good to be consistent where the space is. Now the middle line both starts and ends with a space. First and second ending with a space would be a bit more clear.

Yeah, that's nitpicking for real. :joy:

src/robot/conf/settings.py
```diff
@@ -607,3 +618,7 @@ def console_output_config(self):
     @property
     def process_empty_suite(self):
         return self['ProcessEmptySuite']
+
+    @property
+    def expand_keywords_args(self):
```
pekkaklarck: The `_args` postfix isn't needed. Others don't have anything like that either.

src/robot/rebot.py
```diff
@@ -201,6 +201,13 @@
                           match (starting from 1).
                           Examples: --tagstatlink mytag:http://my.domain:Title
                           --tagstatlink "bug-*:http://url/id=%1:Issue Tracker"
+    --expandkeywords name:<pattern>|tag:<pattern> *
+                          Matching keywords will be initially expanded in the 
+                          log file.
+                          For pattern matching rules see --removekeywords
+                          option.
+                          Example: --expandkeywords name:CapturePageScreenshot
+                          This is new in RF 3.2.
```
pekkaklarck: This is formatted quite a bit differently than other help texts. Starting the doc after the line with the option itself is fine with this long option, but doc should continue on same line after `log file.`.

src/robot/reporting/expandkeywordmatcher.py
```diff
@@ -0,0 +1,41 @@
...
+        self._namematcher = MultiMatcher(names) if len(names) > 0 else None
+        self._tagmatcher = MultiMatcher(tags) if len(tags) > 0 else None
+
+    def check(self, kw):
+        if (self._namematcher and self._namematcher.match(kw.kwname)) or \
+                (self._tagmatcher and self._tagmatcher.match_any(kw.tags)):
+            self.matched_ids.append(kw.id)
+
+    def active(self):
+        return bool(self._namematcher) or bool(self._tagmatcher)
```
pekkaklarck: Newline missing.

src/robot/reporting/jsbuildingcontext.py
```diff
@@ -36,6 +38,8 @@ def __init__(self, log_path=None, split_log=False, prune_input=False):
         self.split_results = []
         self.min_level = 'NONE'
         self._msg_links = {}
+        expand_matcher = ExpandKeywordMatcher(expand_args)
+        self._expand_matcher = expand_matcher if expand_matcher.active() else None
```
pekkaklarck: Do we need to create `ExpandKeywordMatcher` to check is it actually needed? Wouldn't something like this be enough:
```python
self._expand_matcher = ExpandKeywordMatcher(expand_keywords) \
        if expand_keywords else None
```
The above would remove the need to have the `active()` method. Also, `expand_args` should be renamed to something more explicit like `expand_keywords`.

src/robot/reporting/expandkeywordmatcher.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, expand_args):
+        self.matched_ids = []
+        if not expand_args:
+            expand_args = []
+        elif not is_list_like(expand_args):
+            expand_args = [expand_args]
+        names = [n[5:] for n in expand_args if n[:5].lower() == 'name:']
+        tags  = [p[4:] for p in expand_args if p[:4].lower() == 'tag:']
+        self._namematcher = MultiMatcher(names) if len(names) > 0 else None
+        self._tagmatcher = MultiMatcher(tags) if len(tags) > 0 else None
```
pekkaklarck: I doubt this performance optimization really affects anything. If `--expandkeywords` has been used, either `_namematcher` or `_tagmatcher` need to be used anyway. Always creating them would simplify the `check()` implementation. If this is left, `if len(names) > 0` should be replaced with `if names`.

src/robot/reporting/jsbuildingcontext.py
```diff
@@ -20,13 +20,15 @@
 from robot.utils import (attribute_escape, get_link_path, html_escape,
                          html_format, is_string, is_unicode, timestamp_to_secs,
                          unic)
+from .expandkeywordmatcher import ExpandKeywordMatcher
 
 from .stringcache import StringCache
```
pekkaklarck: The added import should be moved one row down.

src/robot/reporting/jsmodelbuilders.py
```diff
@@ -21,9 +21,9 @@
 
 class JsModelBuilder(object):
 
-    def __init__(self, log_path=None, split_log=False,
+    def __init__(self, log_path=None, split_log=False, expand_args=None,
```
pekkaklarck: `expand_args` -> `expand_keywords`

doc/userguide/src/Appendices/CommandLineOptions.rst
```diff
@@ -56,6 +56,8 @@ Command line options for test execution
   --tagstatcombine <tags:title>  Creates `combined statistics based on tags`_.
   --tagdoc <pattern:doc>  Adds `documentation to the specified tags`_.
   --tagstatlink <pattern:link:title>  Adds `external links`_ to the *Statistics by Tag* table.
+  --expandkeywords <name:pattern|tag:pattern>  automatically `expand keywords`_
+                          in the generated log file.
```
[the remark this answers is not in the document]
DirkRichter: done.

doc/userguide/src/ExecutingTestCases/OutputFiles.rst
DirkRichter: right. robot+rebot. it's a common option, thus we can set these settings via pabot

src/robot/conf/settings.py
DirkRichter: both fixed.

src/robot/rebot.py
DirkRichter: you a re right. this is wanted: i followed the style of removekeywords and flattenkeywords to be consistent inside of rebot.py. i raised a more general *.rst vs. *.py formatting issue to harmonize in more general => https://github.com/robotframework/robotframework/issues/3378

src/robot/reporting/expandkeywordmatcher.py
DirkRichter: fixed

src/robot/reporting/jsbuildingcontext.py
DirkRichter: correct. improved.

src/robot/reporting/expandkeywordmatcher.py
DirkRichter: improved code

src/robot/reporting/jsbuildingcontext.py
DirkRichter: fixed

src/robot/reporting/jsmodelbuilders.py
DirkRichter: fixed

src/robot/conf/settings.py
DirkRichter: fixed

atest/robot/rebot/expand_keywords.robot
```diff
@@ -0,0 +1,43 @@
...
+    fail  TODO: implement
+
+Expand by 3 Tags
+    fail  TODO: implement
+
+Expand by Name and Tag
+    fail  TODO: implement
+
+No expand
+    fail  TODO: implement
```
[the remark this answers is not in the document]
DirkRichter: removed as discussed

pekkaklarck:
Tested this manually and the basic functionality worked well. Noticed that matching by name uses just the keyword name like `Log`, not full name like `BuiltIn.Log` similarly as `--remove/flattenkeywords`. Matching the full name allows pinpointing what keyword exactly to expand and also allows usages like `BuiltIn.*` to expand all keywords from a certain library. This is trivial to change and I can do it after merging the PR. As discussed on Slack, I'll also add some tests for this feature.

Issue #3090, opened 2019-02-06
Type conversion fails in dry-run if value contains variable


AleksandrKE:
Hello.
I've got an issue while running tests in dry run mode after the Robot Framework update.
The problem connected with the automatic conversion of arguments if the method has annotations.
My test:
```
(venv) aleksandr@work-pc:~/my_project$ cat my_test.robot 
*** Settings ***
Library    library.py

*** Test Cases ***
My Test
    ${list}=    Create List    one    two    three

    library.Custom Keyword     ${list}
```
My method:
```
(venv) aleksandr@work-pc:~/my_project$ cat library.py 
# -*- coding: utf-8 -*-

def custom_keyword(lst: list):
    pass
```
I get the error message:
```
(venv) aleksandr@work-pc:~/my_project$ robot --dryrun my_test.robot 
==============================================================================
My Test                                                                       
==============================================================================
My Test                                                               | FAIL |
ValueError: Argument 'lst' got value '${list}' that cannot be converted to list: Invalid expression.
------------------------------------------------------------------------------
My Test                                                               | FAIL |
1 critical test, 0 passed, 1 failed
1 test total, 0 passed, 1 failed
==============================================================================
```
However, the test runs successfully if it runs without the '--dry run' option.
It looks like the robot tries to convert into the list the name of the variable, not value.

My environment:
```
(venv) aleksandr@work-pc:~/my_project$ robot --version
Robot Framework 3.1.1 (Python 3.6.7 on linux)
```

pekkaklarck:
Good catch. This is due to Robot nowadays trying to convert arguments automatically but variables not being resolved in dry-run. It would be best to resolve variables (we got #2322 about that), but that would be a huge task and not likely to happen in the near future. At the moment the best solution is not doing argument conversion in dry-run if the argument contains variable.

Alternatively we could disable automatic argument conversion in dry-run altogether, but what would remove a useful validation. Instead we should add tests for argument validation in dry-run (when not using variables).

A workaround for this problem is moving the logic into a user keyword and adding `robot:no-dry-run` to it.

AleksandrKE:
Thanks for the quick answer.
> At the moment the best solution is not doing argument conversion in dry-run if the argument contains variable.
I think this is a good solution.

pekkaklarck:
Thanks for PR #3126 @aleskarovadi! Hopefully you can it update based on my comments.

PR #3126, opened 2019-03-18
Fixing type conversion in dry-run mode (issue #3090)

aleskarovadi:
Not doing argument conversion in dry-run if the argument contains variable (#3090).


src/robot/running/arguments/argumentconverter.py
```diff
@@ -59,3 +65,8 @@ def _get_type(self, name, value):
         if name in self._argspec.defaults:
             return type(self._argspec.defaults[name]), False
         return None, None
+
+    def _need_conversion(self, value):
+        if self._dry_run and VARIABLE_PATTERN.search(value):
```
pekkaklarck: We actually have `robot.variables.contains_var` utility that could be used here instead. It might even make this `_need_conversion` helper unnecessary.

aleskarovadi: Thank you for pointing this out. I didn't know about `contains_var` utility. This is much better than using a custom regex.

aleskarovadi:
I have added some tests to check type conversion when agrument value contains a variable (for dryrun and regular test excetion). But I am not really sure that I did it correct. Please, tell me if there is something needs changing.

pekkaklarck:
Thanks for updates sorry for late comments from my side. I'm not concentrating on RF 3.1.2 with a goal to have it out in late April and will be more active.

Code looks very good. The only nitpickey comment I have is that the line where `contains_var` is used is longer than recommended by PEP-8 that we mostly follow.

There are some problems with tests:

- New type conversion test with variables is a good idea and test itself looks good too. Because it has been added only to `annotations.robot`, it won't be executed with Python 2, though. I'm not sure what's the best way to handle this, but perhaps it's easiest to add the same test also to `keyword_decorator.robot` that is run both with Python 2 and 3. Definitely no need to add it to all suites related to type conversion.

- We have other dry-run tests in [atest/robot/cli/dryrun](https://github.com/robotframework/robotframework/tree/master/atest/robot/cli/dryrun) and new ones should be added there as well. Probably should add a new suite or suites for type conversion there.

- I'm not sure what's the best way to go validate different type conversion tests in dry-run. Your approach to copy existing tests (`annotations.robot` -> `annotations_dryrun.robot`) certainly works but is pretty verbose and new tests possibly added to `annotations.robot` later won't be validated in dry-run. One possibility would be just running the suite and doing `Should Be Equal    ${SUITE.status}    PASS`, but that hides all the details. A benefit would be that there could be one test for annotations and another for keyword decorator in same suite. Could possibly add few explicit validations for certain tests like handling variables.

Coming up with a good way to test this is not easy. If you'd like to discuss that further, please add a new comment here or as on `#devel` channel on our Slack.

aleskarovadi:
Thanks for your review. I have split too long line and fixed problems with tests according to your comment:
- added the test for type conversion with variables to `keyword_decorator.robot`;
- created `atest/robot/cli/dryrun/type_conversion.robot` test suite to validate type conversion in dry-run. There I used the approach of running `annotations.robot` and `keyword_decorator.robot` suites and checking their statuses.

Becasue there are two tests in `keyword_decorator.robot`, that are expected to fail, I have created two test cases for keyword_decorator: the first runs the tests that should pass, the second runs these negative test cases.

pekkaklarck:
Thanks for the updates! I'm not sure is "Keyword Decorator FAIL tests" test actually needed but I'll merge this now and think about that afterwards.

pekkaklarck:
Noticed that "Keyword Decorator PASS tests" doesn't pass with Python 2 because it runs tests that work only with Python 3. I'll fix that. I'll also remove "Keyword Decorator FAIL tests" because I don't think it's really needed.

Issue #722, opened 2014-06-29
Add `initialize` listener method


spooning:
> _Originally submitted to [Google Code](http://code.google.com/p/robotframework/issues/detail?id=722) by bryan.oakley on 2 Dec 2010_

I would like to see a new method added to the listener interface. This method would be called exactly once prior to any tests being run. I propose the name "initialize".

It is useful for being able to do some initialization prior to the test running (think: telling a remote listening GUI to initialize the display, or a debugger to let let the user set breakpoints before a test runs).

I would like this method to pass as one of the parameters a list of all test suites and their test cases. It doesn't need to include the actual test case data per se, but a tree of names and longnames would be nice. The data needs to be in a format that can be easily parsed to create a hierarchical tree (nested tuples are the obvious choice here)

Again, imagine a test runner or debugger GUI that wants to be able to create a hierarchical tree of all tests, with an indicator beside them for not run yet, running, pass, fail, etc.

pekkaklarck:
Are you @boakley still interested in this? Want to provide a pull request?

pekkaklarck:
No interest related to this lately. The latest comment about actual implementation is actually from 2010. Closing.

PR #3133, opened 2019-03-26
Added settings object to running.model.TestSuite

DanielPBak:
I have a listener library which changes its behavior based on the settings passed to `robot`. Previously it has done this using `sys.argv`.

However, this makes the listener stop working if `robot` is being run programatically through `robot.run`, since settings are passed as arguments, rather than through the command-line.

This MR changes `robot.running.model` to keep the most recent `RobotSettings` used to run in a `settings` attribute, allowing listeners to access them through the `data` argument passed to `start_suite` and `end_suite`.

### DIFF
--- a/src/robot/running/model.py
+++ b/src/robot/running/model.py
@@ -111,7 +111,7 @@ class TestSuite(model.TestSuite):
 
     See the base class for documentation of attributes not documented here.
     """
-    __slots__ = ['resource']
+    __slots__ = ['resource', 'settings']
     test_class = TestCase    #: Internal usage only.
     keyword_class = Keyword  #: Internal usage only.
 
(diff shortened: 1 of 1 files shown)


pekkaklarck:
I can see this change would be useful for you, but there are various reasons I don't like it:

1. `RobotSettings` isn't considered part of the public API and is subject to change in the future.

2. Setting an attribute only when a certain method is called is somewhat suspicious. It should definitely be set already in `__init__`.

3. There are no tests and documentation isn't good enough for anyone else to be able to use this.

4. Settings are part of the execution context and don't really belong to the suite object. Unfortunately we don't currently have that good execution context. I hope this changes in RF 3.3 when we plan to add IF/ELSE support and need to look at that part of the code again.

5. Listeners can already now be configured by passing arguments to them when they are initialized. From the command line you can pass arguments like `Listener:arg1:arg2` and when using `robot.run` you ought to be able to construct the listener first and then pass the listener object to `robot.run`. There's obviously a little duplication if you need to use e.g. `--include foo --listener Listener:foo` compared to the listener automatically checking the value of `--include`, but I don't think that's too bad. If you have a script to start execution, such duplication doesn't matter at all.

In other words, I don't see real need for listeners to be able to inspect settings in the first place. If there really is a use case where that's needed, we need to think about other ways to accomplish that. One possibility is adding a method like `configure` or `initialize` to the listener API that would get settings along with possible other information. That has actually been requested already in 2010 (#722) but nobody has implemented it. If you are interested on that, I can re-open that issue and we can start discussing on the lower level design like what arguments the methods should actually accept.

DanielPBak:
I figured it was unlikely for this to get merged. 

I will look into that PR and your suggestions.

PR #3140, opened 2019-04-02
robottime: Honor the SOURCE_DATE_EPOCH environment variable.

Apteryks:
Honoring the SOURCE_DATE_EPOCH environment variable allows building
the documentation using libdoc reproducibly, by setting the generated
timestamp to a fixed value.

For more background on reproducible builds and the SOURCE_DATE_EPOCH
environment variable, see:
https://reproducible-builds.org/specs/source-date-epoch/.

* src/robot/utils/robottime.py: import `os'.
(TimestampCache._get_epoch): Retrieve date from SOURCE_DATE_EPOCH if
it is defined, otherwise from time.time().
* utest/output/test_logger.py (TestLogger.test_write_to_one_logger):
Check for the existence of a timestamp attribute instead of checking
for its content as the later is easy to break when using the
SOURCE_DATE_EPOCH environment variable.

### DIFF
--- a/src/robot/utils/robottime.py
+++ b/src/robot/utils/robottime.py
@@ -14,6 +14,7 @@
 #  limitations under the License.
 
 import datetime
+import os
 import time
 import re
 
@@ -395,6 +396,8 @@ class TimestampCache(object):
 
     # Seam for mocking
     def _get_epoch(self):
+        if os.getenv('SOURCE_DATE_EPOCH'):
+            return float(os.getenv('SOURCE_DATE_EPOCH'))
         return time.time()
 
     def _use_cache(self, secs, *separators):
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Sounds reasonable. Could you please submit a separate issue about this as well? We use them as our project backlog and also generate release notes based on them.

The implementation itself is simple and ought to be safe. There are some things that should be done still:

1. A dedicated test for using SOURCE_DATE_EPOCH. Unit test ought to be enough. This is important because its likely that time related code is going to be heavily refactored in the future and without tests this feature may be accidentally removed.

2. I'd prefer the updated test to be changed so that it check that the timestamp is in right format, not only that the timestamp exists. This could be done easily with regexps.

3. I think this should be mentioned somewhere in the User Guide. Not sure what's the best place, though.

Apteryks:
Hi @pekkaklarck; sorry for dropping the ball on this.  I've now opened the requested [issue](https://github.com/robotframework/robotframework/issues/4262), rebased my branch and will now try to address the 3 items you mentioned.

Apteryks:
Closing in favor of https://github.com/robotframework/robotframework/pull/4263.

Issue #3151, opened 2019-04-10
unescape method destroyed the json data format


toddlerya:
1. First we determine that such json is correct
`{"key1": 123, "trouble": "{\"inner\": \"hehe\"}"}`
This can be referred to the [json  official document](http://www.json.org/index.html).
Also we can use python check it:
<img width="711" alt="屏幕快照 2019-04-10 22 27 08" src="https://user-images.githubusercontent.com/5418150/55887573-50ecaa00-5be0-11e9-9e17-1acca6cf1a3e.png">

2. We have such a requirement to write test cases:
```
*** Settings ***

*** Test Cases ***
json_test_demo
    ${payload}    Set Variable    {"key1": 123, "trouble": "{\"inner\": \"hehe\"}"}
```

3. But the results of the operation did not meet our expectations：
<img width="648" alt="屏幕快照 2019-04-10 22 28 53" src="https://user-images.githubusercontent.com/5418150/55887489-31558180-5be0-11e9-8b86-f0cd99d51978.png">

4. After a long debug process, I found the problem here. Open `robot/utils/escaping.py`, we can find `return ''.join(self._yield_unescaped(string))` in line 51. This step remove `\`!
Debug Code detail:
<img width="1290" alt="屏幕快照 2019-04-10 22 32 59" src="https://user-images.githubusercontent.com/5418150/55888086-5ac2dd00-5be1-11e9-9a3a-0b7bb97c3326.png">
Robot run output info:
<img width="1119" alt="屏幕快照 2019-04-10 22 40 21" src="https://user-images.githubusercontent.com/5418150/55888260-a7a6b380-5be1-11e9-9f23-770734973076.png">

5. Debug Env:
- Mac os 10.14.4
- Python 3.7.1
- robotframework==3.1.1

6. Repair suggestions:
First judge whether it is a json string, not a json string and then perform normal processing.

pekkaklarck:
Backslash is an escape character in Robot Framework data. You need to double it like `\\` to get a literal backslash. We cannot safely start making exceptions to this rule.

PR #3152, opened 2019-04-10
bug fix [issues#3151] unescape method destroyed the json data format

toddlerya:
https://github.com/robotframework/robotframework/issues/3151


pekkaklarck:
Current behavior is by design and won't be changed.

PR #3177, opened 2019-05-08
Faulty implementation of interval_time in keyword "Wait Until Keyword Succeeds"

Schwaneberg:
The keyword *wait_until_keyword_succeeds* simply adds the interval time after the keyword has failed. This causes unexpected behaviour, if the desired interval time is short in comparison to the keyword runtime.

Therefore, I modified the keyword to measure the time between keyword execution and failure. The difference is then subtracted from the wait time to achieve the desired interval time (keyword runtime + wait time = interval time). 
Furthermore, a warning will be logged whenever the desired interval time is shorter than the actual keyword runtime.

Signed-off-by: Oliver Schwaneberg <oliver.schwaneberg@deuta.de>


pekkaklarck:
As I commented #3209, this would be a backwards incompatible change and it wouldn't even match the documentation of the keyword:

> ``retry_interval`` is the time to wait before trying to run the keyword again after the previous run has failed.

How to interpret the timeout could be made configurable, though.

The change would also need tests and documentation changes.

pekkaklarck:
As I commented #3209, this would be a backwards incompatible change and it wouldn't even match the documentation of the keyword:

> ``retry_interval`` is the time to wait before trying to run the keyword again after the previous run has failed.

How to interpret the timeout could be made configurable, though.

The change would also need tests and documentation changes.

pekkaklarck:
As I commented #3209, this would be a backwards incompatible change and it wouldn't even match the documentation of the keyword:

> ``retry_interval`` is the time to wait before trying to run the keyword again after the previous run has failed.

How to interpret the timeout could be made configurable, though.

The change would also need tests and documentation changes.

pekkaklarck:
Don't have time to look at the changes in detail, but on a very quick look I saw two issues that need to be fixed:

- Tests. This functionality definitely needs them. See `atest/README.rst` for details how to create them.
- A note somewhere that this feature is new in RF 3.2.

Schwaneberg:
I added two acceptance tests in `wait_until_keyword_succeeds.robot` and created a basic release note file for RF 3.2a1 to document the new feature. Please let me know if you want it at another place or with a longer description.

pekkaklarck:
I'm not sure why this PR wasn't re-reviewed two years ago. May have been summer holiday season and then the PR and the related issue #3209 just got buried under other PRs/issues in this tracker. Anyway, the PR looks pretty good but there are few things to change in the code and also the conflicts above need to be sorted out. If you @Schwaneberg are still interested in this PR and the issue, I can do a more thorough review. We can either get this into RF 4.1 in July or to RF 5.0 planned for late 2021 or early 2022. If there's no interest anymore, I assume both the issue and the PR can be closed.

src/robot/libraries/BuiltIn.py
```diff
@@ -2220,8 +2224,17 @@ def wait_until_keyword_succeeds(self, retry, retry_interval, name, *args):
...
+        if match:
+            strict_interval = True
+            retry_interval = retry_interval.lstrip(match.group())
+        else:
+            strict_interval = False
+
+        retry_interval = float(timestr_to_secs(retry_interval))
+        wait_time = retry_interval
         while True:
+            execution_time = time.time()
```
pekkaklarck, line 2237: `start_time` would be a better name than `execution_time`. The latter sounds like it would be the time execution took. This is especially confusing when there's separate `keyword_runtime` variable that is the actual execution time.

`wait_time` could perhaps be renamed to `sleep_time` because that's the exact time used for sleeping but also the former is fine.

src/robot/libraries/BuiltIn.py
```diff
@@ -2220,8 +2224,17 @@ def wait_until_keyword_succeeds(self, retry, retry_interval, name, *args):
...
             message = '%d time%s' % (count, s(count))
-        retry_interval = timestr_to_secs(retry_interval)
+        match = re.search(r'(strict *: *)', retry_interval)
+        if match:
+            strict_interval = True
+            retry_interval = retry_interval.lstrip(match.group())
+        else:
+            strict_interval = False
+
+        retry_interval = float(timestr_to_secs(retry_interval))
```
pekkaklarck, line 2234: Some problems here:

1. `retry_interval` can be given also as an integer or float in which case `re.search` will fail. We clearly don't have any tests for such behavior but I'm absolutely sure someone has used that functionality.
2. Using regexps is an overkill when we just to check does the string start with something and `re.search` is actually wrong because it can find the match anywhere, not only from the beginning.
3. Would be good to support `strict` case-insensitively like we support most similar markers.
4. Unnecessary empty line. We don't generally have them inside methods and for consistency reasons shouldn't use here either.
5. `time_str_to_secs` returns a floating point number so `float()` isn't needed.

This (untested) code ought to handle all above problems:
```python
if is_string(retry_interval) and retry_interval.lower().startswith('strict:'):
    retry_interval = retry_interval.split(':', 1).strip()
    strict_interval = True
else:
    strict_interval = False
retry_interval = timestr_to_secs(retry_interval)
```

src/robot/libraries/BuiltIn.py
```diff
@@ -2232,7 +2245,15 @@ def wait_until_keyword_succeeds(self, retry, retry_interval, name, *args):
...
-                self._sleep_in_parts(retry_interval)
+                keyword_runtime = time.time() - execution_time
+                if strict_interval:
+                    wait_time = retry_interval - keyword_runtime
+                if wait_time >= 0.0:
+                    self._sleep_in_parts(wait_time)
+                else:
+                    logger.warn("Interval violation: retry_interval is {:.3}s"
+                                ", but keyword runtime is {:.3}s."
+                                .format(retry_interval, keyword_runtime))
```
pekkaklarck, line 2256: This warning is logged only when using the strict mode and keyword execution took more time than the retry interval, right? Few things to be enhanced related to that:

1. The logic would be easier to understand if the code would be structured something like this:
   ```python
   if strict_interval:
       wait_time = retry_interval - keyword_runtime
       if wait_time < 0:
           logger.warn(...)
       self._sleep_in_parts(wait_time)
   ```

2. Actual times could be logged in more human readable format by using `secs_to_timestr` utility method. I'm not sure is the actual keyword run time interesting, because you can see that easily from the log file also otherwise. Something more generic like this could be enough:
   ```python
   logger.warn("Interval violation: Keyword execution time was longer than "
               "retry interval %s.' % secs_to_timestr(retry_interval))

3. There needs to be a test validating the warning. Log message validation can be done on the `atest/robot` side using `Check Log Message` like some other tests in the same suite are doing. If you want to log the actual execution time, you need to use patterns in the expected log message because that time cannot be controlled exactly.

src/robot/libraries/BuiltIn.py
```diff
@@ -2199,6 +2202,7 @@ def wait_until_keyword_succeeds(self, retry, retry_interval, name, *args):
         Examples:
         | Wait Until Keyword Succeeds | 2 min | 5 sec | My keyword | argument |
         | ${result} = | Wait Until Keyword Succeeds | 3x | 200ms | My keyword |
+        | ${result} = | Wait Until Keyword Succeeds | 3x | strict: 200ms | My keyword |
```
pekkaklarck, line 2205: Good to add an example!

There should be a note like this at the end of the docstring:
> Support for "strict" retry interval is new in Robot Framework 4.1.

pekkaklarck (changes requested):
Looks good but there are few things to be enhanced in code, docs and tests. See individual comments for details.

Schwaneberg:
Thank you for your review and all suggestions. I just pushed a new commit that addresses all topics you mentioned.

src/robot/libraries/BuiltIn.py
```diff
@@ -2220,8 +2226,14 @@ def wait_until_keyword_succeeds(self, retry, retry_interval, name, *args):
             if count <= 0:
                 raise ValueError('Retry count %d is not positive.' % count)
             message = '%d time%s' % (count, s(count))
+        strict_interval = isinstance(retry_interval, str) \
+            and retry_interval.replace(' ', '').lower().startswith('strict:')
```
pekkaklarck, line 2230: Need to use our `is_string` helper here due to us still for a little time supporting Python 2 where `str` is actually byes.

src/robot/libraries/BuiltIn.py
```diff
@@ -2232,7 +2244,16 @@ def wait_until_keyword_succeeds(self, retry, retry_interval, name, *args):
                     raise AssertionError("Keyword '%s' failed after retrying "
                                          "%s. The last error was: %s"
                                          % (name, message, err))
-                self._sleep_in_parts(retry_interval)
+                keyword_runtime = time.time() - start_time
+                if strict_interval:
+                    sleep_time = retry_interval - keyword_runtime
+                if sleep_time < 0:
```
pekkaklarck, line 2250: It would be a bit more clear that this `if` can be true only if the above `if` is true, if it would be inside the above `if`.

src/robot/libraries/BuiltIn.py
```diff
@@ -2232,7 +2244,16 @@ def wait_until_keyword_succeeds(self, retry, retry_interval, name, *args):
...
+                    logger.warn("Interval violation: retry_interval is {}"
+                                ", but keyword runtime is {}."
+                                .format(secs_to_timestr(retry_interval),
+                                        secs_to_timestr(keyword_runtime)))
```
pekkaklarck, line 2254: This warning still isn't tested.

We also have used the old style `%` formatting elsewhere and for consistency it would be better here as well. Looking forward to the (near) future when we can just use f-strings!

pekkaklarck (approved):
Few things to be enhanced still. I can do that myself.

atest/testdata/standard_libraries/builtin/wait_until_keyword_succeeds.robot
```diff
@@ -130,6 +130,16 @@ Fail With Nonexisting Variable Inside Wait Until Keyword Succeeds
...
     Wait Until Keyword Succeeds    3 times    0s    Access Initially Nonexisting Variable
 
+Pass With Strict Timing
+    [Documentation]    Retry at fixed rate takes at least 20ms to execute and requires exact timing
+    Wait Until Keyword Succeeds    2 times    strict: 50ms    Passes Every Second Time At 50ms Interval
+
+Fail Without Strict Timing
+    [Documentation]    FAIL
+    ...    Keyword 'Passes Every Second Time At 50ms Interval' failed after retrying 2 times. \
+    ...    The last error was: TimeoutError: interval violated
```
pekkaklarck, line 140: This error is wrong, there's not "TimeoutError" in it anymore.

pekkaklarck:
Thanks for the fixes and the PR in general! There are still few things to fix/enhance but I can handle them myself. This PR has been open for too long time already...

Issue #2767, opened 2018-02-06
Syslog, Libdoc, Testdoc and Tidy don't create directory for outputs


bkhouri:
# Summary
If the `ROBOT_SYSLOG_FILE` environment variable is set to a directory and filename, `robot` will fails to create the System log file if the directory does not exist.

# Environment:

- Robot Framework version 3.0.2
- Python version 2.7.10
- OS: Mac OS Sierra version 10.12.6 

# Reproducible
100%

# Steps to Reproduce:
1. Create the following dummy test case named `Foo.robot` 
    ```robot
    *** Test Cases ***
    Foo Test Case
        Log     just logging
    ```
1. Run the following command `ROBOT_SYSLOG_FILE=./garbage/syslog.log robot  Foo.robot`
    ```
    $ ROBOT_SYSLOG_FILE=./garbage/syslog.log robot  Foo.robot
    [ ERROR ] Opening syslog file './garbage/syslog.log' failed: No such file or directory
    [ ERROR ] Opening syslog file './garbage/syslog.log' failed: No such file or directory
    ==============================================================================
    Foo                                                                           
    ==============================================================================
    Foo Test Case                                                         | PASS |
    ------------------------------------------------------------------------------
    Foo                                                                   | PASS |
    1 critical test, 1 passed, 0 failed
    1 test total, 1 passed, 0 failed
    ==============================================================================
    Output:  /Users/******/output.xml
    Log:     /Users/******/log.html
    Report:  /Users/******/report.html
    ```
 
Robot framework should create the directory before attempting to get a filehandler on the SYSLOG file.

pekkaklarck:
I guess this would make sense, especially when we already do it with the normal output files. Interested to provide a PR? It would still have a change to make it into RF 3.0.3 planned to be released this week.

rasjani:
Similar issue also happens in libdoc when trying to create the keyword documentation into a directory that does not exist here: https://github.com/robotframework/robotframework/blob/master/src/robot/utils/robotio.py#L23

pekkaklarck:
OK. I'll add this to RF 3.1.2 scope and we can get both of these fixed.

pekkaklarck:
This is a pretty trivial fix but RF 3.1.2 is also late. Postponing to 3.2.

pekkaklarck:
This was fixed by PR #3186 by @mihaiparvu.

pekkaklarck:
Also Tidy and Testdoc were affected by this issue.

PR #3186, opened 2019-05-14
Ensure creation of missing directories for ROBOT_SYSLOG_FILE and Libdoc outputs

mihaiparvu:
This PR contains a fix for #2767. The acceptance test is only for Libdoc, as I  couldn't find any tests regarding `ROBOT_SYSLOG_FILE`. Ought to be fine though, since the same code is used for the two scenarios.

### DIFF
--- a/src/robot/utils/robotio.py
+++ b/src/robot/utils/robotio.py
@@ -14,12 +14,14 @@
 #  limitations under the License.
 
 import io
+import os.path
 
 from .platform import PY3
 
 
 def file_writer(path=None, encoding='UTF-8', newline=None):
     if path:
+        _ensure_destination_directory_exists(path)
         f = io.open(path, 'w', encoding=encoding, newline=newline)
     else:
         f = io.StringIO(newline=newline)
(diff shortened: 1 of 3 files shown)


pekkaklarck:
Thanks! Is this important for you to have this in RF 3.1.2 ot could it wait until 3.2? I just got all tests run for 3.1.2 and wouldn't like to do even simple code changes unless really needed.

mihaiparvu:
Not in a hurry to have this in RF 3.1.2, it can wait until the next release.

pekkaklarck:
Thanks @mihaiparvu!

src/robot/utils/robotio.py
```diff
@@ -41,3 +43,17 @@ def binary_file_writer(path=None):
     getvalue = f.getvalue
     f.getvalue = lambda encoding='UTF-8': getvalue().decode(encoding)
     return f
+
+
+def _ensure_destination_directory_exists(path):
+    if not os.path.exists(path):
+        base, ext = path.rsplit(os.sep, 1)
```
pekkaklarck: This line actually has a bug. If the path is given as just a file name like `syslog.txt`, the above line fails with `ValueError: need more than 1 value to unpack`. Better to use `os.path.dirname(path)` and then also check `if dirname and not os.path.exists(dirname)`. I'll fix this myself while doing some other cleanup to creating destination directories.

Issue #3167, opened 2019-04-29
Set ROBOT_LIBRARY_SCOPE dynamically


manosnoam:
Is there a way to define ROBOT_LIBRARY_SCOPE = 'GLOBAL' when creating TestCase dynamically ?
Currently I have to define ROBOT_LIBRARY_SCOPE in the class of my library, but I'd like to have it defined during the creation of the test case.

For example, I tried this:
```
suite.resource.imports.library('MyLib.py')
# Add to Library a class property:
suite.resource.imports[-1].__class__.ROBOT_LIBRARY_SCOPE = 'GLOBAL'
setattr(suite.resource.imports[-1].__class__, 'ROBOT_LIBRARY_SCOPE', 'GLOBAL')

testCase = suite.tests.create('MyTest') 
# Add to the TestCase a class property:
setattr(testCase.__class__, 'ROBOT_LIBRARY_SCOPE', 'GLOBAL')
testCase.__class__.ROBOT_LIBRARY_SCOPE = 'GLOBAL'
```

But when running this testCase, it ignores these class properties, and it has no effect on ROBOT_LIBRARY_SCOPE.

Thanks!

pekkaklarck:
This isn't possible. Could you explain your use case?

manosnoam:
When dynamically creating Test suite with an external python script, It usually required to import some libraries too, and it is done in that external script. 
So it would be useful to set ROBOT_LIBRARY_SCOPE = 'GLOBAL', outside the library class definition, and avoid the need to add scope in every library, that might be imported later.
That's why it would be a nice option to set the library scope, when importing a library dynamically.

pekkaklarck:
Could you please clarify? Also, are you interested to implement this functionality if we agree it would be a good idea?

manosnoam:
I think to implement it, we'll need to set for example
lib._libcode.ROBOT_LIBRARY_SCOPE = 'GLOBAL'

inside _import_library of namespace.py.

But it is only possible during runtime, not during suite.resource.imports.library.

How can we import namespace, and use import_library() method, before suite.run() ?

manosnoam:
I figure how to implement it, and I can add a new pull request. Which branch to clone, and how do I add my commit (git push) ?

manosnoam:
Hi @pekkaklarck, did you see my PR https://github.com/robotframework/robotframework/pull/3189
?

pekkaklarck:
[...]

PR #3189, opened 2019-05-21
Option to set ROBOT_LIBRARY_SCOPE dynamically (feature request #3167)

manosnoam:
Allowing to set ROBOT_LIBRARY_SCOPE when importing custom libraries in the API.
It is an alternative way, to the hard-coded definition of ROBOT_LIBRARY_SCOPE,
in your custom library class.

For example, you'll be able to call:
suite.resource.imports.library('MyLib.py', scope='TEST SUITE')

Instead of setting ROBOT_LIBRARY_SCOPE in MyLib class:
ROBOT_LIBRARY_SCOPE = 'TEST SUITE'

### DIFF
--- a/src/robot/model/imports.py
+++ b/src/robot/model/imports.py
@@ -23,7 +23,7 @@ from .itemlist import ItemList
 class Import(object):
     ALLOWED_TYPES = ('Library', 'Resource', 'Variables')
 
-    def __init__(self, type, name, args=(), alias=None, source=None):
+    def __init__(self, type, name, args=(), alias=None, source=None, scope=None):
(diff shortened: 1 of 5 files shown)


pekkaklarck:
I haven't looked at the code that closely, but this PR lacks tests and documentation. Much more importantly, I'm not convinced is this a useful functionality to begin with. We can continue discussing about that in #3167.

pekkaklarck:
As I [commented the related issue](https://github.com/robotframework/robotframework/issues/3167#issuecomment-505813759), I don't consider this a useful feature.

Issue #3091, opened 2019-02-08
Add `Set Local Variable` keyword


mihaiparvu:
Below is a request from library developers in my organization that would benefit from the addition of a new keyword to set a local scope variable.

I will act as a proxy on their behalf and here is some context from their side:

------------------------------------------------------------------------------------------------------------
Currently there are BuiltIn keywords for setting variables for the scopes `global`, `suite` and `test`:

``` robotframework
Set Global Variable  ${global_var}  global_value
Set Suite Variable  ${suite_var}  suite_value
Set Test Variable  ${test_var}  test_value
```

but no keyword:
``` robotframework
Set Local Variable  ${local_var}  local_value
```
for the scope `local`.

Instead this case is covered by a different syntax:
``` robotframework
${local_var}=  Set Variable  local_value
``` 
A consequence of this is, that the `local` scope case cannot be covered by `Run Keyword If`, but requires a separate keyword `Set Variable If`, which has a different syntax (see `ELSE IF`/`ELSE`).

Furthermore, as a library keyword developer, I want to be able to get/set Robot variables of all scopes from Python. Currently this can be achieved by calling:
```python
BuiltIn().get_variable_value()
BuiltIn().set_global_variable()
BuiltIn().set_suite_variable()
BuiltIn().set_test_variable()
BuiltIn()._variables["${local_var}"] = obj
```
The last call for setting a local variable requires to use a protected attribute of `BuiltIn()`, which as such is not part of a stable public API. If there would be a keyword `Set Local Variable` in BuiltIn, then it would be possible to set a local Robot variable from a library keyword by calling `BuiltIn().set_local_variable()` in Python.

There might be different opinions about whether it is a good idea to call Robot keywords from Python or whether a keyword should have a side effect on a Robot variable, but in practice we have such cases.

---------------------------------------------------------------------------------------------------------------
Our team will happily provide a pull-request for this keyword if it is decided to be included in a future release.

pekkaklarck:
My initial reaction was that on test data it's cleaner to only have assignment syntax for this purpose and for libraries we should have some better API for variable handling. On the other hand, `Set Local Variable` would be consistent with existing `Set Test/Suite/Global Variable` and adding this single keyword is a lot easier than creating a new API for variables. An added benefit of this keyword would be that it ought to support creating variable using a variable in the name like

    Set Local Variable    ${Field_Number_${i}}    value

[...]

PR #3191, opened 2019-05-22
Set Local Variable Keyword

cristii006:
PR for #3091 .


src/robot/libraries/BuiltIn.py
```diff
@@ -1428,6 +1428,21 @@ def set_variable(self, *values):
...
+    @run_keyword_variant(resolve=0)
+    def set_local_variable(self, name, *values):
+        """ Makes a variable available everywhere within the local scope.
+
+        Variables set with this keyword are available within the
+        local scope of the currently executed test case or in the local scope
+        of the keyword in which they are defined. For example, if you set a
+        variable in a user keyword, it is available only in that keyword. Other
+        test cases or keywords will not see variables set with this keyword.
+        """
```
pekkaklarck: 1. It would be good to mention, probably with an example, that this keyword is equivalent to a normal variable assignment based on a keyword return value.
2. Related to the above, it would be good to mention that the use case is to be able to use this dynamically with `Run Keyword If` and by external test libraries that want to set local variables. Were there some other important use case?
3. A note that this is new in RF 3.2 is a must. See other keywords for examples.
4. There's an unnecessary space after the opening `"""`.

src/robot/libraries/BuiltIn.py
```diff
@@ -1428,6 +1428,21 @@ def set_variable(self, *values):
...
+        name = self._get_var_name(name)
+        value = self._get_var_value(name, values)
+        self._variables.set_local_variable(name, value)
+        self._log_set_variable(name, value)
```
pekkaklarck: At least one of the acceptance tests should check that the message is logged correctly. That can be done on the `atest/robot` side by using the `Check Log Message` keyword.

atest/testdata/standard_libraries/builtin/setting_variables/variables.robot
```diff
@@ -33,6 +33,66 @@ Set Variable With More Or Less Than One Value
...
+    Should Be True    @{list} == ['1','2','3']
+    Set Local Variable    @{new}    This    is    ok
+    Should Be True    @{new} == ['This','is','ok']
+
+Set Local Variable - Dicts
+    Should Be True    &{DICT} == {'key': 'value', 'foo': 'bar'}
+    Set Local Variable    \&{DICT}    hello=world
+    Should Be True    &{DICT} == {'hello': 'world'}
+    Should Be Equal    ${DICT.hello}    world
+    Set Local Variable    &{dict}    a=${1}    ${2}=b
```
miktuy: maybe ```&{DICT}``` here?

Issue #2962, opened 2018-09-18
Support variables in test case names


kanchi240:
Here is the test case, which name contains the variable `${amount}`,

*** Variables ***
${amount}         ${5000000}

*** Test Cases ***
amount cannot larger than ${amount}
    no operation

and below is the log, test case name still contains the `${amount}`, could test case name display to `amount cannot larger than 5000000` in the log?
![15371782599106](https://user-images.githubusercontent.com/4908696/45658478-a3ba2380-bb21-11e8-9863-d3efdd6bbccb.png)

pekkaklarck:
Variables in test case names being resolved, when they exist, sounds good to me, but I don't think that really applies to suite names. Anyone interested to implement this? Shouldn't be too hard but I don't habe time for that. Can onviously help and review pull requests.

pekkaklarck:
The proposed change is to some extend backwards incompatible. Not that badly that it couldn't be implemented, but badly enough that such a change can only be done in a major release. Now before RF 3.1 would be a good time.

pekkaklarck:
PR #3193 looks very good but I still need to think is this a functionality we really want. I added this to RF 3.2 scope to make sure this issue isn't forgotten.

michalanglart:
Hi,

I've just seen this issue and while wondering how would it influence RED editor two questions came to my mind:

- how would it influence the way `--test` selector works?
- what will be the test name reported by `start_test`/`end_test` events of listener API?

pekkaklarck:
Good questions @michalanglart! I think it should go like this:

- `--test` is used before starting execution and variables cannot be resolved at that time. It thus could only work with the original name that contains the variable.
- With the listener API v3 we only send both data object and result object to `start_test` and `end_test`. We in general don't want to modify the data object at execution time, so it would have the original name that contains the variable. The result object should then have the final name as that's the one we wanted to show in reports and logs.
- With the listener API v2 we only send `name`. I don't actually know should it be the original or the final name. Does anyone else have any preferences?

Notice also that I still haven't made a decision is this a good feature to add in general or not. This isn't too high priority issue anyway, and want to have more important RF 3.2 features ready before thinking about this more. If others feel strongly about this either way, please vote the original description with :+1:/:-1: or add a separate comment.

pekkaklarck:
[...]

PR #3193, opened 2019-05-23
 Support variable in test case name (#2962) 

yamatoRT:
Hello,
this PR fix issue #2962.

I added new test file and I changed the name for a lot of test.
But i am not sure my change are fine for this file atest/testdata/output/names_needing_escaping.robot, I feel that I am not understand the goal of test.

Regards,

Richard


pekkaklarck:
On a quick look this PR looks very good. Still need to make the final decision about the actual functionality. I add #2962 into RF 3.2 scope to not forget about it.

pekkaklarck:
@michalanglart pointed out in the [comments of the original issue](https://github.com/robotframework/robotframework/issues/2962#issuecomment-551978253) that we need to decide how to handle the name passed to `start/end_test` listener methods. This is the current thinking:

1. With listener v3, the original test data object should preserve the original name but the result object should have the name so that variables are replaced. This is a must because we in general don't want to change the test data object during execution, but the result object should reflect the actual results.

2. With listener v2 we could either use the original or the resolved name, but we  use the original name. I don't have much preferences, but @michalanglart told using the original would be easier for them. If nobody objects, let's do that. I guess we could add a new item to the `attributes` dict that shows the name with variables resolved, but I doubt it is worth the effort. Let's add it if someone actually needs it.

I'm fairly certain the current code already works so that the listener v3 behaves as designed above and there's a good change that also listener v2 behaves like designed. Explicit tests should be added for both of these usages, though. In practice `atest/robot/output/support_variables_in_test_case_name.robot` should be enhanced to run tests on the `atest/testdata` side so that v2 and v3 listeners are enabled. Listeners can either directly verify test case names or write names stderr (or some file) where `atest/robot` side can validate them. I can help with this if needed.

yamatoRT:
I add test for listener v3 and v2. I don't succeed to keep unresolved name for the listener v2 for the name argument. I think it is possible but i need search more. o you want I continue to search?
I append documentation is release note, i hope it is enough, i don't know what to say.

pekkaklarck:
I'm on mobile so only quick comments:

- Issue v/ listener v2: see where this is used with tests and try setting priority: https://github.com/robotframework/robotframework/blob/master/src/robot/running/runner.py#L204

- No need to update release notes.

yamatoRT:
I removed the documentation in release note.
The names of listener v2 is not resolved now and i add in attribute entry 'resolved_name'.
I updated utest, atest and documentation of listener attributes.

src/robot/output/listenerarguments.py
```diff
@@ -72,7 +72,8 @@ def _get_version2_arguments(self, item):
         attributes = dict((name, self._get_attribute_value(item, name))
                           for name in self._attribute_names)
         attributes.update(self._get_extra_attributes(item))
-        return item.name, attributes
+        name = getattr(item, 'unresolved_name', None) or item.name
```
pekkaklarck: The real name should be passed here instead of conditionally checking what name to use.

Also, if I understand the logic correctly, `name` and `unresolved_name` end up being exactly the same with tests. That wouldn't be useful at all. Having `name` with the original name and separate `resolved_name` would potentially be useful, but I think `resolved_name` can be added later if needed.

src/robot/running/runner.py
```diff
@@ -112,13 +112,13 @@ def visit_test(self, test):
...
         self._executed_tests[test.name] = True
-        result = self._suite.tests.create(name=test.name,
+        result = self._suite.tests.create(name=self._resolve_setting(test.name),
                                           doc=self._resolve_setting(test.doc),
                                           tags=self._resolve_setting(test.tags),
                                           starttime=get_timestamp(),
                                           timeout=self._get_timeout(test))
         self._context.start_test(result)
-        self._output.start_test(ModelCombiner(test, result))
+        self._output.start_test(ModelCombiner(test, result, unresolved_name=test.name))
```
pekkaklarck: As I commented above, I don't think `unresolved_name` is a useful attribute to add to the listener interface. `resolved_name` could be, but I think it can be added later if really needed.

The way to give the original name to listener, without needing the `unresolved_name` later, is just using `name=test.name` here.

doc/userguide/src/CreatingTestData/CreatingTestCases.rst
```diff
@@ -523,7 +523,18 @@ test suite should have unique names.  Pertaining to this, you can also
...
+in result logs.
+
+.. sourcecode:: robotframework
+
+    *** Variables ***
+    ${amount}    ${5000000}
+
+    *** Test Cases ***
+    amount cannot larger than ${amount}
+        no operation
```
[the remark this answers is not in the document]
pekkaklarck: If needed, I can do these changes myself as well.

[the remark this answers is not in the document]
yamatoRT: i prefer you modify doc if you can because i haven't a lot of time this week.

src/robot/output/listenerarguments.py
yamatoRT: initially, i haven't this logic. The argument `name`  of listener v2 and the `result.name` of listener v3 had the resolved test cases name.

But we want the argument `name`  of listener v2 have the unresolved test cases name and the `result.name` of listener v3 had the resolved test cases name and i would like keep the same logic.

A other solution is let the name of the result unresolved, add attribute `resolved_name` on model of test case result and just change in `robot/output/xmllogger.py` the printed value by `start_test`.

pekkaklarck: Do you mean using `ModelCombiner(test, result, name=test.name)` like I proposed below breaks listener v3 or doesn't work otherwise?

yamatoRT: Your proposition works, it is so simple. I will apply change tomorrow.

doc/userguide/src/ExtendingRobotFramework/ListenerInterface.rst
```diff
@@ -166,6 +167,7 @@ it. If that is needed, `listener version 3`_ can be used instead.
    |                  |                  |                                                                |
    |                  |                  | * `id`: Same as in `start_test`.                               |
    |                  |                  | * `longname`: Same as in `start_test`.                         |
+   |                  |                  | * `resolved_name`: Test name with resolved variables           |
```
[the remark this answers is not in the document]
yamatoRT: it is remove

doc/userguide/src/ExtendingRobotFramework/ListenerInterface.rst
```diff
@@ -152,6 +152,7 @@ it. If that is needed, `listener version 3`_ can be used instead.
    |                  |                  |   the beginning is the parent suite id and the last part       |
    |                  |                  |   shows test index in that suite.                              |
    |                  |                  | * `longname`: Test name including parent suites.               |
+   |                  |                  | * `resolved_name`: Test name with resolved variables           |
```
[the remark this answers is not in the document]
yamatoRT: it is remove

src/robot/running/runner.py
yamatoRT: I apply your proposition for ListenerArguments, I had misunderstood how works ListenerArguments. But now, the testcases name in the output.xml  is not resolved. 
It's due to `def start_test(self, test)` of `Logger` who take in argument the `ModelCombiner` class and when the method call `test.name`, it's a unresolved testcase name who is returned.
I don't kwow how we can have unresolved name in listenerv2 and the output.xml. Do you want really the unresolved name in listenerv2? Otherwise, i think we need to modify more code. What do you think?

pekkaklarck: Good point that if we change the name used by ModelCombiner then also others who use it see that name. We have three possibilities what to do with listener v2:

1. Just use the resolved name. Easiest to implement but @michalanglart [commented](https://github.com/robotframework/robotframework/issues/2962#issuecomment-551978253) earlier that this would cause some problems for them.

2. Use the resolved name but also add `originalname` to the listener attributes. This would be easy to implement as well and would give @michalanglart an easy way to handle this situation in their code. Would also have the benefit of showing both names.

3. Try to handle this special case in the listener side. ModelCombiner has access to both the original and the resolved names so this ought to be doable. Name is handled differently than other attributes, so this would be a bit more work than option 2.

Option 2. starts to sound best for me. On the listener side it shouldn't require more changes than this:

```diff
--- a/src/robot/output/listenerarguments.py
+++ b/src/robot/output/listenerarguments.py
@@ -117,7 +117,8 @@ class StartTestArguments(_ListenerArgumentsFromItem):
 
     def _get_extra_attributes(self, test):
         return {'critical': 'yes' if test.critical else 'no',
-                'template': test.template or ''}
+                'template': test.template or '',
+                'originalname': test.data.name}
```

Notice that I used `originalname`, not `original_name` on purpose for consistency reasons. We've used that style with other attributes as well (which may not have been a great idea but cannot really be changed).

yamatoRT: I applied the modification, i updated test and the documentation for listeners v2.

atest/robot/output/listener_interface/support_variables_in_test_case_name_with_listener_v2.robot
```diff
@@ -0,0 +1,32 @@
...
+${MESSAGE FILE}    %{TEMPDIR}${/}messages.txt
+
+*** Test Cases ***
+Test Case Name With Variable
+    Check Test Case    ${TESTNAME}
+    Check Stdout Contains    SEPARATOR=\n
+    ...    [Start] [name] ${TESTNAME} [original name] Test Case \${NAME} With \${VARIABLE}
+    ...    [END] [name] ${TESTNAME} [original name] Test Case \${NAME} With \${VARIABLE}
+
+Test Case Name With ['My', 'List']
```
pekkaklarck: Test names on the `atest/robot` side should be easy to understand. These tests are great otherwise and I can change these myself after merging the PR.

doc/userguide/src/ExtendingRobotFramework/ListenerInterface.rst
```diff
@@ -152,6 +152,8 @@ it. If that is needed, `listener version 3`_ can be used instead.
    |                  |                  |   the beginning is the parent suite id and the last part       |
    |                  |                  |   shows test index in that suite.                              |
    |                  |                  | * `longname`: Test name including parent suites.               |
+   |                  |                  | * `originalname`: Original test name with unresolved           |
+   |                  |                  |   variables.                                                   |
```
pekkaklarck: Needs a note that this is new in RF 3.2. I can add it.

Issue #3194, opened 2019-05-28
`ExecutionResult` should support input as bytes


lubrst:
According to the documentation only paths to xml output files are allowed as input parameters of the `ExecutionResult` class. So there are the following possible enhancements:

1. Strings can also be also used as input parameters, so the documentation could be more precise here.

2.  It would be great if byte strings could also be allowed as input parameters of the `ExecutionResult` class, as in this Python3 example: `result = ExecutionResult(b'xml-string')`.
At the moment the following exception is thrown `robot.errors.DataError: Reading XML source '<in-memory file>' failed: Invalid argument` if a byte string is used.
As I see in the code, the `_open_source_if_necessary` function in the `ETSource` class is responsible for converting input to file objects, so it might be a small change to add support for byte strings.

If these improvements are okay, I can work on a pull request for them.

pekkaklarck:
Do you have a use case to use bytes as paths?

pekkaklarck:
Also, "paths" in the doc actually mean strings. I don't know do actual `pathlib.Path` objects even work. If not, that should be fixed. Should then also look are there other similar APIs that should be fixed.

lubrst:
Yes, I have a use case. I need to read output.xml files from a MongoDB GridFS to analyze them and they are always byte strings. Right now I have to decode the byte strings into strings first, so the function `_open_source_if_necessary` has to encode them back into byte strings. This unnecessary conversion could be avoided by directly supporting byte strings.

In addition, `ExecutionResult` are working with `pathlib.Path` parameters.

pekkaklarck:
OK. Encoding on the client side shouldn't be a big problem, but I don't see problems fixing this either. If I remember correctly, file system APIs in Python 3 still support byte paths so most likely everything should work fine if such paths are just passed to them as-is. I'd assume byte paths work already now on Python 2 and with Python 3 the problem would be due to some other check we do for paths. Did you already debug the problem further?

The fix should contain also unit tests to validate the behavior. It would be great to add tests also to the `pathlib.Path` use case.

lubrst:
Yes, the problem seems to be in the `not is_string(self._source)` condition in the following function:
```    
def _open_source_if_necessary(self):
    if self._source_is_file_name() or not is_string(self._source):
        return None
    if IRONPYTHON_WITH_BROKEN_ETREE:
        return StringIO(self._source)
    return BytesIO(self._source.encode('UTF-8'))
```
[...]

PR #3197, opened 2019-05-29
Support byte strings as ExecutionResult input (#3194)

lubrst:
This PR fix issue #3194.

I have updated and added some unit tests. But I have not added unit tests for `pathlib.Path` objects because `pathlib` requires Python 3.4 and these tests would not run with lower interpreter versions.

### DIFF
--- a/src/robot/utils/etreewrapper.py
+++ b/src/robot/utils/etreewrapper.py
@@ -74,8 +74,11 @@ class ETSource(object):
                 and not self._source.lstrip().startswith('<')
 
     def _open_source_if_necessary(self):
-        if self._source_is_file_name() or not is_string(self._source):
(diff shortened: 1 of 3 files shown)


pekkaklarck:
You can use

```python
if PY3:

    def test_xxx(self):
         # ...
```

to create tests that run only on Python 3.

lubrst:
Added unit tests for pathlib.Path as ExecutionResult input

pekkaklarck:
The PR looks good. There are small things that should be fixed like `_open_source_if_necessary` returning implicitly `None` if the input isn't string or bytes (it could be an opened file object). I'll fix those myself and also try to add support for paths specified as bytes.

Issue #3195, opened 2019-05-28
Feature request: add older YAML file extension (.yml - Variable file as YAML)


hunkisslaszlo:
Hi!

We want to use robotframework for testing out project. One of our test would be a YAML structure validation. 

I know, that the suggested file extension for YAML files is `.yaml` ([YAML faq](https://yaml.org/faq.html)) but several other projects (ex. Ansible) use it with extension `.yml`. 

I know you guys mention this in your [documentation](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#variable-file-as-yaml) but is it possible to add support older `.yml` extension?

robotframework/src/robot/variables/filesetter.py (line 44)
```
- if path_or_variables.lower().endswith('.yaml'):
+ if path_or_variables.lower().endswith('.yaml') or path_or_variables.lower().endswith('.yml'):
```

pekkaklarck:
Sounds good to me. Would require a PR with
- code changes,
- new test or just change to existing test to use `.yml` and not `.yaml`, and
- documentation update.

pekkaklarck:
There were some problems with tests and docs in PR #3200. It took me so long to review it that instead of asking @hunkisslaszlo to update the PR, I can do changes myself.

PR #3200, opened 2019-05-30
Add yml file extension support

hunkisslaszlo:
The original issue was #3195 .


pekkaklarck:
Thanks for the PR. It's great it had tests, but duplicating all same tests as with the `.yaml` extension is perhaps a bit overkill. In my opinion it's enough to just add tests for the new '.yml` extension to the existing suite. Also in the docs there should be a note that `.yml` is supported only with RF 3.2+.

It took me so much time to review this PR that I don't want to make you do the above changes. I can instead do them myself.

pekkaklarck:
I cannot easily find a way to merge only the code changes from this PR. Changes are so simple that it's easier if I just close this PR and do the change along with test and doc updates discussed above.

Issue #3027, opened 2018-12-14
Read signature (argument names, defaults, types) from "wrapped" keywords correctly


rainmanwy:
As discussed in #3024 , when keyword use user defined decorator and decorator may mask arguments of origin function/method, and Type Information Validation will report error.

Only solution is get origin function from "`__wrapped__`" attribute(if user use "functools.wraps"), so that could get origin arguments. This only works with *python3* and *"functools.wraps" is used in decorator*.

I guess this should be documented in User Guide or somewhere.

rainmanwy:
@pekkaklarck , it is ok to me.
Thanks for the information.

pekkaklarck:
After thinking this a bit more, decided to move it to 3.1.2 scope. Feature is very useful, shouldn't cause problems, and is easy to implement by using [inspect.unwrap](https://docs.python.org/3/library/inspect.html#inspect.unwrap). Decided to add `backwards incompatible` because the keyword signature that Robot sees changes and there's a tiny possibility for backwards incompatible problems. In general seeing correct signature should obviously be useful.

pekkaklarck:
Moved back to 3.2 scope. Want to to have 3.1.2 ready ASAP and avoid all unnecessary backwards compatibility problems.

PR #3206, opened 2019-06-07
Read signature from "wrapped" keywords correctly

mihaiparvu:
PR for #3027 

I feel that more acceptance tests are needed for this fix, but I would need some hints on this.


pekkaklarck:
Wouldn't it be enough to just add some test that use a library that uses a decorator that uses [functools.wraps](https://docs.python.org/3/library/functools.html#functools.wraps)? I don't think that too many tests are needed because we don't really need to test that [inspect.unwrap](https://docs.python.org/3/library/inspect.html#inspect.unwrap) works correctly.

After thinking this a bit more, I don't see much need to support this on Python 2 at all. `functool.wraps` doesn't add `__wrapped__` on it so anyone using this functionality needed to add it manually. That would include our acceptance tests. I'd rather encourage everyone to upgrade to Python 3.

pekkaklarck:
Notice that I won't be merging PRs related to RF 3.2 until we have merged the [newparser](https://github.com/robotframework/robotframework/tree/newparser) branch to master. Due to the starting holiday season that may happen only in August, but that doesn't mean we should stop working with this PR until that.

atest/testdata/keywords/type_conversion/keyword_decorator_with_wraps.robot
```diff
@@ -0,0 +1,10 @@
+*** Settings ***
+Force Tags        require-py3
+Library           LibraryWithKeywordWraps.py
+
+*** Test Cases ***
+Keyword Decorator With Wraps
+    Keyword With Wraps    42
+
+Keyword Decorator With Wraps Mismatched Type
+    Run Keyword And Expect Error    *Argument 'arg' got value 'string' that cannot be converted to integer.    Keyword With Wraps    string
```
pekkaklarck: We've used `[Documentation]` to specify the expected error instead of using `Run Keyword And Expect Error`.

atest/testdata/keywords/type_conversion/keyword_decorator_with_wraps.robot
```diff
@@ -0,0 +1,10 @@
+*** Settings ***
+Force Tags        require-py3
```
pekkaklarck: `require-py3` isn't needed here, can be used, but is needed on the `atest/robot` side.

doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
```diff
@@ -1307,9 +1307,10 @@ Python's decorators. However, decorators modify function signatures,
 and can confuse Robot Framework's introspection when determining which
 arguments keywords accept. This is especially problematic when creating
 library documentation with Libdoc_ and when using RIDE_. To avoid this
-issue, either do not use decorators, or use the handy `decorator module`__
-to create signature-preserving decorators.
+issue, you could use `functools.wraps`__ on Python 3, or use the handy
+`decorator module`__ to create signature-preserving decorators.
 
+__ https://docs.python.org/3/library/functools.html#functools.wraps
```
pekkaklarck: A note that `functools.wraps` works like this only with RF 3.2 needs to be added.

src/robot/running/arguments/argumentparser.py
```diff
@@ -23,10 +23,11 @@
 if PY2:
     from inspect import getargspec, ismethod
 
+
     def getfullargspec(func):
         return getargspec(func) + ([], None, {})
 else:
-    from inspect import getfullargspec, ismethod
+    from inspect import getfullargspec, ismethod, unwrap
```
pekkaklarck: Dummy `unwrap` implementation is needed also in the `PY2` block.

src/robot/running/arguments/argumentparser.py
```diff
@@ -23,10 +23,17 @@
...
+
     def getfullargspec(func):
-        return getargspec(func) + ([], None, {})
+        return getargspec(unwrap(func)) + ([], None, {})
+
+
+    def unwrap(func):
+        while hasattr(func, '__wrapped__'):
+            func = func.__wrapped__
+        return func
```
pekkaklarck: I'm not sure is this a good idea. In my opinion we should unwrap keywords only when using Python 3 where it's supported by Python itself. If we were to support Python 2, we needed to add some tests for it.

We do need `unwrap` implementation here, though, because it's used by `PythonArgumentParser` below. It's implementation can just be:

```python
def unwrap(func):
    return func
```

mihaiparvu:
Thanks for reviewing @pekkaklarck! I did all the changes from your comments. I'm pretty sure the tests I added are not what you requested, but I'm out of ideas on how to do them.

src/robot/running/arguments/argumentparser.py
```diff
@@ -21,12 +21,10 @@
 
 
 if PY2:
-    from inspect import getargspec, ismethod
-
-    def getfullargspec(func):
-        return getargspec(func) + ([], None, {})
```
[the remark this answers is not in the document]
mihaiparvu: Sorry, was a bit overzealous with code removal. I ran the PY2 tests with the initial commit, but forgot to run with this change 😅. Ought to work now.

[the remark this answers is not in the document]
pekkaklarck: Great! There seems to be some unnecessarily added/removed lines still but I can fix them myself.

pekkaklarck:
Thanks for the PR @mihaiparvu! It's been now merged but I'll add some more tests and may do a little cleanup as well.

PR #3212, opened 2019-06-19
testdoc: fix typo in the execution examples

fabo-at-noona:
The test filename passed as an argument should be my_test.robot, instead
of my_test.html.

Signed-off-by: Fathi Boudra <fathi.boudra@varian.com>


pekkaklarck:
Good catch! I'll wait until we get the `newparser` branch merged into master before merging this one. Due to holiday season that may take a little time.

Issue #2699, opened 2017-10-26
Support literal list/dict syntax when using `@{list}` and `&{dict}` as embedded arguments


zachauten:
Currently you can pass lists and dictionaries as arguments to embedded keywords but there is no way to ensure an argument will be a list or dictionary.

VVasyaev:
why don't you like the
${type}    Evaluate    type($arg)
?

pekkaklarck:
Could you please clarify what kind of change are you actually proposing? I think I know what you want, but issues should be more explicit.

ZachAuten:
Currently the only way to embed arguments into a custom library keyword is as a scalar.
```python
@keyword("test ${var} keyword")
def test(self, *var):
```
If you pass a list variable into the keyword (`test @{list} keyword`) it works fine, but it would be nice to be able to embed lists like `test firstarg secondarg thirdarg keyword`.

To do this you could allow someone writing a decorator for a custom keyword to specify if a variable is a list or a dictionary like `@keyword(test @{list} keyword)` or `@keyword(test &{dict} keyword)` instead of only scalars `@keyword(test ${var} keyword)`

pekkaklarck:
If you'd have a keyword with name `test @{list} keyword` and it would be used like `test firstarg secondarg thirdarg keyword`, how could Robot know how to split argument `firstarg secondarg thirdarg` into a list? Just using a space as a separator wouldn't work too well if the keyword would then also be used like `test first arg second arg third arg keyword`.

ZachAuten:
Sorry, I've been trying to find out more about how Robot parses embedded arguments but I'm not that familiar with it yet. But in the case of just one list being embedded into a keyword couldn't you just use a regex to separate the keyword from the list?

pekkaklarck:
This isn't a technical problem. The problem is deciding about the list or dictionary syntax. You proposed that arguments specified as a string `firstarg secondarg thirdarg` are parsed to a list `["firstarg", "secondarg", "thirdarg"]`, but then `first arg second arg third arg` is parsed to `["first", "arg", "second", "arg", "third", "arg"]` which clearly is wrong.

pekkaklarck:
[...]

PR #3218, opened 2019-06-26
List/dict syntax when using `@{list}` and `&{dict}` as embedded arguments

mihaiparvu:
PR for #2699

### DIFF
--- a/src/robot/running/userkeywordrunner.py
+++ b/src/robot/running/userkeywordrunner.py
@@ -20,7 +20,7 @@ from robot.errors import (ExecutionFailed, ExecutionPassed, ExecutionStatus,
                           PassExecution, ReturnFromKeyword,
                           UserKeywordExecutionFailed, VariableError)
 from robot.result import Keyword as KeywordResult
-from robot.utils import getshortdoc, DotDict, prepr, split_tags_from_doc
+from robot.utils import getshortdoc, DotDict, prepr, split_tags_from_doc, is_list_like, is_dict_like
 from robot.variables import is_list_var, VariableAssignment
 
 from .arguments import DefaultValue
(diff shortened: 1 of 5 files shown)


src/robot/running/userkeywordrunner.py
```diff
@@ -235,7 +235,19 @@ def _resolve_arguments(self, args, variables=None):
...
+        for n, v in self.embedded_args:
+            try:
+                v = eval(v)
+                if is_dict_like(v):
+                    yield n, DotDict(v)
+                elif is_list_like(v):
+                    yield n, variables.replace_list(v)
+            except:
+                pass
+            yield n, variables.replace_scalar(v)
```
pekkaklarck: Blindly using eval and especially empty `except:` are both pretty ugly. Also, wouldn't this code turn embedded argument values that can be evaluated in Python to the resulting value? For example, arguments like `True` and `1 + 2` (both strings) would be converted to Python `True` and integer `3`.

Evaluation should *only* be done if the original argument name starts with `@` or `&`. Then it should be validated that if it starts with `@`, the result should be a list, and if it starts with `&` it should be a dict. Also, evaluation should rather be done using `ast.literal_eval` than `eval`.

It would probably be a good idea to implement dedicated utility functions for converting string to lists and dicts. In the beginning these utils could just use `ast.literal_eval` (and validate the result is list/dict), but in the future we could try coming up with other, more readable formats as discussed in #2699.

pekkaklarck:
It seems more and more likely that #3278 will be implemented instead of #2699. It is best not to spent more time with this PR until a decision has been made.

mihaiparvu:
I commited the changes I had in progress for this PR in https://github.com/robotframework/robotframework/pull/3218/commits/34b68255ae32cd343503e5f03c609e61bb22930f. The code is still not finished yet, but if it's decided at some point in the future that this is needed I will continue the implementation.

Issue #3019, opened 2018-12-02
`@library` decorator that supports configuring and forces using `@keyword` to mark keywords


yahman72:
I guess this would be a new feature: have a look an let me know what you think about it.

The "Problem":
Sometimes you have python code with class structure like this
```
             BaseClass
             foo()
                |
       ---------------------
       |                   |
   SubClassA           SubClassB
   bar_a()             bar_b()
```

And you would like to make a RF Library out of it like this:
```
             BaseClass
             foo()
                |
       ---------------------
       |                   |
   SubClassA           SubClassB
   bar_a()             bar_b()
       |                   |
       |                   |
   RFLibA              RFLibB
   foo_bar()           bar_foo()
```

But when trying to use those `RFLibA` and `RFLibB` libraries you'll run into to trouble because e.g. `foo()` is in both libraries.

It would be good to be able to hide certain KWs in a library.

I had a look at the code and the `@keyword` decorator looks like nice fit for this i.e. reserve a value for the name argument that is treated as "hide this keyword" - for example the string "_" (could also be something else

I.e. in `RFLibA` I would have this for hiding the `foo()` KW
```
    @keyword(name="_")
    def foo(self):
        super(RFLibA).foo()
```
This could be handled e.g. in the `robot.running.testlibraries._BaseTestLibrary` e.g in the `_create_handlers()`

I could also do the same (i.e. have `RFLibA` without `foo()`) without this "hide" feature also by creating the libraries in a different way:
- `RFLibA` is a wrapper that is not inherited from `SubClassA`, but creates an instance of `SubClassA` and calls it's methods
- override `__dir__()` in `RFLibA` and explicitly list the KWs I want to expose --> a bit of a hack
 
But in my current case (much more complex class hiearchy) neither of this is desirable - that "hide" would be the nicest & cleanest way.

pekkaklarck:
First of all, I highly recommend avoiding inheritance hierarchies like this and using composition instead. You already mentioned that `RFLibA` doesn't necessarily need to inherit `SubClassA` but instead it could use it internally.

There are, however, also other reasons to expose only certain methods of a class as keywords. Using the `@keyword` decorator to mark which methods should be considered keywords is a good idea. For backwards compatibility reasons we cannot change the library API so that others would be ignored automatically, though.

[...]

PR #3219, opened 2019-06-26
Disable public methods

cristii006:
This is the PR for the first part of #3019 as discussed on Slack https://robotframework.slack.com/archives/C0K0240NL/p1560209961022000. The second PR related to this issue is in progress.

### DIFF
--- a/doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
+++ b/doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
@@ -470,13 +470,13 @@ For example, the library below implements only keywords
 __ https://docs.python.org/tutorial/modules.html#importing-from-a-package
 
 In order to avoid all public methods of a class that are not decorated with
-`@keyword` decorator becoming keywords, class attribute `ROBOT_AUTO_LIBRARY`
+`@keyword` decorator becoming keywords, class attribute `ROBOT_AUTO_KEYWORDS`
 should be set to False.
 
  .. sourcecode:: python
 
      class MyLibrary:
-        ROBOT_AUTO_LIBRARY = False
+        ROBOT_AUTO_KEYWORDS = False
 
          def my_keyword(self, name):
             return name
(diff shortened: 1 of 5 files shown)


pekkaklarck:
Cool! I don't have time to look at the PR now and due to holiday season and various other tasks it may take some time before I have that time.

It might be a good idea to have a separate issue that covers only this part of the #3019. What do you think?

cristii006:
Created issue #3221 for this PR.

doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
```diff
@@ -469,6 +469,18 @@ For example, the library below implements only keywords
...
 __ https://docs.python.org/tutorial/modules.html#importing-from-a-package
 
+In order to avoid all public methods of a class that are not decorated with
+`@keyword` decorator becoming keywords, class attribute `ROBOT_AUTO_LIBRARY`
+should be set to False.
+
+ .. sourcecode:: python
+
+     class MyLibrary:
+        ROBOT_AUTO_LIBRARY = False
```
bkhouri: should this be `ROBOT_AUTO_KEYWORDS` instead of `ROBOT_AUTO_LIBRARY`

cristii006: Good catch! Thanks!

pekkaklarck:
Thanks for a great PR @cristii006!

Few things that should still be done:

- It would be good to add an explicit test for using this functionality with modules. It should just work, but I'd like to make it sure because the forthcoming `@library` decorator won't work with modules and `ROBOT_AUTO_KEYWORDS` is the only way to enable this functionality.

- The docs also need a note that this is a new feature in RF 3.2. They also need a note that with classes using `@library` is generally a better idea than setting this attribute explicitly, but that cannot really be added until the `@library` PR is merged.

This PR is otherwise in so good shape that I think I'll just merge it and do those changes myself.

PR #3222, opened 2019-06-28
Library decorator

cristii006:
PR for #3019.


src/robot/api/deco.py
```diff
@@ -68,3 +68,43 @@ def decorator(func):
         func.robot_types = types
         return func
     return decorator
+
+
+def library(scope=None, version=None, robot_auto_kws=False):
```
pekkaklarck: 1. `robot_auto_kws` is inconsistent compared to other arguments. `auto_keywords` would be better.
2. `auto_keywords` should be `True` by default! There should be a test that it can be disabled, though.
3. We also have two other attributes libraries can have: `ROBOT_LIBRARY_DOC_FORMAT` and `ROBOT_LIBRARY_LISTENER`. Should add matching `doc_format` and `listener` arguments here as well.
4. Hard to say what would be the best order. Perhaps this: `scope, version, doc_format,  listener, auto_keywords`

src/robot/api/deco.py
```diff
@@ -68,3 +68,43 @@ def decorator(func):
...
+            def public_method_is_not_keyword():
+                print('This method will not become keyword')
+     """
+
+    def lib_decorator(func):
+        func.ROBOT_LIBRARY_SCOPE = scope
+        func.ROBOT_LIBRARY_VERSION = version
+        func.ROBOT_AUTO_KEYWORDS = robot_auto_kws
+        return func
+    return lib_decorator
```
pekkaklarck: 1. This is a class decorator so argument shouldn't be named `func` but `cls`. The decorator itself could just be named `decorator`, not `lib_decorator`.

2. Need to decide does this decorator or attributes set explicitly have higher precedence. I'll add a separate comment about that.

3. This implementation seems to only work if you pass arguments to the decorator like `@library(version='1.1')`, not if you just use `@library`. See `@keyword` for how to handle that. Tests for just using `@library` are needed as well.

doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
```diff
@@ -469,6 +489,17 @@ For example, the library below implements only keywords
...
+decorator which by default set its parameter `robot_auto_kws` to False.
+
+ .. sourcecode:: python
+
+     @library(robot_auto_kws=False)
+    class MyLibrary:
+
+         def my_keyword(self):
+            print('This method will not become keyword)
+
```
pekkaklarck: This documentation can be fine tuned once PR #3219 adding `ROBOT_AUTO_KEYWORDS` is merged. That will be soon.

atest/testdata/test_libraries/LibraryDecorator.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __init__(self):
+        self.invalid = 'This method is not a keyword.'
+
+    def library_decorator_disables_public_methods(self):
+        print(self.invalid)
+
+    @keyword(name="Decorated Method Is Keyword")
+    def decorated_method(self):
+        print(self.invalid)
```
pekkaklarck: It's a bit strange to print `self.invalid` here when this is valid usage. I doubt `self.invalid` is needed in general. Could just print something relevant here and in the other use something like `raise RuntimeError("Should not be executed!")`.

pekkaklarck:
Thanks for another great PR @cristii006! Do you have time/interest to continue with this one? I can do needed changes to #3219 and merge it.

pekkaklarck:
I enhanced documentation related to limiting what methods become keywords in d020725902bb20c3888acb58e786be953776d4c7. `@library` decorator needs to be mentioned there under `Class based libraries` as well.

Need to think how we should cover `@library` in general. Should we only mention it as an option in places where features like the library scope and version are discussed, or should we have a separate section dedicated to it and `@keyword`? Should anyway mention `@library` when scope, version, etc. are discussed, but having a dedicated section would mean all info related to `@library` and `@keyword` would be in same place. If you @cristii006 aren't interested in this, I can take care of documentation once this PR is otherwise ready.

cristii006:
Thanks for reviewing @pekkaklarck and for all the tips! I managed to do the requested changes from the comments.

PS. Sorry for the mess caused by the commits! :)

pekkaklarck:
Thanks for fixes @cristii006! Some things that should still be enhanced:

1. The code looks good otherwise, but the decorator currently unconditionally sets attributes to the class and overrides possible values set in the class definition. I commented about this [already earlier](https://github.com/robotframework/robotframework/pull/3222#pullrequestreview-316407653) and proposed setting values only if they are given with an exception to `ROBOT_AUTO_KEYWORDS` that should always be set.

2. There should be dedicated tests for setting all these attributes, not only for setting the new `ROBOT_AUTO_KEYWORDS`. The features provided by the existing attributes are already tested thoroughly and new tests don't need to test them again in that level. There should, however, be some test for setting each attribute and preferably also for (optionally) overriding existing attributes. I guess it would be enough to test scope, version, and doc format with Libdoc. If you need help figuring out where and what kind of tests to add, ping me on Slack. I can also add these tests myself if needed.

3. Documentation needs to be extended. I already promised to take care of that so no need to worry about it.

pekkaklarck:
Do you @cristii006 have time to continue with this PR this week? I'd like to have RF 3.2 beta out early next week and definitely want to have this feature in. If you are busy, I can continue the work myself.

pekkaklarck:
Thanks for the PR @cristii006! Docs and tests still need to be enhanced a bit but I'll do that myself.

Issue #3224, opened 2019-07-01
Forgotten Debug Statement ("Length Is")


tgoncuoglu:
There is a forgotten 

`self.log('Length is %d' % length)`

statement in **BuiltIn.py** at line **1213** that ends up in logs and pollutes output when a condition is used such as `Should Not Be Empty` etc. Solution is to simply remove the line.

Thanks in advance...

pekkaklarck:
The message is there on purpose. What problems does it cause?

tgoncuoglu:
Those "length is..." messages end up in output.xml as well as log.html in places where it does not convey any useful information to the reader of log; human or otherwise. For instance after "Should Not be Equal  A  B"  you get a "Length is 16", which does not mean anything to robot user. That is why I assume it was a debug statement, to see if the _get_length sub-call actually works during development, and then forgotten there.

Our situation is similar, we take the output.xml and put it through a pipeline to fetch information out of it, but after many statements we get additional "Length is" information for python-level variables that are not necessarily seen from outside the python script itself, therefore meaningless, and also disruptive since they come up in between CLI commands and their output.

Yes, we could try to clean the xml first in pipeline, but message really looks like a forgotten debug info statement. 

If you are not willing to remove it, could you at least do something like showing it only if a debug tag was given ?

pekkaklarck:
Are you certain you get this message with `Should Not Be Equal`? That would be odd.

The message is in `Get Length` to make it possible to just do

```
Get Length    ${something}
```

to see the length in the log file. Some other length related keywords use `Get Length` as a helper and thus get the message as well. In most of these cases the message is, in my opinion, pretty useful, and it others I don't see it doing any harm.

If this is really a big problem for you, we could consider making the log level configurable and change some of the keywords that use `Get Length` to use debug level. That would require a PR with (changed) tests, but first we should agree what keywords should actually be changed.

pekkaklarck:
The message is there on purpose. It could be made configurable, but apparently there's not enough interest for that.

PR #3225, opened 2019-07-01
Forgotten debug statement removed

tgoncuoglu:
There is a log info statement in BuiltIn.py that outputs to log "Length is NN" messages when many built-in functions are called. This is most probably forgotten there and should be removed. This commit removes that debug statement.


pekkaklarck:
As I explained in comments to issue #3224, the log message is there on purpose.

PR #3246, opened 2019-07-23
Enable ${} translation for keywords when no assignment applies

gerard:
If a keyword cannot be parsed where it should, try translating any
variables enclosed in ${}.  This is useful to provide "method-like"
syntax for an object.  An obvious usecase would be objects which have
been used as arguments to "Import Library", similar to this:

    Import Library      library     ${object}     WITH NAME    object.library
    <some method setting ${object.library} to "object.library">

    Run Keyword         ${object.library}.My Keyword

This change enables you to write this instead of the last line:

    ${object.library}.My Keyword

Note that the following is already possible before this change:

    object.library.My Keyword

However, this does not work once you try passing ${object} to a
keyword, since the keyword cannot possibly know the original name that
was given to ${object}.

### DIFF
--- a/src/robot/running/namespace.py
+++ b/src/robot/running/namespace.py
@@ -215,7 +215,7 @@ class Namespace(object):
 
     def get_runner(self, name):
         try:
-            return self._kw_store.get_runner(name)
+            return self._kw_store.get_runner(name, self)
         except DataError as error:
             return UserErrorHandler(error, name)
 
@@ -255,8 +255,8 @@ class KeywordStore(object):
                 return lib
         self._no_library_found(instance)
 
-    def get_runner(self, name):
-        runner = self._get_runner(name)
+    def get_runner(self, name, ns):
+        runner = self._get_runner(name, ns)
         if runner is None:
             self._raise_no_keyword_found(name)
         return runner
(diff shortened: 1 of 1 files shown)


gerard:
So, I was just wondering if something like this would be accepted.  It lets you drop the "Run Keyword" when the keyword name needs to be "resolved".  I've neglected writing tests for the feature since I wanted to get your feedback first.

pekkaklarck:
We cannot support
```
${keyword}    argument
```
syntax because it is interpreted as
```
${result}    Keyword
```
but supporting
```
${x}.keyword    argument
```
or generally
```
${x}kw    argument
kw${x}    argument
k${x}w    argument
```
would be technically possible. The question is, what would we benefit from this new syntax?

The use case in the description is about being able to use `${lib}.Keyword` syntax and apparently the idea is that you could change the value of `${lib}` and get different keywords executed. There are, however, features like importing different library or resource file based on a variable (e.g. `Library    ${lib}    WITH NAME    Lib`) and using `BuiltIn.Set Library Search Order` that provide same functionality. Adding new syntax always makes the framework more complicated and there should be concrete benefits from it.

For further discussion about this, it would be good to submit an issue.

pekkaklarck:
No updates in two years. Closing.

Issue #3232, opened 2019-07-08
Telnet: Global Library Scope or Cleanup


bbara:
Hey guys,

I just realized that the scope of the Telnet Library is not set to GLOBAL but to TEST_SUITE.
As explained [here](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#test-library-scope), it might be useful for connection-based tests, as there might be multiple test suites and test cases that require the same connection and "re-connecting" for each test introduces quite some overhead.

If this is not possible, I would suggest something like adding an implicit closing of connections into the "destructor" of the library, so something like:

```
def __exit__(self, exc_type, exc_value, traceback):
    self.close_all_connections()
```

which could also be option-based to keep backward-compatibility.

What do you think about it?

Cheers,
BB

pekkaklarck:
I don't remember why we've used the test suite scope and not global and don't see why it couldn't be changed. I need to run tests with Telnet anyway before RF 3.2 rc 1 and can test the PR at the same time.If there are no problems, I'll merge the PR and let users test it during the release candidate period.

pekkaklarck:
I have also been thinking that Telnet should be turned to an external library. It would allow developing it further without being tied to Robot Framework's release schedule.

pekkaklarck:
Unfortunately PR #3250 had problems making it impossible to merge it now before RF 3.2.

I have also decided it's better to make the Telnet library an external library and submitted #3537 about that. As the result I'm closing all Telnet related issues and PRs (including this one) and linking them from #3537 as a reference. If you are interested in becoming maintainer of the new library and adding this functionality there, please comment #3537 or contact me otherwise.

PR #3250, opened 2019-07-29
TelnetLib: switch to GLOBAL library scope

bbara:
Hey all,

this commit changes the LIBRARY_SCOPE of the Telnet Library to GLOBAL
allowing the user to keep a persistent Telnet connection.
Beside of that, the behavior is exactly the same.

Targeting my issue #3232.
This is useful if multiple test suites are based on the same connection and
the connection handling is not part of the test specification.

Please feel free to inform me if additional tasks are required.
Tested on Debian Buster w/ Python 3.7.3.

Cheers,
BB

### DIFF
--- a/src/robot/libraries/Telnet.py
+++ b/src/robot/libraries/Telnet.py
@@ -286,7 +286,7 @@ class Telnet(object):
     Considering string ``NONE`` false is new in Robot Framework 3.0.3 and
     considering also ``OFF`` and ``0`` false is new in Robot Framework 3.1.
     """
-    ROBOT_LIBRARY_SCOPE = 'TEST_SUITE'
+    ROBOT_LIBRARY_SCOPE = 'GLOBAL'
     ROBOT_LIBRARY_VERSION = get_version()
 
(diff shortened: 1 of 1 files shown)


drchopper:
Upvote for GLOBAL scope in Telnet Lib.

pekkaklarck:
I agree the global scope would make more sense. Unfortunately the approach used in this PR masks keyword documentation from Robot Framework and from Libdoc making it impossible to create library documentation.

The library uses the hybrid library API and earlier actual keywords seen by Robot came from the `TelnetConnection` class and thus their docstrings were used. The problem could be easily fixed by copying/moving docstrings to the `Telnet` class, but it would probably be a good idea to change the library to use the normal static library API at that point too. I believe it would ease maintaining the library in general as well. Unfortunately doing big changes like this isn't possible anymore before RF 3.2. 

I have also decided it's better to make the Telnet library an external library and submitted #3537 about that. As the result I'm closing all Telnet related issues and PRs (including this one) and linking them from #3537 as a reference. If you are interested in becoming maintainers of the new library and adding this functionality there, please comment #3537 or contact me otherwise.

Issue #3275, opened 2019-08-21
Why no automated testing? Travis CI, CircleCI, AppVeyor, Azure Pipelines, etc?


cclauss:
It would be cool to see robotframework use one of these free software testing environments.

https://docs.travis-ci.com/user/languages/python/#running-python-tests-on-multiple-operating-systems

pekkaklarck:
The main reason is that free options typically support only Python on Linux (haven't checked lately), but we'd also need Windows, Jython, IronPython and PyPy. Without them CI doesn't really give much benefits because Python/Linux combination I can run on my own laptop faster than on CI. I can run everything else as well (I have Windows virtual machine) but that takes time and there CI would help.

Another reason is that I don't want to maintain a CI system myself. We've been talking that Robot Framework Foundation should take care of it, but it hasn't been done yet. Need to discuss about that again. If someone from the community would be interested to help, that would definitely be appreciated. The foundation can take care of possible hosting fees.

cclauss:
Script for testing every code change on Linux, macOS, and Windows on Travis CI: https://docs.travis-ci.com/user/languages/python/#running-python-tests-on-multiple-operating-systems  Similar things can be done on the others or natively with GitHub Actions CI.

MaciejWiczk:
Github actions seems like a good, cross platform choice. Currently there is a need to signup for beta, but it's going to be public at november. I could give it a try, when possible.

bollwyvl:
On #3045, @rasjani (_hej, Jani!_) demonstrated the whole matrix listed [above](https://github.com/robotframework/robotframework/issues/3275#issuecomment-523659647) quite robustly with azure pipelines, i believe. Open source projects are only limited by [10 parallel jobs](https://azure.microsoft.com/en-us/blog/announcing-azure-pipelines-with-unlimited-ci-cd-minutes-for-open-source/), so that PR (14 jobs) would saturate it for a good amount of time, but it would no doubt be worth it vs future contributor review time (and, presumably, platform-specific edge case regression issues) saved.

If queue wait times become a problem, it can certainly make sense to "gate" the big matrix with high-impact, low-cost jobs like linting/static analysis and a "fail fast" reference configuration, e.g. Linux with cpython37, and then proceed. Those are pretty micro-optimizations, and might not even be necessary... :euro: can't buy happiness, but it sure can buy additional CI assets.

rasjani:
[...]

PR #3276, opened 2019-08-21
Travis CI: Lint code for Python syntax errors and undefined names

cclauss:
Also, follow Python porting best practice [__use feature detection instead of version detection__](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: python
+python:
+  - 2.7
+  - 3.7
+matrix:
+  allow_failures:
+    - python: 2.7
+    - python: 3.7
+install: pip install flake8
+script: flake8 . --count --exclude=./.*,./atest/testdata --select=E9,F63,F7,F82 --show-source --statistics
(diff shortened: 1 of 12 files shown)


cclauss:
The remaining [flake8 issues on Python 3](https://travis-ci.org/robotframework/robotframework/jobs/574843324) should be fixed in #3277

pekkaklarck:
1. It would be good to have linting on CI but that should be done when setting CI in general. My comment to #3275 explains why that's not so easy.

2. I don't like `# noqa` comments too much in general. Is there a standard so that such comments work with different linters and also disable warnings with IDEs? If yes, then we could add them. If all tools need their own magic comments, I don't think they are worth it.

3. Our code uses both version detection and feature detection depending on the context. I don't think the latter is always better and definitely don't think it should be changed everywhere. Also, I wouldn't like want to touch our Python 2/3 compatibility code before we can strip Python 2 compatibility out for good. Finally, our version detection works with possible Python 4.

4. Enabling CI, adding `# noqa` comments and possible code changes should all have their own PRs.

pekkaklarck:
Closing this because we are using GitHub actions, not Travis, for other CI needs. I'm still not how much benefits we'd got from linting either. Would be interesting to see a sample reports.

PR #3277, opened 2019-08-21
Use print() function in both Python 2 and Python 3

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

### DIFF
--- a/doc/api/code_examples/check_test_times.py
+++ b/doc/api/code_examples/check_test_times.py
@@ -8,6 +8,7 @@ took longer than given amount of seconds to execute.
 Optional `outpath` specifies where to write processed results. If not given,
 results are written over the original file.
 """
+from __future__ import print_function
 
 import sys
 from robot.api import ExecutionResult, ResultVisitor
@@ -34,4 +35,4 @@ if __name__ == '__main__':
     try:
         check_tests(*sys.argv[1:])
     except TypeError:
-        print __doc__
+        print(__doc__)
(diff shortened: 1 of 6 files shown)


pekkaklarck:
We've change all `print` statements to functions in our actual source code, but apparently there are some of them still in examples and in our development scripts. Especially making examples Python 3 compatible is important.

I don't like adding the `from __future__ import print_function` line into the examples, though. It makes them more complicated, and because `print` usage in them is so simple it's OK to just use `print('message')` and it works both in Python 2 and 3. It doesn't seem to be necessary with the devel scripts either, but leaving it to them is OK.

cclauss:
Python porting best practice says that we should [__prevent compatibility regressions__](https://docs.python.org/3/howto/pyporting.html#prevent-compatibility-regressions).  This import ___changes the way that Python 2 operates___ so that __print__ statements are now syntax errors while __print()__ function continues to work as expected.  This prevents Python 2 developers from writing code that breaks compatibility with Python 3.

Proof:
* $ python2 -c "print 'This works...'"
* $ python2 -c "from __future__ import print_function ; print '...but this raises a syntax error.'"

The import is also __required__ in any __print__ statement that contains an unquoted comma.

$ __python2__
```
>>> print "hello", "world"
hello world
>>> print("hello", "world")
('hello', 'world')
```

With 132 days left until the end of life of Python 2, we should be teaching Python 2 developers not to break Python code.

pekkaklarck:
I know all this. But I also know I don't want our examples to have future imports that aren't needed. It would be fine for them to be only Python 3 compatible.

pekkaklarck:
Thanks for the PR! It also reminded we have a totally outdated Python tutorial under 'doc/python`. Probably best to remove it altogether. We don't have resources to update it, and nowadays there are so many great Python tutorials available that it's not really needed.

Issue #3291, opened 2019-09-03
Document making .robot files executable


sebastianciupinski:
It is possible to run *.robot files directly in shell by changing access rights to +x and adding:
```#!/usr/local/bin/robot``` as first line, but such execution way is not documented.

eternalconcert:
This is not a Robot Specific behaviour: https://en.wikipedia.org/wiki/Shebang_(Unix)

sebastianciupinski:
Lists, loops, returns aren't to, however they are documented. For me,
documenting this is quick win, but I may be wrong and this should be kept
as a secret.

wt., 3 wrz 2019, 15:22 użytkownik eternalconcert <notifications@github.com>
napisał:

eternalconcert:
Don't wanted to be offensive, but I think there is a major difference between lists/loops and a Shebang: 

The list and loop syntax is Robot specific, the Shebang is not.
And it only works on Unix/Linux systems.

pekkaklarck:
Yeah, this trick isn't Robot specific but oddly enough it never occurred to me it works also with Robot! I don't see any harm documenting it, on the contrary, so lets do it. Have you @sebastianciupinski looked where in the User Guide would be a suitable place? Interested to provide a PR?

Notice that instead of using `#!/usr/local/bin/robot`, it's probably better to use `#!/usr/bin/env robot` to make the shebang independent on the exact installation location. That's the recommended approach with Python as well.

eternalconcert:
I think, the right place would be somewhere around [executing-robot-framework](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#executing-robot-framework)

eternalconcert:
I updated my previous PR accordingly: https://github.com/robotframework/robotframework/pull/3294

pekkaklarck:
I don't think this is so important that it should be explained as part of installation instructions. I proposed a better place, and gave some other comments, in by review to #3294.

pekkaklarck:
Have you @eternalconcert had time to look at my comments to your PR?

pekkaklarck:
I wanted to get this issue closed and added a section to the User Guide about this in 3ba2d3776e5205d9ed5ec5a0d0edaae1c4c3ab17 myself.

Issue #3214, opened 2019-06-25
Document that the position of the `[Return]` setting does not affect its usage


rajakarthikeya:
Set of instructions which are written after [Return] statement are getting executed. Robot Framework should through an error or should not consider the keywords written after [Return] keyword

Ex:
```robotframework
***Keywords***
Custom Keyword
    [Return]    hyyyy
    Return From Keyword    hyyyy2222    

***Test Cases***
Test1  
   ${var_rt_ky2}=    Custom Keyword
   Log To Console    ${var_rt_ky2}
```
Output:
hyyyy2222 

In other case
```robotframework
***Keywords***
Custom Keyword
    Return From Keyword    hyyyy2222    
    [Return]    hyyyy

***Test Cases***
Test1  
   ${var_rt_ky2}=    Custom Keyword
   Log To Console    ${var_rt_ky2}
```
Output:
hyyyy

In second case working as expected. Where as in the first case it is working wrong. Please give explanation if I missed something

pekkaklarck:
The `[Return]` setting just defines what the keyword should return after all keywords it contains have been executed. It doesn't work the same way as the `return` statement in languages like Python or Java. Settings can, in fact, be in whatever order in Robot Framework data.

Because `[Return]` works like it does, separate `Return From Keyword` has been added to BuiltIn. It or its `Return From Keyword If` variant should be used if there's a need to return without executing the whole keyword.

Now that we are planning to add more programmatic functionality such as `IF/ELSE` (#3074) and `TRY/EXCEPT` (#3075), it would be good to add also separate `RETURN` functionality that would work like the `return` statement in other languages. We could submit an issue about it, but I don't think it's as important as the aforementioned features.

So, the `[Return]` statement works as designed and instead of changing the design we could add separate `RETURN` functionality in the future. At the moment the only thing that could be changed is making sure that `[Return]` is documented adequately in the User Guide. Could someone check that?

rajakarthikeya:
Thank you for getting back. Is documentation updates by Robot Framework official team members?

pekkaklarck:
Could you @rajakarthikeya clarify what you mean? If the question is who can update documentation, the answer is "anyone" via pull requests.

eternalconcert:
It is documented in the: [User Guide - Using [Return] setting](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#using-return-setting)

I will add some code documentation as well (maybe during the next days) and send a pull request.

[...]

Issue #3269, opened 2019-08-13
Support any file extension when explicitly running file and when using `--extension`


pekkaklarck:
RF 3.2 parses only `.robot` files (see #2820 for discussion) by default and requires using `--extension` if test data files use some other extension. Earlier the valid extensions were limited to `.robot`, `.txt`, `.tsv`, `.html` (not supported at all anymore), `.rest` and `.rst`, but we don't need to have that restriction anymore. We can simply parse all files that match whatever is configured with `--extension` and only `.robot` files if that options isn't used at all.

When execution a simple file, we can parse it regardless the extension. Just need to make sure we use reST parser when execution `.rest` or `.rst` files. When executing a directory, `--extension` is needed.

PR #3294, opened 2019-09-03
Updated UserGuide #3214 and #3291

eternalconcert:
…xecuting (#3214)

If necessary, I will provide some code documentation as well. Please let me known.


pekkaklarck:
Some comments about this PR:

1. The documentation to make `*.robot` files executable looks pretty good, but I don't think its important enough to be explained as part of the installation instructions. A better place would be under the [Executing test cases > Basic usage](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#id172) section, probably after 3.1.6 Creating start-up scripts.

2. I don't think example using `chmod` is that relevant, especially when many use `chmod 755` or similar instead of `chmod +x`. Just mentioning that the file must be made executable, with possibly a link to some good external resource, ought to be enough.

3. RF 3.2 makes it possible to use *any* extension, or no extension whatsoever, with test case files when you execute a single file (#3269). That means the file could actually be just `tests` or `do_something` in the RPA use case. This could be worth mentioning, but there should be a note that it only works with RF 3.2+.

4. This PR has also documentation updates related to `[Return]`. They look pretty good as well, but would be better to have a separate PR for them. In this particular case changes are so small, and all changes look pretty good, that I guess I could merge both in one PR.

pekkaklarck:
This is not needed anymore after 3ba2d3776e5205d9ed5ec5a0d0edaae1c4c3ab17.

PR #3299, opened 2019-09-09
Remove state sharing between calls in XML-RPC client

TChatzigiannakis:


pekkaklarck:
As I commented #3300, I'd like to know a bit more why this change is needed. Assuming it is, I think it would make the change simpler to change `_server` to a property. Probably should at the same time make it public and rename it to `server`.

TChatzigiannakis:
> I think it would make the change simpler to change `_server` to a property. Probably should at the same time make it public and rename it to `server`.

I didn't put a lot of thought into the decision of method versus property, it just seemed more appropriate to make it a method to indicate that the object returned is always new (and that it's a good idea to close it).

I will amend the PR to make this a property.

pekkaklarck:
Making the server a context manager is a good idea. Was there a reason `run_keyword` didn't use it with `with` statement?

It's true that `get_server_proxy` makes it more explicit that a new proxy is created. I don't feel strongly either way.

If we use a property, it could itself be a context manager and handle converting `socket.error` and `xmlrpclib.Error` to `TypeError` as well. It would remove lot of duplication but should perhaps done separately.

TChatzigiannakis:
> Making the server a context manager is a good idea. Was there a reason `run_keyword` didn't use it with `with` statement?

Just an oversight because I re-created the commit on a different machine than where I initially found it. (For this example at least, everything passes even without the `with` statement, but I suppose that in other scenarios the unreleased resources would cause problems.)

> If we use a property, it could itself be a context manager and handle converting `socket.error` and `xmlrpclib.Error` to `TypeError` as well. It would remove lot of duplication but should perhaps done separately.

I agree about removing the duplication. In other languages, I'd have extracted the duplicated code into a separate method and used a lambda to pass the inside of the `try` block. But I'm quite new to Python, so I didn't know what's the most idiomatic way to do that.

pekkaklarck:
Something like this (untested) could work:

TChatzigiannakis:
Do you think the latest commit I pushed is okay in removing the code duplication?

pekkaklarck:
Due to RoboCon 2020 organization work and some training courses I've run I haven't had time for actual Robot Framework development or reviewing pull requests lately. I hope to return to that soon. I added #3300 to RF 3.2 scope that I won't forget it. Thanks for patience!

pekkaklarck:
I tested the performance locally and have both good and bad news. 

Good news are that there were no performance issues. I created a test that did 10k remote calls and it run in average 16.9s both with the old and new version. The remote server was running on the same machine, but I doubt that would affect the relative performance. Not

The bad news is that ServerProxy can be used as a context manager only with Python 3.5 and newer. We still support Python 2.7 and 3.4 (the latter support could be dropped) so the code needs to be changed. Luckily it seems that changing the code in the `_server` property to this fixes the issue:

```python
        server = xmlrpclib.ServerProxy(self.uri, encoding='UTF-8',
                                       transport=transport)
        try:
            yield server
        except (socket.error, xmlrpclib.Error) as err:
            raise TypeError(err)
        finally:
            server('close')()
```

Can you @TChatzigiannakis check does the fix work also in your context and update the PR if it does?

TChatzigiannakis:
Yes, the `try`/`finally` block works as well. I've updated the PR.

Issue #3301, opened 2019-09-09
Libdoc: Support converting docs to HTML with XML outputs


Snooz82:
Some tools uses the xml output from libdoc to import Keywords as teststeps into their repository.
Language Server implementation could also benefit from this feature.

It would be very helpful if there could be a way to get the "rendered" docstrings into the <doc> node in the xml.

Libdoc always puts the raw docformat into the xml.
If the input format from library is HTML Libdoc put HTML into xml.

The attribute `format` of the node `keywordspec` is set to HTML. <keywordspec format="REST">

When now Libdoc would be able to set the output format to "rendered" HTML however the input format would be, it would be possible to import Keywords into other tools or show the docstrings in IDEs.

The HTML generation of libdoc based on the XML input file would still work, because HTML doc is a valid input now also.

kinde regards
René

Snooz82:
Hi Pekka,

i looked at the code and it would be easy if...

remember the discussion how to call this kind of format?

Option 1:
`python -m robot.libdoc --format XML --xmldocformat HTML SeleniumLIbrary SL.xml`
an extra option that only works if --format would be XML. (or extension...)

Option 2:
`python -m robot.libdoc --format HTMLinXML  SeleniumLIbrary SL.xml`
(we may call it different)
Two advantages. 
1. No dependencies between options.
2. supereasy to implement.

code:
``def LibdocWriter(format=None):
    format = (format or 'HTML').upper()
    if format == 'HTML':
        return LibdocHtmlWriter()
    if format == 'XML':
        return LibdocXmlWriter()
    raise DataError("Format must be either 'HTML' or 'XML', got '%s'." % format)``

here we could add an Argument to LibdocXmlWriter() like LibdocXmlWriter(output_doc_format='raw') or we could implement a third Writer.
i would prefer to add an argument to switch the output_doc_format because then we only need to add a ahnd full of line to the existing LibdocXmlWriter.

Opinions?

pekkaklarck:
1. I agree this would be a useful feature making it easier for external tools processing Libdoc's XML specs to handle them correctly.

2. Implementation in PR #3302 looks good in high level. Some tests and docs would obviously be needed and the are some things to be enhanced in the code as well.

[...]

PR #3302, opened 2019-09-10
Libdoc: Embedding HTMl doc into XML output format. #3301

Snooz82:


Snooz82:
Hi Pekka,

I finished the implementation. Documentation is lean but should be enough.
i added a few atests and extended some others. 

i would be happy if this could be reviewed quickly. ;-)
We need this feature for QS-Tag and want to show rendered Keyword Docs from Robot in our GUI.

kind regards
René

pekkaklarck:
In general looks very good. Quick comments:
- Docs should have a note this is new in 3.2.
- At least UG docs should be extended a little.
- There's some unnecessary whitespace changes (trailing spaces, unnecessary empty lines, possibly inconsistent newlines).
- The whole gven format should be validated in a one go. Initially accepting `XML:whatever` just complicates code elsewhere.
- I got some small comments to code otherwise but don't have time for them right now.

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -15,33 +15,51 @@
...
-        writer.element('namedargs', 'yes' if libdoc.named_args else 'no')
-        writer.element('doc', libdoc.doc)
-        self._write_keywords('init', libdoc.inits, writer)
-        self._write_keywords('kw', libdoc.keywords, writer)
-        writer.end('keywordspec')
-        writer.close()
-
-    def _write_keywords(self, type, keywords, writer):
+        self._formatter = DocToHtml(libdoc.doc_format) if self._html_doc_format else self.raw_format
+        self._writer = XmlWriter(outfile)
```
pekkaklarck: I don't like setting values to `self` when they could just be passed to functions as arguments. The latter approach worked fine earlier with `writer` and I don't see why it wouldn't work with `formatter` as well. Using `self` for this purpose feels like misusing object state and `writer.start(...)` also reads better than `self._writer.start(...)`.

Also, we try to keep the code PEP-8 compliant and that includes keeping lines less than 80 characters long. The line with `formatter` should either be split or a separate `DocFormatter` implemented and logic moved there.

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -15,33 +15,51 @@
...
+                self._writer.element('tag', tag)
+            self._writer.end('tags')
+            self._writer.end(type)
+
+    def _write_end(self):
+        self._writer.end('keywordspec')
+        self._writer.close()
+
+    def raw_format(self, doc):
+        return doc
```
pekkaklarck: This method doesn't belong to the public API of this class. I'd use `lambda doc: doc` where this is used instead (unless a separate `DocFormatter` would be implemented).

doc/userguide/src/SupportingTools/Libdoc.rst
```diff
@@ -41,9 +41,16 @@ Synopsis
...
+  -f, --format <html|xml|xml:html>
+                           Specifies whether to generate HTML or XML output.
                            If this options is not used, the format is got
                            from the extension of the output file.
+                           Embedding HTML documentation in XML output:
+                           If XML should have embedded HTML in <doc> as
+                           documentation, it is possible to configure
+                           this with ``XML:HTML``. This could be used if the XML
+                           files are used to be imported into other systems
+                           that require HTML text to present this to a human.
```
pekkaklarck: I don't think it's adequate to explain this kind of a feature only in the command line usage. Features should be documented so that the docs can be found easily.

The problem is that Libdoc documentation doesn't seem to have a good place to explain this. Probably the best place would be under the existing `Generating documentation` section, but that section should also be enhanced to explain HTML and XML formats a bit more thoroughly. I can do that if you want.

There should also be a note that this is a new feature in RF 3.2.

src/robot/libdoc.py
```diff
@@ -64,9 +64,16 @@
...
+ -f --format HTML|XML|XML:HTML
+                          Specifies whether to generate HTML or XML output.
                           If this options is not used, the format is got
                           from the extension of the output file.
+                          Embedding HTML documentation in XML output:
+                          If XML should have embedded HTML in <doc> as
+                          documentation, it is possible to configure 
+                          this with 'XML:HTML'. This could be used if the XML
+                          files are used to be imported into other systems
+                          that require HTML text to present this to a human.
```
pekkaklarck: Same comments about docs I wrote related to the User Guide documentation apply also to this help text. The help text doesn't need to be as thorough as User Guide, though.

src/robot/writer/filewriters.py
```diff
@@ -30,7 +30,7 @@ def FileWriter(context):
     """Creates and returns a ``FileWriter`` object.
 
     :param context: The type of the returned ``FileWriter`` is determined based
-        on ``context.format``. ``context`` is also passed to created writer.
+        on ``context.format``. ``context`` is also passed to created _writer.
```
pekkaklarck: This doesn't look relevant. Automatic refactoring by an IDE gone wrong?

src/robot/reporting/resultwriter.py
```diff
@@ -36,8 +36,8 @@ class ResultWriter(object):
     or enabled using ``settings`` or ``options`` passed to the
     :meth:`write_results` method. The latter is typically more convenient::
 
-        writer = ResultWriter(result)
-        writer.write_results(report='custom.html', log=None, xunit='xunit.xml')
+        _writer = ResultWriter(result)
+        _writer.write_results(report='custom.html', log=None, xunit='xunit.xml')
```
pekkaklarck: This doesn't look relevant. Automatic refactoring by an IDE gone wrong?

atest/robot/libdoc/doc_format.robot
```diff
@@ -67,6 +86,9 @@ Test Format In XML
...
     Keyword Doc Should Be    0    *bold* or <b>bold</b> http://example.com
 
-Format should be
-    [Arguments]    ${expected}
-    Element Attribute Should Be    ${LIBDOC}    format    ${expected}
+Test Format In XML:HTML
+    [Arguments]    ${expected}    ${cli}=    ${lib}=DocFormat.py
+    ${lib} =    Join Path    ${TESTDATADIR}    ${lib}
+    Run Libdoc And Parse Output     ${cli} ${lib}
+    Format should be    ${expected}
```
pekkaklarck: Isn't this format always HTML? If yes, it should be hard coded here and the `${expected}` argument removed.

atest/robot/libdoc/doc_format.robot
```diff
@@ -44,12 +44,31 @@ Format in XML
...
 
+Format from XML:HTML spec
+    [Template]    NONE
+    Run Libdoc    -F ROBOT --format XML:HTML ${TESTDATADIR}/DocFormat.py ${OUTXML}
+    Copy File    ${OUTXML}    ${OUTPREFIX}-2.xml
+    Test Format In XML:HTML    HTML    lib=${OUTPREFIX}-2.xml
+
+Compare HTML from XML:HTML
+    [Template]    NONE
+    Run Libdoc    -F ROBOT --format XML:HTML ${TESTDATADIR}/DocFormat.py ${OUTPREFIX}_xml_with_html.xml
```
pekkaklarck: I think you could use just `${OUTXML}` instead of `${OUTPREFIX}_xml_with_html.xml` here.

atest/robot/libdoc/doc_format.robot
```diff
@@ -44,12 +44,31 @@ Format in XML
...
+    Test Format In HTML    <b>bold</b> or &lt;b&gt;bold&lt;/b&gt; ${EXAMPLE LINK}    
+    ...                    ${EMPTY}    
+    ...                    ${OUTPREFIX}_xml_with_html.xml
```
pekkaklarck: Instead of passing `${EMPTY}` as the second argument, you could use the named argument syntax like `lib=<value>`.

src/robot/libdocpkg/xmlwriter.py
Snooz82: done

src/robot/writer/filewriters.py
Snooz82: correct

src/robot/reporting/resultwriter.py
Snooz82: correct

src/robot/libdoc.py
Snooz82: I think it is not disturbing the user, when it is very briefly explained here.
often Manpages are too lean. Imho.

i would let it here because in the moment it has no negative effect.

doc/userguide/src/SupportingTools/Libdoc.rst
Snooz82: I would really appreciate, if you could add this to Libdoc documentation.
I wanted to do this but as you mentioned i could not fit this into the existing structure.
I was hard for me to "restructure" your docs.

I would recomment only to add the "new in RF 3.2" comment only in the libdoc CLI doc.
I think it really overloads the manpage without adding value to the user who is already using 3.2 when he reads this information.

Snooz82: added "``XML:HTML`` is new in RF 3.2." into libdoc.rst

atest/robot/libdoc/doc_format.robot
Snooz82: thats correct.
i added the `${expected}` to the arguments to have a better self similarity.
That a person who tries to understand the test not always habe to look into Keyword definition.

If you insist, i can remove it.

atest/robot/libdoc/doc_format.robot
Snooz82: choosen this name for better understandability. 
Reader should understand that this OUTXML contains HTML as doc.

atest/robot/libdoc/doc_format.robot
Snooz82: done

atest/robot/libdoc/doc_format.robot
pekkaklarck: An argument that isn't really used feels, well, useless for me. It also gives an impression that XML:HMTL format would support some other doc formats than HTML. I'd prefer this to be changed but don't feel strongly. If you prefer this format, leave it like that.

atest/robot/libdoc/doc_format.robot
pekkaklarck: For me a generic and pre-existing name would feel a bit better but I don't feel strongly about it. We can leave it like this as well.

atest/robot/libdoc/doc_format.robot
pekkaklarck: An argument that isn't really used feels, well, useless for me. It also gives an impression that XML:HMTL format would support some other doc formats than HTML. I'd prefer this to be changed but don't feel strongly. If you prefer this format, leave it like that.

src/robot/libdoc.py
pekkaklarck: In my opinion it should be explicitly documented in the body of the help text. Usage should be short.

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -15,33 +15,50 @@
...
         writer.element('namedargs', 'yes' if libdoc.named_args else 'no')
-        writer.element('doc', libdoc.doc)
-        self._write_keywords('init', libdoc.inits, writer)
-        self._write_keywords('kw', libdoc.keywords, writer)
-        writer.end('keywordspec')
-        writer.close()
+        writer.element('doc', formatter(libdoc.doc))
 
-    def _write_keywords(self, type, keywords, writer):
+    def _write_keywords(self, writer, type, keywords, formatter):
```
pekkaklarck: Why was position of the `writer` argument changed? Especially having `writer` and `formatter` at the different ends of the signature feels a bit odd.

pekkaklarck:
Thanks @Snooz82! I'll enhance docs separately.

src/robot/libdocpkg/xmlwriter.py
bkhouri: why isn't `writer` an instance attribute?

pekkaklarck: That's actually how @Snooz82 implemented it in the beginning. :smile:

Snooz82: When coding object oriented i like to use instance attributes over handing over and returning stuff. As Pekka said, its kind of a personal style.

Because in Python the `self.` decreases the readability i have no real preferences here.

PR #3313, opened 2019-09-21
Travis CI: Lint code for Python syntax errors

cclauss:
This PR is like #3276 but with fewer tests so that it avoids several issues that were mentioned in the review of that PR.  It does not yet provide the array of platform tests that are discussed at https://github.com/robotframework/robotframework/issues/3275#issuecomment-523659647, most of them can be satisfied with future PRs that build on this one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: python
+install: pip install flake8
+script: flake8 . --count --select=E9,F63 --show-source --statistics


pekkaklarck:
Closing this because we are using GitHub actions, not Travis, for other CI needs. I'm still not how much benefits we'd got from linting either. Would be interesting to see a sample reports.

PR #3315, opened 2019-09-23
Update typos in the "Specifying library to import" section

Cuck00:


pekkaklarck:
Thanks! I think also the old wording was OK but agree with the new one being better.

PR #3327, opened 2019-10-04
Small changes to BuiltIn documentation

stdedos:
* Fix typo
* Add an example for checking `None`


pekkaklarck:
Thanks! I think the added example should be moved down a bit to have same examples first with `${x}` and `$x` syntax. I'll do that myself.

Issue #3182, opened 2019-05-10
Support variable index access like `${var}[2]` with all sequences (incl. strings and bytes)


pekkaklarck:
RF 3.1 added support for accessing variable items using syntax `${var}[item]` (incl. nested access `${var}[x][y][z]`, #2601) but *only* when variable is a list or a dictionary. This is a nice syntax and it would be good to extend it to cover all objects, including strings. Since RF 3.1. this syntax causes an error with other objects than lists or dicts (see e.g. #3028), so changing it to mean item access should not cause backwards compatibility issues. Should at the same time add support also for slice access `${var}[2:]` similarly as lists support (#2973).

One problem is that with custom objects we don't know should `${var}[42]` mean item access with `42` as a string or as an integer. We probably should just use strings with objects that aren't list-like and let users use `${var}[${42}]` is the really need an integer. Support with custom objects isn't likely that important anyway, the most important functionality is access characters and slices in strings.

-----

**UPDATE:** This was implemented so that accessing an item from a variable like `${var}[item]` so that `item` is an index now works with all sequences (incl. lists and bytes), not only with objects Robot considers list-like normally. The item is not considered an index *only* if the variable is a dictionary. In other words, the `${var}[item}` syntax works with custom objects *only* if these custom objects expect items to be indices.

pekkaklarck:
It's probably better to wait a bit more until we change the meaning of `${var}[item]` when `${var}` isn't a list or dictionary. This syntax currently causing an error makes it easier to spot places that still use this syntax to mean variable `${var}` followed with a literal string `[item]`. If we implement this functionality in RF 3.2, those who migrate from RF 3.0.x or older to it won't see an error but instead the behavior changes and understanding what happens can be hard.

pekkaklarck:
We have #3350 implementing this feature. Added this issue back to RF 3.2 scope.

pekkaklarck:
In the end decided to move this back to RF 3.3 for the following reasons:

- Although PR #3350 is in pretty good shape there are still changes to be done. See my review and comments after that for more information.

- I would like to have RF 3.2 out still this year (getting unlikely) or at least before [RoboCon](https://robocon.io) in mid January. This means I don't have time for further reviews even if PR #3350 would be developed further.

[...]

PR #3350, opened 2019-10-21
get item by index for byte like object

bollenn:
I would like to also allow bytes like objects to get item(s) by index.


pekkaklarck:
We have issue #3182 about item access syntax with all types. It was recently moved from RF 3.2 scope to RF 3.3 partly for backwards reasons and partly to get already overdue RF 3.2 ready sooner. If there's interest implementing it, we can still add it back to RF 3.2 scope. That would include:
- Support all types, not only bytes.
- Tests.
- Documentation.

bollenn:
Thanks for your feedback, for me having bytes support is very valuable, I will do some extra commits in here to add other types, lets see how far we make it....

pekkaklarck:
You don't need to explicitly test other types. Basically should just replace the current `is_list_like` with something like `is_iterable` (doesn't exist yet) to know should item be converted to integer (or slice). Then just try accessing the item directly and handle possible errors.

bollenn:
Updated the code according your recommendations, also fixed the tests which started failing because of my changes....

src/robot/variables/finders.py
```diff
@@ -73,9 +73,9 @@ def _get_match(self, variable):
 
     def _validate_value(self, value, identifier, name):
         if identifier == '@':
-            if not is_list_like(value):
-                raise VariableError("Value of variable '%s' is not list or "
-                                    "list-like." % name)
+            if not is_iterable(value):
+                raise VariableError("Value of variable '%s' is not iterable."
+                                    % name)
```
pekkaklarck: I don't think this particular change is a good idea. In my opinion we should still require variables that are used like `@{var}` to be list-like, not any iterable (i.e. not a string or a file). The original issue is about using something like `${var}[1]` and that working if if `${var}` is any iterable ought to be fine.

This change is also the reason for so many tests failing.

src/robot/variables/replacer.py
```diff
@@ -139,31 +139,31 @@ def _get_variable_item(self, match, value):
...
             if is_dict_like(value):
                 value = self._get_dict_variable_item(name, value, item)
-            elif is_list_like(value):
-                value = self._get_list_variable_item(name, value, item)
+            elif is_iterable(value):
+                value = self._get_iterable_variable_item(name, value, item)
             else:
                 raise VariableError(
-                    "Variable '%s' is %s, not list or dictionary, and thus "
+                    "Variable '%s' is %s, which is not iterable, and thus "
```
pekkaklarck: This error message could be enhanced a little. We explicitly support dictionaries and iterables, and while dicts actually are iterable, in this context they are accessed by key, not by index like iterables. Something like `not a dictionary or iterable` or perhaps `does not support item access` might be better. If you don't come up with something that you like, just leave it and I can think about it later.

src/robot/variables/finders.py
bollenn: reverted...

src/robot/variables/replacer.py
bollenn: Updated the error message in f47f4a582

bollenn:
@pekkaklarck I pulled master and fixed the merge conflicts, anything blocking this PR which I need to fix?

pekkaklarck:
Sorry, don't have time to review until at the end of this week or possibly only next week. I'll add #3182 back to RF 3.2 scope so this won't be forgotten.

bollenn:
Thx!

src/robot/variables/replacer.py
```diff
@@ -139,31 +139,31 @@ def _get_variable_item(self, match, value):
...
             if is_dict_like(value):
                 value = self._get_dict_variable_item(name, value, item)
-            elif is_list_like(value):
-                value = self._get_list_variable_item(name, value, item)
+            elif is_iterable(value):
+                value = self._get_iterable_variable_item(name, value, item)
             else:
                 raise VariableError(
-                    "Variable '%s' is %s, not list or dictionary, and thus "
+                    "Variable '%s' is %s, not a dictionary or iterable, and thus "
```
pekkaklarck: I'm slightly worried is "dictionary or iterable" (or the old "list or dictionary") clear for an average user. Should we consider rephrasing this somehow? One alternatively could be something like "Variable '%s' does not support key or index access", but I'm not sure is that much better. Opinions? Better ideas?

Another thing to consider is enhancing the error to tell how this problem can be avoided by escaping the `[item]` part. Should we add something like "To use '[%s]' as a literal value, it needs to be escaped like '\\[%s]'.", where `%s` are placeholders for `item`, to the end of the error message?

I know changing error messages mean extra work updating existing tests, but we've got lot of questions and bug reports since we added the `${var}[item]` syntax in RF 3.1 and I think making the situation more clear to users is a good thing.

src/robot/variables/replacer.py
```diff
@@ -139,31 +139,31 @@ def _get_variable_item(self, match, value):
...
 
-    def _get_list_variable_item(self, name, variable, index):
+    def _get_iterable_variable_item(self, name, variable, index):
         index = self.replace_string(index)
         try:
-            index = self._parse_list_variable_index(index, name[0] == '$')
+            index = self._parse_iterable_variable_index(index, name[0] == '$')
         except ValueError:
-            raise VariableError("List '%s' used with invalid index '%s'."
+            raise VariableError("Iterable '%s' used with invalid index '%s'."
```
pekkaklarck: This is another place where we could enhance the error message. The biggest problem is that you end up here if you have data like `${xpath}[predicate]` and didn't mean to use the item access syntax. Something like "To use '[%s]' as a literal value, it needs to be escaped like '\\[%s]'." might be a good addition. This was discussed already above, and possibly both situations could be handled by same code.

Another issue that the word "iterable" may not be that familiar for all users. Not sure what else to use, though, as "list" or even "list-like" aren't accurate here anymore.

pekkaklarck:
One thing to decide is how to handle custom objects implementing `__getitem__` when used like `${x}[item]`. The current implementation considers them iterable (`iter(x)` returns an iterable if `x` has `__getitem__`) and thus we convert `item` to an index (i.e. integer). This means that a custom object accepting keys won't work *unless* that object implements [Mapping](https://docs.python.org/library/collections.abc.html#collections.abc.Mapping) in which case it is considered a dictionary and there's no conversion to the `item`.

Should we care about custom objects having `__getitem__` that aren't list-like but not explicitly mappings? If not, the current implementation in this PR is fine. If yes, then we needed to figure out how to separate such objects from other iterables and that may not be easy. For me the current functionality is good enough, especially if nobody comes up with a simple way to different different iterables reliably. We can also revisit this in the future if there are concrete needs, although later changing this may cause backwards compatibility problems.

pekkaklarck:
I've been thinking this a bit more and think that we should convert the item to an index (or possible slice) *only* if the accessed variable is a [sequence](https://docs.python.org/3/glossary.html#term-sequence). With other objects we can then pass the item unchanged.

This means that sequences work nicely like `${x}[1]` or `${x}[1:]` as with the current implementation. An enhancement would be that if a custom object implements only `__getitem__`, it is possible to use it like `${x}[item]`. With the current implementation this would fail for `ValueError` when converting `item` to an index. If such a custom object would need the item as an index, it is possible to use something like `${x}[${1}]`.

src/robot/variables/replacer.py
JasperCraeghs: I am changing "not a dictionary or iterable" to "which is not subscriptable" and enhancing the error message as you suggested. Would this extension to the error message also be useful in case of a subscriptable variable that is not a sequence? I am not familiar with the use cases of `\[%s]`.

src/robot/variables/replacer.py
JasperCraeghs: I'm changing "iterable" to "sequence" and extending the error message as you suggested.

src/robot/variables/replacer.py
JasperCraeghs: Shouldn't `\[%s]` be added to the [user guide](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#escaping)?

src/robot/utils/__init__.py
```diff
@@ -67,7 +67,7 @@
 from .robottypes import (FALSE_STRINGS, Mapping, MutableMapping, TRUE_STRINGS,
                          is_bytes, is_dict_like, is_falsy, is_integer,
                          is_list_like, is_number, is_string, is_truthy,
-                         is_unicode, type_name, unicode)
+                         is_unicode, type_name, unicode, is_iterable)
```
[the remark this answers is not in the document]
JasperCraeghs: This has been resolved by 15d42ba7095a4af068849ba78d3c247280ea4ea9

bollenn:
Pulled master, fixed utests... ready for another review

pekkaklarck:
It's getting bit late to get this into RF 3.2. The change shouldn't be backwards incompatible and isn't that big so I'll give it a change. Due to my winter holiday I won't have time for a real review now, but I add the original issues to RF 3.2 scope so that this won't be forgotten.

src/robot/variables/replacer.py
```diff
@@ -137,46 +137,59 @@ def _get_variable_item(self, match, value):
...
-    def _get_list_variable_item(self, name, variable, index):
+    def _get_sequence_variable_item(self, name, variable, index):
         index = self.replace_string(index)
         try:
-            index = self._parse_list_variable_index(index, name[0] == '$')
+            index = self._parse_sequence_variable_index(index, name[0] == '$')
         except ValueError:
-            raise VariableError("List '%s' used with invalid index '%s'."
-                                % (name, index))
+            raise VariableError("Sequence '%s' used with invalid index '%s'. "
```
pekkaklarck: "Sequence" is better than "List" in this context but perhaps `type_name(variable)` would be even better. In here we needed to capitalize the returned word, though, and should take into account that e.g. `"MyClass".capitalize()` yields `"Myclass"` which isn't what we want. Probably would be best to add `capitalize=False` argument to `type_name()` to handle this stuff there, but I'm not certain should it be done as part of this PR or should it be done afterwards.

src/robot/variables/replacer.py
```diff
@@ -137,46 +137,59 @@ def _get_variable_item(self, match, value):
...
-    def _get_dict_variable_item(self, name, variable, key):
-        key = self.replace_scalar(key)
+    def _get_subscriptable_variable_item(self, name, variable, key):
+        if not isinstance(key, (int, slice)):
+            key = self.replace_scalar(key)
         try:
             return variable[key]
         except KeyError:
-            raise VariableError("Dictionary '%s' has no key '%s'."
+            raise VariableError("Subscriptable '%s' has no key '%s'."
```
pekkaklarck: Yet another place where to use `type_name(variable, capitalize=True)`. In addition to that, perhaps it would be better to change `key` to `item`. Both ought to be OK with dictionaries, but the latter works better with custom objects.

src/robot/variables/replacer.py
```diff
@@ -137,46 +137,59 @@ def _get_variable_item(self, match, value):
...
+            if not isinstance(key, (int, slice)):
+                try:
+                    key = self._parse_sequence_variable_index(
+                        key, name[0] == '$')
+                except:
+                    pass
+                else:
+                    return self._get_subscriptable_variable_item(
+                        name, variable, key)
+            raise VariableError("Subscriptable '%s' used with invalid key: %s"
```
pekkaklarck: Why is this done? Is the idea to automatically convert strings like "2" to an integers and "2:" to a slice? If yes, I don't think that's a good idea. We already do that with sequences and don't see much benefits with other objects. It's already now possible to do e.g. `${dict}[${2}]` to use non-string keys. That works also with other types than integers such as `${dict}[${1.2}]` and `${dict}[${True}]` while this code only supports integers and slices.

If there is a good use case this feature, or I have understood the code wrong, then  the plain `except:` should still be replaced with something that doesn't blindly swallow all exceptions.

src/robot/variables/search.py
```diff
@@ -83,7 +83,7 @@ def __nonzero__(self):
     def __unicode__(self):
         if not self:
             return '<no match>'
-        items = ''.join('[%s]' % i for i in self.item) if self.items else ''
+        items = ''.join('[%s]' % i for i in self.items) if self.items else ''
```
pekkaklarck: Good catch, there's clearly a bug in the existing code! Luckily this is new code in RF 3.2.

utest/variables/test_variables.py
```diff
@@ -222,7 +226,7 @@ def test_math_with_internal_vars_does_not_work_if_first_var_is_float(self):
         assert_raises(VariableError, self.varz.replace_scalar, '${${1.1} * ${2}}')
         assert_raises(VariableError, self.varz.replace_scalar, '${${1.1}/${2}}')
 
-    def test_list_variable_as_scalar(self):
+    def var_variable_as_scalar(self):
```
pekkaklarck: This looks like an accidental change.

src/robot/variables/replacer.py
pekkaklarck: The above said, error handling would need to be changed a bit. This is because with custom objects `variable[key]` can fail with _any_ exception, not only with `TypeError`. Perhaps something like this:

```python
except:
    raise VariableError("Accessing item '%s' from %s '%s' failed: %s"
                        % (key, type_name(variable), name, get_error_message())
```

Notice that the plain `except:` is fine here because `get_error_message()` automatically re-raises exceptions such as `SystemExit` that we shouldn't catch.

src/robot/variables/replacer.py
pekkaklarck: Hmm, perhaps I should enhance `type_name()` in this regard separately. It could then be used here after merging changes from master.

src/robot/variables/replacer.py
JasperCraeghs: The idea was to support integers, slices and non-string keys for custom subscriptable classes, which is what this implementation is capable of. I now understand that you don't want this. I have simplified this implementation according to your suggestions. 2afca766be3f4a2058f4225895890babe4999c31

JasperCraeghs: I have simplified the implementation by using your suggestions in 2afca766be3f4a2058f4225895890babe4999c31. Updated documentation in 4ef6eb43757b0bfc0c38c31535a8504fa62c266b.

src/robot/variables/replacer.py
```diff
@@ -137,46 +137,59 @@ def _get_variable_item(self, match, value):
...
                         "is deprecated. Use '$%s' instead." % (var, var[1:]))
         for item in match.items:
-            if is_dict_like(value):
-                value = self._get_dict_variable_item(name, value, item)
-            elif is_list_like(value):
-                value = self._get_list_variable_item(name, value, item)
+            if is_sequence(value):
+                value = self._get_sequence_variable_item(name, value, item)
+            elif is_subscriptable(value):
+                value = self._get_subscriptable_variable_item(name, value, item)
```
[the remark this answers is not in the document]
JasperCraeghs: Right, but after changing to the simplified implementation in 2afca766be3f4a2058f4225895890babe4999c31 custom classes that don't inherit from `Sequence` are no longer subscriptable by index/slice (as string). This is a problem for class `ExecutionErrors`, for example. Acceptance tests fail because of this. I am in favor of reverting the aforementioned commit and  tweaking error handling. What do you think?

src/robot/variables/replacer.py
JasperCraeghs: The tests no longer pass after this simplified implementation. See explanation in https://github.com/robotframework/robotframework/pull/3350#discussion_r386965038

src/robot/variables/replacer.py
JasperCraeghs: Done in d7110621c16a371e72b541b70f01cbea7f0f4c8e

src/robot/variables/replacer.py
JasperCraeghs: Done in d7110621c16a371e72b541b70f01cbea7f0f4c8e

src/robot/variables/replacer.py
```diff
@@ -137,46 +137,59 @@ def _get_variable_item(self, match, value):
...
+            raise VariableError("Sequence '%s' used with invalid index '%s'. "
+                                "To use '[%s]' as a literal value, it needs "
+                                "to be escaped like '\\[%s]'."
+                                % (name, index, index, index))
         try:
             return variable[index]
         except IndexError:
-            raise VariableError("List '%s' has no item in index %d."
+            raise VariableError("Sequence '%s' has no item in index %d."
                                 % (name, index))
```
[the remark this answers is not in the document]
JasperCraeghs: Done in d7110621c16a371e72b541b70f01cbea7f0f4c8e

src/robot/variables/replacer.py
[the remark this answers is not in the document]
JasperCraeghs: Done this in d7110621c16a371e72b541b70f01cbea7f0f4c8e

src/robot/variables/replacer.py
JasperCraeghs: I refactored the first implementation in d7110621c16a371e72b541b70f01cbea7f0f4c8e. Existing tests pass and I added some additional unit tests.

utest/variables/test_variables.py
JasperCraeghs: Reverted by 312360d873a4ab47b366f79fe9bf9a288f47fdea

src/robot/variables/replacer.py
```diff
@@ -137,46 +137,65 @@ def _get_variable_item(self, match, value):
...
+            raise VariableError("%s '%s' has no item '%s'."
+                                % (type_name(variable, capitalize=True),
+                                   name, item))
+        except Exception as err:
+            # Try to treat custom class as a Sequence but don't raise error
+            # on failure
+            try:
+                return self._get_sequence_variable_item(name, variable, item)
+            except:
+                pass
```
pekkaklarck: I don't think it is a good idea to first try dictionary access and then sequence access if that fails. It's just too complicated and error prone compared to the benefits. Better to decide which access to use based on the variable type. Just need to decide what algorithm to use for that. Let's discuss more about that in the comments of the original issue or on Slack.

doc/userguide/src/CreatingTestData/Variables.rst
```diff
@@ -284,62 +284,67 @@ are imports, setups and teardowns where dictionaries can be used as arguments.
...
+       Keyword    ${SEQUENCE}[2:-1]
 
    Step
-       Keyword    ${LIST}[::2]
-       Keyword    ${LIST}[1:-1:10]
+       Keyword    ${SEQUENCE}[::2]
+       Keyword    ${SEQUENCE}[1:-1:10]
 
 .. note:: The slice syntax is new in Robot Framework 3.1 and does not work
           with the old `@{var}[index]` syntax.
```
pekkaklarck: Need a note here about accessing items from strings/bytes/others being new in RF 3.2.

src/robot/utils/robottypes.py
```diff
@@ -57,3 +57,7 @@ def is_truthy(item):
 def is_falsy(item):
     """Opposite of :func:`is_truthy`."""
     return not is_truthy(item)
+
+
+def is_subscriptable(item):
+    return hasattr(item, '__getitem__')
```
pekkaklarck: I'm not certain is this helper needed or should we just inline it. Perhaps it will be useful in the future so leaving it makes sense.

doc/userguide/src/CreatingTestData/Variables.rst
```diff
@@ -284,62 +284,67 @@ are imports, setups and teardowns where dictionaries can be used as arguments.
...
```
pekkaklarck: Need a note here about accessing items from strings/bytes/others being new in RF 3.2.

src/robot/variables/replacer.py
```diff
@@ -139,39 +139,44 @@ def _get_variable_item(self, match, value):
...
+                value = self._get_sequence_variable_item(name, value, item)
             else:
                 raise VariableError(
-                    "Variable '%s' is %s, not list or dictionary, and thus "
-                    "accessing item '%s' from it is not possible."
-                    % (name, type_name(value), item)
+                    "Variable '%s' is %s, which is not subscriptable, and "
+                    "thus accessing item '%s' from it is not possible. To use "
+                    "'[%s]' as a literal value, it needs to be escaped like "
+                    "'\\[%s]'." % (name, type_name(value), item, item, item)
```
pekkaklarck: I like the enhanced message with explanation about escaping.

src/robot/variables/replacer.py
```diff
@@ -139,39 +139,44 @@ def _get_variable_item(self, match, value):
...
         if ':' not in index:
             return int(index)
         if index.count(':') > 2 or not support_slice:
             raise ValueError
         return slice(*[int(i) if i else None for i in index.split(':')])
 
-    def _get_dict_variable_item(self, name, variable, key):
-        key = self.replace_scalar(key)
+    def _get_dict_variable_item(self, name, variable, item):
+        key = self.replace_scalar(item)
```
pekkaklarck: I like `key` better than `item` in this context.

utest/variables/test_variables.py
```diff
@@ -15,12 +15,16 @@
...
+class PythonObject(object):
     def __init__(self, a, b):
         self.a = a
         self.b = b
+    def __getitem__(self, index):
+        return (self.a, self.b)[index]
     def __str__(self):
         return '(%s, %s)' % (self.a, self.b)
+    def __len__(self):
+        return (self.a, self.b).__len__()
```
pekkaklarck: Why not just `return 2` here?

JasperCraeghs: Good point, changed by 898921a8052d058bec881ee807e3585c584819a1

src/robot/variables/replacer.py
JasperCraeghs: Changed by 6548ba3a4a32cc367d582653c16aca8f6f502ced

doc/userguide/src/CreatingTestData/Variables.rst
JasperCraeghs: Note added in b0000035fa46c7a9519c8e4deddfebf1b869afae

doc/userguide/src/CreatingTestData/Variables.rst
```diff
@@ -284,62 +284,70 @@ are imports, setups and teardowns where dictionaries can be used as arguments.
...
+''''''''''''''''''''''''
 
-It is possible to access a certain item of a list variable with the syntax
-`${var}[index]`, where `index` is the index of the selected value. Indices
-start from zero, negative indices can be used to access items from the end,
-and trying to access an item with too large an index causes an error.
-Indices are automatically converted to integers, and it is also possible to
-use variables as indices. List items accessed in this manner can be used
-similarly as scalar variables.
+It is possible to access a certain item of a `sequence`__ variable (e.g. list,
```
pekkaklarck: Perhaps `a variable containing a sequence` would be a bit more clear than `a sequence variable`, especially when we use the term `list variable` in context where strings aren't allowed.

doc/userguide/src/CreatingTestData/Variables.rst
```diff
@@ -284,62 +284,70 @@ are imports, setups and teardowns where dictionaries can be used as arguments.
...
    Step
-       Keyword    ${LIST}[::2]
-       Keyword    ${LIST}[1:-1:10]
+       Keyword    ${SEQUENCE}[::2]
+       Keyword    ${SEQUENCE}[1:-1:10]
 
 .. note:: The slice syntax is new in Robot Framework 3.1 and does not work
           with the old `@{var}[index]` syntax.
 
+.. note:: Accessing items from strings/bytes/others is new in Robot Framework
```
pekkaklarck: `from strings/bytes/others` is a bit confusing, especially the `others` part. Perhaps something like `from other objects than lists and tuples, for example strings and bytes,` would be better.

pekkaklarck:
This looks great! I added two comments about the docs in my latest review but I can easily fix those myself after merging this. Thanks for your great work @JasperCraeghs and @bollenn!

Issue #3278, opened 2019-08-23
Variable type conversion


pekkaklarck:
RF 3.1 added automatic argument conversion with library keywords using function annotations (#2890), default values (#2932), and some other means. That has worked very well, and it would be nice if something like that would work also with variables defined in test data. We probably could re-use the underlying argument conversion logic which ought to make it relatively easy to implement this. We also got a similar request related to variables passed from the command line #2946.

Automatically converting numbers like `42` and `3.14` to actual integers/floats would be handy, even when we already support them as variables like `${42}` and `${3.14}`. Being able to use datatime objects, bytes, and other types supported by the argument conversion that don't have direct variable alternatives would naturally be even more useful (but also less commonly needed). I would expect this to be most useful with lists and dictionaries and especially with nested list and dicts that cannot easily be constructed in Robot's data.

My proposed syntax for this functionality is supporting "type postfix" like `:int` in variable names like this:

```robotframework
*** Variables ***
${COUNT:int}         42
${GOOD IDEA:bool}    Yes    # Converted to Boolean True
${ANIMALS:list}      ['cat', 'dog', 'horse']    # Python syntax
${NESTED:dict}       {'a': 1, 'b': [{'x': 2}, {'x': 3}]}    # Great with JSON
```

This syntax is derived from Python's function annotation syntax and also the same syntax that was decided to be used with command line variables in #2946. Created variable names would naturally omit the type suffix and look like `${COUNT}` and `${GOOD IDEA}`.

In addition to being available in the variable section, the new syntax should also be available also when creating variable otherwise, including in keyword arguments:

```robotframework
*** Keywords ***
Example
    [Arguments]    ${count:int}
    ${nested:dict} =    Set Variable    {'a': ${count}, 'b': [{'x': 2}, {'x': 3}]}
    Set Test Variable    ${DATE:datetime}     2019-08-23 12:03:42
```

There are some design decision still to be made like should `${X:dict}` create a `DotDict` instance that allows `${X.key}` syntax and how to handle possible backwards incompatibility issues. For most parts the design looks pretty good, though, and I think this would be a valuable addition to Robot Framework. If there's enough interest, we can consider adding this already to RF 3.2 scope. To show your interest, vote with :+1: or :-1: and/or add comments.

---

**UPDATE:** The final syntax is mostly the same as above, but there needs to be a space after the colon like `${COUNT: int}`. The motivation is making the enhancement less backwards incompatible.

pekkaklarck:
[...]

Issue #2068, opened 2015-07-21
Support automatic argument conversion with dynamic libraries


baudren:
As discussed on the [mailing list](https://groups.google.com/forum/#!topic/robotframework-users/ND1v3wp60m8), it would be nice to define a new (probably optional) keyword for dynamic libraries, in order to specify the type of the arguments needed for a keyword (as a list of strings, maybe?).

Even though RF performs type coercion automatically when calling directly Java libraries, it is not the case when calling a Jython library that itself wraps some Java libraries.

The look-up inside the Java library would use the `getDeclaredMethods()` and `getParamterTypes()`, with a small conversion to perform between, for instance, `double` and `float`.

RF could expect as an output something like `['int', 'float', 'str', 'str']`, to perform a type cast if the method `get_keyword_argument_types` is present on the API, probably before the `run_keyword` is actually called.

I don't know what to do for `list` or `dict` items, though.

I don't know the codebase much. Would [this class](https://github.com/robotframework/robotframework/blob/f9dd7064c5db95cf2e8a7149a06f418ebde0007f/src/robot/running/runner.py) be the place to put it? Or is there a place I missed which calls the external keywords? I could give a shot at a Pull-Request with some minimal starting help, if no one else has time.

Thanks in advance,

pekkaklarck:
Just adding a new method to the dynamic API isn't enough because that type information needs to be used too. The existing argument coercion logic used by Java libraries probably can be reused, but it anyway needs to be hooked up with other libraries.

I also consider adding argument coercion to the normal Python library API more important than adding it to the dynamic API. That ought to be a bit easier too, because there we could just use an attribute like `robot_types` or `robot_argument_types` to specify arguments. That needs a separate issue, though.

baudren:
I agree that adding argument coercion for static Python/Jython library would be nice too - though for my current project, I need the dynamic bit. I can open the issue to trigger a discussion, though, if you want.

I am aware that the type information needs to be used, I just don't know where it would be more appropriate (or I misunderstood your sentence). I am searching for the place where it would make more sense to add it. Since dynamic libraries keywords are already called through their `run_keyword` method, just there would make sense, but which class exactly is reponsible for this?

pekkaklarck:
[...]

PR #3354, opened 2019-10-25
Variable type conversion in arguments (#3278)

yamatoRT:
Hi,
This PR implements a part of functionnality of issues #3278 

I implement only type conversion for arguments of keyword.
I reuse same tests of python function annotations, except for the vargs and the kwargs. Moreover, i remove tests for decorators.
Don't hesitate to challenge the name of class or method, the english is not my  best language.

Best Regards,


atest/testdata/keywords/type_conversion/RobotKeywordAnnotationsWithAliases.robot
```diff
@@ -0,0 +1,87 @@
...
+Double
+    [Arguments]    ${argument: 'Float'}    ${expected}=${Empty}
+    Validate Type     ${argument}    ${expected}
+
+Decimal
+    [Arguments]    ${argument: 'DECIMAL'}    ${expected}=${Empty}
+    Validate Type     ${argument}    ${expected}
+
+Boolean
+    [Arguments]    ${argument: 'Boolean' }    ${expected}=${Empty}
```
miktuy: extra space after ```'Boolean'``` here and in line 34

src/robot/running/arguments/argumentparser.py
```diff
@@ -16,9 +16,11 @@
 from robot.errors import DataError
 from robot.utils import JYTHON, PY_VERSION, PY2
 from robot.variables import is_dict_var, is_list_var, is_scalar_var
+from pydoc import locate
```
miktuy: PEP8 import order
```
Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.
You should put a blank line between each group of imports.
```

pekkaklarck:
First of all, thanks for the PR! This is pretty big task to implement and I've been thinking it's too big for the already overdue RF 3.2, but including it is still possible if we can get this PR into a good enough shape. I won't have time for a full review until I have got RF 3.2 alpha 1 ready, but below are some general comments about this PR after a very quick look:

1. This PR only adds support for argument conversion. Conversion in the Variables section and also when assigning variables based on keyword return values are needed as well. Probably a good idea to get this first feature ready first.

2. I don't think we need to support syntax like `${arg: 'int'}`. Just supporting `${arg: int}` (space being optional) ought to be enough. I also don't see reasons why to support Python's `typing` style usages like `${arg:Mapping[str, int]}` either. Just supporting same string names and aliases that work with the dynamic library API (#2068) ought to be enough.

3. Implementation looks pretty complicated. Most probably that's due to the support for formats like `${arg:Mapping[str, int]}`. I think this could be done pretty easily in [argumentparser.py](https://github.com/robotframework/robotframework/blob/master/src/robot/running/arguments/argumentparser.py): `PythonArgumentParser` already does it with Python based keywords and `UserKeywordArgumentParser` could probably do it with user keywords.

yamatoRT:
Hi,
If we don't need the typing style usages, the code will be simplied.
I will rework my code with your comments.

atest/testdata/keywords/type_conversion/RobotKeywordAnnotationsWithAliases.robot
yamatoRT: I removed extra space.

src/robot/running/arguments/argumentparser.py
yamatoRT: Thank for comments. I change import order

yamatoRT:
I removed typing style usages, abc types, enum and bytesstring type. The other types is already convertible by **TypeConverter**.
Too I put my implementation in  **argumentparser.py** and I removed **typingresolver.py**.

yamatoRT:
I am closing this PR because i will propose new implementation for the issue #3278 who work with variable

Issue #3366, opened 2019-11-04
`Run Process:` Ignore timeout if it is zero, negative or string `None`


stdedos:
Make `timeout` named argument accept the same values as keyword `[Timeout]` does.

Any falsy, zero, or negative values are treated the same way as `[Timeout]` (i.e. ignored aka infinity)

See also #3365

pekkaklarck:
Ignoring timeouts when they are `None` or not positive is a good idea, but I don't think supporting `timeout=False` makes much sense.

pekkaklarck:
Thanks to PR #3367 by @stdedos this has now been implemented. I raised priority from low to medium because there earlier weren't any clean way to specify a timeout value that means timeout shouldn't be activated.

Issue #3365, opened 2019-11-04
Document that zero and negative test/keyword timeout is ignored


pekkaklarck:
Test and keyword timeouts are ignored if they are zero or negative. This doesn't seem to be documented in the User Guide nor adequately tested.

PR #3367, opened 2019-11-04
Normalize `Run Process | ... | timeout=0`

stdedos:
Make `timeout` named argument accept the same values as keyword `[Timeout]` does.

`None` (case-insensitively), falsy, zero or negative values are treated
the same way as `[Timeout]` (i.e. ignored aka infinity)
This is in addition to `${None}` being already ignored.

See also robotframework/robotframework#3365

Resolves robotframework/robotframework#3366

### DIFF
--- a/atest/testdata/standard_libraries/process/run_process_with_timeout.robot
+++ b/atest/testdata/standard_libraries/process/run_process_with_timeout.robot
@@ -4,10 +4,30 @@ Suite Teardown    Remove Files    ${STDOUT}    ${STDERR}
 
 *** Variables ***
 @{COMMAND}        python    ${CURDIR}/files/timeout.py
+${ZERO_TIMEOUT}   0
+${NONE_TIMEOUT}   nONe
+${NEG_TIMEOUT}    -2 hours
 
 *** Test Cases ***
 Finish before timeout
     ${result} =    Run Process    @{COMMAND}
+    Should not be terminated    ${result}
+
+Finish before timeout with nONe timeout
+    ${result} =    Run Process    @{COMMAND}    timeout=${NONE_TIMEOUT}
+    Should not be terminated    ${result}
+
+Finish before timeout with zero timeout
+    ${result} =    Run Process    @{COMMAND}    timeout=${ZERO_TIMEOUT}
+    Should not be terminated    ${result}
+
+Finish before timeout with empty timeout
+    [Documentation]   Verifying that backwards compatibility is honored
+    ${result} =    Run Process    @{COMMAND}    timeout=${EMPTY}
+    Should not be terminated    ${result}
+
+Finish before timeout with negative timeout
+    ${result} =    Run Process    @{COMMAND}    timeout=${NEG_TIMEOUT}
     Should not be terminated    ${result}
 
 On timeout process is terminated by default (w/ default streams)
--- a/src/robot/libraries/Process.py
+++ b/src/robot/libraries/Process.py
@@ -407,7 +407,9 @@ class Process(object):
         given in
         [http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#time-format|
         various time formats] supported by Robot Framework, for example, ``42``,
-        ``42 s``, or ``1 minute 30 seconds``.
+        ``42 s``, or ``1 minute 30 seconds``. Starting from  Robot Framework 3.2
+        ``None``, zero, and negative values is same as not specifying timeout at
+        all.
 
         ``on_timeout`` defines what to do if the timeout occurs. Possible values
         and corresponding actions are explained in the table below. Notice
@@ -439,11 +441,15 @@ class Process(object):
         | ${result} =                 | Wait For Process | timeout=1min 30s | on_timeout=kill |
         | Process Should Be Stopped   |                  |                  |
         | Should Be Equal As Integers | ${result.rc}     | -9               |
+
+        Ignoring Falsy, negative, and zero values is new in Robot Framework 3.2.
         """
         process = self._processes[handle]
         logger.info('Waiting for process to complete.')
-        if timeout:
-            timeout = timestr_to_secs(timeout)
+        if not (is_truthy(timeout) and timeout):
+            timeout = 0
+        timeout = timestr_to_secs(timeout)
+        if timeout > 0:
             if not self._process_is_stopped(process, timeout):
                 logger.info('Process did not complete in %s.'
                             % secs_to_timestr(timeout))
--- a/atest/robot/standard_libraries/process/run_process_with_timeout.robot
+++ b/atest/robot/standard_libraries/process/run_process_with_timeout.robot
@@ -8,6 +8,21 @@ Finish before timeout
     Check Log Message    ${tc.kws[0].msgs[1]}    Waiting for process to complete.
     Check Log Message    ${tc.kws[0].msgs[2]}    Process completed.
 
+Finish before timeout with zero timeout
+    ${tc} =    Check Test Case    ${TESTNAME}
+    Check Log Message    ${tc.kws[0].msgs[1]}    Waiting for process to complete.
+    Check Log Message    ${tc.kws[0].msgs[2]}    Process completed.
+
+Finish before timeout with nONe timeout
+    ${tc} =    Check Test Case    ${TESTNAME}
+    Check Log Message    ${tc.kws[0].msgs[1]}    Waiting for process to complete.
+    Check Log Message    ${tc.kws[0].msgs[2]}    Process completed.
+
+Finish before timeout with negative timeout
+    ${tc} =    Check Test Case    ${TESTNAME}
+    Check Log Message    ${tc.kws[0].msgs[1]}    Waiting for process to complete.
+    Check Log Message    ${tc.kws[0].msgs[2]}    Process completed.
+
 On timeout process is terminated by default (w/ default streams)
     ${tc} =    Check Test Case    ${TESTNAME}
     Check Log Message    ${tc.kws[0].msgs[1]}    Waiting for process to complete.


src/robot/libraries/Process.py
```diff
@@ -407,7 +407,10 @@ def wait_for_process(self, handle=None, timeout=None, on_timeout='continue'):
         given in
         [http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#time-format|
         various time formats] supported by Robot Framework, for example, ``42``,
-        ``42 s``, or ``1 minute 30 seconds``.
+        ``42 s``, or ``1 minute 30 seconds``. ``NONE``, ``0``, or ``0s`` values
+        set timeout to be infinite (as shown
+        [https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#test-case-timeout|here])
+        ``-int`` values are ignored (i.e. set timeout to be infinite).
```
pekkaklarck: This somehow sounds overly complicated and referring to the User Guide is not necessary. Something like this could work better:

> ``NONE`` (case-insensitively) is same as not specifying timeout at all. Starting from Robot Framework 3.2 also zero and negative timeout values are ignored.

src/robot/libraries/Process.py
```diff
@@ -439,11 +442,15 @@ def wait_for_process(self, handle=None, timeout=None, on_timeout='continue'):
         | ${result} =                 | Wait For Process | timeout=1min 30s | on_timeout=kill |
         | Process Should Be Stopped   |                  |                  |
         | Should Be Equal As Integers | ${result.rc}     | -9               |
+        | # Never-ending process      |                  |                  |
+        | ${result} =                 | Wait For Process | timeout=0        | on_timeout=kill |
+        | Process Should Be Stopped   | # Will return ``Fail``              |
```
pekkaklarck: I don't think this is a useful example. It works exactly the same way as not using a timeout at all. I don't see why you'd ever use `timeout=0` or `timeout=NONE` in the first place and having such an example would just cause confusion. Something like `timeout=${timeout}` where the variable value may be `NONE` something else that disables the timeout could be useful, but such an example is too complicated here. In my opinion the earlier explanation about values that are ignored is enough.

src/robot/libraries/Process.py
```diff
@@ -439,11 +442,15 @@ def wait_for_process(self, handle=None, timeout=None, on_timeout='continue'):
...
         """
         process = self._processes[handle]
         logger.info('Waiting for process to complete.')
-        if timeout:
-            timeout = timestr_to_secs(timeout)
+        secs = timestr_to_secs(timeout)
+        if is_truthy(timeout) and secs > 0:
+            timeout = secs
```
pekkaklarck: I'd prefer this:
```python
timeout = timestr_to_secs(timeout) if timeout else 0
if timeout > 0:
    # ...
```

stdedos: Unfortunately, this won't handle

> `NONE` (case-insensitively) is same as not specifying timeout at all.

I'll merge your proposal with my changes to support the aforementioned statement.

src/robot/libraries/Process.py
```diff
@@ -439,11 +440,15 @@ def wait_for_process(self, handle=None, timeout=None, on_timeout='continue'):
...
         | Process Should Be Stopped   |                  |                  |
         | Should Be Equal As Integers | ${result.rc}     | -9               |
+
+        New in Robot Framework ``v3.2``: Falsy, negative, and zero values are also ignored.
         """
         process = self._processes[handle]
         logger.info('Waiting for process to complete.')
-        if timeout:
-            timeout = timestr_to_secs(timeout)
+        if not(is_truthy(timeout) and timeout):
```
pekkaklarck: I don't think ignoring false values in general is a good idea. Specifying `timeout=None` makes sense but `timeout=False` does not. Perhaps creating a helper method like this would be a good idea:

```python
def _timeout_to_secs(self, timeout):
    if not timeout or (is_string(timeout) and timeout.upper() == 'NONE'):
        return 0
    return timestr_to_secs(timeout)
```

Another possibility would be enhancing `timestr_to_secs` to handle Python `None` and string `NONE` (case-insensitively) so that they mean 0. The more than I think about it, the more I like it.

Notice also that our type conversion should actually convert string `NONE` to a real Python `None` with `wait_for_process` based on the argument default value. That doesn't work with `run_process`, though, because timeout is given to it as part of `**configuration` and thus doesn't have a default.

Finally, `not(...)` is not good Python style and should be written as `not (...)` instead.

src/robot/libraries/Process.py
```diff
@@ -407,7 +407,8 @@ def wait_for_process(self, handle=None, timeout=None, on_timeout='continue'):
         given in
         [http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#time-format|
         various time formats] supported by Robot Framework, for example, ``42``,
-        ``42 s``, or ``1 minute 30 seconds``.
+        ``42 s``, or ``1 minute 30 seconds``. ``${None}``, falsy,
+        zero, and negative values is same as not specifying timeout at all.
```
pekkaklarck: - This needs to be changed if we don't accept false values.
- `None` should be used here instead of more complicated `${None}`.
- Could rephrase this like `Starting from Robot Framework 3.2, ...` to avoid the need for a separate note about versions.

src/robot/libraries/Process.py
```diff
@@ -439,11 +440,15 @@ def wait_for_process(self, handle=None, timeout=None, on_timeout='continue'):
         | ${result} =                 | Wait For Process | timeout=1min 30s | on_timeout=kill |
         | Process Should Be Stopped   |                  |                  |
         | Should Be Equal As Integers | ${result.rc}     | -9               |
+
+        New in Robot Framework ``v3.2``: Falsy, negative, and zero values are also ignored.
```
pekkaklarck: Great that you added a note about version when functionality was enhanced. This is slightly different style than we've used elsewhere with such notes, though. As mentioned above, this info could also be explained elsewhere.

src/robot/libraries/Process.py
stdedos: I don't see how your proposed code treats `False` differently :confused: 

`timeout=False` may be questionable, but my code handles also e.g. `timeout='0'` or `timeout='NO'`

(I am sorry for the text-spam. But I cannot attach a file)
```python
import sys

from robot.utils import timestr_to_secs, is_truthy, FALSE_STRINGS, is_string


def uut(timeout):
    if not(is_truthy(timeout) and timeout):
        timeout = 0
    timeout = timestr_to_secs(timeout)

    return timeout


def test_uut():
    print('uut(%s) == 0' % False, file=sys.__stdout__)
    assert uut(False) == 0
    print('uut(%s) == 0' % None, file=sys.__stdout__)
    assert uut(None) == 0
    print('uut(%s) == 0' % 0, file=sys.__stdout__)
    assert uut(0) == 0

    for i in FALSE_STRINGS:
        print('uut(%s) == 0' % i, file=sys.__stdout__)
        assert uut(i) == 0
        print('uut(%s) == 0' % i.title(), file=sys.__stdout__)
        assert uut(i.title()) == 0


def _timeout_to_secs(timeout):
    if not timeout or (is_string(timeout) and timeout.upper() == 'NONE'):
        return 0
    return timestr_to_secs(timeout)


def test_pk():
    print('_timeout_to_secs(%s) == 0' % False, file=sys.__stdout__)
    assert _timeout_to_secs(False) == 0
    print('_timeout_to_secs(%s) == 0' % None, file=sys.__stdout__)
    assert _timeout_to_secs(None) == 0
    print('_timeout_to_secs(%s) == 0' % 0, file=sys.__stdout__)
    assert _timeout_to_secs(0) == 0

    for i in FALSE_STRINGS:
        print('_timeout_to_secs(%s) == 0' % i, file=sys.__stdout__)
        assert _timeout_to_secs(i) == 0
        print('_timeout_to_secs(%s) == 0' % i.title(), file=sys.__stdout__)
        assert _timeout_to_secs(i.title()) == 0

#  pytest -rA scratch_6.py
# ========================= test session starts ==========================
# platform linux -- Python 3.7.5, pytest-5.3.0, py-1.8.0, pluggy-0.13.1
# rootdir: /home/sntentos
# collected 2 items
#
# ../../../.PyCharmCE2019.3/config/scratches/scratch_6.py .F                                                                                                                                                                                               [100%]
#
# =============================== FAILURES ===============================
# _______________________________ test_pk ________________________________
#
#     def test_pk():
#         print('_timeout_to_secs(%s) == 0' % False, file=sys.__stdout__)
#         assert _timeout_to_secs(False) == 0
#         print('_timeout_to_secs(%s) == 0' % None, file=sys.__stdout__)
#         assert _timeout_to_secs(None) == 0
#         print('_timeout_to_secs(%s) == 0' % 0, file=sys.__stdout__)
#         assert _timeout_to_secs(0) == 0
#
#         for i in FALSE_STRINGS:
#             print('_timeout_to_secs(%s) == 0' % i, file=sys.__stdout__)
# >           assert _timeout_to_secs(i) == 0
#
# ../../../.PyCharmCE2019.3/config/scratches/scratch_6.py:45:
# _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
# ../../../.PyCharmCE2019.3/config/scratches/scratch_6.py:32: in _timeout_to_secs
#     return timestr_to_secs(timeout)
# _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
#
# timestr = 'FALSE', round_to = 3
#
#     def timestr_to_secs(timestr, round_to=3):
#         """Parses time like '1h 10s', '01:00:10' or '42' and returns seconds."""
#         if is_string(timestr) or is_number(timestr):
#             for converter in _number_to_secs, _timer_to_secs, _time_string_to_secs:
#                 secs = converter(timestr)
#                 if secs is not None:
#                     return secs if round_to is None else roundup(secs, round_to)
# >       raise ValueError("Invalid time string '%s'." % timestr)
# E       ValueError: Invalid time string 'FALSE'.
#
# .venv/lib/python3.7/site-packages/robot/utils/robottime.py:48: ValueError
# ------------------------- Captured stdout call -------------------------
# _timeout_to_secs(False) == 0
# _timeout_to_secs(None) == 0
# _timeout_to_secs(0) == 0
# _timeout_to_secs() == 0
# _timeout_to_secs() == 0
# _timeout_to_secs(FALSE) == 0
# =========================== warnings summary ===========================
# .venv/lib/python3.7/site-packages/robot/utils/robottypes3.py:16
#   /home/sntentos/Documents/Forcepoint/robotframework/.venv/lib/python3.7/site-packages/robot/utils/robottypes3.py:16: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
#     from collections import Mapping, UserString
#
# .venv/lib/python3.7/site-packages/robot/utils/normalizing.py:16
#   /home/sntentos/Documents/Forcepoint/robotframework/.venv/lib/python3.7/site-packages/robot/utils/normalizing.py:16: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
#     from collections import MutableMapping
#
# -- Docs: https://docs.pytest.org/en/latest/warnings.html
# ================================ PASSES ================================
# _______________________________ test_uut _______________________________
# ------------------------- Captured stdout call -------------------------
# uut(False) == 0
# uut(None) == 0
# uut(0) == 0
# uut() == 0
# uut() == 0
# uut(FALSE) == 0
# uut(False) == 0
# uut(0) == 0
# uut(0) == 0
# uut(NO) == 0
# uut(No) == 0
# uut(NONE) == 0
# uut(None) == 0
# uut(OFF) == 0
# uut(Off) == 0
# ======================= short test summary info ========================
# PASSED ../../../.PyCharmCE2019.3/config/scratches/scratch_6.py::test_uut
# FAILED ../../../.PyCharmCE2019.3/config/scratches/scratch_6.py::test_pk - ValueError: Invalid time string 'FALSE'.
# =============== 1 failed, 1 passed, 2 warnings in 0.19s ================
```

src/robot/libraries/Process.py
stdedos: All are good suggestions, and I will handle them after we resolve the other issues

src/robot/libraries/Process.py
stdedos: I don't think I came up with it myself. I remember seeing it in some "Libdoc" generated documentation, hence I tried it. Maybe it wasn't official RF documentation though.

It is also more clean to see expected functionality vs changelog separately, instead of intertwined with each other.

I can amend if you insist.

src/robot/libraries/Process.py
pekkaklarck: My code handles Boolean `False` same way as yours, but there's a difference handling strings `False`, `No`, `OFF`, and so on. Not a big issue, but I don't think it generally makes sense to accept such strings, especially when strings like `True` would anyway cause an error. If we'd leave it like that, this functionality absolutely should not be documented.

We obviously can also argue that accepting the actual Boolean `False` as zero is not a good idea either. That's kind of true, but in Python it is a common idiom to just check the truth value like `if x` and `if not x`, not to check exact values like `if x is None` when the difference doesn't really matter. That means APIs often accept somewhat strange values like `False` or even an empty list with my code. We could change my code to use

```python
if timeout is None or (is_string(timeout) and timeout.upper() == 'NONE'):
    return 0
```

but that would actually be a backwards incompatible change because the old code considered the empty string same as no timeout but this one doesn't. That functionality hasn't been documented, but it is certainly possible someone has noticed that it works and has used it to disable timeouts. We could obviously use

```python
if timeout in ['', None] or (is_string(timeout) and timeout.upper() == 'NONE'):
    return 0
```

but that gets ugly.

src/robot/libraries/Process.py
pekkaklarck: In Robot's docs we've used two approach to specify when a certain feature has been added.
- When explaining the feature, use something like `Starting from Robot Framework 3.2, it is possible to do X.`.
- At the end have a note like `Doing X is a new feature in Robot Framework 3.2.` or possibly a bit shorted `Doing X is new in Robot Framework 3.2.`.

We haven't used  the `v` prefix with the version or added formatting for it. Definitely not a big deal but I really like consistency.

src/robot/libraries/Process.py
pekkaklarck: Notice also that your code doesn't handle `timeout='0'` or any value with quotes around. And it shouldn't. When you use `timeout=0`, the timeout is a string unless Robot's type conversion changes it to something else. If you want a number, you need to use `timeout=${0}`, but we try to design library APIs so that you don't need to. Either keywords can convert values themselves (like we now need to do) or the type conversion can handle that automatically.

stdedos: > Notice also that your code doesn't handle `timeout='0'` or any value with quotes around.

I am getting more confused :confused:.

* On the union of our code's matching input and output ('None', None, int, 0) I think there are no issues
  * Exception is: `timeout == True` (in Python)[1]
* My code also accepts `[j for j in TRUE_STRINGS if j != '1']`, but then `timestr_to_secs` will raise `ValueError` --> Hence, not an issue
* My code also accepts other falsy strings. As you are saying "Not a big issue", therefore boolean False should be okay also (gray area, but again, "Not a big issue")

I understand that `FALSE_STRINGS` are gray area; but I don't understand what else "my code is handling differently".

Would you create a list showing those? Please make sure to "indicate" if e.g. `timeout=1` is in RF or Python syntax (as I get lost when we are talking about which format)

(Again, I cannot attach a file:)
```python
import sys

import pytest

from robot.utils import timestr_to_secs, is_truthy, FALSE_STRINGS, is_string, TRUE_STRINGS


# pytest -p no:warnings -rA /home/sntentos/.PyCharmCE2019.3/config/scratches/scratch_6.py


def fut(timeout):
    if not(is_truthy(timeout) and timeout):
        timeout = 0
    timeout = timestr_to_secs(timeout)

    return timeout


def _timeout_to_secs(timeout):
    if not timeout or (is_string(timeout) and timeout.upper() == 'NONE'):
        return 0
    return timestr_to_secs(timeout)


def template(func):
    print('False values:')
    for i in [*FALSE_STRINGS, False, None, 0, [], {}]:
        if i in ['FALSE', 'OFF', 'NO']:
            continue

        print('%s(%s:%s) == 0' % (func.__name__, i, type(i)), file=sys.__stdout__)
        assert func(i) == 0

        if not callable(getattr(i, 'title', None)):
            continue

        print('%s(%s:%s) == 0' % (func.__name__, i.title(), type(i)), file=sys.__stdout__)
        assert func(i.title()) == 0

    print('Weird values:')
    for i in [j for j in TRUE_STRINGS if j != '1']:
        with pytest.raises(ValueError):
            print('%s(%s:%s) == ???' % (func.__name__, i, type(i)), file=sys.__stdout__)
            assert func(i)

    with pytest.raises(ValueError):
        print('%s(%s:%s) == ???' % (func.__name__, 'True', type('True')), file=sys.__stdout__)
        assert func('True')

    print('%s(%s:%s) == ???' % (func.__name__, True, type(True)), file=sys.__stdout__)
    assert func(True) == timestr_to_secs(True)
    # # IMHO this should be the correct test instead!
    # with pytest.raises(ValueError):
    #     print('%s(%s:%s) == ???' % (func.__name__, True, type(True)), file=sys.__stdout__)
    #     assert func(True)

    print('True values:')
    valid_values = ['1', '1m', '1 hour', 2, 30]
    for i in valid_values:
        expected = timestr_to_secs(i)
        print('%s(%s:%s) == %s' % (func.__name__, i, type(i), str(expected)), file=sys.__stdout__)
        assert func(i) == timestr_to_secs(i)


def test_pk():
    template(_timeout_to_secs)

def test_uut():
    template(fut)

```

[^1]: `timestr_to_secs(True) == 1` because `is_number(True) == isinstance(item, (int, float)) == True` and then I didn't investigate further

src/robot/libraries/Process.py
stdedos: _(Done locally)_

src/robot/libraries/Process.py
pekkaklarck: We are talking how this keyword is used in RF data, so I meant using `timeout=False` there, not in Python. Sorry for not making it obvious.

When you use `timeout=False` in Robot's data, the value given to the `timeout` argument is string `False` (unless there's type information that makes Robot convert it automatically). You can pass Boolean `False` as well, but then you need to use `timeout=${False}`.

If you use `timeout='0'` in Robot's data, the value passed to the `timeout` argument is string `'0'` i.e. string `0` with single quotes around. That's definitely not a valid value. To pass a string without quotes, just use `timeout=0`. To pass an actual integer, use 'timeout=${0}`.

The problem with your code was that it made `timeout=False` (in Robot's data) valid. That's not really a big deal, but I nevertheless don't think it's a good idea. If it was accepted because that would make implementation simpler, at least this functionality shouldn't be documented.

pekkaklarck: In my opinion the valid timeout values should be these:

- Python `None`. It needs to be supported simply because it is the default. It doesn't need to be documented any more than it currently is. Don't think it needs new tests either. This obviously means there's no timeout.
- Strings `NONE`, case-insensitively. This is a new feature (at least with `Run Process`) and should be documented and tested. Also means there's no timeout.
- Empty string as another way to disable timemout. This was accepted earlier and it would be good to support it also now to avoid backwards incompatibility problems. It is possible that someone has used `timeout=` or `timeout=${EMPTY}`. Doesn't need to be explicitly documented but probably would be good to add an explicit test.
- Any value accepted by `timestr_to_secs`. Positive value should enable timeout, others not. Negative values and and zero disabling timeouts is new and needs to be documented and tested. That ought to be done already with positive values.

I don't care much are other objects that are considered false in Python (`False`, empty list, etc.) allowed so that they disable timeout. At the moment they are, because we use `if timeout`, and I don't see any harm doing that. I doubt anyone has used such values, so code changing so that they aren't supported is fine. Either way, these values don't need to be documented or tested.

As I've written earlier, I don't think supporting `timeout=False` as a way to disable timeout makes sense. It doesn't work now and I don't see why it should work in the future. If that simplifies implementation, supporting it is not a big deal. Either way, it shouldn't be documented.

stdedos: > The problem with your code was that it made `timeout=False` (in Robot's data) valid. That's not really a big deal, but I nevertheless don't think it's a good idea. If it was accepted because that would make implementation simpler, at least this functionality shouldn't be documented.

In favor of accepting all the other falsy strings, i.e.:
* `'No'`
* `'Off'` (could be questionable linguistically)
* `''` (definining a currently undefined, but accepted behavior)
* `'0'` (already handled)

I think accepting `'False'` is "an acceptable side effect", and would save us the trouble of redefinining (partially) the "off" behavior for a specific keyword argument.

Specifying in documentation that "falsy" values is acccepted, conveys that unified  RF `is_falsy` behavior, without specifying explicitly that we like `timeout=False` (in RF scope). I can remove it from the `atests` if it is there; any attempt to further mention that "we don't want `'False'`" or declaring that as invalid value it in code would make "more mess" than it would solve IMHO.

I believe we think closely on this one, even though we are standing "on the opposite sides" of the arument. I'll finalize the review tomorrow by "judging" that solving `timeout=False` (RF) loophole is not cost efficient.

If you still insist that it is "absolutely a bad idea", you may push here or otherwise later fix this behavior, e.g.:
```python
def _timestr_to_secs(timeout):
    if timeout.upper() == 'FALSE':
        raise ValueError("'%s' is not a valid timeout value" % timeout)
    if not(is_truthy(timeout) and timeout):
        timeout = 0
    return timestr_to_secs(timeout)
```

I just don't think it's worth packaging 2 lines into 5 lines of a private method of one library. But then again I am not RF's maintainer, neither do I have related or comparable experience.

pekkaklarck: Once more: In my opinion `timeout=False` makes no sense. It's also totally useless if we support `timeout=None` as well as any negative/zero value to disable timeouts. I don't care much if `timeout=False` is supported "accidentally" due to how `timeout` is handled internally, but that functionality definitely shouldn't be documented.

Notice that the reason handling `timeout` is a problem in the first place is that it is passed to `run_process` as part of `**config` and thus cannot get type information. Once we drop Python 2 support, we can change that keyword to accept named-only arguments instead (i..e. `run_process(command, *args, timeout=None)`) and `NONE` as a string (case-insensitively) will be automatically handled by Robot's type conversion. We can then remove all extra conversion code and as the result `timeout=False` won't work anymore. We could obviously then handle that separately, but I don't think it makes sense to add such extra code. This is part of the reason I don't like `timeout=False` being supported now, and this is the primary reason I definitely don't want that support to be documented. I don't want it to be a documented feature we need to in the future.

So, if you insist on using `is_truthy()`, I'm fine with it. It will support string `NONE`, which isn't currently supported, and "accidentally" supporting strings like `False` and `No` isn't a big issue (as long as they aren't documented). As you stated, the benefit of `is_truthy()` is that it avoids a helper method.

src/robot/libraries/Process.py
```diff
@@ -439,11 +441,15 @@ def wait_for_process(self, handle=None, timeout=None, on_timeout='continue'):
...
         """
         process = self._processes[handle]
         logger.info('Waiting for process to complete.')
-        if timeout:
-            timeout = timestr_to_secs(timeout)
+        if not (is_truthy(timeout) and timeout):
+            timeout = 0
+        timeout = timestr_to_secs(timeout)
+        if timeout > 0:
             if not self._process_is_stopped(process, timeout):
```
stdedos: Please check the new changes

stdedos: Please check the new version

src/robot/libraries/Process.py
```diff
@@ -439,11 +441,15 @@ def wait_for_process(self, handle=None, timeout=None, on_timeout='continue'):
         | ${result} =                 | Wait For Process | timeout=1min 30s | on_timeout=kill |
         | Process Should Be Stopped   |                  |                  |
         | Should Be Equal As Integers | ${result.rc}     | -9               |
+
+        Ignoring Falsy, negative, and zero values is new in Robot Framework 3.2.
```
pekkaklarck: Should have `None`, not `Falsy` here, but I can fix that myself.

pekkaklarck:
Thanks for the PR @stdedos!

atest/testdata/standard_libraries/process/run_process_with_timeout.robot
```diff
@@ -4,12 +4,32 @@ Suite Teardown    Remove Files    ${STDOUT}    ${STDERR}
...
 
+Finish before timeout with nONe timeout
+    ${result} =    Run Process    @{COMMAND}    timeout=${NONE_TIMEOUT}
+    Should not be terminated    ${result}
+
+Finish before timeout with zero timeout
+    ${result} =    Run Process    @{COMMAND}    timeout=${ZERO_TIMEOUT}
+    Should not be terminated    ${result}
+
+Finish before timeout with empty timeout
```
pekkaklarck: This test wasn't verified on the `atest/robot` side. I'll add it.

PR #3384, opened 2019-11-21
Added ability to set default value to an environment variable.

VVasyaev:


atest/testdata/variables/environment_variables.robot
```diff
@@ -86,6 +86,14 @@ Empty Environment Variable
...
     Log  %{}
 
+Environment Variable with Default Value
+    ${value}        Set variable    %{NON_EXISTING_VAR=default value}
+                    Should be equal    ${value}    default value
+
+Environment Variable with Empty Default Value
+    ${value}        Set variable    %{NON_EXISTING_VAR=}
+                    Should be equal    ${value}    ${EMPTY}
+
```
pekkaklarck: - You need to add matching tests also to the `atest/robot` side. See CONTRIBUTING.rst and atest/README.rst for details.
- Test with an internal existing variable like `%{NON_EXISTING_VAR=${VAR}}` would be good.
- Should also add a test for a Java properties that can be accessed using this same syntax.
- Testing `=` in the default value would be good as well.
- Please use same style with assignment and other test code as elsewhere.

src/robot/variables/finders.py
```diff
@@ -167,9 +167,12 @@ class EnvironmentFinder(object):
 
     def find(self, name):
         for getter in get_env_var, get_java_property:
-            value = getter(name[2:-1])
+            var_name, is_set, default_value = name.strip('%{').strip('}').partition('=')
+            value = getter(var_name)
             if value is not None:
                 return value
+            if is_set:     # in case if '' is desired default value
+                return default_value
```
pekkaklarck: - Checking does the name contain embedded default value should be done already before the for loop.
- Similarly the default value should be returned outside the loop if neither of the getters return a value.
- Use `name[2:-1]`, not `name.strip('%{').strip('}')`. The latter strips all `%{` and `}` characters from both ends!
- `is_set` could be renamed to `has_default` or something similar.

atest/testdata/variables/environment_variables.robot
VVasyaev: Unfortunately, writing docs is not my best sidde, but i tried to.

src/robot/variables/finders.py
VVasyaev: yeah -- not sure how i did it inside the loop :(
done.

pekkaklarck:
Looks great! The docs could be extended a bit and especially a note about this being a new feature in RF 3.2 is needed. I can do that after merging this. We also nowadays use four spaces as a separator with tests, but it seems files you added tests didn't follow that convention consistently so we can go with two.

PR #3385, opened 2019-11-21
Variable type conversion (#3278)

LeConTesteur:
Hi,
This PR implements functionalities of issues #3278
The functionalities implemented are :
  - type in name of arguments
  - type in name of embedded arguments
  - type in name of assigned variables
  - type in CLI variables
  - type in name of define variables
  - type in name of variable use with *Set Test Variable*

Moreover, the supported type are : int, float, string, boolean, date, datetime, timedelta, list and dict.

Don't hesitate to challenge the code, the development was difficult for me.

Best Regards,

### DIFF
--- a/src/robot/running/arguments/argumentparser.py
+++ b/src/robot/running/arguments/argumentparser.py
@@ -40,6 +40,20 @@ if JYTHON:
     from java.util import List, Map
 
 
+def _get_class_by_type_name(type_):
+    if not type_:
+        return None
+    type_ = type_.strip()
+    if type_.lower() in ('none', 'nonetype'):
+        return type(None)
+    return type_.lower()
+
+def split_variable_name_with_bracket_and_type(name):
+    if not ':' in name:
+        return name, None
+    type_ = name[name.index(':'):name.index('}')]
+    return name.replace(type_, '', 1), _get_class_by_type_name(type_[1:])
+
 class _ArgumentParser(object):
 
     def __init__(self, type='Keyword'):
@@ -202,13 +216,14 @@ class _ArgumentSpecParser(_ArgumentParser):
         spec = ArgumentSpec(name, self._type)
         kw_only_args = False
         for arg in argspec:
+            arg, type_ = self._split_args_and_types(arg)
             arg = self._validate_arg(arg)
             if spec.kwargs:
                 self._raise_invalid_spec('Only last argument can be kwargs.')
             elif isinstance(arg, tuple):
-                arg, default = arg
-                arg = self._add_arg(spec, arg, kw_only_args)
-                spec.defaults[arg] = default
+                arg_, default = arg
+                arg_ = self._add_arg(spec, arg_, kw_only_args)
+                spec.defaults[arg_] = default
             elif self._is_kwargs(arg):
                 spec.kwargs = self._format_kwargs(arg)
             elif self._is_varargs(arg):
@@ -222,6 +237,8 @@ class _ArgumentSpecParser(_ArgumentParser):
                                          'arguments.')
             else:
                 self._add_arg(spec, arg, kw_only_args)
+            if type_:
+                self._add_type_to_arg(spec, arg, type_)
         return spec
 
     def _validate_arg(self, arg):
@@ -248,12 +265,29 @@ class _ArgumentSpecParser(_ArgumentParser):
     def _format_arg(self, arg):
         return arg
 
+    def _get_body_arg(self, arg):
+        if self._is_kwargs(arg):
+            return self._format_kwargs(arg)
+        elif self._is_varargs(arg) or self._is_kw_only_separator(arg):
+            return self._format_varargs(arg)
+        else:
+            return self._format_arg(arg)
+
     def _add_arg(self, spec, arg, kw_only_arg=False):
         arg = self._format_arg(arg)
         target = spec.positional if not kw_only_arg else spec.kwonlyargs
         target.append(arg)
         return arg
 
+    def _split_args_and_types(self, arg):
+        raise NotImplementedError
+
+    def _add_type_to_arg(self, argspec, arg, type_):
+        if argspec.types is None:
+            argspec.types = {}
+        if isinstance(arg, tuple):
+            arg, default = arg
+        argspec.types[self._get_body_arg(arg)] = type_
 
 class DynamicArgumentParser(_ArgumentSpecParser):
 
@@ -288,6 +322,8 @@ class DynamicArgumentParser(_ArgumentSpecParser):
     def _format_varargs(self, varargs):
         return varargs[1:]
 
+    def _split_args_and_types(self, arg):
+        return arg, None
 
 class UserKeywordArgumentParser(_ArgumentSpecParser):
 
@@ -316,3 +352,6 @@ class UserKeywordArgumentParser(_ArgumentSpecParser):
 
     def _format_arg(self, arg):
         return arg[2:-1]
+
+    def _split_args_and_types(self, arg):
+        return split_variable_name_with_bracket_and_type(arg)
--- a/atest/robot/keywords/type_conversion/robot_keyword_annotations.robot
+++ b/atest/robot/keywords/type_conversion/robot_keyword_annotations.robot
@@ -1,0 +1,144 @@
+*** Settings ***
+Suite Setup      Run Tests    ${EMPTY}    keywords/type_conversion/robot_keyword_annotations.robot
+Force Tags       require-py3
+Resource         atest_resource.robot
+
+*** Test Cases ***
+Integer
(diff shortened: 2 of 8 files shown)


pekkaklarck:
Due to me being somewhat busy and there being lot of changes, I don't have time to review this right now. I'll add the underlying issue to RF 3.2 scope so that I won't forget about it. That said, RF 3.2 is already overdue and I definitely want it out this year, so there's a change this won't make it into it anyway. I'll do my best.

pekkaklarck:
I went through the code quickly and realized this is even bigger change than I expected. RF 3.2 is already so much overdue that unfortunately this needs to be descoped. This is a great feature for RF 3.3, but I cannot promise when its active development starts, hopefully sometime 2020.

I won't do a detailed review to this PR now, but I'll go through the code again and add comments to places that I think should be changed. Because this is such a huge feature, I would also recommend splitting it into several PRs implementing independent features separately. These independent features could be:

- User keyword argument type conversion. Could use lot of the existing library keyword type conversion logic, just needed to parse type information from user keyword arguments.
- Conversion in the Variables section.
- Conversion when assigning a variable based on keyword return value.
- Conversion when using `Set Global/Suite/Test/Local Variable`.
- Possible general enhancements to conversion logic such as support for new types. If there are generally useful enhancements, separate issues should be submitted about them.

src/robot/running/arguments/argumentparser.py
```diff
@@ -198,13 +199,15 @@ def parse(self, argspec, name=None):
...
-            elif '=' in arg:
+            elif self._have_default_value(arg):
                 self._add_arg_with_default(spec, arg, kw_only_args)
             elif spec.defaults and not kw_only_args:
                 self._raise_invalid_spec('Non-default argument after default '
                                          'arguments.')
             else:
                 self._add_arg(spec, arg, kw_only_args)
+            if type_:
+                self._add_type_to_arg(spec, arg, type_)
```
pekkaklarck: Is it necessary to get the type information in the beginning? Wouldn't it be easier to just enhance `_add_arg` so that it can check does the name have embedded type information and then act accordingly?

src/robot/running/arguments/embedded.py
```diff
@@ -53,10 +54,8 @@ def parse(self, string):
         return name, args
 
     def _get_name_and_pattern(self, name):
-        if ':' not in name:
-            return name, self._default_pattern
-        name, pattern = name.split(':', 1)
-        return name, self._format_custom_regexp(pattern)
+        name, type_, pattern = split_variable_name_without_bracket_and_type_and_pattern(name)
+        return ':'.join(e for e in [name, type_] if e), self._format_custom_regexp(pattern) if pattern else self._default_pattern
```
pekkaklarck: Didn't realize earlier that handling types with embedded arguments is a bit complicated. They already support `arg:pattern` syntax and we cannot easily differentiate `arg:type` or `arg:pattern:type` syntax from that. This is especially problematic because the `pattern` may also contain literal `:` character.

I'm not sure what's the best way to solve this issue. What I do know is that splitting this PR into smaller parts makes discussing about details like this a lot easier.

src/robot/variables/tablesetter.py
```diff
@@ -45,23 +47,34 @@ def read(self, variables):
                 var.report_invalid_syntax(err)
 
     def _get_name_and_value(self, name, value, error_reporter):
-        return name[2:-1], VariableTableValue(value, name, error_reporter)
+        name, type_ = split_variable_name_with_bracket_and_type(name)
+        return name[2:-1], VariableTableValue(value, name, type_, error_reporter)
 
 
-def VariableTableValue(value, name, error_reporter=None):
+def VariableTableValue(value, name, type_=None, error_reporter=None):
```
pekkaklarck: I got a feeling it would be simpler if `VariableTableValue` still gets only `name` and then internally checks does it contain type information or not. I might be wrong with this. Seeing how all changes affect each others with such a big PR is hard. As I've written earlier, splitting this to smaller pieces would be great.

src/robot/variables/tablesetter.py
```diff
@@ -45,23 +47,34 @@ def read(self, variables):
...
+        self._type = type_
+
+    def _convert(self, value):
+        try:
+            return convert_variable_value('', value, self._type)
+        except ValueError:
+            pass
+        except TypeError:
+            pass
+        return value
```
pekkaklarck: This doesn't look good. Why are errors silenced? If the idea is to avoid backwards incompatibility problems with cases like `${name:with:colons}`, it would be better to already earlier validate is the part of the last colon a type or not. Blindly ignoring all these exceptions means that bugs in the code are easily hidden as well. Also, an API that requires passing `''` isn't that convenient.

src/robot/variables/typeconverters.py
```diff
@@ -122,13 +124,33 @@ def _literal_eval(self, value, expected):
...
+@TypeConverter.register
+class StringConverter(TypeConverter):
+    type = str
+    type_name = 'string'
+    aliases = ('str',)
+
+    def _convert(self, value, explicit_type=True):
+        return str(value)
+
+
```
pekkaklarck: What is this class for? It seems to register a converter for strings, but shouldn't something like that be a generic enhancement deserving its own PR and also an issue?

Notice also that due to differences between Python 2 and 3, this code doesn't work correctly. `str` is actually bytes in Python 2 and there `unicode` should be used instead. Importantly, `str(value)` fails with if the `value` has a non-ASCII string representation.

src/robot/variables/typeconverters.py
```diff
@@ -122,13 +124,33 @@ def _literal_eval(self, value, expected):
...
 
     def _convert(self, value, explicit_type=True):
+        try:
+            return self._string_convert(value)
+        except:
+            try:
+                return bool(value)
+            except:
+                pass
+        raise ValueError
```
pekkaklarck: Why is this code needed?

Should never use plain `except:`, unless the exception handling code checks the exception type, to avoid hiding `SystemExit`, `KeyboardInterrupt`, and other such exceptions.

src/robot/variables/variableconverter.py
```diff
@@ -0,0 +1,22 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from .typeconverters import TypeConverter
+
+def convert_variable_value(name, value, type_):
+        if not type_:
+            return value
+        converter = TypeConverter.converter_for(type_, allow_str_converter=True)
+        return converter.convert(name, value) if converter else value
```
pekkaklarck: Why is a separate converted needed for variable values? Why cannot TypeConverter used directly?

Also, why has the `typeconverters` module been moved under `robot.variables`? I understand having it under `robot.running.arguments` may not be convenient when thinking about variable conversions, but in general I would say having it under variables is even worse. We need to discuss separately what's the best place for the code. Possibly generic parts could be moved under `robot.utils`. That could be done as a separate change before adding variable conversion.

yamatoRT:
@pekkaklarck I will split the merge request but what is a process? I will push my modification in this branch and i keep my merge request or I create a new branch and new merge request?
For the list of independent features, do we write it in issues or do we create a issue for each features?

src/robot/running/arguments/argumentparser.py
yamatoRT: yes it is necessary because _split_args_and_types remove type in args. If we don't split in begin we have problem with kwargs and vargs when the argname will be 'arg:type'

src/robot/running/arguments/embedded.py
yamatoRT: I split the PR after you request, this part of code is not anymore

src/robot/variables/tablesetter.py
yamatoRT: I split the PR after you request, this part of code is not anymore

src/robot/variables/typeconverters.py
yamatoRT: I split the PR after you request, this part of code is not anymore

src/robot/variables/variableconverter.py
yamatoRT: I split the PR after you request, this part of code is not anymore

src/robot/variables/tablesetter.py
yamatoRT: I split the PR after you request, this part of code is not anymore

pekkaklarck:
Sorry for us dropping the ball with this PR @LeConTesteur. I don't remember what happened to it, but it doesn't make sense to resurrect it anymore. Reasons are:
- There are plenty of conflicts.
- It only implements conversion with user keyword arguments.
- PR #5379 implements that in addition to other variable conversions.

PR #3386, opened 2019-11-21
Unit tests & acceptance runs using Github Actions - Option 1

tysonite:
I was just interested in Github Actions, and some project with complex requirements for testing. This is mostly based on work done [here](https://github.com/robotframework/robotframework/issues/3275#issuecomment-531652435), and probably can fit into #3275. There are some issues with running tests on Windows Server platform: might be due misconfiguration of environment, or incorrect Python interpreter versions.

You can see the results of the CI run on [this](https://github.com/tysonite/robotframework/actions) page.

More information about Github Actions can be found [here](https://help.github.com/en/actions/automating-your-workflow-with-github-actions).

PR #3390, opened 2019-11-23
Unit & Acceptances tests run - option 2

tysonite:
In continuation of #3386:
- unit tests run on push and pull request events
- acceptance tests runs on pull request and [each 12 hours](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onschedule)
- run tests on Jython '2.7.2b2', '2.7.0', '2.7.1'

### DIFF
--- a/.github/workflows/acceptance_tests.yml
+++ b/.github/workflows/acceptance_tests.yml
@@ -3,7 +3,7 @@ name: Acceptance tests
 on:
   pull_request:
   schedule:
-    - cron:  '* */12 * * *'
+    - cron: '0 */12 * * *'
 
 jobs:
   test_using_builtin_python:
--- a/.github/workflows/acceptance_tests.yml
+++ b/.github/workflows/acceptance_tests.yml
@@ -69,6 +69,9 @@ jobs:
         java: [ '1.7', '1.8' ]
         os: [ 'windows-latest', 'windows-2016' ]
         jython-version: [ '2.7.2b2', '2.7.1', '2.7.0', '2.5.3' ]
+        exclude:
+          - java: '1.7'
+            jython-version: '2.7.2b2'
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/acceptance_tests.yml
+++ b/.github/workflows/acceptance_tests.yml
@@ -68,10 +68,11 @@ jobs:
       matrix:
         java: [ '1.7', '1.8' ]
         os: [ 'windows-latest', 'windows-2016' ]
+        jython-version: [ '2.7.1', '2.7.0', '2.5.3' ]
 
     runs-on: ${{ matrix.os }}
 
-    name: Jython (Java ${{ matrix.java }}) 2.7.1 on ${{ matrix.os }}
+    name: Jython (Java ${{ matrix.java }}) ${{ matrix.jython-version }} on ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v1
 
@@ -87,10 +88,10 @@ jobs:
           java-version: ${{ matrix.java }}
           architecture: 'x64'
 
-      - name: Setup Jython
+      - name: Setup Jython ${{ matrix.jython-version }}
         run: |
           choco install wget -y --no-progress
-          wget -nv "http://search.maven.org/remotecontent?filepath=org/python/jython-installer/2.7.1/jython-installer-2.7.1.jar" -O jytinst.jar
+          wget -nv "http://search.maven.org/remotecontent?filepath=org/python/jython-installer/${{ matrix.jython-version }}/jython-installer-${{ matrix.jython-version }}.jar" -O jytinst.jar
           java -jar jytinst.jar -s -d jython/
       
       - name: Run acceptance tests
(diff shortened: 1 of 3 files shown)


.github/workflows/acceptance_tests.yml
```diff
@@ -0,0 +1,147 @@
...
+    - cron: '0 */12 * * *'
+
+jobs:
+  test_using_builtin_python:
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ 'ubuntu-16.04', 'ubuntu-18.04', 'macos-latest', 'windows-latest' ]
+        python-version: [ '2.7.x', '3.5.x', '3.6.x', '3.7.x', '3.8.x', 'pypy2', 'pypy3' ]
```
cclauss: Why not lose all the __.x__ clutter? __python-version: ['2.7', '3.5', '3.6', '3.7', '3.8', 'pypy2', 'pypy3']__

tysonite: I am not sure how to answer, I just took that from an [example](https://github.com/actions/setup-python). Looks like `.x` can be removed.

tysonite:
> LGTM but I think the problem that we face with #3045, #3276, #3313, #3386, #3390 is that unfortunately the folks with commit rights to this repo do not believe in CI testing as discussed at [#3276 (comment)](https://github.com/robotframework/robotframework/pull/3276#issuecomment-523666011)

I am pretty sure they will believe. :-)

As project grows up and becomes more mature, a lot of different combinations and permutations shall be verified, and better to do that earlier rather than just before the release as time to figure out issues in a feature after it was completed several months back will be much higher comparing to when you have a fresh context in your mind.

Let's see how it goes. I am interested to see automatic checks running on this project.

Hi-Fi:
I like this one. Not sure if 12 hour interval for acceptance tests is good over PR triggered run, but that's something that's easy to change when moving forward.

tysonite:
> I like this one. Not sure if 12 hour interval for acceptance tests is good over PR triggered run, but that's something that's easy to change when moving forward.

For PRs, acceptance tests will be still triggerred, however if you look into commit history, owners of the project push changes without PRs so still need to verify somehow changes.

hugovk:
> owners of the project push changes without PRs

I'd recommend against this for all but the most trivial changes, especially when the CI is ready to validate PRs.

Hi-Fi:
> > I like this one. Not sure if 12 hour interval for acceptance tests is good over PR triggered run, but that's something that's easy to change when moving forward.
> 
> For PRs, acceptance tests will be still triggerred, however if you look into commit history, owners of the project push changes without PRs so still need to verify somehow changes.

Sorry, missed that the PRs were also there. Then I'm not sure if schedule is needed, but that doesn't hurt I think (as long as there's no charges for data/time in CI).

I think that master pushes also need to be prevented when CI is in place. Of course admins can bypass those things, if no other way by just disabling enforcing momentarily. But still it's better to have checks on other branches than master.

.github/workflows/acceptance_tests.yml
cclauss: Please remove this clutter.

tysonite:
> Sorry, missed that the PRs were also there. Then I'm not sure if schedule is needed, but that doesn't hurt I think (as long as there's no charges for data/time in CI).

The issue with Win platforms for acceptance tests as it takes too much time. We could think about having several independant workflows that runs acceptance tests on different OSes. E.g. [this](https://github.com/tysonite/robotframework/runs/330207775) run takes 2 and half hours.

> I think that master pushes also need to be prevented when CI is in place. Of course admins can bypass those things, if no other way by just disabling enforcing momentarily. But still it's better to have checks on other branches than master.

I suppose process changes shall be separated from CI stuff. Whatever way peopel like to work shall be supported by CI system.

pekkaklarck:
Is there really no way to run tests when changes are committed directly to master? There's typically nobody to review my PRs and creating them would just slow down my work. I like [trunk based development](https://trunkbaseddevelopment.com/) in general and especially on this kind of context. I don't want CI to dictate my workflow.

tysonite:
> Is there really no way to run tests when changes are committed directly to master? There's typically nobody to review my PRs and creating them would just slow down my work. I like [trunk based development](https://trunkbaseddevelopment.com/) in general and especially on this kind of context. I don't want CI to dictate my workflow.

There is a way. In this PR, unit tests will be run when you commit to any branch. But acceptances tests will not run. However, we can separate acceptance tests based on OS (e.g. Linux based tests take much less time to run than Windows) and run only acceptance tests on Linux when you push something to master.

tysonite:
I think you make releases from master. At a point of release, the commit shall be verified against all tests. As well, it would be nice to have restrictions documented in release notes in a user-friendly way so even if some tests fail people know what to expect when they start running Robot on particular OS / Python intrepreter.

pekkaklarck:
Yeah, all acceptance tests should not run with all commits. The old Jenkins based system we had worked so that tests were run on one or two Python versions on Linux and on Windows with every commit (there was some grace period to wait for possible additional commits) and with others every 12 or 24 hours. Then there was also a way to run tests manually if needed e.g. when making a release. Something like that would be awesome.

pekkaklarck:
Yes, we make releases from master. Steps are documented in BUILD.rst.

One step is running tests on all supported interpreters on Windows and Linux. With latest releases I've done it myself on my own Linux laptop where I have Windows on a VM. CI obviously makes that step unnecessary or at least helps a lot - there are some tests we cannot run on CI and sometimes also they should be run. There's typically also at least a week rc period before making the final release when no code changes are done.

PR #3401, opened 2019-12-03
Small changes to .gitignore

stdedos:
* Remove duplicate pattern
* Ignore virtualenv creation locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,12 @@
 tmp
 build
 dist
+venv
+.venv
 *~
 MANIFEST
 atest/results
 utest/jasmine-results
-*~
 .idea
 .vscode
 .*project


pekkaklarck:
Could you clarify these changes? What do you try to get done?

stdedos:
I usually create virtual environments next to the source code they are created for `git clone x ; cd x ; virtualenv --no-site-packages .venv`. Usually I prefer `.venv` folder name, but I have noticed a handful of other patterns (namely, all matches to the comment before the gitignore pattern).

I am trying to ignore all those python-virtualenv folders.
Then, I am adding a small treat to the change (deduplication), so as to "help" in getting it accepted 😏

pekkaklarck:
Configuring a project to match a personal workflow isn't generally a good idea, it's typically better to use personal configs instead. That said, creating `.venv` or `venv` directory to the project root is getting so common that adding them to `.gitignore` ought to be fine.

The problem is that the added `??env` doesn't make it clear it tries to match `.venv` and, more importantly, doesn't even match `venv`. Why not to add `.venv` and `venv` there explicitly on the top after `tmp`, `build` and `dist`? They'd require no additional comments.

The diff only showed removal of `*~`, not that it originally existed there twice. Good riddance.

stdedos:
> The problem is that the added `??env` doesn't make it clear it tries to match `.venv`and, more importantly, doesn't even match `venv`. Why not to add `.venv` and `venv` there explicitly on the top after `tmp`, `build` and `dist`? They'd require no additional comments.

I thought it would make the expression more succint. I also tried it before submitting and it was working ... until now, that I remembered that empty directories are "nothing" to git - not even considered as an untracked file. Oops 😅

pekkaklarck:
Heh, had the exact same problem with empty directories when I tested how these patterns in .gitignore work. Had it before I edited .gitignore so realized quite quickly what the issue was. Anyway, the updated PR has been merged.

PR #3407, opened 2019-12-07
Fixes segmentation fault when using Screenshot with wxPython on Linux and Python 2.7

### LINKED ISSUE #3403: Screenshot library creates Segmentation Fault when using with WXPython 4.0.7 in Linux Mint
On a Virtual Box with Linux Mint Debian edition (and also reported on Ubuntu 18.04), Python 2.7.13, wxPython 4.0.7.post2.
Using Screenshot library works fine with PyGTK but causes segmentation fault when using wxPython.
```
$ python -m robot.libraries.Screenshot example.jpg pygtk
Using 'gtk' to take screenshot.
Taking test screenshot to 'example.jpg'.
Success!
$ python -m robot.libraries.Screenshot example.jpg wxpython
Using 'wx' to take screenshot.
Taking test screenshot to 'example.jpg'.
Segmentation fault
$
```

This problem have the side effect of making RIDE blocked, even if the import is using other than wxPython screenshot_module. [RIDE #2070](https://github.com/robotframework/RIDE/issues/2070)

HelioGuilherme66:
Fixes #3403

### DIFF
--- a/src/robot/libraries/Screenshot.py
+++ b/src/robot/libraries/Screenshot.py
@@ -31,6 +31,7 @@ elif sys.platform == 'cli':
 else:
     try:
         import wx
+        wx_app = wx.App(False)  # Linux Python 2.7 must exist on global scope
     except ImportError:
         wx = None
     try:
@@ -251,7 +252,6 @@ class ScreenshotTaker(object):
     def __init__(self, module_name=None):
         self._screenshot = self._get_screenshot_taker(module_name)
         self.module = self._screenshot.__name__.split('_')[1]
-        self._wx_app_reference = None
 
     def __call__(self, path):
         self._screenshot(path)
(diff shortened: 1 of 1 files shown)


src/robot/libraries/Screenshot.py
```diff
@@ -31,6 +31,7 @@
 else:
     try:
         import wx
+        wx_app = wx.App(False)  # Linux Python 2.7 must exist on global scope
```
pekkaklarck: I'm slightly worried will this be slow or reserve unnecessary resources if wx isn't in the end used for taking screenshots. Could you test would just using `wx_app = None` here and using the code below later work as well? We'd only create the app reference when actually needed.

```python
def _wx_screenshot(self, path):
    global wx_app
    if wx_app is None:
        wx_app = wx.App(False)
    # ...
```

HelioGuilherme66: It does not work. Only having the  App created in global scope, like I propose.

pekkaklarck: What I proposed above _does_ create a global app reference.

HelioGuilherme66: I know, I even tried declaring it in the import wx block, but the problem still happened.

pekkaklarck: This sounds super strange and I'm still somewhat worried about the global `wx.App` reference. Because the app references is only created if `wx` can be imported I guess that's ok. We also use wxPython for taking the screenshot if it's available by default.

pekkaklarck:
This commit caused issue #3798 and has been now reverted in RF 4.1.

PR #3412, opened 2019-12-10
Process: Add DEVNULL option to disable stdout and stderr

### LINKED ISSUE #3397: `Process`: Add option to disable stdout and stderr
The option should be added to 'Run Process' arguments to use stdout=NULL or stdout=NONE to disable writing stdout/stderr in general. Same that you can do in shell with > /dev/null.

adrianyorke:
Fixes #3397


pekkaklarck:
Thanks for the PR @adrianyorke! Third time’s the charm. :smile: 

I think it would have been enough to just keep committing to the original PR and let GitHub squash everything to a single commit on merge. That's not a big deal and great that this is now done!

PR #3426, opened 2019-12-23
Upgrade GitHub actions/checkout@v1 --> v2

cclauss:
https://github.com/actions/checkout/releases


tysonite:
There is one more instance of checkout action in unit_tests/test_using_jython and also in commented section for IronPython. Let's make this change to all of them.

PR #3429, opened 2019-12-27
Libdoc toc

### LINKED ISSUE #3280: Libdoc: Support automatic generation of table of contents when using "robot format"
It would be nice enhancement if `libdoc` could automatically generate table of contents in the keyword documentation. Although it's small amount of work to maintain, it's still something to be remembered when updating `__intro__` documentation.

It could be good idea to parse all first level headings (meaning `= heading one =`) from the `__intro__` documentation and turn those heading as toc. It could problematic to please the toc in the right spot, but that could be done by adding a marker in the documentation, example like `%TOC%`. 

SeleniumLibrary made an automatic [toc generation](https://github.com/robotframework/SeleniumLibrary/pull/1433/files#diff-40c471e982a520769c6721ee04a812cdR499) and there is a [marker](https://github.com/robotframework/SeleniumLibrary/pull/1433/files#diff-40c471e982a520769c6721ee04a812cdR499) where the toc is placed.

For more details about the SeleniumLibrary implementation, please see https://github.com/robotframework/SeleniumLibrary/issues/1430

aaltat:
Fixes #3280


aaltat:
Not yet ready but I needed to ask one design thing and therefor made a PR. See the comment for more details.

src/robot/libdocpkg/model.py
```diff
@@ -27,19 +28,25 @@ class LibraryDoc(object):
...
         self.keywords = []
 
     @setter
     def doc_format(self, format):
         return format or 'ROBOT'
 
+    @property
+    def doc(self):
+        if self.doc_format == 'ROBOT':
+            return add_toc(self._doc, self.inits)
```
aaltat: The ToC looked wrong if it would always create importing section, like [SL has](https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Importing), so I made it configurable. But the `LibraryDoc` does not yet in this stage does not know, does library has a importing section or not. It is set later, like in here the [robotbuilder](https://github.com/robotframework/robotframework/blob/master/src/robot/libdocpkg/robotbuilder.py#L38). Therefore I did wonder, is my solution good one? Or should make bigger change and  change `LibraryDoc` __init__ to take in also the inits?

pekkaklarck: Good question! My first reaction was that we could change `LibraryDocBuilder._get_doc` instead, but we then also needed to change other doc builders so probably changing model is better. 

What's the problem of using a property? If someone accesses `LibraryDoc.doc` before setting inits, then getting wrong TOC is kind of a user error. That's especially true if the property always recreates the TOC (i.e. doesn't set `sefl._doc`) and accessing `.doc` after inits have been set returns the correct TOC.

The above said, `LibraryDoc.__init__` getting new arguments `inits=None, keywords=None` would be fine. Even then someone could update `.inits` afterwards, but a property would always work correctly based on the current data.

atest/robot/libdoc/toc_generation_lib_init.robot
```diff
@@ -0,0 +1,61 @@
+*** Settings ***
+Suite Setup       Run Libdoc And Parse Output    ${TESTDATADIR}/TocLibraryInitRobot.py
+
```
pekkaklarck: Unnecessary empty line here and in other test case files.

atest/robot/libdoc/toc_generation_lib_init.robot
```diff
@@ -0,0 +1,61 @@
...
+    ...
+    ...    = 2 Heading =
+    ...
+    ...    Second chapter text there.
+    ...
+    ...    Second paragraph.
+
+Version
+    Version Should Be               ${EMPTY}
+
```
pekkaklarck: Do we actually need to test other information from the library than the documentation? If we'd only check that, there could be just one test case file that could have separate tests for lib w/ and w/o toc, resource file, etc.

atest/testdata/libdoc/TocLibraryInitRobot.py
```diff
@@ -0,0 +1,23 @@
+class TocLibraryInitRobot(object):
+    """Simple library
+
+    %TABLE_OF_CONTENTS%
```
pekkaklarck: I think I proposes `%TABLE_OF_CONTENTS%` instead of `%TOC%` but now that I see it live it looks a bit ugly. What do you think, should we use `%TOC%` anyway?

src/robot/libdoc.py
```diff
@@ -121,11 +121,11 @@
 Examples:
 
   python -m robot.libdoc Dialogs list
-  python -m robot.libdoc Selenium2Library list browser
+  python -m robot.libdoc SeleniumLibrary list browser
```
pekkaklarck: Good catch! Unrelated to this PR otherwise but so small change that including it here is fine.

src/robot/libdocpkg/model.py
```diff
@@ -17,6 +17,7 @@
 
 from robot.model import Tags
 from robot.utils import getshortdoc, Sortable, setter
+from robot.utils.table_of_content import add_toc
```
pekkaklarck: I don't think this Libdoc related util belongs to `robot.utils`. If it was to be added there, then it should be exposed directly from `robot.utils`, not from a submodule. Also, Python module names in general shouldn't use `_`.

It would be best to move that `table_of_content` module directly under `libdocpkg` and rename it to `tocbuilder` or something like that.

src/robot/libdocpkg/model.py
```diff
@@ -27,19 +28,25 @@ class LibraryDoc(object):
...
                  scope='', named_args=True, doc_format=''):
         self.name = name
-        self.doc = doc
+        self.doc_format = doc_format
+        self._doc = doc
         self.version = version
         self.type = type
         self.scope = scope
         self.named_args = named_args
-        self.doc_format = doc_format
```
pekkaklarck: I don't see a reason this to be moved. The property works fine regardless.

src/robot/libdocpkg/model.py
pekkaklarck: I already replied this elsewhere. I don't see a problem using a property.

src/robot/libdocpkg/model.py
```diff
@@ -27,19 +28,25 @@ class LibraryDoc(object):
...
         self.inits = []
         self.keywords = []
 
     @setter
     def doc_format(self, format):
         return format or 'ROBOT'
 
+    @property
+    def doc(self):
+        if self.doc_format == 'ROBOT':
```
pekkaklarck: Could perhaps have `and '%TOC%' in self._doc` here as well.

src/robot/utils/table_of_content.py
```diff
@@ -0,0 +1,23 @@
+import re
+
+
+def add_toc(doc, importing):
```
pekkaklarck: `importing` is a Boolean, right? Since we cannot use type hints, could perhaps add a default value here to make it more explicit. Could also/alternatively rename it to something more clear. For example, `include_importing_section` would be very explicit but somewhat long.

utest/utils/test_toc.py
```diff
@@ -0,0 +1,103 @@
...
+Other text
+"""
+
+
+class TestTableOfContentLibdoc(unittest.TestCase):
+
+    def test_no_toc(self):
+        test_doc = 'Library doc'
+        doc = add_toc(test_doc, True)
+        self.assertEqual(doc, test_doc, 'Doc should not have been created.')
```
pekkaklarck: We've typically used our own `robot.utils.asserts.assert_equal` to avoid `self.` and using camelCase. Using the standard `self.assertEqual` is fine as well, though.

src/robot/utils/table_of_content.py
```diff
@@ -0,0 +1,23 @@
...
+        doc = doc.replace('%TABLE_OF_CONTENTS%', '\n'.join(toc))
+    return doc
+
+
+def _get_headings(doc):
+    match = re.findall(r'(^= )(.*)( =$)', doc, re.MULTILINE)
+    return ['- `%s`' % heading[1] for heading in match]
+
+
+def _create_toc(headings, importin):
```
pekkaklarck: Typo in `importin`.

src/robot/utils/table_of_content.py
```diff
@@ -0,0 +1,23 @@
...
+    toc = ['== Table of contents ==', '']
+    toc.extend(headings)
+    if importin:
+        toc.extend(['- `Importing`'])
+    toc.extend(['- `Shortcuts`', '- `Keywords`'])
+    return toc
```
pekkaklarck: Kind of a nitpick, but for me it would feel a bit more natural if `_create_toc` would return the toc as a string, not as a list the caller needs to know how to format. In other words, `return '\n'.join(toc)` instead of `return toc`.

atest/testdata/libdoc/TocLibraryInitRobot.py
aaltat: I am happy with both ways, but I agree that `%TOC%` looks better. Will change, not a big thing.

src/robot/utils/table_of_content.py
aaltat: Yes, it's a used as a boolean. If list is empty init is not populated. Adding default is good option, should I actually use `[]` ?

pekkaklarck: If it's used as a Boolean, possible default should be a Boolean as well. On the calling side I'd also use `bool(...)` but that's not such a big deal.

atest/robot/libdoc/toc_generation_lib_init.robot
aaltat: Well, perhaps not. I did not read that code so deeply and just took examples from other libdoc tests.

aaltat:
Yes, the doc is missing. Will write once I get all the other things done. 

It reminds me one one thing, now the tests are done with documentation using RF documentation format. There should be tests for `.rst` format (which is missing) but with the latest parser changes, should I test other formats too and if yes what?

pekkaklarck:
I don't think we need TOCs for anything else than Robot's own format. This is why:

- Nobody has requested that. Can be added later if needed.
- reST has its own directive for generating a TOC.
- Implementation would be totally different with different formats.

aaltat:
Sounds good, let's go with Robot format only. I somewhat busy to get the Robocon stuff ready, so let's see when I can get time to do this.

pekkaklarck:
There will be RF 3.2 beta 1 this week (needed for RoboCon as well) but earliest on Friday. If you get this ready by then, great, if not, there's still time to get it into the final release.

One thing to do is taking the TOC functionality into use with Robot's own standard libs. Can be done as part of this PR or separately.

aaltat:
I totally missed the part of updating standard libraries (Eating your own dog food), good catch. Will do.

src/robot/utils/table_of_content.py
pekkaklarck: Should also add a section for tags if keywords have them.

pekkaklarck:
This PR was never finished and the functionality has now been implemented by acdea400d00caa052e878245eda2431bef2f77f4.

PR #3441, opened 2020-01-21
Fixed typo in documentation for `Get Slice From List`.

miktuy:
If the `end` of slice is not explicit, the bug will follow.
image

E.g. ```Get Slice From List  ${L5}  end=-2``` is correct:
image


pekkaklarck:
Good catch! This probably worked when using the HTML format (and may still work with the pipe separated format).

Issue #3440, opened 2020-01-19
Allow setting a tag to highlight keywords with the tag via query string to docs generated with libdoc


rasjani:
Scenario:

Libdoc generated keyword documation has ability to scope to certain tags within generated html documentation. When adding plugins (or extending existing libraries) it would be helpful for the reading to automatically highlight those added keywords. Currently, this can be done by selecting "tags" in the generated documentation but this selection process cannot be done directly by linking to the keyword document.

Proposal:

add a bit of javascript code into libdoc.html template that has querystring handler that checks if certain tag should be highlighted.

PR #3442, opened 2020-01-21
Issue 3440

rasjani:
Adds functionality to automatically select tags in libdoc html documantetion if query string "tag=$tag_to_select" is passed to it.

Another commit where around this same topic as the element title said "tests" but at least in libdoc context, they where referring to document tags.

#3440

### DIFF
--- a/src/robot/htmldata/libdoc/libdoc.html
+++ b/src/robot/htmldata/libdoc/libdoc.html
@@ -322,7 +322,7 @@
            title="Show tests with this tag">${$value}</a> &middot;
         {{/each}}
         <a href="javascript:resetKeywords()" class="normal-first-letter"
-           title="Show all tests">[Reset]</a>
+           title="Show all tags">[Reset]</a>
     </div>
 </script>
 
--- a/src/robot/htmldata/libdoc/libdoc.html
+++ b/src/robot/htmldata/libdoc/libdoc.html
@@ -49,6 +49,12 @@
         }
         renderTemplate('keywords', libdoc);
         renderTemplate('footer', libdoc);
+        params = util.parseQueryString(window.location.search.slice(1));
+        if ("tag" in params) {
+            $(document).ready(function() {
+              tagSearch(params["tag"]);
+            })
+        }
         scrollToHash();
         $(document).bind('keydown', handleKeydown);
     });
@@ -92,6 +98,7 @@
         markMatches(tag, include);
         highlightMatches(tag, include);
         $('#keywords-container').find('.kw-row').addClass('hide-unmatched');
+        history.replaceState && history.replaceState(null, '', location.pathname + "?tag=" + tag + location.hash);
     }
 
     function doSearch() {
@@ -217,6 +224,7 @@
         }
         $('#match-count').hide();
         $('#altogether-count').show();
+        history.replaceState && history.replaceState( null, '', location.pathname + location.hash);
     }
 
     function setMatchVisibility() {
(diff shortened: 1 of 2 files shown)


src/robot/htmldata/libdoc/libdoc.html
```diff
@@ -376,5 +378,15 @@ <h2 id="Keywords">Keywords</h2>
...
 </script>
 
+
+<script type="text/javascript">
+    params = utils.parseQueryString(window.location.search.slice(1))
+    if ("tag" in params) {
+        $(document).ready(function() {
+          tagSearch(params["tag"])
+        })
+    }
```
pekkaklarck: 1. Shouldn't it be `util`, not `utils`?

2. Could this be move to the top of this file where the main logic is? Something like this:

    ```
    if ("tag" in params) {
        tagSearch(params["tag"]);
    else
        scrollToHash();
    ```

    Not sure is this possible (haven't looked at this code for ages) but it would feel better to have the main logic in same place.

3. Our JS style (which hasn't been explicitly documented anywhere) has been including `;` at the end of lines.

src/robot/htmldata/libdoc/libdoc.html
```diff
@@ -322,7 +324,7 @@ <h2 id="Tags">Tags</h2>
            title="Show tests with this tag">${$value}</a> &middot;
         {{/each}}
         <a href="javascript:resetKeywords()" class="normal-first-letter"
-           title="Show all tests">[Reset]</a>
+           title="Show all tags">[Reset]</a>
```
pekkaklarck: :+1:

pekkaklarck:
Didn't yet have time to test this live but I like the idea. Here are some comments/questions in addition to comments I already added to the code:

- You wrote that the syntax is `tag=$tag_to_select`, but since the code uses the existing `parseQueryStringHelper`, isn't it actually `tag?$tag_to_select`?
- I'm impressed you found that helper! :smile:

rasjani:
>     * You wrote that the syntax is `tag=$tag_to_select`, but since the code uses the existing `parseQueryStringHelper`, isn't it actually `tag?$tag_to_select`?

? in url is separator for url and params part -- rfc3986 😉  afaik, correct way to pass this is "key=value", first separator from "file" part is ? and then sub separators are either & or ; but even semicolon might have some issues .. 
 
>     * When clicking a tag in the Tags section, it would be great if the url line would be updated to match the selection. I'm worried nobody will find this feature otherwise.

That Actually Should Be Handled. If you link to from outside url, tagSearch() will be called. tagSearch() then will check the appropriate feature (history.replaceState) and use it to update the url to include "tag=chosentag" if its not there already. Or if you land on the keyword doc page without the tag parameter and then click on any of the tags, tagSearch() will be called then too, which again checks for history api and will update the url if its there.

Only occasion this url update has some issues is when clicking reset on the search popup, it will reset url and remove the tag=chosetag parameter in the process ..  bug or feature, fifty sixty..

src/robot/htmldata/libdoc/libdoc.html
rasjani: 1. yes.
2. yes but..

location for this code is definitely better suited to where you point  yes... And there's few things i need to work on so that current functionality work get broken.

3. yes boss.

rasjani: went thru 2. and if i understand the rfc correctly, something like this still works fine in my usecase: `?tag=plugin#Plugin%3A SeleniumTestability`  .. eg, in my case the link would be: https://rasjani.github.io/robotframework-seleniumtestability/index.html?tag=plugin#Plugin%3A%20SeleniumTestability

src/robot/htmldata/libdoc/libdoc.html
rasjani: context where these are selected has head of "Tags" so, reset's title makes more sense than "keywords" - imho

pekkaklarck:
Our report.html actually uses anchors as query strings like `#tags?t1` or `#search?suite=&test=foo&include=t1&exclude=` and I thought all that was implemented by out `util.parseQueryString` function. I was wrong and stripping `#` was done elsewhere.

I don't see problems Libdoc using "real" query strings here so I guess this PR is fine. There are few small things that might still be worth tweaking (e.g. should reset also clear possible anchor containing keyword name) but I can look at them separately after merging this PR.

Issue #3460, opened 2020-02-10
plural_or_not method return "s" for value "-1"


IlfirinPL:
robot/utils/misc.py 
method plural_or_not  return "s" for value -1

pekkaklarck:
Could you clarify what's the problem?

IlfirinPL:
When ask method plural_or_not for values  (0) or (-1) return "s" , 
in example message size is
-1 element plural_or_not(-1) -> "-1 elements", and i think it should be 
"-1 element"

pekkaklarck:
I'm not a native English speaker, but AFAIK you need to use the plural form with zero items.

IlfirinPL:
OK, maybe I jump the gun, its more complex then i was previously expected
there is zero element
http://mathworld.wolfram.com/ZeroElement.html
also 
https://www.grammar-monster.com/lessons/singular_plural_number.htm

I think it should stay as is.

pekkaklarck:
Yeah, the behavior with zero shouldn't change. The situation with -1 is a lot less clear, but apparently it is more often considered singular. I'm fine that being changed.

IlfirinPL:
update in code #3461

pekkaklarck:
I commented the PR. Can you please also update the issue title?

PR #3461, opened 2020-02-10
improve plural_or_not

IlfirinPL:
#3460

### DIFF
--- a/src/robot/utils/misc.py
+++ b/src/robot/utils/misc.py
@@ -105,7 +105,7 @@ def _is_camel_case_boundary(prev, char, next):
 
 def plural_or_not(item):
     count = item if is_integer(item) else len(item)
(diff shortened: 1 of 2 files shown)


pekkaklarck:
- This needs a unit test.
- `count in (1, -1)` would read a bit better.

pekkaklarck:
I was certain this util already had unit tests and it would be enough to just add some negative numbers there but apparently I was wrong. The added tests looks good but there could be another negative case like -2. I'll add it myself.

Thanks for the PR!

pekkaklarck:
On a more detailed look tests are actually buggy! You are using strings, not numbers as an input to `plural_or_not` and it is looking at string length. The test actually fails when I run it locally. I'll fix it myself but please run tests before commit next time.

IlfirinPL:
Thank you for feedback

PR #3462, opened 2020-02-10
Add support for svg image link

osrjv:
This PR adds .svg extensions to the list of image types that are automatically converted into &lt;img&gt; tags in documentation fields.

I modified the only unittest I found for this functionality. There didn't seem to be any existing acceptance tests.

Related to issue: #3464


pekkaklarck:
Looks good. Two more small things should be done:
- A note in the User Guide saying this is new in 3.2. There are similar examples there.
- An issue for tracking purposes. We create release notes templates based on them.

osrjv:
Done and done. Hopefully that note is what you had in mind.

pekkaklarck:
Looks great, thanks!

PR #3469, opened 2020-02-12
List ignore

bbpatel2001:
1.  As list is mutable and if we use sort, it will alter original list.  Original list should not be altered so used sorted function instead of sort method. 

2. Instead of modify more lines in the _yield_list_diffs method to handle `name` option given by user, just passed sorted list to get the diff as we should support name argument/parameter as well.

### DIFF
--- a/atest/testdata/standard_libraries/collections/list.robot
+++ b/atest/testdata/standard_libraries/collections/list.robot
@@ -331,6 +331,12 @@ Lists Should Be Equal With Named Indices As Dictionary With Too Few Values
     ${names} =    Create Dictionary    0=a    2=c
     Lists Should Be Equal    ${L3}    ${L3B}    names=${names}
 
+Lists Should Be Equal Ignore Order
+    [Documentation]    PASS list are equal
+    ${names1} =    Create List    A    B	C	D
+    ${names2} =    Create List    D    B	C	A
+    Lists Should Be Equal    ${names1}    ${names2}    ignore_order=True
+    
 List Should Contain Sub List
     List Should Contain Sub List    ${LONG}    ${L4}
 
--- a/atest/testdata/standard_libraries/collections/list.robot
+++ b/atest/testdata/standard_libraries/collections/list.robot
@@ -332,7 +332,7 @@ Lists Should Be Equal With Named Indices As Dictionary With Too Few Values
     Lists Should Be Equal    ${L3}    ${L3B}    names=${names}
 
 Lists Should Be Equal Ignore Order
-    [Documentation]    PASS list are equal
+    [Documentation]    FAIL Lists are different:
     ${names1} =    Create List    A    B	C	D
     ${names2} =    Create List    D    B	C	A
     Lists Should Be Equal    ${names1}    ${names2}    ignore_order=True
--- a/atest/testdata/standard_libraries/collections/list.robot
+++ b/atest/testdata/standard_libraries/collections/list.robot
@@ -332,8 +332,8 @@ Lists Should Be Equal With Named Indices As Dictionary With Too Few Values
     Lists Should Be Equal    ${L3}    ${L3B}    names=${names}
 
 Lists Should Be Equal Ignore Order    
-    ${names1} =    Create List    A    B	C	D
-    ${names2} =    Create List    D    B	C	A
+    ${list1} =    Create List    A    B    C    D
+    ${list1} =    Create List    D    B    C	A
     Lists Should Be Equal    ${names1}    ${names2}    ignore_order=True
     
 List Should Contain Sub List
--- a/atest/testdata/standard_libraries/collections/list.robot
+++ b/atest/testdata/standard_libraries/collections/list.robot
@@ -331,8 +331,7 @@ Lists Should Be Equal With Named Indices As Dictionary With Too Few Values
     ${names} =    Create Dictionary    0=a    2=c
     Lists Should Be Equal    ${L3}    ${L3B}    names=${names}
 
-Lists Should Be Equal Ignore Order
-    [Documentation]    FAIL Lists are different:
+Lists Should Be Equal Ignore Order    
     ${names1} =    Create List    A    B	C	D
     ${names2} =    Create List    D    B	C	A
     Lists Should Be Equal    ${names1}    ${names2}    ignore_order=True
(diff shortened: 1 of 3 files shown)


src/robot/libraries/Collections.py
```diff
@@ -387,7 +387,12 @@ def lists_should_be_equal(self, list1, list2, msg=None, values=True,
         default = 'Lengths are different: %d != %d' % (len1, len2)
         _verify_condition(len1 == len2, default, msg, values)
         names = self._get_list_index_name_mapping(names, len1)
-        diffs = list(self._yield_list_diffs(list1, list2, names))
+        if not ignore_order:
+            diffs = list(_yield_list_diffs(list1, list2, names))
```
tysonite: Some automatic checks are failing. I suspect, it is because of `self.` missed before `_yield_list_diffs` call.

bbpatel2001: my Bad! Correcting it.

bbpatel2001: Corrected and pushed code. Thanks @tysonite and @pekkaklarck  for quick review.

bbpatel2001:
I am working on fixing the tests,

bbpatel2001:
@finally all tests passed. Today I learn new thing on test expected value!!!

bbpatel2001:
@tysonite @pekkaklarck can you please review.

src/robot/libraries/Collections.py
```diff
@@ -346,7 +346,7 @@ def list_should_not_contain_duplicates(self, list_, msg=None):
                                  '%s found multiple times.' % seq2str(dupes))
 
     def lists_should_be_equal(self, list1, list2, msg=None, values=True,
-                              names=None):
+                              names=None,ignore_order=False):
```
laurentbristiel: missing space between comma and ignore_order

bbpatel2001: Resolved. learning many things in 2nd PR. thanks.

bbpatel2001:
is this RF 3.2 Candidate ?

src/robot/libraries/Collections.py
```diff
@@ -387,7 +387,12 @@ def lists_should_be_equal(self, list1, list2, msg=None, values=True,
         default = 'Lengths are different: %d != %d' % (len1, len2)
         _verify_condition(len1 == len2, default, msg, values)
         names = self._get_list_index_name_mapping(names, len1)
-        diffs = list(self._yield_list_diffs(list1, list2, names))
+        if not ignore_order:
+            diffs = list(self._yield_list_diffs(list1, list2, names))
+        else:
+            sorted_list1 = sorted(list1)
+            sorted_list2 = sorted(list2)
+            diffs = list(self._yield_list_diffs(sorted_list1, sorted_list2, names))
```
pekkaklarck: This looks a bit more complicated than needed. AFAIK this would be enough:

```python
if ignore_order:
    list1 = sorted(list1)
    list2 = sorted(list2)
```

src/robot/libraries/Collections.py
```diff
@@ -346,7 +346,7 @@ def list_should_not_contain_duplicates(self, list_, msg=None):
                                  '%s found multiple times.' % seq2str(dupes))
 
     def lists_should_be_equal(self, list1, list2, msg=None, values=True,
-                              names=None):
+                              names=None, ignore_order=False):
         """Fails if given lists are unequal.
 
         The keyword first verifies that the lists have equal lengths, and then
```
pekkaklarck: The new option needs to be explained in the documentation, possibly with an example. A note about the option being new in Robot Framework 3.2 is needed as well.

atest/testdata/standard_libraries/collections/list.robot
```diff
@@ -331,6 +331,11 @@ Lists Should Be Equal With Named Indices As Dictionary With Too Few Values
     ${names} =    Create Dictionary    0=a    2=c
     Lists Should Be Equal    ${L3}    ${L3B}    names=${names}
 
+Lists Should Be Equal Ignore Order    
+    ${names1} =    Create List    A    B	C	D
+    ${names2} =    Create List    D    B	C	A
```
pekkaklarck: It looks a bit strange that the separator between arguments is different. Using four spaces consistently would be better. These values aren't names either so probably variables should be just `${list1}` and `${list2}`.

pekkaklarck:
Yeah, this can easily be included to RF 3.2. Please take a look at the review comments. I hope you can fix them all so that I can simply merge this online without any further changes.

atest/testdata/standard_libraries/collections/list.robot
```diff
@@ -331,6 +331,11 @@ Lists Should Be Equal With Named Indices As Dictionary With Too Few Values
     ${names} =    Create Dictionary    0=a    2=c
     Lists Should Be Equal    ${L3}    ${L3B}    names=${names}
 
+Lists Should Be Equal Ignore Order    
+    ${list1} =    Create List    A    B    C    D
+    ${list2} =    Create List    D    B    C	A
```
bbpatel2001: Somehow formatting of this shown differently but I just checked checked in file here and it 
shows correct. 

image

bbpatel2001:
Changes done @pekkaklarck

atest/testdata/standard_libraries/collections/list.robot
bbpatel2001: Somehow formatting of this shown differently but I just checked checked in file here and it
shows correct.
https://user-images.githubusercontent.com/33742811/74623859-3ffbb600-516c-11ea-86ac-902a73a13862.png

atest/testdata/standard_libraries/collections/list.robot
pekkaklarck: You probably have a tab at least between `C` and `A`.

pekkaklarck:
Looks good to me. Documentation could perhaps be enhanced a bit but I can do that after merging. Thanks!

rainydew:
well... is that work all time?

test in python3.6.8

>>> sorted([{1,2}, {3,4}]) == sorted([{4,3}, {2,1}])
False

but since set is orderless, i expect the above can pass the compare

pekkaklarck:
That seems to be some Python oddity:
```python
>>> sorted([{1, 2}, {3,4}])
[{1, 2}, {3, 4}]
>>> sorted([{3, 4}, {1, 2}])
[{3, 4}, {1, 2}]
```

My guess is that sets cannot compared. Anyway, nothing we really can do anything.

Issue #3423, opened 2019-12-21
Support JSON output files as part of execution


teaglebuilt:
Problem:

Prefer to use a NOSQL database to store test reports. NOSQL databases typically require json format because its a relational database.

Solution:

Provide robotframework with the ability to use json rather than xml in the output file. 

For example:

```
           # flag --output
          robot --output output.json ./tests
```

teaglebuilt:
If its ok with the community, I would like to work on this. I created a pull request for tracking progress and questions along the way

pekkaklarck:
Being able to use output.json instead of output.xml may be useful. Biggest possible benefits I see are:

- JSON is a lot less verbose than XML, so output.json file would be a lot smaller than output.xml.
- Because output.json would be smaller, writing it to disk during execution would be faster as well.
- It is possible that processing output.json would be faster than processing output.xml. This is not certain, though, as our XML processing code is pretty well optimized and the underlying XML modules are pretty fast.

If processing output.json is faster than processing output.xml, then this feature is definitely a good idea. If it isn't, then we need to think are the other benefits big enough. This is something we can only know after prototyping this a bit.

I don't thing possibility to add data to a NOSQL database or to any other system is a good reason to add anything like this. It is already now possible to do that by parsing the information from output.xml using the standard XML modules or the `robot.api.ExecutionResult` API. Data being stored in JSON instead of XML doesn't really make a big difference.

pekkaklarck:
The are two design decisions to be made:

1. How to enable this feature. I don't think any new command line option is needed. We just need to make it possible to say `--output output.json` and detect the type automatically. Also Rebot needs to be enhanced to allow using `rebot output.json`.

2. The output.json format. Because it would be a new external interface, it needs to be designed carefully. No need to think about that too much in the beginning, though. It is more important to test are there performance benefits or not. If there are, then we need to design the format more carefully.

mkorpela:
There is also another option to reduce output size: write directly to tar.gz.
I think both reading and writing of output.xml would be fairly easy to modify to slso being able to read and write (the same xml) to output.tar.gz

mkorpela:
https://docs.python.org/3/library/tarfile.html
(Zip and others also could work)

teaglebuilt:
@mkorpela one reason why i am set on an json option is the possibilities of using NoSQL databases to easily store robotframework results.

pekkaklarck:
Can you @teaglebuilt explain how this feature would help storing results in a db?

[...]

PR #3475, opened 2020-02-20
#3423 Robot Framework JSON output

Lemonlemmings:


src/robot/conf/settings.py
```diff
@@ -203,7 +204,7 @@ def __getitem__(self, name):
     def _get_output_file(self, option):
         """Returns path of the requested output file and creates needed dirs.
 
-        `option` can be 'Output', 'Log', 'Report', 'XUnit' or 'DebugFile'.
+        `option` can be 'Output', 'Log', 'Report', 'XUnit', 'Junit', or 'DebugFile'.
```
laurentbristiel: Junit => Json

src/robot/output/jsonlogger.py
```diff
@@ -0,0 +1,241 @@
...
+            parent_item = self._current_suite
+        elif destination == 'test':
+            parent_item = self._current_test
+        else:
+            parent_item = self._item_stack[-1]
+
+        if 'kw' not in parent_item:
+            parent_item['kw'] = list()
+        parent_item['kw'].append(self._current_keyword)
+        # Mark the current keyword as being none
```
laurentbristiel: not sure these 2 comments have much added value. The code is clear enough.

src/robot/output/jsonlogger.py
```diff
@@ -0,0 +1,241 @@
...
+            self._current_suite['tests'] = list()
+        self._current_suite['tests'].append(self._current_test)
+
+    def start_suite(self, suite):
+        # If there is an "open" suite, this will be placed inside its suites
+        if self._current_suite:
+            # Push this onto the stack
+            self._item_stack.append(self._current_suite)
+
+        # Default the current suite with t
```
laurentbristiel: with t ?

src/robot/conf/settings.py
Lemonlemmings: Good catch! Resolved.

src/robot/output/jsonlogger.py
Lemonlemmings: I've removed the comments

src/robot/output/jsonlogger.py
Lemonlemmings: Nice catch, resolved!

pekkaklarck:
Thanks for your efforts with this PR @Lemonlemmings. Unfortunately I didn't have time to review it fully back when it was created and now there are so bad conflicts that I think it's best to just close it.

Good news is that there has been a lot of progress with JSON support otherwise:
- The execution model got JSON serialization support in RF 6.1 (#3902).
- The result model got it in RF 7.0 (#4847).
- The result model will be enhanced a bit in RF 7.2 (#5160).

The final missing part is supporting JSON as part of the execution. That enhancement is covered by #3423 and it's currently targeted for RF 7.2 as well. With all earlier work that shouldn't anymore be too big a task, but due to time constraints there's a risk it will be moved to RF 7.3.

If you are still interested to help with this enhancement, let me know and we can collaborate in getting the aforementioned final piece done. I also totally understand if you don't want to spend more time for this as the previous PR didn't get anywhere and you obviously may not be working with Robot Framework anymore either.

PR #3477, opened 2020-02-21
Code coverage metrics on codecov.io

tysonite:


.github/workflows/unit_tests.yml
```diff
@@ -33,11 +33,17 @@ jobs:
...
           python -m coverage run --branch utest/run.py -v
 
-      - name: Prepare HTML coverage report
+      - name: Prepare HTML/XML coverage report
         run: |
           python -m coverage html -i -d cov_html
+          python -m coverage xml -i
         if: always()
 
+      - uses: codecov/codecov-action@v1
```
Hi-Fi: Isn't this now failing the build at forks if the token is not available?

Also, external actions should use SHAs, not tags (as those tags are mutable).

tysonite: Nope. See Slack, I did a test, if token is not here, this action does not fail.
I removed tag and added SHA.

tysonite: BTW, see results of checks run in this PR, they are not failing, however token is not there :)

PR #3480, opened 2020-02-25
Ignore utest and atest directories on codecov.io

tysonite:


tysonite:
I've tried to put `codecov.yml` to another directory (e.g. .github) and reference it from the action, but this doesn't work :-(

PR #3481, opened 2020-02-26
Using seps by default when converting timestamp to secs

JustasM11:
It was noticed, that sometimes Robot fails with the following trace:

```
...snip...
  File "/usr/local/lib/python3.7/site-packages/robot/reporting/jsmodelbuilders.py", line 66, in <genexpr>
    model = tuple(self._build_keyword(k) for k in kws)
  File "/usr/local/lib/python3.7/site-packages/robot/reporting/jsmodelbuilders.py", line 140, in build
    self._get_status(kw),
  File "/usr/local/lib/python3.7/site-packages/robot/reporting/jsmodelbuilders.py", line 53, in _get_status
    self._timestamp(item.starttime),
  File "/usr/local/lib/python3.7/site-packages/robot/reporting/jsbuildingcontext.py", line 58, in timestamp
    millis = int(timestamp_to_secs(time) * 1000)
  File "/usr/local/lib/python3.7/site-packages/robot/utils/robottime.py", line 301, in timestamp_to_secs
    raise ValueError("Invalid timestamp '%s'." % timestamp)
```
Logging shows timestamp that cannot be parsed: '2020-02-26 16:34:28.227'
So we use default seps to avoid such failure

### DIFF
--- a/src/robot/utils/robottime.py
+++ b/src/robot/utils/robottime.py
@@ -294,7 +294,7 @@ def get_timestamp(daysep='', daytimesep=' ', timesep=':', millissep='.'):
     return TIMESTAMP_CACHE.get_timestamp(daysep, daytimesep, timesep, millissep)
 
 
-def timestamp_to_secs(timestamp, seps=None):
+def timestamp_to_secs(timestamp, seps=(' ', ':', '-', '.')):
     try:
         secs = _timestamp_to_millis(timestamp, seps) / 1000.0
     except (ValueError, OverflowError):


pekkaklarck:
How does this happen? Those items should have timestamp in format like "20200226 16:34:28.227" i.e. without hyphens. Have you created/modified output.xml programmatically or possibly modified results with pre-Rebot modifiers or otherwise?

JustasM11:
> How does this happen? Those items should have timestamp in format like "20200226 16:34:28.227" i.e. without hyphens. Have you created/modified output.xml programmatically or possibly modified results with pre-Rebot modifiers or otherwise?

Nope, I did not created/modified output.xml/result programmatically or otherwise.

I noticed several places with format like:

```
'%Y-%m-%d %H:%M:%S.%f'
```

, so one of such place may cause issue. Issue is very elusive and does not reproduce often, but sometimes happens.

pekkaklarck:
I'd need to know a bit more about the problem. Preferably there should be a test that demonstrates it and is fixed by this change.

pekkaklarck:
A problem with the proposed change is that `seps` not being `None` would mean that code would normalize all timestamps. This code is run _very_ often so any such extra work would slow down execution. We simply cannot do it just in case it would avoid problems in some strange case. Much better to analyze why that problem occurs and fix the root cause.

Issue #3492, opened 2020-03-06
Dialogs library bring to front doesn't work in Windows Server 2016


jpeltonen:
Dialog was left minimized after calling "Execute Manual Step". 

Could fix by modifying C:/Python368/Lib/site-packages/robot/libraries/dialogs_py.py (commented lines are original which didn't work)

Original code:
```
    def _bring_to_front(self):
        self.attributes('-topmost', True)
        self.attributes('-topmost', False)
```
Fix that seem to work:

```
    def _bring_to_front(self):
        self.lift()
        self.attributes('-topmost', True)
        self.after_idle(self.attributes, '-topmost', False)
```
Pull request to fix this #3496

pekkaklarck:
Do you mean this can be fixed by commenting code? If not, please provide a diff showing the actual changes or, even better, a pull request.

jpeltonen:
No this is just issue. Not pull request yet. Don't know if it depends on operating system also and how it works with Linux.

pekkaklarck:
This was fixed by PR #3496. Thanks @jpeltonen!

PR #3496, opened 2020-03-07
Update dialogs_py.py bring to front

jpeltonen:
Bring to front wasn't working in Windows server 2016

this commit fixes issue #3492


pekkaklarck:
I tested this locally on Linux and Windows and everything worked fine. I cannot reproduce the original problem but I trust you that the fix fixes it.

PR #3504, opened 2020-03-15
Acceptance tests on Jython 2.7.2 with Java >= 9

tysonite:
When running Jython >= 2.7.2 on Java >= 9, unexpected printouts are present on stdout, set JAVA_OPTS to `--add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/sun.nio.ch=ALL-UNNAMED` to avoid that.

References #3501.

### DIFF
--- a/atest/interpreter.py
+++ b/atest/interpreter.py
@@ -31,6 +31,7 @@ class Interpreter(object):
         self.name = name
         self.version = version
         self.version_info = tuple(int(item) for item in version.split('.'))
+        self.java_version = self._get_java_version()
 
     def _get_interpreter(self, path):
         path = path.replace('/', os.sep)
@@ -47,6 +48,23 @@ class Interpreter(object):
         name = name if 'PyPy' not in output else 'PyPy'
         version = re.match(r'\d+\.\d+\.\d+', version).group()
         return name, version
+
+    def _get_java_version(self):
+        if not self.is_jython:
+            return (-1, -1)
+        try:
+            # platform.java_ver() returns Java version in a format:
+            # ('9.0.7.1', 'Azul Systems, Inc.', ('OpenJDK 64-Bit Server VM', '9.0.7.1+1', 'Azul Systems, Inc.'), ('Windows 10', '10.0', 'amd64'))
+            # ('11.0.6', 'Ubuntu', ('OpenJDK 64-Bit Server VM', '11.0.6+10-post-Ubuntu-1ubuntu118.04.1', 'Ubuntu'), ('Linux', '4.4.0-18362-Microsoft', 'amd64'))
+            # ('1.8.0_121', 'Oracle Corporation', ('Java HotSpot(TM) 64-Bit Server VM', '25.121-b13', 'Oracle Corporation'), ('Windows 10', '10.0', 'amd64'))
+            script = 'import platform; print(platform.java_ver()[0])'
+            output = subprocess.check_output(self.interpreter + ['-c', script],
+                                             stderr=subprocess.STDOUT,
+                                             encoding='UTF-8')
+        except (subprocess.CalledProcessError, FileNotFoundError):
+            raise ValueError('Invalid interpreter: %s' % self.path)
+        major, minor = output.strip().split('.', 2)[:2]
+        return (int(major), int(minor))
 
     @property
     def os(self):
@@ -117,6 +135,19 @@ class Interpreter(object):
         return tools_jar
 
     @property
+    def java_opts(self):
+        if not self.is_jython:
+            return None
+        java_opts = os.environ.get('JAVA_OPTS')
+        if self.version_info[:3] >= (2, 7, 2) and self.java_version[0] >= 9:
+            if not java_opts:
+                java_opts = ''
+            # https://github.com/jythontools/jython/issues/171
+            if '--add-opens' not in java_opts:
+                java_opts += ' --add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/sun.nio.ch=ALL-UNNAMED'
+        return java_opts
+
+    @property
     def is_python(self):
         return self.name == 'Python'
 
@@ -173,7 +204,11 @@ class Interpreter(object):
         return self.interpreter + [join(ROBOT_PATH, 'tidy.py')]
 
     def __str__(self):
-        return '%s %s on %s' % (self.name, self.version, self.os)
+        if self.is_jython:
+            java_version = '(Java %s) ' % '.'.join(str(ver_part) for ver_part in self.java_version)
+        else:
+            java_version = ''
+        return '%s %s %son %s' % (self.name, self.version, java_version, self.os)
 
 
 class StandaloneInterpreter(Interpreter):
--- a/atest/interpreter.py
+++ b/atest/interpreter.py
@@ -31,7 +31,7 @@ class Interpreter(object):
         self.name = name
         self.version = version
         self.version_info = tuple(int(item) for item in version.split('.'))
-        self.java_version = self._get_java_version()
+        self.java_version_info = self._get_java_version_info()
 
     def _get_interpreter(self, path):
         path = path.replace('/', os.sep)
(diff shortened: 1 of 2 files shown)


atest/interpreter.py
```diff
@@ -48,6 +49,23 @@ def _get_name_and_version(self):
...
+            # ('11.0.6', 'Ubuntu', ('OpenJDK 64-Bit Server VM', '11.0.6+10-post-Ubuntu-1ubuntu118.04.1', 'Ubuntu'), ('Linux', '4.4.0-18362-Microsoft', 'amd64'))
+            # ('1.8.0_121', 'Oracle Corporation', ('Java HotSpot(TM) 64-Bit Server VM', '25.121-b13', 'Oracle Corporation'), ('Windows 10', '10.0', 'amd64'))
+            script = 'import platform; print(platform.java_ver()[0])'
+            output = subprocess.check_output(self.interpreter + ['-c', script],
+                                             stderr=subprocess.STDOUT,
+                                             encoding='UTF-8')
+        except (subprocess.CalledProcessError, FileNotFoundError):
+            raise ValueError('Invalid interpreter: %s' % self.path)
+        major, minor = output.strip().split('.', 2)[:2]
+        return (int(major), int(minor))
```
MaciejWiczk: I believe parentheses around `(int(major), int(minor))` are redundant

tysonite: Yeah. I always forget about that. Fixed.

atest/interpreter.py
```diff
@@ -48,6 +49,23 @@ def _get_name_and_version(self):
...
         return name, version
 
+    def _get_java_version(self):
+        if not self.is_jython:
+            return -1, -1
+        try:
+            # platform.java_ver() returns Java version in a format:
+            # ('9.0.7.1', 'Azul Systems, Inc.', ('OpenJDK 64-Bit Server VM', '9.0.7.1+1', 'Azul Systems, Inc.'), ('Windows 10', '10.0', 'amd64'))
+            # ('11.0.6', 'Ubuntu', ('OpenJDK 64-Bit Server VM', '11.0.6+10-post-Ubuntu-1ubuntu118.04.1', 'Ubuntu'), ('Linux', '4.4.0-18362-Microsoft', 'amd64'))
+            # ('1.8.0_121', 'Oracle Corporation', ('Java HotSpot(TM) 64-Bit Server VM', '25.121-b13', 'Oracle Corporation'), ('Windows 10', '10.0', 'amd64'))
```
pekkaklarck: I don't think this comment is needed. At least it doesn't need to be this verbose but I'd say those interested in `platform.java_ver()` return value can check the docs.

atest/interpreter.py
```diff
@@ -116,6 +134,19 @@ def classpath(self):
...
         return tools_jar
 
+    @property
+    def java_opts(self):
+        if not self.is_jython:
+            return None
+        java_opts = os.environ.get('JAVA_OPTS')
+        if self.version_info[:3] >= (2, 7, 2) and self.java_version[0] >= 9:
+            if not java_opts:
+                java_opts = ''
```
pekkaklarck: The default could be given to `os.environ.get()`. If the idea is to return `None` with Jython if not 2.7.2 on Java 9+, then it would be good to do that explicitly. I guess returning an empty string in that case is fine, though.

tysonite: The idea was to append `--add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/sun.nio.ch=ALL-UNNAMED` in case `JAVA_OPTS` contains something already given by user, or only set in case it is empty (or not set).

atest/interpreter.py
```diff
@@ -48,6 +49,23 @@ def _get_name_and_version(self):
...
+            # ('11.0.6', 'Ubuntu', ('OpenJDK 64-Bit Server VM', '11.0.6+10-post-Ubuntu-1ubuntu118.04.1', 'Ubuntu'), ('Linux', '4.4.0-18362-Microsoft', 'amd64'))
+            # ('1.8.0_121', 'Oracle Corporation', ('Java HotSpot(TM) 64-Bit Server VM', '25.121-b13', 'Oracle Corporation'), ('Windows 10', '10.0', 'amd64'))
+            script = 'import platform; print(platform.java_ver()[0])'
+            output = subprocess.check_output(self.interpreter + ['-c', script],
+                                             stderr=subprocess.STDOUT,
+                                             encoding='UTF-8')
+        except (subprocess.CalledProcessError, FileNotFoundError):
+            raise ValueError('Invalid interpreter: %s' % self.path)
+        major, minor = output.strip().split('.', 2)[:2]
+        return int(major), int(minor)
```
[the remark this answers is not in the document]
tysonite: There are 3 different cases I know about at this point of time:
* `'9.0.7.1`
* `'11.0.6'`
* `1.8.0_121`

I prefer to cut everything off after seconds dot because of following reasons:
- we don't need that, it just doesn't make sense to invest more into figuring out all possible combinations and parsing them.
- still unclear what kind of patch-level format can exist in other cases

atest/interpreter.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, path, name=None, version=None):
         self.name = name
         self.version = version
         self.version_info = tuple(int(item) for item in version.split('.'))
+        self.java_version = self._get_java_version()
```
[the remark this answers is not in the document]
tysonite: Fixed.

atest/interpreter.py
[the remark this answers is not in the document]
tysonite: Reduced it a bit.

atest/interpreter.py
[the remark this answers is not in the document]
pekkaklarck: I'm fine ignoring patch level info. Simply getting the third component and splitting the part after possible `_` woudln't be too complicated, though.

atest/interpreter.py
pekkaklarck: I meant that instead of
```python
java_opts = os.environ.get('JAVA_OPTS')
if not java_opts_
    java_opts = ''
```
we could just have
```python
java_opts = os.environ.get('JAVA_OPTS', '')
```
Appending `--add-opens` to possibly existing `JAVA_OPTS` is a good idea.

tysonite: Yeah. I did that in this way. Please, have a look at last commit.

atest/interpreter.py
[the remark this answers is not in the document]
tysonite: I though about this a bit more. I am still not confident that this worth the effort taking into account unknown variations of version info after second dot. Let's take this part as is now, and return back only if needed.

I also think that this is really Jython responsibility to provide an API for returning various parts of Java version in order 3rd party applications can deal with that information in a common way.

pekkaklarck:
Looks good to me. Thanks @tysonite!

pekkaklarck:
By the way, does this only occur with Jython 2.7.2 and not with 2.7.1 or 2.7.0? I've been thinking this is only related to Java version but I can be wrong. In that case it would be interesting to know what has changed on Jython side.

tysonite:
> By the way, does this only occur with Jython 2.7.2 and not with 2.7.1 or 2.7.0? I've been thinking this is only related to Java version but I can be wrong. In that case it would be interesting to know what has changed on Jython side.

Jython 2.7.2 is a first version that could be run on Java >=9. On Java 8, both Jython 2.7.1 and 2.7.2 pass tests.

pekkaklarck:
I see. JAVA_OPTS could then be enabled regardless Jython version. Not a big deal.

tysonite:
Not really. As 2.7.1 does not support Java >= 9 (means runs on Java <= 8), setting JAVA_OPTS lead to Java error as Java does know nothing about `--add-opens`. And as 2.7.2 can still be run on Java 8, there will be the same error from Java. The only safe way to set it on Java >= 9 and Jython >= 2.7.2

pekkaklarck:
I meant we could set JAVA_OPTS if Java > 8 without checking what is the Jython version.

tysonite:
Oh. Got it. Yes, you are right.

Issue #3451, opened 2020-02-03
Expose test line number via listener API v2


pekkaklarck:
Nowadays executed `TestCase` objects have line numbers available (#549). This means they are automatically exposed via listener API v3 and we should also expose them via listener API v2.

Unfortunately we cannot easily expose keyword line numbers via listeners. API v3 doesn't even have the `start/end_keyword` methods due to technical reasons (#3296) and same reasons prevent exposing them via API v2.

boakley:
_"Unfortunately we cannot easily expose keyword line numbers via listeners."_ - this seems to make no sense. If you have the line numbers in the object, and the objects are available when the listener is called, why can't they be exposed to the listener? Why can't you just add additional keys in the `attributes` argument? 

This is disappointing, because it makes the addition of line numbers virtually useless.

boakley:
I did some quick experimenting, and it _seems_ that to support listerner v2 only requires changing a dozen or so lines of code. Admittedly, I have a very, very shallow understanding of the internals of robot so maybe I'm missing something. However, I was able to write a v2 listener that could log the source file and line number of both a library keyword and user keywords in the test file and in a resource file. 

Of course, that "dozen lines" will balloon into a hundred or so when we add in tests and documentation of course, but at it seems like a very straightforward change at first blush. So I'm wondering if you're open to a PR, or are you simply opposed to exposing line numbers to a v2 listener for a philosophical reason.

pekkaklarck:
There's absolutely no philosophical reason why line numbers weren't added to `start/end_keyword` methods of v2 listeners. When I looked it at earlier, it just didn't look too easy. The main reason is that the code creating the `attributes` dictionary gets results objects that don't have line numbers. It probably wouldn't be too hard to add line numbers to those objects but I don't think _that_ is a good idea. It would be strange if keyword result objects had line numbers during execution but not if they are read from output.xml. And I don't think it makes sense to add that info there. Anyway, if you want to work on this it would be great.

boakley:
I may submit a PR, though the only solution seems to be to add line numbers to the result objects and you seem opposed to that. I don't want to put in a bunch of effort just to have it rejected.

pekkaklarck:
There are two ways how to implement this:

1. Add line numbers to result objects. They should be added to all result objects, not only to keywords. They should then also be added to output.xml to be able to construct same result objects based on it. The problem is that this will increase the output.xml size and slow down processing it.

[...]

Issue #3296, opened 2019-09-04
Support keywords and control structures with listener version 3


bergstenarn:
I would like to see support for start_keyword and end_keyword in listener version 3.

pekkaklarck:
The problem implementing this is that on the execution side there currently isn't any place where we could call `start/end_keywords` so that they would get data and result objects similarly as other listener v3 methods get. Changing that would require quite big changes to the code and doing them just to get this feature implemented hasn't been considered worth the effort.

In RF 4.0 the plan is to add new features like IF/ELSE (#3074) that require big changes to the execution side. When doing them, this issue definitely needs to be taken into account. Thus assigned this one to the v4.0 scope.

Snooz82:
i really like this feature. 
including line numbers for start and end keyword!
This would be a huge enabler for debuggers!

dakaizou:
Hi @bergstenarn and @Snooz82 ,
I'm not sure what those problems mentioned by @pekkaklarck are. Maybe there are some potential issues, but I patched the keyword runners to get something similar, and it worked fine for me:

```python
from functools import wraps

import robot
from robot.running.librarykeywordrunner import LibraryKeywordRunner
from robot.running.userkeywordrunner import UserKeywordRunner

ROBOT_LISTENER_API_VERSION = 3


def start_keyword(kw, context):
    print(kw.lineno)
    print(kw.source)


def end_keyword(kw, context):
    print(kw)


def patch_kwrunner(runner):
    old_run = runner.run

    @wraps(runner)
    def wrapper(self, kw, context):
        start_keyword(kw, context)
        try:
            if False:
                # bypassing this keyword
                return self.dry_run(kw, context)
            return old_run(self, kw, context)
        finally:
            end_keyword(kw, context)

    runner.run = wrapper


def start_suite(suite, result):
    patch_kwrunner(LibraryKeywordRunner)
    patch_kwrunner(UserKeywordRunner)


if __name__ == "__main__":
    robot.run("main_suite.robot", listener=__file__)


```

mkorpela:
Related draft pull request #3672

damies13:
Hi,

rfswarm agent uses the following version 2 api methods:

- start_suite
- log_message
- end_keyword

So not having "end_keyword" in the version 3 listener API is preventing me upgrading rfswarm from the version 2 API, fortunately the version 2 API works in Robot Framework V3 and hopefully won't be depreciated be in Robot Framework V4?

[...]

PR #3518, opened 2020-03-24
Add 'lineno' and 'source' attributes to V2 listener methods

boakley:
This is related to issue #3451, even though that issue was closed. 

This does _not_ add the source and lineno information to output.xml. While that might be useful, it would bloat the size of the xml considerably. I think it would be best to get the logic necessary to include the information in the listener, and then make the problem of saving the information to output.xml as a separate issue.

### DIFF
--- a/atest/testresources/listeners/attributeverifyinglistener.py
+++ b/atest/testresources/listeners/attributeverifyinglistener.py
@@ -14,8 +14,8 @@ OUTFILE = open(os.path.join(os.getenv('TEMPDIR'), 'listener_attrs.txt'), 'w')
 START = 'doc starttime '
 END = START + 'endtime elapsedtime status '
 SUITE = 'id longname metadata source tests suites totaltests '
-TEST = 'id longname tags critical template originalname lineno '
-KW = ' kwname libname args assign tags type '
+TEST = 'id longname tags critical template originalname source lineno '
+KW = ' kwname libname args assign tags type lineno source '
 EXPECTED_TYPES = {'tags': [basestring], 'args': [basestring],
                   'assign': [basestring], 'metadata': {basestring: basestring},
                   'tests': [basestring], 'suites': [basestring],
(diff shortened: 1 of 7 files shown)


pekkaklarck:
Getting source information via the listener interface would be very valuable, but unfortunately the approach used by this PR is not something I want to use. Adding source information to the result side objects has some major problems:

- Unless we save them also to output.xml and read them there, they would only be available during execution and not when processing results after execution. This would be highly inconsistent behavior.

- Saving source information to output.xml is pretty much out of the question. The main problem is that it would increase the output.xml size considerably and reading this information would also slow down processing them. Another problem is that source information, especially paths, make sense only on the machine where tests have been run. If using relative paths, they would be broken also locally if files are moved.

- As far I can see, there are no other benefits in adding source information to the result objects other than making it possible to use it with `start/end_keyword` listener methods.

I believe the only reason the result objects have been used for this purpose is that at the moment there is no other simple way to get the source information to `start/end_keyword` methods. The reason for that is that the execution code related to this is somewhat messed up, and the same reason prevents also adding `start/end_keyword` methods to the listener API v3. A much better solution to this problem is cleaning up that mess. That won't happen in RF 3.2, but it is a high priority enhancement planned for RF 4.0 (#3296).

It can be argued that we should anyway add source info to result objects to get this important enhancement implemented. Possibly it could then be removed when there's a better way to get this feature implemented. After thinking this quite hard, I've nevertheless decided not to do that. It would just add extra work (incl. deprecation) to further releases. I rather spent that time and energy for cleaning up the execution side mess. If it turns out that this feature would be very important, we can even consider creating RF 3.3 with #3296 and this feature as the main enhancements.

pekkaklarck:
I submitted issue #3538 about adding this feature in RF 4.0 (or possibly already in RF 3.3). I linked this PR from there to give those people who need this feature badly a possibility to take it into use. It would be great if you @boakley could merge the changes from master to this branch to get also the latest changes from there. If you want, you can also make your version more easily available for possibly interested users.

pekkaklarck:
This PR won't be merged in the current form so I close it.

boakley:
There's no point in me merging features from master and making it more readily available if this is a dead end.

Issue #3525, opened 2020-03-30
Update LibDoc XML_Schema to typed Schema 


Snooz82:
Hi Pekka,

i reviewed you schema file and decided to rewrite it as typed schema.
it is not as deep nested and easier to maintain.
I also added enumerations for some values.
These should be reviewed.

your original schema had some bugs.

- `agruments` tags could only have one `arg` tag
- `tags` tags could only have one `tag` tag
-  `init` was not supported

i also added two schema files.
one for Libdocs until 3.1.2 (libdoc.01.xsd) and one for 3.2 (libdoc.02.xsd)

please check the enums and if i missed anything.

i was not sure if `init` could also have a `source` like you defined in keywords or only a `lineno`

Cheers

Snooz82:
This should fix issues with #3523

pekkaklarck:
Great that we have schema expert here! My knowledge about them is somewhat limited and I just created the original schema with some online tool and edited it. Due to my limited knowledge I'm not going to review your versions and just trust they are good (and then blame you if someone reports bugs later :smiling_imp:).

Good catches with those bugs. Also `init` can have `source`. The code writing the spec files is here: https://github.com/robotframework/robotframework/blob/master/src/robot/libdocpkg/xmlwriter.py

I'll re-open #3523 and close this one. Could you @Snooz82 create a PR with your updates?

Issue #3523, opened 2020-03-28
Add spec version to Libdoc spec files


pekkaklarck:
We are making several changes to Libdoc spec file elements and attributes (#3522, #3507, #3498) in RF 3.2 and are also going to create a schema for them (#3520). To make it easier for external tools to handle future changes, we should add the spec version to the spec files as well.

My thinking is that we should add `specversion` attribute to the top level `keywordspec` element. The value should be a positive integer that is incremented between RF versions when changes are made. The version with RF 3.2 should be `2`.

PR #3526, opened 2020-03-30
Rewrote LibDoc XML-Schema files for specversion 1 and 2 for #3525

Snooz82:
please see issue #3523


MaciejWiczk:
> It would be good to have a comment telling that "scope" and "namedargs" elements will be removed in the future and attributes with same names should be used instead. I had such comment in my version.

@pekkaklarck why namedargs and scope are going to be removed? I recall that lately scope was added to library decorator.

doc/schema/libdoc.01.xsd
```diff
@@ -0,0 +1,76 @@
...
+      <xs:pattern value="\d{8} ([0-1]\d|2[0-3]):[0-5]\d:[0-5]\d" />
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:simpleType name="SupportNamedArgs">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="yes" />
+      <xs:enumeration value="no" />
+    </xs:restriction>
+  </xs:simpleType>
+</xs:schema>
```
pekkaklarck: Newline needed.

doc/schema/libdoc.02.xsd
```diff
@@ -0,0 +1,93 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+  <xs:element name="keywordspec" type="KeywordSpec" />
+  <xs:complexType name="KeywordSpec">
+    <xs:sequence>
+      <xs:element name="version" type="xs:string" />
+      <xs:element name="scope" type="LibraryScope" minOccurs="0" />
+      <xs:element name="namedargs" type="SupportNamedArgs" minOccurs="0" />
```
pekkaklarck: This comment from by original schema should be added here before the above two lines:

``` 
<!-- Both "scope" and "namedargs" elements are deprecated.
      Use "scope" and "namedargs" attributes instead. -->
```

Otherwise it's impossible to know why we have "scope" and "namedargs" elements and attributes and which one to use.

doc/schema/libdoc.02.xsd
```diff
@@ -0,0 +1,93 @@
...
+    </xs:sequence>
+    <xs:attribute name="name" type="xs:string" use="required" />
+    <xs:attribute name="type" type="LibraryType" use="required" />
+    <xs:attribute name="format" type="DocumentationFormat" use="required" />
+    <xs:attribute name="scope" type="LibraryScope" use="required" />
+    <xs:attribute name="namedargs" type="xs:boolean" use="required" />
+    <xs:attribute name="generated" type="GenerationDate" use="required" />
+    <xs:attribute name="specversion" type="SpecVersion" use="required" />
+    <xs:attribute name="source" type="xs:string" use="optional" />
+    <xs:attribute name="lineno" type="xs:positiveInteger" use="optional" />
```
pekkaklarck: There could be a comment before the above two lines about "source" and "lineno" being omitted if getting them from the library has failed for some reason.

doc/schema/libdoc.02.xsd
```diff
@@ -0,0 +1,93 @@
...
+  <xs:complexType name="Keyword">
+    <xs:sequence>
+      <xs:element name="arguments" type="Arguments" />
+      <xs:element name="doc" type="xs:string" />
+      <xs:element name="tags" type="Tags" />
+    </xs:sequence>
+    <xs:attribute name="name" type="xs:string" use="required"/>
+    <xs:attribute name="deprecated" type="xs:boolean" use="optional" />
+    <xs:attribute name="source" type="xs:string" use="optional" />
+    <xs:attribute name="lineno" type="xs:positiveInteger" use="optional" />
```
pekkaklarck: Here we could have a comment that "source" and "lineno" are omitted if getting them from the library has failed and "source" is also omitted if it is same as the library has. Especially the latter piece of information is important when creating tools that use this info for something.

doc/schema/libdoc.02.xsd
```diff
@@ -0,0 +1,93 @@
...
+      <xs:maxInclusive value="2" />
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:simpleType name="SupportNamedArgs">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="yes" />
+      <xs:enumeration value="no" />
+    </xs:restriction>
+  </xs:simpleType>
+</xs:schema>
```
pekkaklarck: Newline needed.

doc/schema/libdoc.02.xsd
```diff
@@ -0,0 +1,93 @@
...
+  </xs:simpleType>
+  <xs:simpleType name="LibraryType">
+    <xs:restriction base="xs:string">
+      <xs:enumeration value="library" />
+      <xs:enumeration value="resource" />
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:simpleType name="GenerationDate">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="\d{8} ([0-1]\d|2[0-3]):[0-5]\d:[0-5]\d" />
```
pekkaklarck: IMHO this is overly complicated and just leaves readers scratch their head. `\d{8} \d\d:\d\d:\d\d` would be a lot easier to understand. Notice that this schema is mainly for documentation purposes and thus more for humans than for machines. Do you have any use case where the exact time validation would bring benefits?

pekkaklarck: Alternatively, should we change the generation time to a date time that is valid `xsd:dateTime`? I seriously doubt anyone uses the time for anything at the moment, so changing it ought to be pretty well backwards compatible. I wouldn't change it in a minor release but in this major release with other spec changes that ought to be fine.

pekkaklarck: Changed the date format in #3528. This part of the spec should now be updated accordingly. I f I understand this correctly, the whole `GenerationDate` element can be removed and `xsd:dateTime` be used in its place.

This change obviously doesn't affect the other schema. In there I'd still simplify the validation so that it's just `\d{8} \d\d:\d\d:\d\d`.

doc/schema/libdoc.01.xsd
Snooz82: Done

doc/schema/libdoc.02.xsd
Snooz82: could you please put a comment there? You have write rights. I don't get the need for this comment, because these elements are optional, so the user have to expect that these can be not there.. for what reason ever.

doc/schema/libdoc.02.xsd
Snooz82: please put your comment there. This would definitely have a better wording as if i do it.

doc/schema/libdoc.02.xsd
Snooz82: i like this change!
done!

Snooz82:
Validates with the new ISO dateTime format.

doc/schema/libdoc.02.xsd
pekkaklarck: I can do that. I also agree that here on the library level a comment doesn't add so much value as source and lineno are already marked optional. On the keyword level knowing that source is omitted also if it's the same as on the library level (which is the common case) is valuable information. I also just realized we should mention that the source path is relative to the location where the spec file is created to.

pekkaklarck:
Thanks for great PR @Snooz82!

PR #3527, opened 2020-03-31
Create lint_python.yml

cclauss:
As requested at https://github.com/robotframework/robotframework/pull/3313#issuecomment-606458304  This is a exact replica of #3421

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,0 +1,10 @@
+name: lint_python
+on: [push, pull_request]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v1
+      - run: pip install flake8
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
--- a/utest/utils/test_normalizing.py
+++ b/utest/utils/test_normalizing.py
@@ -194,9 +194,9 @@ class TestNormalizedDict(unittest.TestCase):
 
     def test_unicode(self):
         nd = NormalizedDict({'a': u'\xe4', u'\xe4': 'a'})
-        if PY2:
+        try:
             assert_equal(unicode(nd), "{'a': u'\\xe4', u'\\xe4': 'a'}")
-        else:
+        except NameError:
             assert_equal(str(nd), u"{'a': '\xe4', '\xe4': 'a'}")
 
     def test_update(self):
--- a/atest/testdata/keywords/Annotations.py
+++ b/atest/testdata/keywords/Annotations.py
@@ -1,5 +1,7 @@
+# flake8: noqa
+
 def annotations(arg1, arg2: str):
     return ' '.join(['annotations:', arg1, arg2])
 
 def annotations_with_defaults(arg1, arg2: 'has a default' = 'default'):
-   return ' '.join(['annotations:', arg1, arg2])
+    return ' '.join(['annotations:', arg1, arg2])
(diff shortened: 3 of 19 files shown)


pekkaklarck:
It's good that CI runs this automatically. The [run fails](https://github.com/robotframework/robotframework/pull/3527/checks?check_run_id=548030879) and those failures should obviously be fixed before this PR could be accepted. On a quick look things that should be done are:

- Exclude atest and utest directories altogether (it seems this may have already been addressed while I'm writing this).
- Use `unicode` via `robot.utils` everywhere. We already expose it there but apparently don't use in all places.
- Add `long` to `robot.utils` same way as `unicode`. Use it where appropriate.
- Add magic comments to `unicode` and `long` under `robot.utils` to ignore the linting error if needed.

.github/workflows/lint_python.yml
```diff
@@ -7,4 +7,6 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
       - run: pip install flake8
-      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: |
+          EXCLUDE=./.*,./atest/testdata/*
+          flake8 . --count --exclude=$EXCLUDE --select=E9,F63,F7,F82 --show-source --statistics
```
pekkaklarck: Could this just be run for `src/robot` instead of of trying to exclude other directories?

cclauss: Do we just ignore problems that appear in `setup.py` for instance?  I think it is safer to only exclude certain test directories where you are consciously trying to break things.  Also, should we just paper over the fact that this code does not follow the Python porting best practice [___use feature detection instead of version detection___](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection)?

utest/utils/test_normalizing.py
```diff
@@ -194,9 +194,9 @@ def test_str(self):
 
     def test_unicode(self):
         nd = NormalizedDict({'a': u'\xe4', u'\xe4': 'a'})
-        if PY2:
+        try:
             assert_equal(unicode(nd), "{'a': u'\\xe4', u'\\xe4': 'a'}")
-        else:
+        except NameError:
```
pekkaklarck: **NO!** We definitely absolutely never ever want to have `try/except` in test code like this! It can too easily cause false positives. Also, `if PY2` makes it a lot more clear why we have two different assertions.

The whole `utest` directory should be excluded from linting altogether.

.github/workflows/lint_python.yml
pekkaklarck: 1. If you think it's important to test `setup.py`, run linter for it in addition to `src/robot`. I want linting to be opt-in, not opt-out, in general and I definitely don't want it for *any* code under atest or utest directories.

2. I'm personally even sure does it make sense to lint `setup.py`. Do you have examples of problems that linter could find that wouldn't be found otherwise? Now it only found a non-bug and an annoying magic comment is needed to silence it. 

3. This code base uses both version detection and feature detection, depending on which is more practical. Most of the Python 2/3 differences are such that version detection is 100% safe, results with cleaner code, and is also a bit faster (`try/except` isn't free). Our `PY3` constant is defined so that it will be true also with possible Python 4, so as long it doesn't break compatibility with Python 3 we are safe. And if it breaks, we need to fix our code anyway. Finally, this stuff is definitely not going to change before RF 4.0 when we remove Python 2 support altogether.

pekkaklarck:
I got here to mention that I simplified `setup.py` so that it doesn't anymore need `exec()` that caused the linting error in e32aee98db1e5e0639e09fdbe466776303be83a8, but since you closed the PR I guess that doesn't anymore matter.

PR #3530, opened 2020-03-31
Collect acceptance tests coverage data

tysonite:
Collect coverage data on CPython, excluding PyPy as it has some issues running acceptance tests with coverage tool as well as it takes too much time.

### DIFF
--- a/atest/interpreter.py
+++ b/atest/interpreter.py
@@ -15,16 +15,16 @@ def get_variables(path, name=None, version=None):
     return {'INTERPRETER': interpreter, 'UNICODE PREFIX': u}
 
 
-def InterpreterFactory(path, name=None, version=None):
+def InterpreterFactory(path, name=None, version=None, coverage=False):
     if path.endswith('.jar'):
-        return StandaloneInterpreter(path, name, version)
-    return Interpreter(path, name, version)
+        return StandaloneInterpreter(path, name, version, coverage)
+    return Interpreter(path, name, version, coverage)
 
 
 class Interpreter(object):
 
-    def __init__(self, path, name=None, version=None):
-        self.path = path
+    def __init__(self, path, name=None, version=None, coverage=False):
+        self.path = self._get_interpreter_path(path, coverage)
         self.interpreter = self._get_interpreter(path)
         if not name:
             name, version = self._get_name_and_version()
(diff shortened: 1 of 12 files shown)


atest/interpreter.py
```diff
@@ -39,7 +39,7 @@ def _get_interpreter(self, path):
 
     def _get_name_and_version(self):
         try:
-            output = subprocess.check_output(self.interpreter + ['-V'],
+            output = subprocess.check_output([self.interpreter[0]] + ['-V'],
```
pekkaklarck, line 42: This doesn't look right. I'm fairly certain it breaks running tests on Windows like `atest/run.py "py -2" atest/robot/...`.

tysonite: Might be. What would be the best way to solve that?

I can only think about looking for a “python -m” or “coverage” pattern with self.interpreter and use self.interpreter[0] if above present. 

BTW, what is “py -2”?

pekkaklarck: I don't have any immediate good ideas how to handle this. My priority at the moment is getting RF 3.2 rc 1 ready, but I can think more about this one after that. Might be better to use Slack for further discussion, though.

`py -2` is how you run Python 2 using the [Python launcher for Windows](https://docs.python.org/3/using/windows.html#python-launcher-for-windows).

tysonite: This now addressed via `--coverage` option.

atest/run.py
```diff
@@ -53,9 +57,9 @@
 '''.strip()
 
 
-def atests(interpreter, *arguments):
+def atests(coverage, interpreter, *arguments):
```
pekkaklarck, line 60: I'd prefer `def atest(interpreter, *arguments, coverage=False)` here. This is code is Python 3 only so keyword-only arguments are fine.

atest/run.py
```diff
@@ -119,9 +123,13 @@ def dos_to_long(path):
 
 
 if __name__ == '__main__':
+    coverage = False
+    if '--coverage' in sys.argv:
+        sys.argv.remove('--coverage')
+        coverage = True
```
pekkaklarck, line 129: It probably doesn't matter, but modifying global `sys.argv` directly feels a bit bad. Perhaps this could be changed to something like this:

```python
args = sys.argv[1:]
if not args or '--help' in args:
    print(__doc__)
    rc = 251
elif '--coverage' in args:
    args.remove('--coverage')
    rc = atests(*args, coverage=True)
else:
    rc = atests(*args)
sys.exit(rc)
```

atest/interpreter.py
```diff
@@ -15,16 +15,16 @@ def get_variables(path, name=None, version=None):
...
+    return Interpreter(path, name, version, coverage)
 
 
 class Interpreter(object):
 
-    def __init__(self, path, name=None, version=None):
-        self.path = path
+    def __init__(self, path, name=None, version=None, coverage=False):
+        self.path = self._get_interpreter_path(path, coverage)
         self.interpreter = self._get_interpreter(path)
```
pekkaklarck, line 28: Is it on purpose that `self.path` gets coverage configuration but `self.interpreter` doesn't? I don't remember how all this works together, but I thought `self.interpreter` is what is actually used when running tests.

pekkaklarck:
Coverage information for unit tests wasn't really used for anything and I removed collecting it some time ago. I don't think it makes sense to try collect coverage info for acceptance tests either.

PR #3543, opened 2020-04-11
Remove code coverage artifacts to speed up unit tests workflows

tysonite:


tysonite:
I think some checks are failing here because of https://github.com/robotframework/robotframework/commit/6a2155a5b68fd97172d2e492e1db8d0d3c0a86a9. Also, canceled checks are due to temporary GHA issue.

PR #3557, opened 2020-04-27
YAMLWarning fix. Deprecated function load() replaced with safe_load()…

jhanka:
YAMLLoadWarning fix: Use safe_load function insted of unsafe load() function to load YAML file.

PyYAML's load function has been unsafe since the first release in May 2006.
More info:
https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

### DIFF
--- a/src/robot/variables/filesetter.py
+++ b/src/robot/variables/filesetter.py
@@ -80,7 +80,7 @@ class YamlImporter(object):
                             'to be installed. Typically you can install it '
                             'by running `pip install pyyaml`.')
         if yaml.__version__.split('.')[0] == '3':
-            return yaml.load(stream)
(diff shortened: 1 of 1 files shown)


pekkaklarck:
We already use ` yaml.load()` or ` yaml.full_load()` depending on the PyYAML version. Does this affect only the former or should both be changed?

jhanka:
> We already use ` yaml.load()` or ` yaml.full_load()` depending on the PyYAML version. Does this affect only the former or should both be changed?

It affect only the yaml.load() function. yaml.full_load() works correctly.

pekkaklarck:
Great! Then this is already fixed when using newish PyYAML versions.

Issue #3238, opened 2019-07-11
Keyword arguments not used when resolving its documentation and tags


stdedos:
I have this test:
```robot
*** Settings ***

*** Variables ***
${test}   foobar
&{dict}   foo=bar   bar=baz   baz=foo

*** Test Cases ***
Test Kwargs
  ${my_user} =        Set Variable  damn
  Log My User         ${my_user}  ${test}
  Log My Other User   ${my_user}  ${test}

*** Keywords ***
Log My User
  [Documentation]   Log ${user} doing ${test}
  [Arguments]       ${user}   ${test}
  Log Many          ${user}   ${test}

Log My Other User
  [Documentation]   Log ${my_user} doing ${test}
  [Arguments]       ${my_user}  ${test}
  Log Many          ${my_user}  ${test}
```

and I see:

![image](https://user-images.githubusercontent.com/133706/61043480-0a753a00-a3df-11e9-9d39-7c5c6f8bb0d8.png)

I would expect that keyword's arguments are taken into account when resolving variables, if all pre-defined variables are taken into account too.

pekkaklarck:
Good catch. Same problem also with keyword tags. Ought to be easy to fix.

pekkaklarck:
I was wrong, this is surprisingly complicated to fix.

The problem is that if we wait until arguments have been added to variables before we resolve doc and tags, we have already called an internal `start_keyword` method that logs doc and tags to output.xml. We could move logging doc and tags to `end_keyword`, but that breaks `--flattenkeywords` because it must found tags before executed keywords in output.xml. This approach probably would work with doc, though.

I guess this could be handled by adding a new method like `kw_info` to the XML logger to log kw doc and tags. It could be called after `start_keyword` but before starting to execute internal keywords. Could also see could calling `start_keyword` be delayed, but that may be impossible due to error handling.

pekkaklarck:
Too much work compared to benefits to be included in RF 3.2.

VernonCrabtree:
Having tried to work on the old lexer, I highly appreciate the work done and was enthusiastic about taking another look at the logging control (particularly due to enforced holidays due to Covid-19).
We had discussed using tags for logging control and it occurred to me that this could, in turn, be controlled via arguments. However, I stumbled on this issue with this approach.
It is a pitty a kw_info method has not been implemented - this could be very helpful for logging control in the future.

[...]

PR #3566, opened 2020-05-01
Keyword arguments not used when resolving its documentation and tags …

VernonCrabtree:
…#3238
The code has been modified to resolve the arguments before executing the keyword. This required splitting the StatusReporter context manager so that any resolution errors could be put into the reporting context.
All tests have passed except Telnet tests.
New tests have been added for tags with parameters.

### DIFF
--- a/atest/robot/tags/tags_with_variables.robot
+++ b/atest/robot/tags/tags_with_variables.robot
@@ -1,0 +1,20 @@
+*** Setting ***
+Suite Setup       Run Tests    ${EMPTY}    tags/tags_with_variables.robot
+Resource          atest_resource.robot
+
+*** Test Case ***
+External variable not resolved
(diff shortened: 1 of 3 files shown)


pekkaklarck:
I don't have time to really look at the PR now, but on a quick look it seems rather complicated. It actually seems so complicated that the functionality provided by it may not be worth the increased maintenance cost. This is especially true because in the next RF major version we plan to refactor execution side code and we most likely need to touch this code as well.

VernonCrabtree:
Oh well - it was fun trying. I had realized it was complex, which is why I deliberately tried to contain the change to one module and reduced the complexity from where I had started. 
Maybe the acceptance tests can be used in the next release?
Also, when looking at refactoring, this code could be worth a quick look to help determine the issues that need to be addressed. It should be easier once python 2 no longer needs to be supported.

PR #3574, opened 2020-05-14
bugfix in log.html: retry if expanding of an element fails

DirkRichter:
On large log files log.html concurrent expanding can fail in rare cases due to conflicting concurrency (esp. if argument --expandkeywords is used). 

After many experiments avoiding such conflicts, i finally found a better, more easy and robust solution: just do a simple retry.


src/robot/htmldata/rebot/log.js
```diff
@@ -54,6 +54,12 @@ function expandElement(item) {
     var children = element.children('.children');
     // .css is faster than .show and .show w/ callback is terribly slow
     children.css({'display': 'block'});
+    // in rare cases on large logs concurrent expanding fails => retry
+    if (children.css('display')!='block') {
```
pekkaklarck, line 58: Please add spaces around `!=` to be consistent with rest of the code.

DirkRichter: done

DirkRichter:
now i restrict to max 3 retries. i also changed the new expand_keywords logic with a delay of 100ms, which also reduced the amount of conflicts.

pekkaklarck:
Looks good to me otherwise, but needing to specify the retry time in all functions that call `expandElement()` is a bit annoying. I'm not a JavaScript expert, but I think something like

pekkaklarck:
Looks good to me. Could you still submit an issue about this for tracking purposes?

DirkRichter:
https://github.com/robotframework/robotframework/issues/3585

pekkaklarck:
Thanks for the PR @DirkRichter!

Issue #3579, opened 2020-05-22
Enhance performance of selecting tests using `--include` and `--exclude`


vrpolakatcisco:
Hello.

In FD.io CSIT project we are running many robotframework tests. Imagine single robot invocation selecting hundreds out of tens of thousands test cases. Due to both historic inertia and desire for clarity, we ended up with a scheme where we construct really long tag expressions. Basically each test case is selected by a separate --include option, each ANDing ~8 single tags.

The downside is in robot taking too long to figure out which tests are selected.
Here is an example console output [0] with al the gory details (using robotframework==3.1.2 and CPython3.6.9.final.0-64 in case tat matters).
See the time gap between
  14:04:31  +++ robot
and
  15:07:45  ===

In middle term we are working on a solution which uses --test instead of --include.
But I think in long run it would be good if robotframework was faster also when dealing with tags.

[0] https://logs.fd.io/production/vex-yul-rot-jenkins-1/csit-vpp-perf-mrr-daily-master-2n-skx/898/console-timestamp.log.gz

vrpolakatcisco:
I have some ideas how to speed things up. I will upload a draft of a solution as a pull request, but I doubt I will have the skill (and time) to make it through the review process.

vrpolakatcisco:
So I managed to create https://github.com/robotframework/robotframework/pull/3580
I forgot about markdown rendering of commit messages, so you do not see which items have plus sign (improvements) and which have minus sign (missing features or possible bugs).

vrpolakatcisco:
My main question is whether it is an attempt that goes in a good direction, or some other approach is preferred.

pekkaklarck:
I don't have time to look at this in more detail right now so only quick comments:
- I'm really surprised that selecting tests is so slow that it causes problems.
- It would be nice if you could explain the problem in a bit more detail so that we didn't need to open some external log file. Most importantly, how much time is actually spent for selecting tests?
- If you already know where the performance issue is, please share that as well.
- I took a quick look at #3580 and was surprised how much changes were needed. I hope we could come up with something simpler.
- RF 3.2.2 is planned for early July so there's still plenty of time to get this fixed and included in it.

skhomuti:
We have the same problem. It takes about a minute to select even one test by a unique tag. Choosing a test by name did not improve the situation in any way, even using the full test name.
[...]

PR #3580, opened 2020-05-22
Issue #3579: Incomplete draft of a solution

vrpolakatcisco:
Employ memoization to speed up tag filtering.

- This is a first draft.
- Unit tests not edited at all.
- Not sure about storing to self._memo everywhere.
- Not sure if normalization is applied everywhere needed.
+ Normalization applied in most places.
+ Memo dictionary is passed from callers, created empty if needed.
+ Memo dictionary uses normalized pattern or tag as the key.
+ Multiple copies of the same pattern for SingleTagPattern do not duplicate matchers.
+ Each such matcher has memo dictionary to store (normalized) tag match results.
+ Calls to the same matcher with the same tag return the memoized result.
- Matcher memo should probably have different name than result memo.

Signed-off-by: Vratko Polak <vrpolak@cisco.com>

### DIFF
--- a/src/robot/model/filter.py
+++ b/src/robot/model/filter.py
@@ -40,12 +40,13 @@ class EmptySuiteRemover(SuiteVisitor):
 class Filter(EmptySuiteRemover):
 
     def __init__(self, include_suites=None, include_tests=None,
-                 include_tags=None, exclude_tags=None):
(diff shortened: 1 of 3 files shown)


pekkaklarck:
Based on the comment in #3579 this attempt didn't work.

Issue #2086, opened 2015-08-05
Relative order of messages and keywords is not preserved in log


pekkaklarck:
If a keyword has both messages and child keywords, they are always show in logs so that keywords are first. The reason this issue isn't typically seen is that most of the time keywords only have either child keywords or messages, not both. There are, however, at least two cases where both keywords and messages exists:
- When using `--loglevel trace`, all keywords get trace level messages logging arguments and return values. With user keywords both of these messages are listed after child keywords although arguments are logged before executing them.
- `Wait Until Keyword Succeeds` and some other special keywords that execute other keywords also log messages. All these messages are shown after all executed keywords.

Notice that a related problem that Rebot messed up the order when reading output.xml (#1867) has already been fixed. Fixing logs may unfortunately be a lot harder.

pekkaklarck:
No time for this in 3.0. Changes to log aren't backwards incompatible so this can be considered also in 3.0.x minor versions. The main problem is that required changes to Javascript code are quite big and thus need to be tested with various browsers and on various operating systems. Good thing is that we ought to be able to forget about IE 8 at this point.

testautomation:
I added below screenshots to visualize the effect of above described issue.

When a keyword has many nested keywords it can be hard to find the right trace/debug in log.html.  Thus it would be better if debug and trace information in log.html were right below the keyword they belong to instead of at the end of all nested keywords.

![image](https://user-images.githubusercontent.com/38438046/74331584-479a1400-4d94-11ea-89a7-980848b70820.png)

![image](https://user-images.githubusercontent.com/38438046/74331747-99db3500-4d94-11ea-9fc0-679349edf51b.png)

stdedos:
Well ... not exactly.

Why not have them at the order they arrived i.e. Arguments first, Return last, and everything else in-between?

pekkaklarck:
Of course kws/msgs should be in the order they occurred, that's what this issue is all about, but it's hard to fix and nobody has done it yet. Changing logic so that msgs are before kws would be trivial.

stdedos:
> Changing logic so that msgs are before kws would be trivial.

_I guess_ it could be a workaround; I'd prefer, however, that applying the workaround doesn't mean that this issue is abandoned.

pekkaklarck:
[...]

PR #3602, opened 2020-06-20
Fix incorrect suite teardown display position in log.html

jnhyperion:
Suite teardown should be dislpayed in the end of current suite block.

### DIFF
--- a/src/robot/htmldata/rebot/log.js
+++ b/src/robot/htmldata/rebot/log.js
@@ -32,13 +32,30 @@ function populateChildren(elementId, childElement, childrenNames) {
 
 function drawCallback(element, childElement, childrenNames) {
     return function () {
-        util.map(childrenNames, function (childName) {
-            var children = element[childName + 's']();
-            var template = childName + 'Template';
-            util.map(children, function (child) {
-                $.tmpl(template, child).appendTo(childElement);
-            });
-        });
+      var renderChildren = function (children, template) {
+          util.map(children, function (child) {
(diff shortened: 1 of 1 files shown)


jnhyperion:
Before fix:
Screen Shot 2020-06-20 at 4 36 08 PM

After fix:
Screen Shot 2020-06-20 at 4 36 39 PM

Snooz82:
I like this.
Would be nice to get thin into 3.2.2

pekkaklarck:
The current behavior is by design. If suite teardown is in its correct chronological position, it is nit seen when you look at the general suite info on top. I don't want to change that.

This has been discussed earlier few times and someone once proposed making this configurable with a cli option. I personally don't think that's worth the effort and I'm also little concerned that lot if configuration options increase complexity. That said, if someone provides a PR making this configurable I'd be happy to merge it. The first step would be submitting an issue about it.

pekkaklarck:
I would also expect implementation ro be a lot simpler. I didn't look the code now, but simply modifying suite template in log.html ought to be enough.

jnhyperion:
@pekkaklarck you're right, we have some similar issue for `TRACE` log: https://github.com/robotframework/robotframework/issues/2086

We should use a unified solution to fix them, my PR can only fix suite teardown which is not good enough.

I'll close my PR.

pekkaklarck:
Relative order of keywords and messages being wrong (#2086) is simply a bug and needs to be fixed (but is hard to). Suite teardown being shown at the top along with other suite information is by design.

Issue #3353, opened 2019-10-24
Tidy should have option to leave whitespaces as they are


mnojek:
I have my robot files formatted in a way that they are easy to read (I aligned code in columns). 
I need now to update the files (migration from 3.0.2 to 3.1.1 RF version) to include new FOR loop syntax but when I use tidy, all my custom spaces between the cells (keywords, vars etc.) are getting unified and the whole code is difficult to read. It would be great to have a switch in tidy for turning off the making off whitespaces so that I can preserve the ones I have. I didn't find solution for that. Maybe there is another script that does that (updates the code to the new syntax)?

pekkaklarck:
Unifying whitespace usage is one of Tidy's main features, but I can see disabling that and letting Tidy handle other cleanup could sometimes be useful. Tidy has been basically rewritten for RF 3.2 because the new parser (#3076) broke the old implementation totally. With the new Tidy implementation it shouldn't be too complicated to add this functionality, but at least I don't have time for that. Are you interested to look at that yourself?

There would be some some things to agree related to design:
- Should all whitespace be preserved or only whitespace inside settings, variables, test cases and keywords?
- How to enable that? One possibility is using the existing `--spacecount` like `--spacecount preserve`.

nojas10:
Thanks for the response, @pekkaklarck .
I can look into that but I cannot ensure that I will succeed.
Regarding your questions I will try to answer them when I will have time to investigate it and think about which approach to use it is the best.

pekkaklarck:
Great! It's probably best to start by looking at the current Tidy code. It will probably still be modified in the near future, but the overall design shouldn't change. We can also discuss this further on `#devel` channel on our Slack.

mnojek:
@pekkaklarck I finally found some time and created PR #3612 that implements this change. Please feel free to review it :)

pekkaklarck:
I finally had time to fully review PR #3612, including testing it locally. I like the functionality in general but there are some problems that still need to be addressed as I explain in my [comment to the PR](https://github.com/robotframework/robotframework/pull/3612#issuecomment-681829998).

pekkaklarck:
There are still [some problems with the PR](https://github.com/robotframework/robotframework/pull/3612#issuecomment-748620030) that should be addressed. It would be a lot simpler if spaces just weren't touched at all when they ought to be preserved.

[...]

PR #3612, opened 2020-07-12
Enable tidy to preserve whitespaces

mnojek:
I implemented changes in tidy according to issue #3353  
Now tidy accepts value `preserve` for option `--spacecount` as well as integers like before. The new value tells tidy to not mess with whitespaces and leave them as they are when producing output file. It is very helpful for RF coders that use whitespaces to align code vertically so that it is easier to read. I updated also help message to include this change and added corresponding unit and acceptance tests. Some update to documentation may also be required. I fixed also some typos.
The implementation may seem not very clever and pretty, but I am open to all criticism :)

### DIFF
--- a/src/robot/tidypkg/transformers.py
+++ b/src/robot/tidypkg/transformers.py
@@ -369,12 +369,12 @@ class Aligner(ModelTransformer):
 
     def __init__(self, short_test_name_length,
                  setting_and_variable_name_length,
-                 pipes_mode, space_count):
+                 pipes_mode, preserve_spaces):
         self.short_test_name_length = short_test_name_length
         self.setting_and_variable_name_length = \
             setting_and_variable_name_length
         self.pipes_mode = pipes_mode
-        self.space_count = space_count
+        self.preserve_spaces = preserve_spaces
 
     def visit_TestCaseSection(self, section):
         if len(section.header.data_tokens) > 1:
@@ -400,6 +400,8 @@ class Aligner(ModelTransformer):
         return statement
 
     def _should_be_aligned(self, tokens):
+        if self.preserve_spaces:
+            return False
         if not tokens:
             return False
         if len(tokens) == 1:
--- a/src/robot/tidypkg/transformers.py
+++ b/src/robot/tidypkg/transformers.py
@@ -218,9 +218,14 @@ class SeparatorNormalizer(ModelTransformer):
 
     def _normalize_spaces(self, index, token, line_length):
         if token.type == Token.SEPARATOR:
-            spaces = self.space_count * self.indent \
-                if index == 0 else self.space_count
-            token.value = ' ' * spaces
+            if self.space_count != 'preserve':
+                spaces = self.space_count * self.indent \
+                    if index == 0 else self.space_count
+                token.value = ' ' * spaces
+            else:
+                if index == 0:
+                    token.value = ' ' * 4 * self.indent
+
         # The last token is always EOL, this removes all dangling whitespace
         # from the token before the EOL
         if index == line_length - 2:
@@ -363,11 +368,13 @@ class ColumnWidthCounter(ModelTransformer):
 class Aligner(ModelTransformer):
 
     def __init__(self, short_test_name_length,
-                 setting_and_variable_name_length, pipes_mode):
+                 setting_and_variable_name_length,
+                 pipes_mode, space_count):
         self.short_test_name_length = short_test_name_length
         self.setting_and_variable_name_length = \
             setting_and_variable_name_length
         self.pipes_mode = pipes_mode
+        self.space_count = space_count
 
     def visit_TestCaseSection(self, section):
         if len(section.header.data_tokens) > 1:
@@ -386,9 +393,10 @@ class Aligner(ModelTransformer):
                             not in (Token.SEPARATOR, Token.EOL)]
             if self._should_be_aligned(value_tokens):
                 first = value_tokens[0]
-                first.value = first.value.ljust(
-                    self.setting_and_variable_name_length
-                )
+                if self.space_count != 'preserve':
+                    first.value = first.value.ljust(
+                        self.setting_and_variable_name_length
+                    )
         return statement
 
     def _should_be_aligned(self, tokens):
--- a/src/robot/tidypkg/transformers.py
+++ b/src/robot/tidypkg/transformers.py
@@ -225,7 +225,6 @@ class SeparatorNormalizer(ModelTransformer):
             else:
                 if index == 0:
                     token.value = ' ' * 4 * self.indent
-
         # The last token is always EOL, this removes all dangling whitespace
         # from the token before the EOL
         if index == line_length - 2:
@@ -393,7 +392,7 @@ class Aligner(ModelTransformer):
                             not in (Token.SEPARATOR, Token.EOL)]
             if self._should_be_aligned(value_tokens):
                 first = value_tokens[0]
-                if self.space_count != 'preserve':
+                if not self.preserve_spaces:
                     first.value = first.value.ljust(
                         self.setting_and_variable_name_length
                     )
--- a/utest/tidy/test_argument_validation.py
+++ b/utest/tidy/test_argument_validation.py
@@ -17,7 +17,7 @@ class TestArgumentValidation(unittest.TestCase):
         assert_equal(opts['spacecount'], 42)
 
     def test_invalid_space_count(self):
-        error = '--spacecount must be of value "preserve" or an integer greater than 1.'
+        error = '--spacecount must be either "preserve" or an integer greater than 1.'
         self._validate(spacecount='not a number', error=error)
         self._validate(spacecount='1', error=error)
         self._validate(spacecount='preserv', error=error)
--- a/utest/tidy/test_argument_validation.py
+++ b/utest/tidy/test_argument_validation.py
@@ -17,9 +17,14 @@ class TestArgumentValidation(unittest.TestCase):
         assert_equal(opts['spacecount'], 42)
 
     def test_invalid_space_count(self):
-        error = '--spacecount must be an integer greater than 1.'
+        error = '--spacecount must be of value "preserve" or an integer greater than 1.'
         self._validate(spacecount='not a number', error=error)
         self._validate(spacecount='1', error=error)
+        self._validate(spacecount='preserv', error=error)
+
+    def test_valid_alternative_space_count(self):
+        opts, _ = self._validate(spacecount='preserve')
+        assert_equal(opts['spacecount'], 'preserve')
 
     def test_inplace_and_recursive_cannot_be_used_together(self):
         self._validate(inplace=True, recursive=True,
--- a/atest/robot/tidy/tidy.robot
+++ b/atest/robot/tidy/tidy.robot
@@ -8,6 +8,9 @@ Tidying single test case file spaces -> spaces
     Run tidy with golden file and check result    ${EMPTY}    golden.robot
     Run tidy with golden file and check result    --spacecount 2    golden_two_spaces.robot
 
+Tidying single test case file with preserved spaces
+    Run tidy with golden file and check result    --spacecount preserve    golden.robot    input=golden.robot
+
 Tidying single test case file spaces -> pipes
     Run tidy with golden file and check result    --usepipes    golden_pipes.robot
 
--- a/doc/userguide/src/SupportingTools/Tidy.rst
+++ b/doc/userguide/src/SupportingTools/Tidy.rst
@@ -40,8 +40,9 @@ Options
                   option is used. Does not process referenced resource files.
  -p, --usepipes   Use a pipe character (`|`) as a column separator in the plain
                   text format.
- -s, --spacecount <number>
+ -s, --spacecount <number|preserve>
                   The number of spaces between cells in the plain text format.
+                  To leave whitespaces unchanged use value `preserve`.
                   Default is 4.
  -l, --lineseparator <native|windows|unix>
                   Line separator to use in outputs. The default is *native*.
--- a/src/robot/tidy.py
+++ b/src/robot/tidy.py
@@ -75,8 +75,9 @@ Options
                  are processed in-place similarly as when --inplace option
                  is used. Does not process referenced resource files.
  -p --usepipes   Use pipe ('|') as a column separator in the plain text format.
- -s --spacecount number
+ -s --spacecount number|preserve
                  The number of spaces between cells in the plain text format.
+                 To leave whitespaces unchanged use value 'preserve'.
                  Default is 4.
  -l --lineseparator native|windows|unix
                  Line separator to use in outputs. The default is 'native'.
@@ -258,7 +259,7 @@ class ArgumentValidator(object):
             if spacecount < 2:
                 raise ValueError
         except ValueError:
-            raise DataError('--spacecount must be an integer greater than 1.')
+            raise DataError('--spacecount must be of value "preserve" or an integer greater than 1.')
         return spacecount
 
 
--- a/src/robot/tidy.py
+++ b/src/robot/tidy.py
@@ -259,7 +259,7 @@ class ArgumentValidator(object):
             if spacecount < 2:
                 raise ValueError
         except ValueError:
-            raise DataError('--spacecount must be of value "preserve" or an integer greater than 1.')
+            raise DataError('--spacecount must be either "preserve" or an integer greater than 1.')
         return spacecount
 
 
--- a/src/robot/tidy.py
+++ b/src/robot/tidy.py
@@ -166,7 +166,7 @@ class Tidy(SuiteStructureVisitor):
         Aligner(self.short_test_name_length,
                 self.setting_and_variable_name_length,
                 self.use_pipes,
-                self.space_count).visit(model)
+                self.space_count == 'preserve').visit(model)
         model.save(output)
 
     def visit_file(self, file):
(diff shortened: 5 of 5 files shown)


src/robot/tidypkg/transformers.py
```diff
@@ -231,9 +231,14 @@ def _remove_consecutive_separators(self, line):
...
-            spaces = self.space_count * self.indent \
-                if index == 0 else self.space_count
-            token.value = ' ' * spaces
+            if self.space_count != 'preserve':
+                spaces = self.space_count * self.indent \
+                    if index == 0 else self.space_count
+                token.value = ' ' * spaces
+            else:
+                if index == 0:
+                    token.value = ' ' * 4 * self.indent
```
bhirsz, line 240: so preserve option preserve spaces but only between arguments? Let's say I have ident of 8 spaces in my test, I would assume that preserve option would not change this:

```
My Test
[Template]  Some Template
            1             2
            3             4
```

utest/tidy/test_argument_validation.py
```diff
@@ -17,9 +17,14 @@ def test_valid_space_count(self):
         assert_equal(opts['spacecount'], 42)
 
     def test_invalid_space_count(self):
-        error = '--spacecount must be an integer greater than 1.'
+        error = '--spacecount must be of value "preserve" or an integer greater than 1.'
```
MaciejWiczk, line 20: Howa about : `--spacecount must be either "preserve" or an integer greater than 1.'`?

src/robot/tidypkg/transformers.py
mnojek: This is due to that tidy sometimes adds a token (e.g. `END` when changing old FOR loop syntax to the new one) and it needs some spacing to not be justified to the left without whitespaces but you're right that someone may use other indentation than 4 spaces so I need to test it more and come up with an idea to use the same spacing as it is in the file.

tysonite:
Nice addition!

pekkaklarck:
Looks good based on a very quick look on mobile. I'm on holiday and won't have time for real review now, but I'll assign the original issue to RF 3.2.2 scope it won't be forgotten.

src/robot/tidypkg/transformers.py
mnojek: Ok, I did some testing and it seems that tidy works like this. It has some rules about parsing templates and some default indentation. I think it's good that it does some automatic parsing of strange things like too wide indent and it should stay like this thus I will not make changes in this area. I picked 4 spaces as a default value because there need to be some default value 😉  and I also think that it's the most popular indentation style in space separated robot files so maybe it's good to keep it like this.

pekkaklarck:
This PR has a conflict due to fix to #3618 in 8e2e56cb8e028202dc52b67b25acd64cf1a7f6f3. This is easy to resolve by adding

```python
if self.space_count == 'preserve':
    return False
```

to the new `_should_be_aligned` helper and leaving code in `visit_Statement` as it's in the master. I'm not certain is that the best way to handle aligning in this particular case:

- If I understand the code right (@yanne possibly can comment if I don't), `visit_Statement` is only called with the Settings and the Variables sections. Uniformly aligning data in these sections sounds like a good idea to me even if spaces were preserved with keyword calls.

- If we want to preserve all spaces, then the code should be updated to handle also `Aligner.visit_TestCaseSection` that even with this PR aligns columns. I believe keeping this code as-is and also removing other changes to `Aligner` would be better and obviously a lot easier. This could be revisited later if needed.

- `Aligner` getting `space_count` argument that is only used for checking should spaces be preserved is a bit weird. If we decide to keep this behavior, it should be changed to a Boolean named something like `preserve_spaces`. Alternatively the code could be just modified so that the whole `Aligner` isn't used when spaces should be preserved. That would actually resolve the above problem as well.

antonpaa:
I tested the changes especially running with the `-s preserve`  option with the .robot-codebase of couple of projects I had available. Nothing weird came up when comparing the delta, so all in all the whitespace preserving seems to work in multiple cases that I just had to offer.

Although I don't regularly do it, if I would use `--usepipes` option the output is perhaps a surprise at first. In e.g. a `*** Settings ***` section with the old implementation the line
```    
Library     BuiltIn
```
would change to
```
| Library        | BuiltIn |
```
where as with these changes if I run the same _with_ `-s preserve` it would change to 
```
| Library | BuiltIn |
```
Imo _with_ `-s preserve` it should output
```
| Library    | BuiltIn |
```
and without it should look like 
```
| Library | BuiltIn |
```

Re: @pekkaklarck 's comments:

2. I agree, a boolean such as `preserve_spaces` makes it more readable and conforms better to the other attributes.

3. I disagree, but not strongly. The primary reason why I would agree with this is because the aligning has worked there really well and certainly improved readability basically always. Yet for various reasons users may want to have that part of the files untouched too, which is why I would preserve their whitespaces as well.

pekkaklarck:
Thanks for comments @antonpaa. Quick replies to them:

1. Good point about interoperability with pipes. I think it should work like earlier by default but need to think how pipes should work when preserving spaces.

2. I don't care strongly are spaces preserved in the Settings and the Variables sections.

3. Aligning data-driven tests must be fixed. As I commented earlier, this might be easiest by reverting all changes to Aligner and then just not using Aligner at all when spaces are preserved.

4. As discussed above, this PR is mostly ready and works well. I hope someone can finalize it.

mnojek:
Thanks all for your input @antonpaa @pekkaklarck ! I finally had time to dive into it again.
Replies to what has been discussed (in random order):

1. I have no experience in pipes (honestly, I don't like them) but I will try to test it. @antonpaa Any help appreciated 😉 
2. I prefer preserving spaces in Variables and Settings thus I will not revert changes to the Aligner...
3. So I need to fix templated tests. This was difficult to locate the issue but now I have the place: `ColumnAligner.align_statement` method. I still do not have a solution, but I am testing it locally (it works in a very strange way). Will publish the code when it's done. High changes that it will happen in the upcoming week.
4. I changed `space_count` in `Aligner` to `preserve_spaces` boolean as you suggested (still on my local branch).
5. Hopefully this week I will finish it.

pekkaklarck:
Have you considered changing the code so that Aligner isn't used at all when spaces are preserved?

mnojek:
I disabled `Aligner` and started testing it but I think I found a bug in Robot API in the meantime (I checkout to master to reject the possibility of my changes affecting how it works).

When I have `Test Template` used in a code, e.g.:
```
*** Settings ***
Test Template  Login To Page

*** Test Cases ***      USERNAME        PASSWORD
Valid Login             robot           secret
Invalid Login           robot           wrong
Another Login           robot2          pass
```

Robot API parses each line under `TestCaseSection` as separate `TestCase` (which is fine) but it injects `EmptyLine` after it (which is strange). So after using tidy on it I end up with:
```
*** Settings ***
Test Template     Login To Page

*** Test Cases ***    USERNAME    PASSWORD
Valid Login           robot       secret

Invalid Login         robot       wrong

Another Login         robot2      pass
```
It works well with this example of using Template:
```
*** Test Cases ***
My Test
    [Template]  Some Template
    1    2
    3    4
```
and after `tidy` it looks the same because it does not add `EmptyLine` statement between each data-driven case.

Is it expected or is it a bug? If I want to support `tidy` to preserve spaces for test templates than I really need to do some strange ifs and weird conditions which I'd rather not do. Why this `EmptyLine` is added where there isn't one between cases mentioned in first code example?

mnojek:
@pekkaklarck @antonpaa Hey! Can someone continue discussion with me? :) I would be glad to finalize it

src/robot/tidypkg/transformers.py
```diff
@@ -405,6 +412,8 @@ def visit_Statement(self, statement):
         return statement
 
     def _should_be_aligned(self, tokens):
+        if self.space_count == 'preserve':
+            return False
```
pekkaklarck, line 416: 1. Are you sure disabling alignment is a good idea? It is done only when test case sections has additional headers like

    ```robotframework
    *** Test Cases ***    USER     PWD
    Valid login           robot    secret
    Invalid login         robot    bad
    ```
    and in that case data in columns is aligned consistently. We can argue that explicitly specified headers should always enable this behavior, but then again explicitly used `preserve` should also preserve spaces. I guess not aligning columns in this case is fine.

2. In this class this attribute is used only to control is data aligned or not and thus `space_count` is pretty bad name. It should be changed to something like `no_align` or `preserve_spaces`.

3. The sole purpose of the Aligner class is to align data. If that's not desired, it would actually be better to not use the Aligner at all.The current

    ```python
    Aligner(...).visit(model)
    ```
    in `Tidy._tidy` should be changed to
    ```python
    if self.space_count != 'preserve':
        Aligner(...).visit(model)
    ```

pekkaklarck (changes requested):
Looks good but some smallish changes should still be done:

1. Instead of configuring Aligner to not do anything if `space_count == 'preserve'`, it should simply not be used at all. See my separate comment for more details.

2. `preserve` should be case-insensitive. Now it is partly which probably mainly causes problems. A test, atest or utest, for this functionality is needed.

3. Docs should mention that preserving spaces is a new feature in Robot Framework 4.0.

src/robot/tidypkg/transformers.py
mnojek: First of all, sorry for the long comment 😉

For this sample code:
```
*** Test Cases ***      USERNAME        PASSWORD
Valid Login             robot           secret
Invalid Login           robot           wrong
```
I get this, when `Aligner` is enabled and my changes from this PR are included:
```
*** Test Cases ***      USERNAME        PASSWORD
Valid Login           robot              secret

Invalid Login         robot              wrong

```
I get this, when `Aligner` is disabled for `preserve` value of `spacecount` (according to your latest suggestion):
```
*** Test Cases ***      USERNAME        PASSWORD
Valid Login    robot           secret

Invalid Login    robot           wrong

```
So the templated test cases are ignored entirely when I want to preserve spaces which is not the intended solution.

And I get this, when not using `spacecount` option at all (so basically this is how it works on master right now):
```
*** Test Cases ***    USERNAME    PASSWORD
Valid Login           robot       secret

Invalid Login         robot       wrong

```
**The problem is** that every time there is a new line after each templated test case. I tried to debug this but I couldn't find the cause of this issue. What I found is that inside each `TestCase` object body there are 2 other objects: `TemplateArguments` and `EmptyLine` but there is also an `EOL` token at the end of the `TemplateArguments` so the **new line is duplicated**. Visualization:
```
Valid Login           robot       secret  # TemplateArguments(SEPARATOR, 'robot', 'SEPARATOR', 'secret', '\r\n') object
# EmptyLine object
Invalid Login         robot       wrong # TemplateArguments(SEPARATOR, 'robot', 'SEPARATOR', 'wrong', '\r\n') object
# EmptyLine object
```

Whatever I do, the problem is there, regardless of the suggestions to this PR. The best results I get with using tidy when I keep the current changes from this PR.

**What do I propose?**
We have different options:
- Merge the changes from this PR like they are
- Debug the issue, fix it and then merge (I cannot promise I will do it)
- Discard the changes and forget about them
- Rewrite the tidy as a new external tool
I suggest merging this PR and giving a chance for the last one: rewriting the tidy.

**Why to rewrite the tidy?**
- Keeping such tools (like tidy) as a part of the RF package may lead to maintenance nightmare - by removing them we don't need to worry about implementing the latest changes to RF in all internal tools
- It can be better - it has some flaws and issues for some specific situations
- Maintaining of RF would be cheaper because the tool will be stored as an external package
- Many more opportunities e.g. we can introduce simple configuration files which will give the developers the power of configuring the formatting for each project

**Who will do it?**
I can volunteer on it. And of course I am open to any help 😀 I already have some experience with new robot API (gained when creating [Robocop](https://github.com/MarketSquare/robotframework-robocop)) and I know how much such tools are helpful in my current projects where several developers create tests and they are struggling to follow the guidelines - refactoring is sometimes a very time-consuming task.

mnojek:
First of all, sorry for the long comment 😉

For this sample code:
```
*** Test Cases ***      USERNAME        PASSWORD
Valid Login             robot           secret
Invalid Login           robot           wrong
```
I get this, when `Aligner` is enabled and my changes from this PR are included:
```
*** Test Cases ***      USERNAME        PASSWORD
Valid Login           robot              secret

Invalid Login         robot              wrong

```
I get this, when `Aligner` is disabled for `preserve` value of `spacecount` (according to your latest suggestion):
```
*** Test Cases ***      USERNAME        PASSWORD
Valid Login    robot           secret

Invalid Login    robot           wrong

```
So the templated test cases are ignored entirely when I want to preserve spaces which is not the intended solution.

And I get this, when not using `spacecount` option at all (so basically this is how it works on master right now):
```
*** Test Cases ***    USERNAME    PASSWORD
Valid Login           robot       secret

Invalid Login         robot       wrong

```
**The problem is** that every time there is a new line after each templated test case. I tried to debug this but I couldn't find the cause of this issue. What I found is that inside each `TestCase` object body there are 2 other objects: `TemplateArguments` and `EmptyLine` but there is also an `EOL` token at the end of the `TemplateArguments` so the **new line is duplicated**. Visualization:
```
Valid Login           robot       secret  # TemplateArguments(SEPARATOR, 'robot', 'SEPARATOR', 'secret', '\r\n') object
# EmptyLine object
Invalid Login         robot       wrong # TemplateArguments(SEPARATOR, 'robot', 'SEPARATOR', 'wrong', '\r\n') object
# EmptyLine object
```

Whatever I do, the problem is there, regardless of the suggestions to this PR. The best results I get with using tidy when I keep the current changes from this PR.

**What do I propose?**
We have different options:
- Merge the changes from this PR like they are
- Debug the issue, fix it and then merge (I cannot promise I will do it)
- Discard the changes and forget about them
- Rewrite the tidy as a new external tool
I suggest merging this PR and giving a chance for the last one: rewriting the tidy.

**Why to rewrite the tidy?**
- Keeping such tools (like tidy) as a part of the RF package may lead to maintenance nightmare - by removing them we don't need to worry about implementing the latest changes to RF in all internal tools
- It can be better - it has some flaws and issues for some specific situations
- Maintaining of RF would be cheaper because the tool will be stored as an external package
- Many more opportunities e.g. we can introduce simple configuration files which will give the developers the power of configuring the formatting for each project

**Who will do it?**
I can volunteer on it. And of course I am open to any help 😀 I already have some experience with new robot API (gained when creating [Robocop](https://github.com/MarketSquare/robotframework-robocop)) and I know how much such tools are helpful in my current projects where several developers create tests and they are struggling to follow the guidelines - refactoring is sometimes a very time-consuming task.

pekkaklarck:
Two quick comments:

- I need to study the problem discussed above a bit more. I'll comment it separately.

- Having Tidy as an external tool would be awesome. Being able to develop it further independently would be especially good. I think we should still have Tidy as part of RF 4, but the external project could be created already before that. Let's discuss this more on Slack!

MaciejWiczk:
@mnojek , count me in for Tidy as separate project

bhirsz:
> @mnojek count me in for Tidy as separate project

Something like my https://github.com/bhirsz/robotframework-clean (beware its only small project) but mor elike current tidy ;) (although I do like feature in my tool that I can actually select what lines I want to format, which makes it easy to integrate with IDEs)

pekkaklarck:
> Something like my https://github.com/bhirsz/robotframework-clean (beware its only small project) but mor elike current tidy ;) (although I do like feature in my tool that I can actually select what lines I want to format, which makes it easy to integrate with IDEs)

This tools seems to have some nice features. I'm sure they could be incorporated with the new Robot Tidy!

pekkaklarck:
Looking at the Tidy code, it seems empty lines are added between tests by `NewlineNormalizer`. I agree they shouldn't be added when using the data-driven style, or at least there should be an option to configure that, but that's a subject to a separate issue and possible PR.

Column alignment shouldn't be messed up by this PR, though. Considering that the idea of this PR is to preserve spaces, I think it would be best to leave alignment as it was. Alternatively column alignment should be done always done if there are headers.

mnojek:
I don't know if I will have time to finish this PR in a foreseeable future. As I remember I couldn't figure out what was the issue causing bad aligning when using DDT. Since this issue wasn't reported by many people I think it's not very critical. We have several options:
- abandon this PR since new Tidy is being developed
- wait for me (or other person) to finish it
- merge and close it as it is

@pekkaklarck Share your thoughts.

pekkaklarck:
I looked at this a bit more closely now and here are my findings:

1. There's a bug in Tidy when used with data like this:

pekkaklarck:
I looked at this a bit more closely now and here are my findings:

1. I find it odd that this PR preserves spaces in some places but still changes them in others. It would be much more logical to simply not touch spaces at all when they ought to be preserved. Such design would also be easier. As I've commented already earlier, the whole `Aligner` could be used only if spaces aren't preserved and there are also some other places where this logic would be simpler. Are you @mnojek interested to change the design that way or should we just close this and concentrate on the new Tidy?

2. There's a bug in Tidy when aligning data-driven tests with headers and test case name is short. This should be fixed separately.

mnojek:
Thanks for the answer, @pekkaklarck.
1. I tried to resolve this issue but each time I witnessed some other related issues. I would suggest closing this PR since it should work fine in new Tidy. I am figuring out how to reserve some time for Tidy development, so I think I will be able to help with that project soon. The good news is that this PR was the seed on which the whole idea of new Tidy grew 🌱
2. I will create a bug for the weird additional new line when using tidy on DDT.

pekkaklarck:
I agree it's better to concentrate on the new Tidy nd it's true that this issue/PR was the main motivation for it!

In general I think it's best that we consider the old Tidy to be in maintenance mode meaning we fix bugs but don't add new features. Submitting a but about the DDT issue is a good idea in that regard.

mnojek:
I created an issue related to problems with tidy run on data-driven tests here: https://github.com/robotframework/robotframework/issues/3805

Issue #3611, opened 2020-07-10
Argument conversion with enums should work with normalized names


Snooz82:
We are using a lot of enums in Browser Library and would like to have the automatic conversion from strings in RF to enum members in Python being able to match members based on a normalized string.

Because members of enums can not contains spaces, like method names in python, but spaces are typically quite common in RobotFramework we want to be able to allow spaces in member names.
Same with case sensitivity. 
Case sensitivity with values is very rare.

Lets assume we have the following enum:
```python
AssertionOperator = Enum(
    "AssertionOperator",
    {
        "noassertion": "NO_ASSERTION",
        "NO_ASSERTION": "NO_ASSERTION",
        "equal": "==",
        "==": "==",
        "shouldbe": "==",
        "inequal": "!=",
        "!=": "!=",
        "shouldnotbe": "!=",
        "lessthan": "<",
        "<": "<",
        "greaterthan": ">",
        ">": ">",
        "<=": "<=",
        ">=": ">=",
        "contains": "*=",
        "*=": "*=",
        "starts": "^=",
        "^=": "^=",
        "shouldstartwith": "^=",
        "ends": "$=",
        "shouldendwith": "$=",
        "$=": "$=",
        "matches": "$",
        "validate": "validate",
        "then": "then",
    },
)
```

The values in robot `shouldnotbe`, `Should Not Be` and `should_not_be` etc all match to the member `shouldnotbe`

When documenting the enum in libdoc it should also document the "robotic" name `Should Not Be`

This is related to #3607 but imho with a higher priority.

pekkaklarck:
Enum member matching being case, space and also underscore insensitive like matching keyword names sounds good to me. It also ought to be so easy to do and backwards compatible that we can consider it in RF 3.2.2.

How Libdoc should show enum members is a different question and part of #3607.

mkorpela:
This will actually be somewhat backwards incompatible as currently there could be lower and uppercase etc enum values that are different. They will be same after this change.
But I do ’t know how much trouble this might be. I would assume that not so much.

pekkaklarck:
Yeah, it's possible someone would be affected by this change but those changes are pretty small compared to benefits.

The bigger question is, do we want to have such an enhancement in a patch release in the first place. A library using this functionality would then work differently with RF 3.2.1 and 3.2.2 which needed to be documented. Then again, there's the same issue between RF 3.2 and 3.3 if we do the change later.

[...]

PR #3632, opened 2020-08-12
Enum conversion with normalized names. Implements #3611

willemvanoort:
I have implemented the enhancement proposed in issue #3611. The changes are backwards compatible as the member is first matched without normalization.

### DIFF
--- a/atest/testdata/keywords/type_conversion/Annotations.py
+++ b/atest/testdata/keywords/type_conversion/Annotations.py
@@ -11,7 +11,7 @@ from robot.api.deco import keyword
 class MyEnum(Enum):
     FOO = 1
     bar = 'xxx'
-
+    foo = 'yyy'
 
 class Unknown(object):
     pass
--- a/atest/testdata/keywords/type_conversion/KeywordDecorator.py
+++ b/atest/testdata/keywords/type_conversion/KeywordDecorator.py
@@ -18,6 +18,7 @@ from robot.utils import PY2, unicode
 class MyEnum(Enum):
     FOO = 1
     bar = 'xxx'
+    foo = 'yyy'
 
 
 class Unknown(object):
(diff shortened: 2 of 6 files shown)


willemvanoort:
Not sure why "Acceptance tests (CPython + PyPy) / Python 3.7 on ubuntu-latest (pull_request)" has failed here.

pekkaklarck:
Quick comments on mobile:
- Code changes look good. First trying the exact march is a good idea.
- This new functionality needs to be tested.
- Something is needed in the User Guide as well.
- Don't know why tests fail, could be something unrelated as well, and cannot check right now.

pekkaklarck:
Looking very good. Two minor things to do, though:

- Would be good to add a test that using `BAR` matches member `BAR` even if there would be members like `bar` or `B_A_R`.
- User Guide doc needs a note that case/space/underscore insensitivity is new in RF 3.2.1. Could either add `Starting from RF 3.2.1` before the newly added text or add a separate note about that elsewhere.

Do you @willemvanoort want to do the above or should I just merge this PR now and do them myself?

willemvanoort:
Updated last commit with the minors. One thing that I realized when updating the tests is that in python27 dir(Enum) is called to collect enum members, alphabetizing the members, whereas in python3 the Enum.__members__ property is read, which returns an OrderedDict (in the order of definition). Hence, conversion might in some very specific cases return a different member when running in different environments. Just dropping this here in case you'd want to smooth this out.

pekkaklarck:
In my opinion it's important, and enough, if exact match matches the correct member. I doubt it's common to have members with names that have same normalized form and if you do, it's better to then use exact arguments anyway. Having members like `FOO` and `foo` and then using `Foo` doesn't make much sense.

The PR looks very good and I'll merge it. Tests seem to cover everything important, but there could bel separate tests for normalized member matching. This goes to nitpicking so I'll split these tests myself. :smile: 

Thanks for the PR!

pekkaklarck:
After thinking this a bit more, decided to make it an error to use a non-exact match if there are multiple members with same normalized name. Explicit is better than implicit.

PR #3634, opened 2020-08-19
Libdoc html in-browser switch to list one shortcut per line

Rudolf-AT:
Hi,
when reading the library documentation in a web browser, I would like to have a list of the "shortcuts" (keywords) one per line, because for me it seems easier to read.

With the attached pull request, the HTML page created by libdoc looks like this:

RF_libdoc_after_

A click on "One per line" switches the list to one keyword per line:
RF_libdoc_after_one-per-line_

The new optional "One per line" shortcuts list style takes more vertical place. However, it may be easier to spot a specific keyword than in the compact list. It is possible to switch back to the original style in the browser.

Tested for the libraries from Robot Framework source and for SeleniumLibrary in Google Chrome 84, Mozilla Firefox 79 and Internet Explorer 11.

Best Regards,
Rudolf

### DIFF
--- a/src/robot/htmldata/libdoc/libdoc.css
+++ b/src/robot/htmldata/libdoc/libdoc.css
@@ -39,25 +39,37 @@ a:hover {
     font-weight: bold;
     letter-spacing: 0.1em;
 }
-.shortcuts-style-select {
-    border: 0;
-    padding: 0;
-}
-.shortcuts-style-select label {
+label.switch {
     background-color: white;
-    border: 1px solid #c30;
-    border-radius: 2px;
-    color: #c30;
     cursor: pointer;
+    position: relative;
 }
-.shortcuts-style-select input[type="radio"] {
+input.switch {
     display: none;
 }
-.shortcuts-style-select input[type="radio"]:checked ~ span {
-    background: #c30 !important;
-    border: 1px solid #c30;
-    border-radius: 2px;
-    color: white;
+.slider {
+    background-color: grey;
+    display: inline-block;
+    height: 20px;
+    position: relative;
+    top: 5px;
+    width: 40px;
+}
+.slider:before {
+    background-color: white;
+    content: "";
+    height: 14px;
+    left: 3px;
+    position: absolute;
+    top: 3px;
+    width: 14px;
+}
+input.switch:checked + .slider::before {
+    background-color: white;
+    left: 22px;
+}
+input.switch:checked + .slider {
+    background-color: #c30;
 }
 .shortcuts .one-per-line {
     display: none;
--- a/src/robot/htmldata/libdoc/libdoc.css
+++ b/src/robot/htmldata/libdoc/libdoc.css
@@ -39,6 +39,29 @@ a:hover {
     font-weight: bold;
     letter-spacing: 0.1em;
 }
+.shortcuts-style-select {
+    border: 0;
+    padding: 0;
+}
+.shortcuts-style-select label {
+    background-color: white;
+    border: 1px solid #c30;
+    border-radius: 2px;
+    color: #c30;
+    cursor: pointer;
+}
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Interesting idea. Could you please submit an issue about this enhancement idea for tracking purposes? You could possibly also ask comments from the community on out Slack or mailing list.

Did you think about other alternatives for the widget controlling the behavior? Could it perhaps be some king of a [toggle switch](https://www.w3schools.com/howto/howto_css_switch.asp)?

Rudolf-AT:
Thanks for the feedback. I like that kind of switch and changed the code of this pull request to implement it.
I created an issue: [https://github.com/robotframework/robotframework/issues/3635](url)

pekkaklarck:
Tested this locally and I like it! Some fine tuning could still be done, though:

- The toggle styles could be enhanced. There could possibly be text on both sides (not sure what yet) and possibly the text and the toggle could have some light background.

- Instead of rendering shortcuts twice and hiding one set based on the toggle position, it ought to be possible to render them once and then change `&middot;` with `<br>` depending on the toggle.

I'll experiment with both of these locally.

Rudolf-AT:
Maybe a text like:
"Compact list" - "Expanded list"
"Terse" - "Enlarged"
"Deflated" - "Inflated"
"Minimum height" - ?
"All in one line" - "One per line"
"Only one line" - "Multiple lines"
"Crammed" - "Easy to read"

pekkaklarck:
I've been doing some fine tuning locally. The basic functionality is exactly the same, but these have been changed:

- Both the slider and the text around it are a bit smaller. 
- Background color stays the same regardless the switch position.
- Text is now `List style: Compact [x ] Expanded` but been trying also `Compact list [x ] Expanded list`.
- Also tags have same functionality as shortcuts.
- The separator between items is toggles between `&middot;` and `<br>` instead of rendering both lists and toggling which is shown.

I really like the original idea and think the small enhancements above make it even better. I still have an idea to see how easily the toggle position could be saved persistently in browsers local storage. Our reports have such functionality and if the code is easily reusable it would be nice addition to this feature as well. After checking that I'll commit the changes. I'll either update this PR and then merge it to master or merge the PR first and do changes then.

pekkaklarck:
Probably easiest for me to just merge this PR and then do changes afterwards. I hope you @Rudolf-AT can review the changes once they are committed. And thanks for the PR!

Rudolf-AT:
I have to admit that I'm happy that you like this idea :)

One very minor thing:
if a libdoc html has only two or three keywords, meaning no switch, the shortcuts are shown in the expanded style if the persistet state is expanded.
Steps to reproduce:
1. load "BuiltIn.html"
2. set "expanded"
3. load "Screenshot.html" (has three keywords)
-> Shortcuts are shown expanded, but should be compact.
Maybe ignore. Or maybe change the ".length > 3" to ".length > 1" for shortcuts and tags. Easy, no surprises, why not show the switch for 2 keywords as well.

Everything else seems to be ok/as expected.

- BuiltIn: in compact list: shortcut links to anchor 
- BuiltIn: in expanded list: shortcut links to anchor
- SeleniumLibrary: all shortcuts are in compact list
- SeleniumLibrary: all shortcuts are in expanded list
- All shortcuts are shown if only three keywords exist
- State is persistent: open html-1 and switch to expanded, open html-2 in another tab: expanded list is shown
- Quick general check ok for: Firefox 80, Google Chrome 85, Internet Explorer 11
- Quick check in browser console: no error

pekkaklarck:
Thanks for review! Showing expanded list and no switch with less than four keywords is definitely a bug. I agree showing the toggle when there's more than one keyword is the best solution. Hiding it with 2-3 kws doesn't bring any real benefits and would require some extra logic.

Issue #3457, opened 2020-02-07
Remove Python 2 and Python 3.5 support


pekkaklarck:
Python 2 was [officially retired in April 2020](https://www.python.org/psf/press-release/pr20191220/) and even security fixes won't be released afterwards. Robot Framework continuing to support it much long does not make sense, and the [Robot Framework Foundation](https://robotframework.org/foundation/) has decided that it will not sponsor Robot Framework development targeting Python 2 anymore in 2021. The main reason we don't want to be more aggressive removing Python 2 support is that it would also drop Jython support. 

When Python 2 support is removed, also support for Python 3.5 and older will be removed. This eases development by making it possible to take into use newer Python features, most notably f-strings. Python 3.5 will also [reach its end-of-life in September 2020](https://devguide.python.org/#status-of-python-branches), a lot before the expected Robot Framework 5.0 release.

The current Python version support plan is as follows:

- Robot Framework 3.2 supports Python 2.7 as well as Python 3.4 and newer. This covers also all RF 3.2.x bug fix releases.

- Robot Framework 4.0 planned for late 2020 still supports Python 2.7 as well as Python 3.5 and newer. This covers also RF 4.0.x bug fix releases and possible RF 4.x feature releases.

- Robot Framework 5.0 will support only Python 3.6 and newer.

pekkaklarck:
The decision to make the version where Python 2 support is removed Robot Framework 4.0 means that there most likely won't be RF 3.3.  Such a release _may_ be done if there are needs for changes before RF 4.0 that are too big for RF 3.2.x minor releases. This could, for example, include enhancements to the new parser to make LSP development easier. Possible RF 3.3 would nevertheless be a pretty small releases and all big enhancements (IF/ELSE, nested loops, etc.) have been moved from [RF 3.3](https://github.com/robotframework/robotframework/milestone/51) to [RF 4.0](https://github.com/robotframework/robotframework/milestone/52).

pekkaklarck:
For projects using Robot Framework transitioning from Python 2 to Python 3 may require some work. Luckily most of the important Robot Framework libraries and tools have support Python 3 for some time, so for projects using only ready-made libraries  and tools upgrading shouldn't be too hard.

If projects have their own libraries, they obviously need to be ported to Python 3 as well. With smallish projects it may be possible to just go directly from Python 2 to Python 3, but with bigger ones it is likely that at leas some of the tooling needs to work with both versions for some time. 

[...]

PR #3653, opened 2020-09-06
Remove EOL Python 3.4 from officially supported versions in setup.py and README

### LINKED ISSUE #3577: Drop Python 3.4 support
Python 3.4 has reached its end-of-life already in 2019. RF 3.2.x still supports it but there's no point in RF 3.3 or newer supporting it anymore. What this mean in practice:

- Remove Python 3.4 from officially supported versions in setup.py and README.
- Stop testing with Python 3.4. We actually haven't tested it on CI at all, but I've had Python 3.4 on my old laptop and tested it there.
- No immediate code changes. As long as we still support Python 2.7 and 3.5 (#3457) we cannot really take modern Python features into use anyway.

If someone happens to still need Python 3.4 support, please test RF in your environment and submit pull requests if we break something. PRs fixing Python 3.4 support will be accepted assuming they are simple and don't cause problems with other interpreters.

hugovk:
Fixes #3577.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -117,11 +117,13 @@ a good place to start is http://python.org. There you can download a suitable
 installer and get more information about the installation process and Python
 in general.
 
-Robot Framework 3.2 supports Python 2.7 and Python 3.4 and newer, but the `plan
+Robot Framework 4.0 supports Python 2.7 and Python 3.5 and newer, but the `plan
 is to drop Python 2 support soon`__ and require Python 3.6 or newer. If you
-need to use older Python versions, Robot Framework 3.0 supports Python 2.6,
-Robot Framework 2.5-2.8 support Python 2.5, and Robot Framework 2.0-2.1
-support Python 2.3 and 2.4.
+need to use older Python versions,
+Robot Framework 3.2 supports Python 3.4,
+Robot Framework 3.0 supports Python 2.6,
+Robot Framework 2.5-2.8 support Python 2.5,
+and Robot Framework 2.0-2.1 support Python 2.3 and 2.4.
 
 After installing Python, you probably still want to configure PATH_ to make
 Python itself as well as the ``robot`` and ``rebot`` `runner scripts`_
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -118,12 +118,7 @@ installer and get more information about the installation process and Python
 in general.
 
 Robot Framework 4.0 supports Python 2.7 and Python 3.5 and newer, but the `plan
-is to drop Python 2 support soon`__ and require Python 3.6 or newer. If you
-need to use older Python versions,
-Robot Framework 3.2 supports Python 3.4,
-Robot Framework 3.0 supports Python 2.6,
-Robot Framework 2.5-2.8 support Python 2.5,
-and Robot Framework 2.0-2.1 support Python 2.3 and 2.4.
+is to drop Python 2 support soon`__ and require Python 3.6 or newer.
 
 After installing Python, you probably still want to configure PATH_ to make
 Python itself as well as the ``robot`` and ``rebot`` `runner scripts`_
--- a/README.rst
+++ b/README.rst
@@ -15,7 +15,7 @@ Python or Java.
 
 Robot Framework is operating system and application independent. The core
 framework is implemented using `Python <http://python.org>`_, supports both
-Python 2 and Python 3, and runs also on `Jython <http://jython.org>`_ (JVM),
+Python 2.7 and Python 3.5+, and runs also on `Jython <http://jython.org>`_ (JVM),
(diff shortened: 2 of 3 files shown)


pekkaklarck:
Thanks! Could you also check what INSTALL.rst says about versions?

We may have some Py 3.4 related code and tests but I'm not certain is trying to find them worth the effort. They can be removed once we later increase minimum version to Py 3.6.

hugovk:
> Thanks! Could you also check what INSTALL.rst says about versions?

It says:

> Robot Framework 3.2 supports Python 2.7 and Python 3.4 and newer, but the plan
is to drop Python 2 support soon and require Python 3.6 or newer. If you
need to use older Python versions, Robot Framework 3.0 supports Python 2.6,
Robot Framework 2.5-2.8 support Python 2.5, and Robot Framework 2.0-2.1
support Python 2.3 and 2.4.

I've updated it in71ee807cf3f15ab1d22379cf9433d18befabcd2b, to change the start to say:

```diff
-Robot Framework 3.2 supports Python 2.7 and Python 3.4 and newer
+Robot Framework 4.0 supports Python 2.7 and Python 3.5 and newer
```
And add:
```diff
+Robot Framework 3.2 supports Python 3.4,
```

Should it include a lower RF version number that supports 3.4?

As this paragraph gets longer, would it be clearer as a table?

---

`INSTALL.rst` also has this 3.4 mention, I think it can stay (but could also say "3.5 and newer"):

> With Python 3.4 and newer pip is officially part of the
interpreter and should be automatically available.

---

> We may have some Py 3.4 related code and tests but I'm not certain is trying to find them worth the effort. They can be removed once we later increase minimum version to Py 3.6.

Agreed, I intentionally didn't remove any 3.4-specific stuff, as #3577 says:

> * No immediate code changes. As long as we still support Python 2.7 and 3.5 (#3457) we cannot really take modern Python features into use anyway.

pekkaklarck:
A table listing what Python versions different RF versions support might be a good idea but I'm not certain is that information really that relevant. Python 2 support is certainly important information but I seriously doubt anyone needs info about Python 2.6 or older. What do you think about this?

A problem with updating README and INSTALL is that this info is visible on the project pages even before RF 4.0 is released. Trying to change wording so that it works well both before and after RF 4.0 is probably somewhat complicated. Perhaps it's easiest just to merge this PR once we are closer to the release.

hugovk:
> A table listing what Python versions different RF versions support might be a good idea but I'm not certain is that information really that relevant. Python 2 support is certainly important information but I seriously doubt anyone needs info about Python 2.6 or older. What do you think about this?

I agree, 2.6 and older aren't really relevant. Here's the pip installs for robotframework from PyPI for August 2020:

| category | percent | downloads |
|----------|--------:|----------:|
| 2.7      |  28.15% |   117,964 |
| 3.7      |  23.91% |   100,197 |
| 3.6      |  23.79% |    99,686 |
| 3.8      |  15.84% |    66,392 |
| 3.5      |   4.07% |    17,076 |
| null     |   3.98% |    16,664 |
| 3.9      |   0.18% |       759 |
| 3.4      |   0.07% |       274 |
| 2.6      |   0.01% |        47 |
| 3.3      |   0.00% |         1 |
| Total    |         |   419,060 |

Source: `pip install -U pypistats && pypistats python_minor robotframework --last-month`

Shall we remove 2.6 and older from that paragraph?

> Robot Framework 4.0 supports Python 2.7 and Python 3.5 and newer, but the plan is to drop Python 2 support soon and require Python 3.6 or newer. If you need to use older Python versions, Robot Framework 3.2 supports Python 3.4.

---

> A problem with updating README and INSTALL is that this info is visible on the project pages even before RF 4.0 is released. Trying to change wording so that it works well both before and after RF 4.0 is probably somewhat complicated. Perhaps it's easiest just to merge this PR once we are closer to the release.

Read the Docs solves this problem by building two sets of docs: `stable` and `latest`. You can point end users to `stable` by default, but `latest` is also available where needed. Perhaps RF could point to a tagged version of the docs, or a stable tag? Although this does complicate the release process a bit, waiting is easiest :)

pekkaklarck:
Yeah, let's remove info about old versions. Not certain is even "If you need to use older Python versions, Robot Framework 3.2 supports Python 3.4." necessary.

We use Read the Docs and I think the default is to show stable. The problem with README and INSTALL is here on GitHub where you always see the latest code (and docs) by default. Probably best to wait a bit before merging but we should nevertheless get this PR ready.

hugovk:
Updated to:

> Robot Framework 4.0 supports Python 2.7 and Python 3.5 and newer, but the plan is to drop Python 2 support soon and require Python 3.6 or newer.

pekkaklarck:
Thanks for contribution @hugovk!

PR #3662, opened 2020-09-11
Remove criticality

yanne:


mkorpela:
Only 81 files 🤣

mkorpela:
Some notes about criticality use cases to check if we can cover these also in the future:

Static cases:
- [ ] _not-ready tag_ where a test/task is executed on automated system to ensure it almost works but should not affect the outcome)
- [ ]._flaky tag_ where a test/task is known or thought to have some problems and is under investigation

Dynamic cases:
- [ ] _environment-error tag_ where a test/task detects during execution that it can not be executed (SKIP case)

Any other?

mkorpela:
I think `--skiponfailure` will cover the two dynamic cases I had in mind

src/robot/running/status.py
```diff
@@ -82,7 +82,7 @@ def teardown_executed(self, failure=None):
             self.exit.failure_occurred(failure)
 
     def critical_failure_occurred(self):
-        self.exit.failure_occurred(critical=True)
+        self.exit.failure_occurred()
```
pekkaklarck, line 85: Should probably remove `critical_` from the method name.

atest/robot/cli/runner/status_rc.robot
```diff
@@ -11,19 +11,15 @@ Zero RC when all tests pass
     ${EMPTY}                    ${PASSING}      rc=0
     --statusrc                  ${PASSING}      rc=0
 
-Zero RC when all critical tests pass
-    --critical pass             ${FAILING}      rc=0
-    -c pass --status            ${FAILING}      rc=0
-
-Non-zero RC when critical tests fail
+Non-zero RC when ests fail
```
pekkaklarck, line 14: ests -> tests

pekkaklarck (changes requested):
Looks great but I found few smallish things that could be still addressed. Assuming tests pass, this PR is otherwise ready to be merged.

src/robot/rebot.py
```diff
@@ -339,6 +339,8 @@ def __init__(self):
     def main(self, datasources, **options):
         settings = RebotSettings(options)
         LOGGER.register_console_logger(**settings.console_output_config)
+        if settings['XUnitSkipNonCritical']:
+            LOGGER.warn("Command line option --xunitskipnoncritical has been deprecated.")
```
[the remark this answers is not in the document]
yanne: For some reason, when I added the deprecation warning there, the warning message was echoed twice to console. Something to do with `register_console_logger` I suppose.

PR #3671, opened 2020-09-15
Topic/debugger api

mkorpela:
Lsp debugger needs for a stable API for a debugger.
Key requirements for the (listener) API are:
1. Source + linenumber for currently executed step. To stop in debugger to a specific point OR to show where we are in robot code.
2. Variables in scope. To observe variables and their current values.
3. Execution stack (suite source + line, test source + line, keyword source + line). To show also higher levels of execution.
4. Keyword / variable evaluation in context. When debugger is stopped to certain position, to be able to execute a keyword.

This DRAFT Pull request now contains an example DebuggerListener with call stack and line numbers exposed through Listener API 2.

To test it:
`./rundevel.py --listener robot.api.DebuggerListener YOURTESTS`


mkorpela:
Pruning my drafts to get space for the really important ones.

PR #3672, opened 2020-09-16
Sketching how to make keyword start, end and lineno work on Listener v3 and v2

mkorpela:
Some random code.


mkorpela:
Test template lineno is still None.

PR #3673, opened 2020-09-17
failure exceptiontype in output.xml

mkorpela:
Based on discussions on #devel channel.

I got requirement that perform auto analysis on failures. I have did this but I felt if robot provide few info in error message this will help more in my task with high accuracy
Example: Let us assume test case consist keyword Should Be True  "Demo" != "Demo1" and it got failed, we will get error message like "Demo" != "Demo1"
In above error message =! is my key and I update testcase analysis comment that its application issue since validation got failed.
Note: Accuracy may not be 100% but we have started and will improve this.
Here I need help from robotframework that, Let us have custom robotframework exceptions as prefix in error message
Example:
 1) Robotframework: If any Should * keyword got failed error message should be like ValidationException: <actual error>
 2) Selenium: If any Wait * keyword got failed error message should be ElementException: <actual selenium throwed exception & error>
So above approach gives confident on which type of error we got based on exception type and we can take a call on this.
Note: This is draft idea and im not sure is this possible or not, is it valid?, how it helps robotframework, how other libraries owner react to it if possible.
Since we are working on CI error message like showing error line number I have proposed this.

### DIFF
--- a/src/robot/output/loggerhelper.py
+++ b/src/robot/output/loggerhelper.py
@@ -49,30 +49,31 @@ class AbstractLogger(object):
     def warn(self, msg):
         self.write(msg, 'WARN')
 
-    def fail(self, msg):
+    def fail(self, msg, error_type=None):
         html = False
         if msg.startswith("*HTML*"):
             html = True
             msg = msg[6:].lstrip()
-        self.write(msg, 'FAIL', html)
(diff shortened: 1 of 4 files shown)


adiralashiva8:
Implementation of #3674 issue

Issue #2294, opened 2016-01-11
Run Keyword And Warn On Failure keyword


dimanhsb:
Hello,

A useful feature (that we have already implemented in our framework, but might be useful for everyone) is a "Run Keyword And Warn On Failure" keyword - same thing as Run Keyword And Continue On Failure, but KW passes, and issues a warning.

our current implementation in Robot Framework is:

```
Run Keyword And Warn On Failure
    [Arguments]  ${kw}  @{args}
    [Documentation]    When keyword fails, this passes, but a warning is issued.
    ${state}  ${output}=  Run Keyword And Ignore Error  ${kw}  @{args}
    Run Keyword If  '${state}'!='PASS'  Log  Warning: Keyword failed: ${kw} @{args}{\n}${output}  WARN
    [Return]  ${output}
```

pekkaklarck:
This should work like `Run Keyword And Ignore Error` but also report possible failure as a warning, right? I guess it would be useful where you absolutely don't want the keyword to fail but don't want it fail silently either.

I want to be fairly conservative when accepting new BuiltIn keywords, but perhaps this would be useful enough. An alternative is making it somehow possible to configure the existing `Run Keyword And Ignore Error` to support this. What do others think? If we agree this keyword should be added, would someone be interested to provide a pull request implementing it?

HelioGuilherme66:
I think the best keyword to add the warning option is the `Run Keyword And Continue On Failure` not the `Run Keyword And Ignore Error`.

pekkaklarck:
I don't see why this would be so useful with `Run Keyword And Continue On Failure`. It doesn't hide the failure. It just allows execution to continue afterwards. If we decide to support this functionality by configuring existing keywords, adding such support also to that keyword wouldn't hurt. Adding a new keyword, like originally proposed, may be easier and clearer, though.

dimanhsb:
Yes, I agree that there might be some confusion on a keyword that would be `Run Keyword And Ignore Error But Warn` or the like.

If it is to be added, I think it's quite clear what  `Run Keyword And Warn On Failure` means, so it should probably be added as-is.

dimanhsb:
I have implemented the keyword in python, but I haven't been able to understand the testing framework on how to create relevant unit tests (other than a simple one issuing a failure and expecting the test case to pass). Should I do the pull request, and wait for others to fill in?

pekkaklarck:
Did you read atest/README.rst?

[...]

PR #3679, opened 2020-09-22
Add new keyword `Run Keyword And Warn On Failure`

cristii006:
This PR implements the new keyword `Run Keyword And Warn On Failure` as described in #2294.


boakley (commented):
I think this would be more useful if we could optionally supply a custom error message.

pekkaklarck:
The PR looks good one a very quick look. I don't have time for real review but I added the issue to RF 4.0 scope so this won't be forgotten.

I agree with @boakley that being able specify a custom message would be handy. How should that work in practice?

pekkaklarck:
This is a great PR with code, tests and docs. Only smallish problem I see is that tests use somewhat inconsistent separators but I don't think that's a reason not to merge this.

PR #3684, opened 2020-09-26
Issue 3651: String: `Should Be Tittle Case` is not consistent with `C…

miktuy:
Try to fix bug by using keyword ```convert_to_title_case``` internal
image

### DIFF
--- a/atest/testdata/standard_libraries/string/should_be.robot
+++ b/atest/testdata/standard_libraries/string/should_be.robot
@@ -70,7 +70,7 @@ Should Be Titlecase Positive
 
 Should Be Titlecase Negative
     [Template]    Run Keyword And Expect Error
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Thanks for PR! The new functionality needs tests and we perhaps should also think what's the best way to handle bytes.

pekkaklarck:
I already earlier mentioned that this new functionality would need tests. I looked at the changes a bit more thoroughly and noticed that documentation is somewhat misleading especially related to the `exclude` argument. The docs should also mention that the logic has changed in RF 4.0. Are you @miktuy interested to continue with this PR or should I fix the aforementioned issues?

miktuy:
Sorry. I couldn't find time to fix this issue totally(

pekkaklarck:
Ok, no problem.

PR #3686, opened 2020-09-29
Topic/libdoc keyword wall

mkorpela:
- keyword count and match count shown in the header
- ability to toggle between overview that lists keywords and a "keyword wall"
Screenshot 2020-09-29 at 10 58 30
Screenshot 2020-09-29 at 10 58 50


pekkaklarck:
Looks good but the UG should get a note that linking to Shortcuts and Tags sections doesn't anymore work in RF 4.0.

pekkaklarck:
UG docs have typo but otherwise this seems to be good to be merged.

PR #3694, opened 2020-10-02
Topic/libdoc

Snooz82:


atest/robot/libdoc/invalid_usage.robot
```diff
@@ -62,6 +62,9 @@ Invalid output file
     ...    Remove Directory    ${OUT HTML}    AND
     ...    Remove Directory    ${OUT XML}
 
+invalid Spec File version
+    ${TESTDATADIR}/OldSpec.xml ${OUT XML}    Invalid spec file version 'None'. RF >= 4.0 requires XML specversion 3.
+
```
Snooz82, line 67: Added Test. New Libdoc does not accept older XML Versions as input

atest/robot/libdoc/libdoc_resource.robot
```diff
@@ -131,63 +126,69 @@ Keyword Arguments Should Be
...
+    [Arguments]    ${index}   ${type}=keywords/kw
     ${kws}=    Get Elements    ${LIBDOC}    xpath=${type}
-    ${args}=    Get Elements Texts   ${kws}[${index}]    xpath=arguments/arg
+    #${args}=    Get Elements Texts   ${kws}[${index}]    xpath=arguments/arg
+    ${args}=    Create List
+    ${arg_elems}=    Get Elements    ${kws}[${index}]    xpath=arguments/arg
+    FOR    ${elem}    IN    @{arg_elems}
+        ${arg}=    Get Element Attribute    ${elem}    string_repr
+        Append To List    ${args}    ${arg}
+    END
```
Snooz82, line 137: temporary solution until atests are changed to test separate argument nodes

atest/robot/libdoc/spec_library.robot
```diff
@@ -94,10 +94,6 @@ Keyword Source Info
     Run Libdoc And Parse Output    %{TEMPDIR}/Example.libspec
     Test Everything
 
-Old spec format
-    Run Libdoc And Parse Output    ${TESTDATADIR}/OldSpec.xml
-    Test Everything
-
```
Snooz82, line 100: not needed anymore. Current Version rejects old versions

src/robot/libdocpkg/htmlwriter.py
```diff
@@ -48,39 +48,6 @@ def write_data(self):
...
-        return {
-            'name': kw.name,
-            'args': kw.args,
-            'doc': self._doc_formatter.html(kw.doc),
-            'shortdoc': ' '.join(kw.shortdoc.splitlines()),
-            'tags': tuple(kw.tags),
-            'matched': True
-        }
-
-
```
Snooz82, line 83: this "external" converter for the model is now included in the model.
Closer relation between classes and its dictionary conversion.

src/robot/libdocpkg/model.py
```diff
@@ -40,6 +48,24 @@ def __init__(self, name='', doc='', version='', type='LIBRARY',
...
+            'doc_format': self.doc_format,
+            'source': self.source,
+            'lineno': self.lineno,
+            'inits': [init.to_dictionary() for init in self.inits],
+            'keywords': [kw.to_dictionary() for kw in self.keywords],
+            'generated': get_timestamp(daysep='-', millissep=None),
+            'all_tags': tuple(self.all_tags),
+            'contains_tags': bool(self.all_tags)  # ToDo: delete that thing when "mikko" fixed the front end
+        }
+
```
Snooz82, line 68: new dictionary conversion embedded in the model

src/robot/libdocpkg/model.py
```diff
@@ -97,3 +147,104 @@ def deprecated(self):
...
+            if prev_kind != arg.kind:
+                if prev_kind == 'POSITIONAL_ONLY':
+                    arg_strings.append('/')
+                if arg.kind == 'KEYWORD_ONLY' and prev_kind != 'VAR_POSITIONAL':
+                    arg_strings.append('*')
+            arg_strings.append(str(arg))
+            prev_kind = arg.kind
+        if prev_kind == 'POSITIONAL_ONLY':
+            arg_strings.append('/')
+        return arg_strings
```
Snooz82, line 165: temporary solution until after your ideas. that works but may be smarter.

src/robot/libdocpkg/model.py
```diff
@@ -97,3 +147,104 @@ def deprecated(self):
...
+        while len(members) > 3 and len(' | '.join(members)) > 42:
+            members[-2:] = ['...']
+        return ' | '.join(members)
+
+    def _format_default(self):
+        if not self.default:
+            return ''
+        default_str = ' = ' if self.type else '='
+        default_str += str(self.default)
+        return default_str
```
Snooz82, line 241: The top methods have originally been in robotbuilder.py to create the list of strings.
They are now used to convert the ArgumentDoc into a string

src/robot/libdocpkg/model.py
```diff
@@ -97,3 +147,104 @@ def deprecated(self):
...
+
+
+class DefaultValue(object):
+
+    def __init__(self, value=None):
+        self.value = value
+
+    def __str__(self):
+        return str(self.value)
```
Snooz82, line 250: Class DefaultValue is used to store a NoneType and check the difference between NoneType and No Default

src/robot/libdocpkg/specbuilder.py
```diff
@@ -45,29 +45,17 @@ def _parse_spec(self, path):
...
         namedargs = spec.get('namedargs')
-        if namedargs == 'true':
+        if namedargs.lower() == 'true':
             return True
-        if namedargs == 'false':
-            return False
-        # RF < 3.2 has "namedargs" element with text 'yes' or 'no'.
-        namedargs = spec.find('namedargs').text
-        return namedargs == 'yes'
+        return False
```
Snooz82, line 58: Scope and NamedArgs have been deprecated and removed now.

atest/robot/libdoc/LibDocLib.py
```diff
@@ -16,11 +16,18 @@
...
+        self.schema = XMLSchema(join(ROOT, 'doc', 'schema', 'libdoc.03.xsd'))
+
+    @property
+    def libdoc(self):
+        return self.interpreter.libdoc
+
+    @property
+    def encoding(self):
+        return SYSTEM_ENCODING \
+                if not self.interpreter.is_ironpython else CONSOLE_ENCODING
```
Snooz82, line 30: changed LibDocLib do be documentable by LibDoc to use it in with LSP

src/robot/htmldata/libdoc/libdoc.html
```diff
@@ -401,9 +407,17 @@ <h2>
...
-                <span>${$value}</span>{{if $index < args.length-1}}<br>{{/if}}
+                  <li>
+                      <div class="arg ${$value.kind} {{if $value.required}}arg-required{{else}}arg-optional{{/if}}">
+                        <span class="arg-name">{{if $value.kind === 'VAR_POSITIONAL'}}*{{/if}}{{if $value.kind === 'VAR_KEYWORD'}}**{{/if}}${$value.name}</span>
+                        {{if $value.type}}: <span class="arg-type">${$value.type}</span>{{/if}}
+                        {{if $value.default}} = <span class="arg-default-value">${$value.default}</span>{{/if}}
+                      </div>
+                  </li>
                 {{/each}}
+                </ul>
```
Snooz82, line 420: Changes so that html accesses the new argument structure

src/robot/libdocpkg/model.py
```diff
@@ -74,21 +100,45 @@ def save(self, output=None, format='HTML'):
...
 
+    def convert_doc_to_html(self):
+        formatter = DocFormatter(self.keywords, self.doc, self.doc_format)
+        self._doc = formatter.html(self.doc, intro=True)
+        self.doc_format = 'HTML'
+        for init in self.inits:
+            init.doc = formatter.html(init.doc)
+        for keyword in self.keywords:
+            keyword.doc = formatter.html(keyword.doc)
+
```
Snooz82, line 111: With that "builtin" html doc converter it is much easier to generate html containing docs.
XmlWriter gets easier and HTMLWriter as well

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -111,17 +128,3 @@ def _get_start_attrs(self, kw_type, kw, lib_source, writer):
...
-    def __init__(self, doc_format, force_html=False):
-        if force_html:
-            self._formatter = DocToHtml(doc_format)
-            self.format = 'HTML'
-        else:
-            self._formatter = lambda doc: doc
-            self.format = doc_format
-
-    def __call__(self, doc):
-        return self._formatter(doc)
```
Snooz82, line 127: DocFormatter could be completely removed due to new "builtIn" html converter in model

pekkaklarck:
This pull request contains at leas these changes:
- Libdoc model changes to preserve defaults and types.
- XML spec changes to store defaults and types separately.
- JSON spec functionality.
- Escaping backslashes and other such characters in default values.

All these are great features and on a quick look the code looks fine as well, but having this many unrelated changes makes reviewing the PR pretty much impossible. The PR must be split to smaller, independent pieces.

This must obviously start from Libdoc model changes. This PR introduced `ArgumentDoc` object that stores information related to each argument and changed `Libdoc.args` to be a list of these objects. Using a list may not be enough, though, and a separate object containing `ArgumentDoc`s would probably be needed. We do, however, already have `ArgumentSpec` object for storing information about all arguments of a keyword and introducing a new model object is not a good idea. We can, instead, enahnce `ArgumentSpec` so that it allows getting a list of `ArgumentDoc` objects (or whatever name the object would have). Then Libdoc could simply use `ArgumentSpec`. I have already prototyped enhancing `ArgumentSpec` like that and results are really promising. Let's talk how to proceed on Slack or via phone.

Snooz82:
Great!

As mentioned.
The JSON writer and builder can easily be a separate PR.

We should first solve the model.
If you want to extend ArgumentSpec i am fine, but would request you to make an example!

The escaping of string defaults. Is pretty simple and can be a separate pr as well.

But i think model with type and default plus xml and json are very related to each other.
Because we could generate the "old" xml with the new model, because we would just compress information but reading old xml and create the new model wolf be somewhat not possible/unnecessary much effort.

So new model and the both writer and the spec reader should imho in one PR

PR #3696, opened 2020-10-04
Removed unused escaped example.

lijinsyam:
${ESCAPED} gets the value "quotes and spaces" , removed as its not quoted in example.


pekkaklarck:
Good catch. I assume there's been an example using the now removed `--escape` option and this part of the example hasn't been removed.

PR #3699, opened 2020-10-05
Topic/union converter

mkorpela:
Union conversion.
Depends on https://github.com/robotframework/robotframework/pull/3713.

### DIFF
--- a/atest/testdata/keywords/type_conversion/unions.robot
+++ b/atest/testdata/keywords/type_conversion/unions.robot
@@ -1,0 +1,31 @@
+*** Settings ***
+Library                  unions.py
+Force Tags               require-py3
+
+*** Test Cases ***
+Union testing
+     With optional argument  1  ${1}
+     With optional argument  None  ${None}
+
+Optional is removed when None is default
+     [Documentation]    FAIL ValueError: Argument 'arg' got value 'None' that cannot be converted to float.
+     Unescaped optionalism   1.0
+     Unescaped optionalism   None
+
+Multitype union works in order
+    Union of int float and string   1  ${1}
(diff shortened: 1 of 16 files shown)


mkorpela:
Todolist:

- [x] string "None" is not converted by default
- [x] Union is handled
- [x] Types inside union are matched even if there is no converter for the specific type
- [ ] float -> int converter
- [ ] int -> float converter
- [ ] number -> timedelta converter
- [ ] testing

atest/testdata/keywords/type_conversion/unions.robot
```diff
@@ -0,0 +1,31 @@
...
+    Union of int float and string   ${-110}  ${-110}
+
+Custom type inside of union
+    ${myobject}=  Create my object
+    Custom type in union  myobject     <class 'str'>
+    Custom type in union  ${myobject}  <class 'unions.MyObject'>
+
+Unexpected object is just passed when in union
+    ${object}=  Create unexpected object
+    Custom type in union  ${object}     <class 'unions.UnexpectedObject'>
```
mkorpela, line 29: I'm really unsure how this should really work. I would prefer an exception but now it passes.

mkorpela:
This also has enum None conversion fix.

pekkaklarck (approved):
The general approach looks. Only functional thing I want to change is how unrecognized types are handled. When used like `arg: MyThing`, we don't do any conversion and just pass the value as-is. For consistency reasons `arg: Union[int, MyThing]` shouldn't cause an error and instead values that cannot be converted to an integer should be passed as-is. I'll change that logic after merging this PR and may do some fine-tuning/cleanup also otherwise.

Issue #3701, opened 2020-10-05
If doc_format of a library or libdoc xml is HTML, the shortdoc of Keyword contains all doc


Snooz82:
The `getshortdoc` method only works when the doc is "normal" text.
and line breaks are `\n`.

For HTML docs the first paragraph, which is the shortdoc is within a <p>.

i would propose to get the content of the <p> and removes all <b/>, <i/> and <code/> for better readability.

pekkaklarck:
I believe this issue is real but the description formatting is so messed up that understanding it fully is pretty hard. Also, do you have a concrete example demonstrating the problem?

PR #3702, opened 2020-10-05
fixes html shortdoc #3701

Snooz82:


pekkaklarck:
Things to do/consider:

1. This PR needs tests. It would probably be best to test `get_shortdoc_from_html` extensively with unit tests but some acceptance tests that demonstrate that the problem has been fixed are needed as well.

2. Keywords should know their doc format so that we could keep the `shortdoc` property and just make it dynamically check how to get the doc. This ought to be easy to implement by adding `parent` argument to `KeywordDoc.__init__` and passing the `LibraryDoc` instance to it.

3. I'm not sure is `get_shortdoc_from_html` needed by other code than Libdoc. If it isn't, then it doesn't belong to `robot.utils`.

4. We could consider enhancing `getshortdoc` utility so that it handles also HTML. It could get an optional argument like `html=False` or `format='text'` that could be used to control how to handle the doc. Possibly it could even auto-detect HTML usage an handle that case automatically.

Snooz82:
@pekkaklarck 

This should be "ready" for review and hopefully merge.
The utest have to groups.
The second one tests the roundtrip with the model.
like `REST to HTML to Shortdoc` for all three initial formats.

PR #3703, opened 2020-10-07
Topic/libdoc shortdoc

Snooz82:


utest/libdoc/test_libdoc.py
```diff
@@ -81,19 +81,6 @@ def test_shortdoc_with_multiline_robot_format(self):
         exp = "Writes the *message* to _the_ ``console``."
         verify_keyword_shortdoc('ROBOT', doc, exp)
 
-    def test_shortdoc_with_multiline_reST_format(self):
```
tysonite: Should be fine as long as you add docutils into utest/requirements.txt.

Snooz82: @pekkaklarck should/could we add docutils to utest requirements?
Then we could run utests on KeywordDoc with reST

pekkaklarck: We just added`utest/requirements.txt` and we certainly can add docutils there. Unfortunately it doesn't work with IronPython so it needs to be added like

```
docutils >= 0.9; platform_python_implementation != 'IronPython'
pygments; platform_python_implementation != 'IronPython'
```

and this test needs to be run only if docutils is installed. That needs to be done with something like

```python
try:
    import docutils
except ImportError:
    docutils = None

...


    if docutils:

        def test_whatever_that_needs_docutils(self):
            ...
```

src/robot/libdocpkg/htmlwriter.py
```diff
@@ -142,25 +142,58 @@ def _get_formatter(self, doc_format):
...
+        'b': '*',
+        'i': '_',
+        'strong': '*',
+        'em': '_',
+        'code': '``'
+    }
+
+    single_chars = {
+        '<br>': '\n',
+        '<br/>': '\n',
```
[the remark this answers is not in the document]
pekkaklarck: Should `<br>` be replaced with `\n\n`? In other words, should `<p>first<br>second</p>` yield `first` or `first second` as the short doc? I assume we nowadays get the latter but the former could be better as there's an explicit break. Probably should also test `first<br>second` i.e. without wrapping `<p>`.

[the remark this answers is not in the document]
pekkaklarck: Should we also `<hr */?>` here? It's supported by Robot's own doc format syntax so I guess there's a possibility we see it. It could be replaced e.g. with `\n\n---\n\n` which would then mean that `first<hr>second` would give us `first` as the short doc.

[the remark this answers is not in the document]
pekkaklarck: `single_chars` is a bit misleading attribute name already now and gets totally inadequate if we support e.g. `<hr>`. Something like `html_to_text` would be better.

src/robot/libdocpkg/htmlwriter.py
```diff
@@ -142,25 +142,58 @@ def _get_formatter(self, doc_format):
...
+        match = re.search(r'<p.*?>(.*?)</p>', doc, re.DOTALL)
+        if match:
+            doc = match.group(1)
+        doc = self.html_to_plain_text(doc)
+        return doc
+
+    def html_to_plain_text(self, doc):
+        for tag, repl in self.html_tags.items():
+            doc = re.sub('(<%(tag)s>)(.*?)(</%(tag)s>)' % {'tag': tag},
+                         lambda m: repl + m.group(2) + repl, doc, flags=re.DOTALL)
```
pekkaklarck, line 196: No need to group tags or to use `lambda` here. I find this variant easier to read:
```python
            doc = re.sub(r'<%(tag)s>(.*?)</%(tag)s>' % {'tag': tag},
                         r'%(repl)s\1%(repl)s' % {'repl': repl}, 
                         doc, flags=re.DOTALL)
```

src/robot/libdocpkg/htmlwriter.py
```diff
@@ -142,25 +142,58 @@ def _get_formatter(self, doc_format):
...
+        '<br/>': '\n',
+        '&amp;': '&',
+        '&lt;': '<',
+        '&gt;': '>',
+        '&quot;': '"',
+        '&apos;': "'"
+    }
+
+    def get_shortdoc_from_html(self, doc):
+        match = re.search(r'<p.*?>(.*?)</p>', doc, re.DOTALL)
```
pekkaklarck, line 187: It's not certain text is wrapped with `<p>` or that there's `<p>` in the first place. It's fairly common to use HTML when writing Javadocs, the tooling AFAIK doesn't automatically add `<p>` around the docs, and doc writers in general cannot be expected to add `<p>`s themselves.

pekkaklarck (changes requested):
Looks very good but there are some places where the code should be simplified. I also added some comments about possible enhancements like changing how `<br>` is handled and adding handling for `<hr>`. I'm fine leaving them as-is if you don't agree changes are a good idea.

Another thing to consider is should we do something for tags in general. Converting recognized tags like `<b>` -> `*` is great, but what about `<div>`, `<li>`, etc.? I propose we just remove these tags. We could use

```python
doc = re.sub(r'<(.*?)>(.*?)</\1>', r'\2', doc, re.DOTALL)
```
to cover cases like `<div>foo</div>`
but perhaps
```python
doc = re.sub(r'</?(.*?)>', '', doc)
```
would be better as it would also handle cases where there's no closing tag. This would also avoid the current special handling for `<p>`.

src/robot/libdocpkg/htmlwriter.py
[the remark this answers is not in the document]
Snooz82: one <br> is one line break, so it is part of the short doc and will be replaced with a space later.
That is the current rule.
more than one line break is ending the short doc.
There are utest for that.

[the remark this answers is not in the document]
Snooz82: > single_chars is a bit misleading attribute name already now and gets totally inadequate if we support e.g. <hr>. Something like html_to_text would be better.

i renamed them to `html_chars` because this dict are "html encoded characters" even the line break <br> is a single character later.

I would go so far to replace all possible html situations.
Just the most common ones and those that we create. The shortdoc is later only relevant if users uses HTML as doc.
And if they use funny stuff there, it possibly ends up with some `<hr>` as part of the short doc. I am fine with that as long as the most basic formats like <b> and <strong> are readable fine.

src/robot/libdocpkg/htmlwriter.py
```diff
@@ -142,25 +142,58 @@ def _get_formatter(self, doc_format):
...
+            doc = match.group(1)
+        doc = self.html_to_plain_text(doc)
+        return doc
+
+    def html_to_plain_text(self, doc):
+        for tag, repl in self.html_tags.items():
+            doc = re.sub('(<%(tag)s>)(.*?)(</%(tag)s>)' % {'tag': tag},
+                         lambda m: repl + m.group(2) + repl, doc, flags=re.DOTALL)
+        doc = re.sub("|".join(map(re.escape, self.single_chars.keys())),
+                     lambda match: self.single_chars[match.group(0)], doc)
```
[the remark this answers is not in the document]
Snooz82: ok. but it was in one run .... ;-)

changed it to for loop, but using re.sub to be able to use regex as marker.

```
    html_chars = {
        '<br */?>': '\n',
        '&amp;': '&',
        '&lt;': '<',
        '&gt;': '>',
        '&quot;': '"',
        '&apos;': "'"
    }
```
        for html, text in self.html_chars.items():
            doc = re.sub(html, text, doc)
```

File: src/robot/libdocpkg/htmlwriter.py (hunk 4 above)
Comment (Snooz82) [reply src/robot/libdocpkg/htmlwriter.py:187]: Non closing p are now covered

File: src/robot/libdocpkg/htmlwriter.py [hunk 6]
```diff
@@ -113,8 +115,9 @@ def _yield_header_targets(self, introduction):
                 yield match.group(2)
 
     def _escape_and_encode_targets(self, targets):
-        return NormalizedDict((html_escape(key), self._encode_uri_component(value))
-                              for key, value in targets.items())
+        return NormalizedDict(
+            (html_escape(key), self._encode_uri_component(value))
+            for key, value in targets.items())
```
pekkaklarck, line 120: Why was this change done? In my opinion the old formatting was better. Was the line perhaps over 80 chars? If yes, that's fine as I've anyway been thinking that we should make the line length limit 88 chars.

If the formatting would be changed, then the opening and closing parenthesis should be balanced like

```python
return NormalizedDict(
    params,
    more_params
)
```
not like
```python
return NormalizedDict(
    params,
    more_params)
```

src/robot/libdocpkg/htmlwriter.py
```diff
@@ -129,7 +132,8 @@ def html(self, doc, intro=False):
     def _link_keywords(self, match):
         name = match.group(1)
         if name in self._targets:
-            return '<a href="#%s" class="name">%s</a>' % (self._targets[name], name)
+            return '<a href="#%s" class="name">%s</a>' % (
+            self._targets[name], name)
```
pekkaklarck, line 136: This new formatting, more precisely the indentation of the second like, is very badly against PEP-8 and our conventions. In my opinion it would be best to leave it like it was, but if you want to change it you need to use one of these alternatives:

```python
return ('a template that is pretty long'
        % (some, values))

return 'a template that is pretty long' \
        % (some, values)

return 'a template that is pretty long' % (some,
                                           values)
```

pekkaklarck (changes requested):
The fix itself looks great but the unrelated code changes that are against our style conventions need to be reverted or changed as explained in comments.

It might also be a good idea to change HTML parsing to use Python's HTMLParser instead of regexps, but that can be done separately after merging this PR if needed.

src/robot/libdocpkg/htmlwriter.py
```diff
@@ -14,6 +14,7 @@
 #  limitations under the License.
 
 import re
+
```
pekkaklarck, line 17: This empty line is unnecessary.

Snooz82:
format changes had been because line length was more than 80 chars. 
Now changed back and set to 88.

@pekkaklarck Can be merged now.

PR #3712, opened 2020-10-12
None is not by default converted

mkorpela:
Prevent None conversion by default.
Fixes enum conversion.

### DIFF
--- a/doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
+++ b/doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
@@ -1456,8 +1456,8 @@ case-insensitive.
    |             |               |            | Starting from RF 3.2.2, matching members is case-, space-      |        RED = 1                       |
    |             |               |            | and underscore-insensitive.                                    |        GREEN = 2                     |
    |             |               |            |                                                                |        DARK_GREEN = 3                |
-   |             |               |            |                                                                |                                      |
-   |             |               |            |                                                                | | `GREEN` (Color.GREEN)              |
+   |             |               |            | Starting from RF 4.0 `NONE` is not converted and will result   |                                      |
+   |             |               |            | failure if matching enum value is not found.                   | | `GREEN` (Color.GREEN)              |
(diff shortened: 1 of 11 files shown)


mkorpela:
Failing test seems unrelated:
FAIL: Robot.Standard Libraries.Datetime.Get Current Date.Result format epoch
Test 'Result format epoch' was expected to PASS but it FAILED.

Error message:
'0 <= 1602505329.2355149 - 1602505329.235515 < 1' should be true.

doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
```diff
@@ -1456,8 +1456,8 @@ case-insensitive.
    |             |               |            | Starting from RF 3.2.2, matching members is case-, space-      |        RED = 1                       |
    |             |               |            | and underscore-insensitive.                                    |        GREEN = 2                     |
    |             |               |            |                                                                |        DARK_GREEN = 3                |
-   |             |               |            |                                                                |                                      |
-   |             |               |            |                                                                | | `GREEN` (Color.GREEN)              |
+   |             |               |            | Starting from RF 4.0 `NONE` is not converted and will result   |                                      |
+   |             |               |            | failure if matching enum value is not found.                   | | `GREEN` (Color.GREEN)              |
```
pekkaklarck, line 1460: This explanation may not be that clear. Because this is a bug fix and because `NONE` as an enum member is _very_ rare, I'd say we don't need to have any explanation regarding to it.

A bit above this place there's documentation saying that `NONE`, case-insensitively, is converted to  Python `None`. That needs to be changed/removed and a note added that this behavior has changed in RF 4.0.

pekkaklarck (approved):
These changes look good but there are some things that need to be done after merging them:

- This PR prevents `None` conversion also in cases like `arg: list = None`. That's inconsistent with how we handle `arg=None` and makes the change more backwards incompatible than needed. In my opinion we should always do conversion based on the default value but possible explicit type should obviously be used first. This requires a change to the type conversion algorithm and I can take a look at that afterwards.

PR #3713, opened 2020-10-12
Topic/float int conversion

mkorpela:
Convert float and int values.

Depends on https://github.com/robotframework/robotframework/pull/3712

### DIFF
--- a/src/robot/running/arguments/typeconverters.py
+++ b/src/robot/running/arguments/typeconverters.py
@@ -12,7 +12,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-
+import sys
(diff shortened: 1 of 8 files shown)


src/robot/running/arguments/typeconverters.py
```diff
@@ -200,8 +221,11 @@ def _convert(self, value, explicit_type=True):
 @TypeConverter.register
 class ByteArrayConverter(TypeConverter):
     type = bytearray
+    value_types = (unicode, int)
 
     def _convert(self, value, explicit_type=True):
+        if not PY2 and isinstance(value, int):
+            return bytearray(_int_to_bytes(value))
```
pekkaklarck, line 228: Same comment as to bytes above.

pekkaklarck (commented):
Looks great but converting an integer to bytes feels a bit weird. Unless there are some concrete use cases, I think it could be removed.

PR #3717, opened 2020-10-15
Topic/custom converter

mkorpela:
First draft of the custom converter for custom types.


mkorpela:
This is built on top of union converter, int float converter and stricter none conversion

pekkaklarck:
This PR seems to be outdated.

PR #3720, opened 2020-10-18
LibDoc Model Version3

Snooz82:


atest/robot/libdoc/invalid_usage.robot
```diff
@@ -62,6 +62,9 @@ Invalid output file
     ...    Remove Directory    ${OUT HTML}    AND
     ...    Remove Directory    ${OUT XML}
 
+invalid Spec File version
+    ${TESTDATADIR}/OldSpec.xml ${OUT XML}    Invalid spec file version 'None'. RF >= 4.0 requires XML specversion 3.
```
Snooz82, line 66: write our Robot Framework

src/robot/libdocpkg/model.py
```diff
@@ -142,11 +140,23 @@ def _sort_key(self):
...
             'tags': tuple(self.tags),
             'source': self.source,
             'lineno': self.lineno,
             'matched': True
         }
+
+    def convert_arguments(self):
+        return [self.arg_to_dict(arg_info) for arg_info in self.args]
+
+    def arg_to_dict(self, arg_info):
```
Snooz82, line 155: should be private

src/robot/libdocpkg/specbuilder.py
```diff
@@ -45,39 +47,85 @@ def _parse_spec(self, path):
             root = ET.parse(source).getroot()
         if root.tag != 'keywordspec':
             raise DataError("Invalid spec file '%s'." % path)
+        version = root.get('specversion')
+        if version != '3':
+            raise DataError("Invalid spec file version '%s'. "
+                            "RF >= 4.0 requires XML specversion 3." % version)
```
Snooz82, line 53: Write Robot Framework

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -79,21 +74,35 @@ def _get_old_style_scope(self, libdoc):
...
                 writer.end('tags')
             writer.end(kw_type)
+        writer.end(list_name)
+
+    def _write_arguments(self, kw, writer):
+        writer.start('arguments', {'repr': unicode(kw.args)})
+        for arg in kw.args:
+            writer.start('arg', {'kind': arg.kind,
+                                 'required': 'true' if arg.required else 'false',
+                                 'string_repr': unicode(arg)})  # ToDo: string_repr shall be deleted when atests are changed
```
Snooz82, line 97: repr not string_repr

src/robot/running/arguments/argumentspec.py
```diff
@@ -139,12 +140,28 @@ def required(self):
     @property
     def type_string(self):
```
Snooz82, line 141: type_repr instead of type_string

src/robot/running/arguments/argumentspec.py
```diff
@@ -139,12 +140,28 @@ def required(self):
...
-        if issubclass(self.type, Enum):
+            return
+        if isinstance(self.type, Enum):
             return self._format_enum(self.type)
-        return self.type.__name__
+        if isclass(self.type):
+            return self.type.__name__
+        type_name = str(self.type)
+        if type_name.startswith('typing.'):
+            type_name = type_name[len('typing.'):]
```
Snooz82, line 150: move it to different issue

src/robot/running/arguments/argumentspec.py
```diff
@@ -139,12 +140,28 @@ def required(self):
...
+    def default_string(self):
+        if self.default is self.NOTSET:
+            return None
+        if isinstance(self.default, Enum):
+            return self.default.name
+        if isinstance(self.default, str):
+            if self.default == '':
+                return '${Empty}'
+            value_repr = repr(self.default)[1:-1]
+            return re.sub('^(?= )|(?<= )$|(?<= )(?= )', r'\\', value_repr)
```
Snooz82, line 163: move it to other Issue

Snooz82:
@pekkaklarck 
That one should now be ready for review.

JSON Ex/Import and JSONSpec will be done in separate task.

Also the "removal" of the enum values and the real datatype will be a separate task.

atest/robot/libdoc/LibDocLib.py
```diff
@@ -67,3 +68,10 @@ def relative_source(self, path, start):
             return relpath(path, start)
         except ValueError:
             return normpath(path)
+
+    def get_repr_from_arg_model(self, model):
+        arg_info = ArgInfo(kind=model['kind'],
+                           name=model['name'],
+                           type=model['type'] or ArgInfo.NOTSET,
+                           default=model['default'] or ArgInfo.NOTSET)
+        return unicode(arg_info)
```
pekkaklarck, line 77: I think this could just be `return unicode(ArgInfo(**model))`.

atest/robot/libdoc/html_output.robot
```diff
@@ -3,6 +3,14 @@ Resource          libdoc_resource.robot
 Suite Setup       Run Libdoc And Parse Model From HTML    ${TESTDATADIR}/module.py
 Test Template     Should Be Equal Multiline
 
+*** Keywords ***
+Verify Argument Models
+    [Arguments]    ${arg_models}    @{expected_reprs}
+    Should Be True    len($arg_models) == len($expected_reprs)
+    FOR    ${arg_model}    ${expected_repr}    IN ZIP    ${arg_models}    ${expected_reprs}
+       Run Keyword And Continue On Failure    Verify Argument Model    ${arg_model}    ${expected_repr}
+    END
```
pekkaklarck, line 12: We've used order Settings/Variables/Tests/Keywords in our data files. Would be better to follow that convention here as well.

atest/robot/libdoc/invalid_usage.robot
```diff
@@ -62,6 +62,9 @@ Invalid output file
     ...    Remove Directory    ${OUT HTML}    AND
     ...    Remove Directory    ${OUT XML}
 
+invalid Spec File version
+    ${TESTDATADIR}/OldSpec.xml ${OUT XML}    Invalid spec file version 'None'. Robot Framework >= 4.0 requires specversion 3.
```
pekkaklarck, line 66: - "specversion" -> "spec version" or just "version"
- ">=" -> "4.0 and newer"

src/robot/libdocpkg/consoleviewer.py
```diff
@@ -74,7 +72,7 @@ def _show_inits(self, lib):
     def _show_keyword(self, kw, show_name=True):
         if show_name:
             self._header(kw.name, underline='-')
-        self._data([('Arguments', '[%s]' % ', '.join(kw.args))])
+        self._data([('Arguments', '[%s]' % unicode(kw.args))])
```
pekkaklarck, line 75: `unicode()` could be omitted by using `u'[%s]'` as the template. This is fine as well.

src/robot/libdocpkg/model.py
```diff
@@ -142,11 +140,24 @@ def _sort_key(self):
...
             'doc': self.doc,
             'shortdoc': self.shortdoc,
             'tags': tuple(self.tags),
             'source': self.source,
             'lineno': self.lineno,
             'matched': True
         }
+
+    def _convert_arguments(self):
+        return [self._arg_to_dict(a) for a in self.args]
```
pekkaklarck, line 153: Could be inlined.

src/robot/libdocpkg/specbuilder.py
```diff
@@ -25,17 +26,17 @@ class SpecDocBuilder(object):
 
     def build(self, path):
         spec = self._parse_spec(path)
+
```
pekkaklarck, line 29: We try to avoid empty lines inside functions or methods. I don't think this is needed.

src/robot/libdocpkg/specbuilder.py
```diff
@@ -45,39 +46,60 @@ def _parse_spec(self, path):
...
+                setters[arg.get('kind')](name)
+                spec.defaults.update(self._get_default(arg, name))
+                spec.type = self._get_updated_types(arg, name, spec)
+        return spec
+
+    @staticmethod
+    def _get_name(arg):
+        name_elem = arg.find('name')
+        if name_elem is not None:
+            return name_elem.text
```
pekkaklarck, line 89: Should have explicit `return None` as well. Perhaps as ternary like `return name_elem.text if name_elem else None`.

src/robot/libdocpkg/specbuilder.py
```diff
@@ -45,39 +46,60 @@ def _parse_spec(self, path):
...
+            ArgInfo.VAR_POSITIONAL: lambda value: setattr(spec, 'var_positional', value),
+            ArgInfo.NAMED_ONLY: spec.named_only.append,
+            ArgInfo.VAR_NAMED: lambda value: setattr(spec, 'var_named', value),
+        }
+        for arg in elem.findall('arguments/arg'):
+            name = self._get_name(arg)
+            if name:
+                setters[arg.get('kind')](name)
+                spec.defaults.update(self._get_default(arg, name))
+                spec.type = self._get_updated_types(arg, name, spec)
```
pekkaklarck, line 82: The above with the needed helpers is pretty complicated. This would in my opinion be a lot simpler considering it needs no helpers:

```python
name = arg.find('name')
if not name:
    continue
name = name.text
default = arg.find('default')
if default:
    spec.defaults[name] = default.text
type = arg.find('type')
if type:
    if not spec.types:
        spec.types = {}
    spec.types[name] = type.text
```

Could obviously use `name_elem` instead of `name` to avoid using the same variable name.

Notice also that the original code has a bug: `spec.type` should be `spec.types`.

pekkaklarck (changes requested):
Looks great! Most of the comments are about style and I was about to merge the code as-is. Noticed a bug in reading type information from specs in the end so at least the needs to be still fixed.

src/robot/libdocpkg/specbuilder.py
Snooz82: obsolete

src/robot/libdocpkg/specbuilder.py
Snooz82: i changed it.

But you forgot to write the kind with the setter and 
it is essential to check if xml_elem is `None` because also a found xml_element is falsly.
i don't know why, but that is it.

mkorpela:
Done and will work on top this.

src/robot/libdocpkg/model.py
```diff
@@ -150,14 +150,10 @@ def to_dictionary(self):
...
-            'required': arg_info.required,
-            'repr': unicode(arg_info)
-        }
+        return [{'name': a.name,
+                 'type': a.type_repr,
+                 'default': a.default_repr,
+                 'kind': a.kind,
+                 'required': a.required,
+                 'repr': unicode(a)
+                 } for a in self.args]
```
pekkaklarck, line 159: In my earlier comment I meant that `_convert_arguments` could be removed and the list comprehension it contained could be just used where it was called. `_arg_to_dict` was fine as it was. Sorry for not being clear about that. Good thing is that the code is fine like this as well.

src/robot/libdocpkg/specbuilder.py
```diff
@@ -75,31 +74,17 @@ def _create_arguments(self, elem):
...
+                continue
+            name = name_elem.text
+            setters[arg.get('kind')](name)
+            default_elem = arg.find('default')
+            if default_elem is not None:
+                spec.defaults[name] = default_elem.text or ''
+            type_elem = arg.find('type')
+            if type_elem is not None:
+                if not spec.types:
+                    spec.types = {}
```
pekkaklarck, line 88: It occurred to me that this `if` could be avoided by using `spec = ArgumentSpec(types={})` earlier.

src/robot/libdocpkg/specbuilder.py
```diff
@@ -75,31 +74,17 @@ def _create_arguments(self, elem):
...
+                spec.types[name] = type_elem.text
```
pekkaklarck, line 89: Great that the bug with setting type info is fixed. It would have been good to add a test validating the fix as well but I assume you are going to add more round-trip tests in the future.

PR #3727, opened 2020-10-20
added SKIP status and removed critical attribute (fixes #3726)

### LINKED ISSUE #3726: Update and enhance output.xml schema
update RF output.xml schema to reflect v4.0 changes (e.g. SKIP status, critical attribute, and other)

https://github.com/robotframework/robotframework/tree/master/doc/schema

bitcoder:
fixes #3726

### DIFF
--- a/doc/schema/robot-xsd10.xsd
+++ b/doc/schema/robot-xsd10.xsd
@@ -72,14 +72,7 @@
             <xs:restriction base="xs:string">
                 <xs:enumeration value="PASS"/>
                 <xs:enumeration value="FAIL"/>
-            </xs:restriction>
-        </xs:simpleType>
-    </xs:attribute>
-    <xs:attribute name="critical">
-        <xs:simpleType>
-            <xs:restriction base="xs:string">
-                <xs:enumeration value="yes"/>
-                <xs:enumeration value="no"/>
+                <xs:enumeration value="SKIP"/>
             </xs:restriction>
         </xs:simpleType>
     </xs:attribute>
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Comments about the PR:
- Looks good to me in general.
- Critical Tests have been removed under `statistics/total`, but it seems schemas aren't strict enough to notice that.
- We are going to add SKIP as a message level so that should be taken into account as well. Was it already added @yanne? Can you @bitcoder add it to schemas?

While looking at the schemas, I noticed we still have attributes like `html` that has value `yes` or `no`. I'd like to change that to `true/false` (i.e. `xs:boolean`) and possibly see are there others where we could use the standard XSD types. Do you @bitcoder have opinion is that a good idea? Using standard types sounds good in general for my but I'm worried about backwards incompatible change? It would need a separate issue anyway.

bitcoder:
@pekkaklarck ,

-  I've added SKIP to the schemas

> * Critical Tests have been removed under `statistics/total`, but it seems schemas aren't strict enough to notice that.

- well, the schema as it is defines the possible attributes; it doesn't restrict you from using other attributes (including the one that we removed)

- on the `html` yes/no topic, I think the `xs:boolean` type in fact makes more sense; I haven't changed it as it may require a broader decision, as you mentioned, that could affect other attributes

pekkaklarck:
According to @yanne this PR now covers all changes in RF 4.0 so I'll merge it. Changing `yes/no` to `true/false` and doing other such changes can be considered but needs a separate issue.

PR #3728, opened 2020-10-20
Topic/Libdoc - JsonWriter and JsonBuilder (fixes #3730)

### LINKED ISSUE #3730: LibDoc should support JSON output and also JSON input
For now Libdoc can Generate XML Files from any Library or Keyword Resource file.
These XML Files can be used in other Editors or tools as source of information.
Libdoc itself can also read these XML files to generate HTML from that.

JSON output is already embedded in the HTML output.

This Feature should implement JSON as a new output format for Libdoc that can also be used as an input format.

The default documentation output should be HTML.
That means that all input docs from the source are automatically generated into HTML documentation strings.
This should be possible to Switch off.

- [x] Generate JSON output format
- [x] Read from JSON file to generate XML or HTML
- [x] `*.json` output should automatically contain HTML docs
- [x] JSON-Schema is available in docs
- [x] LibDoc Help is updated
- [x] UserGuide is updated
- [x] atests are written/updated
- [x] utests are created for libdoc

Snooz82:


Snooz82:
No Tests yet, but fully functional.

src/robot/libdocpkg/jsonbuilder.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def build_from_dict(self, libdoc_dict):
+        libdoc = LibraryDoc(name=libdoc_dict.get('name'),
+                            doc=libdoc_dict.get('doc'),
+                            version=libdoc_dict.get('version'),
+                            type=libdoc_dict.get('type'),
+                            scope=libdoc_dict.get('scope'),
+                            doc_format=libdoc_dict.get('doc_format'),
+                            source=libdoc_dict.get('source'),
+                            lineno=int(libdoc_dict.get('lineno')))
```
pekkaklarck, line 38: - Does `libdoc_dict` have extra items or only those that are accessed? If not, all this could be replaced with `libdoc = LibraryDoc(**libdoc_dict)`.

- Above proposal wouldn't convert `lineno` to `int` but I would assume it's `int` in JSON as well. I think it may also be `null` in which case `int´ conversion would actually fail.

- `libdoc_dict` might be better named `spec` like it's i `build`. Also `build_from_dict´ could be `build_from_spec`.

- Is the idea that `build_from_dict` is public? If not, prefix with `_` .

src/robot/libdocpkg/jsonbuilder.py
```diff
@@ -0,0 +1,81 @@
...
+        libdoc.inits = self._create_keywords(libdoc_dict.get('inits'))
+        libdoc.keywords = self._create_keywords(libdoc_dict.get('keywords'))
+        return libdoc
+
+    @staticmethod
```
pekkaklarck, line 43: I don't think making helper methods explicitly static adds much value. In my opinion it mainly adds distraction and also makes it harder to turn the method non-static if it later needs `self`. I know IDEs can add a note that a method could be static, but I've just disabled those inspections. Situation is different with public methods that are meant to be used externally in static manner. Anyway, if you prefer having the decorator feel free to leave it.

src/robot/libdocpkg/jsonbuilder.py
```diff
@@ -0,0 +1,81 @@
...
+        return libdoc_dict
+
+    def _create_keywords(self, keywords):
+        return [KeywordDoc(name=kw.get('name'),
+                           args=self._create_arguments(kw.get('args')),
+                           doc=kw.get('doc'),
+                           shortdoc=kw.get('shortdoc'),
+                           tags=kw.get('tags'),
+                           source=kw.get('source'),
+                           lineno=int(kw.get('lineno'))) for kw in keywords]
```
pekkaklarck, line 58: - Unfortunate that args need special handling and this cannot changed just to `KeywordDoc(**kw)`.

- Not all keywords have line number information so `int()` conversion may sometimes fail. If line number is stored in the spec as an integer or null, no conversion is needed.

src/robot/libdocpkg/htmlutils.py
```diff
@@ -1,5 +1,4 @@
-#  Copyright 2008-2015 Nokia Networks
-#  Copyright 2016-     Robot Framework Foundation
+#  Copyright 2020-     Robot Framework Foundation
```
pekkaklarck, line 1: What's this change? Is this a new file where you don't want have old Nokia copyrights? I think it would be better to leave them because I think the file anyway contains old code that has been moved. It's also easier if all files have same copyrights.

src/robot/libdocpkg/jsonbuilder.py
Snooz82: @pekkaklarck 
1. in the json are more information than the constructor accepts.
example: Generated, all_tags, keywords and inits

2. changed the int to int(spec.get('lineno', -1)) as in specbuilder.

3. and 4. Yea the idea was that that could maybe used to generate Libdoc from the same format that JSON creates but as dict, because the `build` method only accepts a path. so if someone/something want to use the jsonbuilder it would need to write a file first.
I did change the variable name to `spec` but i wouldn’t change the method name, because it points out that compared to `build` from a path that one can `build` the model from a dictionary.

i changed all `.get('key')` to normal dictionary syntax

src/robot/libdocpkg/htmlutils.py
Snooz82: can do that.

was just not there before.
But no opinion on that.

doc/userguide/src/SupportingTools/Libdoc.rst
```diff
@@ -41,22 +41,22 @@ Synopsis
...
-   -s --specdocformat <raw|html>
+  -f, --format <html|xml>
+                          Specifies whether to generate an HTML, XML or
+                          JSON output file. The default output `format`
+                          and the default output `specdocformat` is fetched
+                          from the output file extension.
+                          :file:`*.html` -> `-f HTML -s HTML`,
+                          :file:`*.xml` -> `-f XML -s RAW`,
+                          :file:`*.json` -> `-f JSON -s HTML` and
+                          :file:`*.libspec` -> `-f XML -s HTML`
```
pekkaklarck, line 52: Explanation related to default spec doc format doesn't really belong here. It's not a directly related option and isn't even applicable with the HTML format. The end of the documentation could be something like this:

---

The default `format` is got based on the output file extension as follows:

- `*.html` -> HTML
- `*.xml` -> XML
- `*.json` -> JSON
- `*.libspec` -> XML (with HTML spec documentation format)

---

doc/userguide/src/SupportingTools/Libdoc.rst
```diff
@@ -41,22 +41,22 @@ Synopsis
...
+  -s, --specdocformat <raw|html>
                           Specifies for XML outputs whether the keyword
```
pekkaklarck, line 54: Not only XML anymore. I'd use something like "XML and JSON spec files" to use the word spec also in the documentation and not only in the option name.

doc/userguide/src/SupportingTools/Libdoc.rst
```diff
@@ -41,22 +41,22 @@ Synopsis
...
                           documentation is converted to HTML regardless
                           of the original documentation format or kept
-                          raw as in the library or source spec file.
-                          :file:`*.html` files always get HTML documentation
-                          format. Default is based on the format.
-                          `--format HTML` -> `HTML`
-                          `--format XML`  -> `RAW`
+                          raw as in the library, resource or source spec file.
+                          :file:`*.html` files only accept `HTML`.
+                          Default is based on the format.
```
pekkaklarck, line 59: Default values should be explained here. I think the whole doc could be something like this:

---

Specifies the documentation format to use with XML and JSON spec files. `HTML` means converting the documentation to HTML regardless the original documentation format and `RAW` means keeping the original format. The default is got based on the `--format` option so that with XML the default is `RAW` and with JSON it is `HTML`. If `--format` is not used, the final default is got based on the output file extension so that with `*.xml` default is `RAW` and with `*.libspec` and `*.json` it is `HTML`.

---

doc/userguide/src/SupportingTools/Libdoc.rst
```diff
@@ -220,16 +225,36 @@ Robot Framework 4.0 introduced new spec version `3`.
 
 .. note:: Robot Framework 4.0 introduced an updated xml format. It is
           incompatible with former versions. LibDoc can only read and write
-          new XML format.
+          new XML format version 3.
 
-          The `xml:html` format introduced in Robot Framework 3.2. has been
-          replaced by the option `-s --specdocformat <raw|html>`.
+          The `XML:HTML` format introduced in Robot Framework 3.2. has been
+          replaced by the format `LIBSPEC` ot the option combination
```
pekkaklarck, line 231: - Do we really have `LIBSPEC` format? Isn't it only an extension?
- "ot" -> "or"

doc/userguide/src/SupportingTools/Libdoc.rst
```diff
@@ -220,16 +225,36 @@ Robot Framework 4.0 introduced new spec version `3`.
...
+
+Since Robot Framework 4.0 Libdoc can also generate documentation in JSON
+format that is suitable for external tools such as editors or web pages.
+It contains all the same information as the HTML format but in a machine
+readable format.
+
+Similar to XML spec files the JSON spec files contain all information and
+can also be used as input to Libdoc. From that format any other output format
+can be created. By default the library documentation strings are converted
+to HTML format within the json output file.
```
pekkaklarck, line 250: json -> JSON

doc/userguide/src/SupportingTools/Libdoc.rst
```diff
@@ -220,16 +225,36 @@ Robot Framework 4.0 introduced new spec version `3`.
...
+
+The exact JSON spec file format is documented with an `JSON schema`__ (json)
```
pekkaklarck, line 252: What that `(json)` at the end?

src/robot/libdoc.py
```diff
@@ -64,22 +64,22 @@
...
                           documentation is converted to HTML regardless
                           of the original documentation format or kept
-                          raw as in the library or source spec file.
-                          `*.html` files always get HTML documentation
-                          format. Default is based on the format.
-                          `--format HTML` -> HTML
-                          `--format XML`  -> RAW
+                          raw as in the library, resource or source spec file.
+                          `*.html` files only accept `HTML`.
+                          Default is based on the format.
```
pekkaklarck, line 82: Same comments to these docs as to docs in the UG.

src/robot/libdoc.py
```diff
@@ -107,6 +107,8 @@
 
   python -m robot.libdoc src/MyLibrary.py doc/MyLibrary.html
   jython -m robot.libdoc MyLibrary.java MyLibrary.html
+  python -m robot.libdoc src/MyLibrary.py doc/MyLibrary.json
+  python -m robot.libdoc src/MyLibrary.json doc/MyLibrary.html
```
pekkaklarck, line 111: Probably `doc/MyLibrary.json` would be better as the source in the latter example?

src/robot/libdocpkg/htmlwriter.py
```diff
@@ -15,11 +15,16 @@
...
 
+from robot.errors import DataError
 from robot.htmldata import HtmlFileWriter, ModelWriter, LIBDOC
 
 
 class LibdocHtmlWriter(object):
 
+    def __init__(self, spec_doc_format):
+        if spec_doc_format != 'HTML':
+            raise DataError("Spec doc format must be 'HTML' when using 'HTML' output, got '%s'." % spec_doc_format)
```
pekkaklarck, line 26: In my opinion --specdocformat should be invalid with HTML format regardless the value. That could be easily validated when handling arguments in the main Libdoc program.

src/robot/libdocpkg/jsonbuilder.py
```diff
@@ -0,0 +1,82 @@
...
+        return libdoc_dict
+
+    def _create_keywords(self, keywords):
+        return [KeywordDoc(name=kw.get('name'),
+                           args=self._create_arguments(kw['args']),
+                           doc=kw['doc'],
+                           shortdoc=kw['shortdoc'],
+                           tags=kw['tags'],
+                           source=kw['source'],
+                           lineno=int(kw.get('lineno', -1))) for kw in keywords]
```
pekkaklarck, line 58: I'd have just `_create_keyword()` and then call it within a list comprehension like

```python
libdoc.inits = [self._create_keyword(i) for i in spec['inits']]
libdoc.keywords = [self._create_keyword(k) for k in spec['keywords']]
```

src/robot/libdocpkg/jsonbuilder.py
```diff
@@ -0,0 +1,82 @@
...
+            setters[arg['kind']](name)
+            default = arg['default']
+            if default:
+                spec.defaults[name] = default
+            arg_type = arg['type']
+            if arg_type is not None:
+                if not spec.types:
+                    spec.types = {}
+                spec.types[name] = arg_type
+        return spec
```
pekkaklarck, line 82: We could enhance ArgmentSpec to ease building it based on this data. I guess it would be pretty easy to create ArgInfo objects here and then having a classmethods `ArgumentSpec.from_arg_infos` would make it trivial to convert them to an ArgumentSpec. But let's take a look at that ofter merging this PR.

src/robot/libdocpkg/jsonwriter.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class LibdocJsonWriter(object):
+
+    def __init__(self, spec_doc_format):
+        self._spec_doc_format = spec_doc_format
+
+    def write(self, libdoc, outfile):
+        if self._spec_doc_format == 'HTML':
+            libdoc.convert_doc_to_html()
```
pekkaklarck, line 28: I think it would be a bit cleaner if the caller would have called `libdoc.convert_doc_to_html()` if needed already. Then there would be no need to pass `spec_doc_format` here or to other writers.

src/robot/libdocpkg/model.py
```diff
@@ -137,6 +147,10 @@ def deprecated(self):
     def _sort_key(self):
         return self.name.lower()
 
+    def _generate_shortdoc(self):
+        if not self._shortdoc:
+            self.shortdoc = self._get_shortdoc()
```
pekkaklarck, line 152: This doesn't seem to be used anywhere.

src/robot/libdocpkg/specbuilder.py
```diff
@@ -68,8 +69,10 @@ def _create_arguments(self, elem):
         spec = ArgumentSpec()
         setters = {
             ArgInfo.POSITIONAL_ONLY: spec.positional_only.append,
+            ArgInfo.POSITIONAL_ONLY_MARKER: lambda *args: None,
             ArgInfo.POSITIONAL_OR_NAMED: spec.positional_or_named.append,
             ArgInfo.VAR_POSITIONAL: lambda value: setattr(spec, 'var_positional', value),
+            ArgInfo.NAMED_ONLY_MARKER: lambda *args: None,
```
pekkaklarck, line 75: `lambda value: None` would be consistent with other setters.

src/robot/libdocpkg/writer.py
```diff
@@ -17,14 +17,17 @@
...
     format = (format or 'HTML')
     if format == 'HTML':
-        return LibdocHtmlWriter()
+        return LibdocHtmlWriter(spec_doc_format or 'HTML')
     if format == 'XML':
         return LibdocXmlWriter(spec_doc_format or 'RAW')
     if format == 'LIBSPEC':
         return LibdocXmlWriter(spec_doc_format or 'HTML')
+    if format == 'JSON':
+        return LibdocJsonWriter(spec_doc_format or 'HTML')
```
pekkaklarck, line 32: As I commented already with `LibdocJsonWriter`, it would be better to convert docs to HTML if needed already on this side instead of passing `spec_doc_format` to all writers. Could be that doc conversion would be even better done before even calling this factory method.

pekkaklarck (changes requested):
Looks very good! There were some problems with docs that I hope you could address and the code could also be enhanced a bit in some places. Especially converting docs to HTML if needed before calling writers would be good.

atest/requirements.txt
```diff
@@ -19,3 +19,4 @@ lxml; platform_python_implementation == 'CPython' or platform_python_implementat
 pillow < 7; platform_system == 'Windows' and python_version == '2.7'
 pillow < 6; platform_system == 'Windows' and python_version == '3.4'
 pillow >= 7.1.0; platform_system == 'Windows' and python_version >= '3.5'
+jsonschema; platform_python_implementation != 'IronPython' and platform_python_implementation != 'Jython'
```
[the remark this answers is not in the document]
Snooz82: i can try

doc/userguide/src/SupportingTools/Libdoc.rst
Snooz82: i think it is handy to have it here, because file extension and `--format` is directly affecting the `--specdocformat`. So this is the most direct place to say that. As in command line as well.
Users can imho easily understand how these two options affect each other.

But i am totally fine if you want to restructure the docs otherwise.
I think this is more a question of taste so maybe you can put it where you think it is better.

doc/userguide/src/SupportingTools/Libdoc.rst
Snooz82: can you please fix the docs so that they mach your taste, please?

i think

> Default is based on the `format`.

give the user the hint to just look one line to the top and would see the defaults.

doc/userguide/src/SupportingTools/Libdoc.rst
Snooz82: There was this extra special handling of `extension == libspec` then format is `XML:HTML`
That would have to be replace by `*.libspec` would be `format = XML` and `specdocformat = HTML` except the specdocformat was given...

to remove that and create consistency it was imho the best way to just say that when the extension `*.libspec` is allowed, why not just have it as a format as well.

doc/userguide/src/SupportingTools/Libdoc.rst
Snooz82: 🤷🏻‍♂️

It was there because (xsd) was there on xml schema... but you are right. makes no sense

src/robot/libdocpkg/htmlwriter.py
Snooz82: i think that the validation of a format accept that or not can be should be done in the writer itself.
and i definitely not think that `libdoc -f HTML -s HTML` should be invalid!
Because sometimes you want to script that and imho there is not need for a failure here.

So if you would make a batch or shellscript for that you would have to get rid of that `-s` if the other value is HTML wich just make live harder and do not add any value.

imho we could just not fail anyway and always just create HTMl, but you wanted that error.

src/robot/libdocpkg/jsonbuilder.py
Snooz82: yes. different tasks.

src/robot/libdocpkg/jsonwriter.py
Snooz82: you can refactor if you really want, but i think that is easy to understand.

Snooz82: that is not invented in that task, so i would like to keep jsonwriter as all other writers

src/robot/libdocpkg/model.py
Snooz82: It is used. It was just by accident a private method.

That is really important for generating the shortdoc when the keywords list is set to LibraryDoc object.
and it is only generated if the shortdoc is not already set, what means it only happens in Java- and robotbbuilder, but not in json- and xmlbuilder.

src/robot/libdocpkg/writer.py
Snooz82: i think that refactoring is out of scope of that issue.
can be done in a later task.

But still i think the best place for that would be the model.

```python
class LibdocJsonWriter(object):

    def __init__(self, spec_doc_format):
        self._spec_doc_format = spec_doc_format

    def write(self, libdoc, outfile):
        libdoc.convert_doc_to(self._spec_doc_format)
        with file_writer(outfile) as f:
            json.dump(libdoc.to_dictionary(), f, indent=2)
```

pekkaklarck (approved):

PR #3743, opened 2020-10-28
DataTypes in Libdoc Specs

Snooz82:


src/robot/running/arguments/argumentspec.py
```diff
@@ -128,6 +135,25 @@ def __init__(self, kind, name='', type=NOTSET, default=NOTSET):
         self.type = type
         self.default = default
 
+    @setter
+    def type(self, _type):
+        if not _type or _type is self.NOTSET:
+            return tuple()
+        if isinstance(_type, (tuple, list)):
+            return tuple(_type)
```
Snooz82, line 143: just return _type if tuple.
No handling of lists!

src/robot/running/arguments/argumentspec.py
```diff
@@ -138,13 +164,21 @@ def required(self):
...
-        if isclass(self.type):
-            if issubclass(self.type, Enum):
-                return self._format_enum(self.type)
-            return self.type.__name__
-        return unicode(self.type)
+        return ' | '.join(t)
+
+    @property
+    def type_as_repr_list(self):
+        return [self._type_repr(t) for t in self.type]
```
Snooz82, line 174: type_repr_list

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -93,23 +94,48 @@ def _write_arguments(self, kw, writer):
                                  'repr': unicode(arg)})
             if arg.name:
                 writer.element('name', arg.name)
-            if arg.type is not arg.NOTSET:
-                writer.element('type', arg.type_repr)
+            if arg.type and arg.type is not arg.NOTSET:
+                for type_repr in arg.type_as_repr_list:
```
Snooz82, line 98: can arg.type be NOTSET?

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -93,23 +94,48 @@ def _write_arguments(self, kw, writer):
                                  'repr': unicode(arg)})
             if arg.name:
                 writer.element('name', arg.name)
-            if arg.type is not arg.NOTSET:
-                writer.element('type', arg.type_repr)
+            if arg.type and arg.type is not arg.NOTSET:
```
Snooz82, line 97: change to types

src/robot/libdocpkg/specbuilder.py
```diff
@@ -85,9 +86,35 @@ def _create_arguments(self, elem):
...
             if default_elem is not None:
                 spec.defaults[name] = default_elem.text or ''
-            type_elem = arg.find('type')
-            if type_elem is not None:
+            type_elems = arg.findall('type')
+            if type_elems is not None:
                 if not spec.types:
                     spec.types = {}
-                spec.types[name] = type_elem.text
+                spec.types[name] = [t.text for t in type_elems]
```
Snooz82, line 93: make tuple here

src/robot/libdocpkg/model.py
```diff
@@ -183,9 +247,74 @@ def to_dictionary(self):
     def _arg_to_dict(self, arg):
         return {
             'name': arg.name,
-            'type': arg.type_repr,
+            'type': arg.type_as_repr_list,
```
Snooz82, line 250: types

src/robot/libdocpkg/model.py
```diff
@@ -86,28 +147,31 @@ def save(self, output=None, format='HTML'):
...
             'scope': self.scope,
             'doc_format': self.doc_format,
             'source': self.source,
             'lineno': self.lineno,
+            'all_tags': list(self.all_tags),
             'inits': [init.to_dictionary() for init in self.inits],
             'keywords': [kw.to_dictionary() for kw in self.keywords],
-            'generated': get_timestamp(daysep='-', millissep=None),
-            'all_tags': list(self.all_tags)
+            'data_types': [dt.to_dictionary() for dt in self.data_types]
```
Snooz82, line 174: Rename keys in JSON to follow JSON/JS best practice.

src/robot/running/arguments/argumentspec.py
```diff
@@ -173,7 +198,7 @@ def __unicode__(self):
             ret = '*' + ret
         elif self.kind == self.VAR_NAMED:
             ret = '**' + ret
-        if self.type is not self.NOTSET:
+        if self.type:
             ret = '%s: %s' % (ret, self.type_repr)
```
Snooz82, line 202: is type_repr only used here, make the ' | ' joining here

src/robot/libdocpkg/model.py
```diff
@@ -12,13 +12,28 @@
...
+        pass
+
+try:
+    from typing_extensions import TypedDict
+
+    class _TypedDictDummy(TypedDict):
+        pass
+    TypedDictType = type(_TypedDictDummy)
+except ImportError:
+    TypedDictType = None
```
Snooz82, line 32: conditional import of typing or typing_extensions

Snooz82: from typing import TypedDict
type(TypedDict('X', {}))
<class 'typing._TypedDictMeta'>

Snooz82: import typing_extensions with old python versions

src/robot/libdocpkg/model.py
```diff
@@ -38,9 +53,20 @@ def __init__(self, name='', doc='', version='', type='LIBRARY',
         self.doc_format = doc_format
         self.source = source
         self.lineno = lineno
+        self._data_types = dict()
         self.inits = []
```
Snooz82, line 57: check to rename to self.data_types

src/robot/libdocpkg/model.py
```diff
@@ -65,18 +94,50 @@ def doc_format(self, format):
...
+    def _sort_keywords(self, kws):
         for keyword in kws:
+            self._add_types_from_keyword(keyword)
             keyword.parent = self
             keyword.generate_shortdoc()
         return sorted(kws)
 
+    def _add_types_from_keyword(self, keyword):
+        for arg in keyword.args:
+            if arg.type:
```
Snooz82, line 112: use typ instead of _type

get rid of that if

src/robot/libdocpkg/model.py
```diff
@@ -65,18 +94,50 @@ def doc_format(self, format):
...
+                for type_repr, _type in zip(arg.type_as_repr_list, arg.type):
+                    if (type_repr not in self._data_types
+                            and isinstance(_type, type)
+                            and type(_type) is not type):
+                        self._data_types[type_repr] = self._get_type_doc(_type)
```
Snooz82, line 117: explicit check for only supported types

src/robot/libdocpkg/model.py
```diff
@@ -65,18 +94,50 @@ def doc_format(self, format):
...
+    def _get_type_doc(self, arg_type):
+        if isinstance(arg_type, (EnumDoc, TypedDictDoc)):
+            return arg_type
+        elif self._is_TypedDictType(arg_type):
+            return TypedDictDoc(arg_type)
+        elif self._is_EnumType(arg_type):
+            return EnumDoc(arg_type)
+
+    def _is_TypedDictType(self, arg_type):
+        return (TypedDictType
```
Snooz82, line 128: remove the TypedDictType

src/robot/libdocpkg/model.py
```diff
@@ -183,9 +247,74 @@ def to_dictionary(self):
...
+
+    def __init__(self, type_info=None, name='', super='', doc='', items=None,
+                 required_keys=None, optional_keys=None):
+        self.name = name
+        self.super = super
+        self.doc = doc
+        self.items = items or dict()
+        self.required_keys = required_keys or list()
+        self.optional_keys = optional_keys or list()
+        if isinstance(type_info, dict):
```
Snooz82, line 268: use class methods

src/robot/libdocpkg/model.py
```diff
@@ -183,9 +247,74 @@ def to_dictionary(self):
...
+            'name': self.name,
+            'super': self.super,
+            'doc': self.doc,
+            'items': self.items,
+            'required_keys': self.required_keys,
+            'optional_keys': self.optional_keys
+        }
+
+
+class EnumDoc:
```
Snooz82, line 295: class EnumDoc(object):

src/robot/libdocpkg/specbuilder.py
```diff
@@ -85,9 +86,35 @@ def _create_arguments(self, elem):
...
+        args = dict()
+        args['name'] = dt.get('name')
+        args['super'] = dt.get('super')
+        args['doc'] = dt.find('doc').text or ''
+        if args['super'] == 'Enum':
+            args['members'] = [
+                {'name': member.get('name'), 'value': member.get('value')}
+                for member in dt.findall('members/member')]
+            return EnumDoc(**args)
+        elif args['super'] == 'TypedDict':
```
Snooz82, line 109: just if

src/robot/libdocpkg/specbuilder.py
```diff
@@ -85,9 +86,35 @@ def _create_arguments(self, elem):
...
+            args['required_keys'] = list()
+            args['optional_keys'] = list()
+            for item in dt.findall('items/item'):
+                key = item.get('key')
+                args['items'][key] = item.get('value')
+                if item.get('required') == 'true':
+                    args['required_keys'].append(key)
+                else:
+                    args['optional_keys'].append(key)
+            return TypedDictDoc(**args)
```
Snooz82, line 120: raise

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -93,23 +94,48 @@ def _write_arguments(self, kw, writer):
...
                 writer.element('default', arg.default_repr)
             writer.end('arg')
         writer.end('arguments')
 
     def _get_start_attrs(self, kw_type, kw, lib_source, writer):
         if kw_type == 'init':
-            attrs = {}
+            attrs = {'name': kw.name}
         else:
             attrs = {'name': kw.name}
```
Snooz82, line 109: simplify that

atest/robot/libdoc/LibDocLib.py
```diff
@@ -80,3 +80,13 @@ def get_repr_from_json_arg_model(self, model):
...
+
+    def is_typing_extensions(self):
+        try:
+            from typing_extensions import TypedDict
+            logger.info('Using typing_extensions.')
+            return True
+        except ImportError:
+            logger.info('No typing_extensions.')
+            logger.info(PY_VERSION)
+            return PY_VERSION[0] == 3 and PY_VERSION[1] >= 9
```
pekkaklarck, line 92: This check is on wrong. It only checks does the running side Python have this functionality, not does the actually tested Robot have it.

atest/robot/libdoc/libdoc_resource.robot
```diff
@@ -238,3 +238,37 @@ List of Dict Should Be Equal
...
+        FOR    ${item}    IN    @{items}
+            ${cur}=    Get Element Attributes    ${item}
+            Continue For Loop If    $cur['key'] != $exp['key']
+            Should Be Equal    ${cur}[key]         ${exp}[key]
+            Should Be Equal    ${cur}[type]        ${exp}[type]
+            IF    'required' in $exp
+                Should Be Equal    ${cur}[required]    ${exp}[required]
+            END
+            Log    ${cur} == ${exp}
+            Exit For Loop
```
pekkaklarck, line 272: Why is the loop exited unconditionally after the first iteration? If it's on purpose, the algorithm should be simplified somehow. Not it looks really strange.

atest/robot/libdoc/datatypes_json-xml.robot
```diff
@@ -0,0 +1,49 @@
...
+    ...    {"name": "starts","value": "^="}
+    ...    {"name": "^=","value": "^="}
+    ...    {"name": "should start with","value": "^="}
+    ...    {"name": "ends","value": "$="}
+    ...    {"name": "should end with","value": "$="}
+    ...    {"name": "$=","value": "$="}
+    ...    {"name": "matches","value": "$"}
+    ...    {"name": "validate","value": "validate"}
+    ...    {"name": "then","value": "then"}
+    ...    {"name": "evaluate","value": "then"}
```
pekkaklarck, line 33: Do we need this many members in a test? Understand there should be more than one or two and testing aliases is a good idea, but several dozens feels too much.

src/robot/libdocpkg/model.py
```diff
@@ -65,18 +95,23 @@ def doc_format(self, format):
...
-        return self._add_parent(inits)
+        return self._sort_keywords(inits)
 
     @setter
     def keywords(self, kws):
-        return self._add_parent(kws)
+        return self._sort_keywords(kws)
 
-    def _add_parent(self, kws):
+    def _sort_keywords(self, kws):
```
pekkaklarck, line 104: This helper does other things than sorting. Something more generic like `_process_keywords` would be a better name.

src/robot/libdocpkg/model.py
```diff
@@ -190,3 +228,155 @@ def _arg_to_dict(self, arg):
...
+
+class DataTypeCatalog(object):
+
+    def __init__(self):
+        self._enum_catalog = set()
+        self._typed_dict_catalog = set()
+
+    def __iter__(self):
+        for type_dooc in sorted(self._typed_dict_catalog | self._enum_catalog):
+            yield type_dooc
```
pekkaklarck, line 241: Using just `return iter(sorted(...))` ought to be enough.

src/robot/libdocpkg/model.py
```diff
@@ -190,3 +228,155 @@ def _arg_to_dict(self, arg):
             'required': arg.required,
             'repr': unicode(arg)
         }
+
+
+class DataTypeCatalog(object):
+
+    def __init__(self):
+        self._enum_catalog = set()
+        self._typed_dict_catalog = set()
```
pekkaklarck, line 237: I've started to prefer not prefixing attributes with `_` in cases like this and think `_catalog` postfix is unnecessary as well. I'd thus use just `self.enums` and `self.typed_dicts`, but if you prefer the current variants that's fine as well.

src/robot/libdocpkg/model.py
```diff
@@ -190,3 +228,155 @@ def _arg_to_dict(self, arg):
...
+
+    def __iter__(self):
+        for type_dooc in sorted(self._typed_dict_catalog | self._enum_catalog):
+            yield type_dooc
+
+    def __bool__(self):
+        return bool(self._enum_catalog or self._typed_dict_catalog)
+
+    __nonzero__ = __bool__
```
pekkaklarck, line 246: Could also just implement `__nonzero__` and use `@py2to3` class decorator that you can import from `robot.utils`.

src/robot/libdocpkg/model.py
```diff
@@ -190,3 +228,155 @@ def _arg_to_dict(self, arg):
...
+
+    @property
+    def enum_catalog(self):
+        return sorted(self._enum_catalog)
+
+    @property
+    def typed_dict_catalog(self):
+        return sorted(self._typed_dict_catalog)
```
pekkaklarck, line 254: I consider `_catalog` postfix unnecessary here as well.

src/robot/libdocpkg/model.py
```diff
@@ -190,3 +228,155 @@ def _arg_to_dict(self, arg):
...
+
+    def add(self, typ):
```
pekkaklarck, line 256: Could this be `add(self, *types)` and separate `update` then removed? Would also avoid the problem with `type` as an argument name.

src/robot/libdocpkg/model.py
```diff
@@ -190,3 +228,155 @@ def _arg_to_dict(self, arg):
             'required': arg.required,
             'repr': unicode(arg)
         }
+
+
+class DataTypeCatalog(object):
```
pekkaklarck, line 233: This class and the classes used by it could be moved to a separate module to avoid making this module too big.

src/robot/libdocpkg/model.py
```diff
@@ -190,3 +228,155 @@ def _arg_to_dict(self, arg):
...
+        self.doc = doc
+        self.items = items or []
+
+    @classmethod
+    def from_dict(cls, type_doc):
+        if isinstance(type_doc, dict):
+            return cls(name=type_doc['name'],
+                       type=type_doc['type'],
+                       doc=type_doc['doc'],
+                       items=type_doc['items'])
```
pekkaklarck, line 302: Wouldn't `return cls(**type_doc)` work as well? Or does the dict have other keys as well?

src/robot/libdocpkg/model.py
```diff
@@ -190,3 +228,155 @@ def _arg_to_dict(self, arg):
...
+        if isinstance(enum_type, EnumType):
+            return cls(name=enum_type.__name__,
+                       type='Enum',
+                       doc=enum_type.__doc__ or '',
+                       members=[{'name': name,
+                                 'value': unicode(member.value)}
+                                for name, member in enum_type.__members__.items()])
+        raise TypeError(
+            'EnumDoc.from_Enum() requires Enum types but got %s.'
+            % type_name(enum_type))
```
pekkaklarck, line 370: Are you sure this kind of error handling is needed? How likely is it that someone calls `from_TypedDict` or `from_dict` with an invalid argument? If such error handling really is needed, it should be done like

```python
if not valid:
    raise TypeError(...)
# several lines of code
```

not like
```python
if valid:
    # several lines of code
    return
raise TypeError(...)
```

src/robot/libdocpkg/model.py
```diff
@@ -190,3 +228,155 @@ def _arg_to_dict(self, arg):
...
+                              'type': typ,
+                              'required': key in required_keys
+                              if required_keys or optional_keys else None})
+            return cls(name=typed_dict.__name__,
+                       type='TypedDict',
+                       doc=typed_dict.__doc__ if typed_dict.__doc__ else '',
+                       items=items)
+        raise TypeError(
+            'TypedDictDoc.from_TypedDict() requires a TypedDict but got %s.'
+            % type_name(typed_dict))
```
pekkaklarck, line 325: Are you sure this kind of error handling is needed? How likely is it that someone calls `from_TypedDict` or `from_dict` with an invalid argument? If such error handling really is needed, it should be done like

```python
if not valid:
    raise TypeError(...)
# several lines of code
```

not like
```python
if valid:
    # several lines of code
    return
raise TypeError(...)
```

src/robot/libdocpkg/model.py
pekkaklarck: This is actually a pretty strange alternative constructor for this class because you could just use `TypedDictDoct(**type_doc)` to begin with. Separate `from_TypedDict` makes sense but having that and `from_dict` is rather confusing because both have "dict" in their name but the former handles real types and other apparently operates with already parsed data. I guess the latter, i.e. this method, could be named something like `from_spec` but that still doesn't too useful.

src/robot/libdocpkg/model.py
```diff
@@ -190,3 +228,155 @@ def _arg_to_dict(self, arg):
...
+        self.doc = doc
+        self.members = members or []
+
+    @classmethod
+    def from_dict(cls, type_doc):
+        if isinstance(type_doc, dict):
+            return cls(name=type_doc['name'],
+                       type=type_doc['type'],
+                       doc=type_doc['doc'],
+                       members=type_doc['members'])
```
pekkaklarck, line 354: Same comment here as to the above `from_dict`.

src/robot/libdocpkg/model.py
pekkaklarck, line 370: Same comment about error handling here as to `from_TypedDict` above.

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -107,6 +108,40 @@ def _get_start_attrs(self, kw, lib_source, writer):
         self._add_source_info(attrs, kw, writer.output, lib_source)
         return attrs
 
+    def _write_data_types(self, data_types, writer):
+        writer.start('data_types')
+        if data_types.enum_catalog:
+            writer.start('enums')
+            for enum in data_types.enum_catalog:
+                attrs = {'name': enum.name}
```
pekkaklarck, line 116: Seems to me `attrs` could/should be inlined here and with typed dicts.

atest/robot/libdoc/datatypes_xml-json.robot
```diff
@@ -0,0 +1,71 @@
...
+TypedDict Items
+    [Template]    NONE
+    ${longitude}=    Create Dictionary    key=longitude    type=float    required=${True}
+    ${latitude}=     Create Dictionary    key=latitude     type=float    required=${True}
+    ${accuracy}=     Create Dictionary    key=accuracy     type=float    required=${False}
+    FOR    ${exp}    IN    ${longitude}    ${latitude}    ${accuracy}
+        FOR    ${item}    IN    @{Model}[dataTypes][typedDicts][0][items]
+            Continue For Loop If    $exp['key'] != $item['key']
+            Dictionaries Should Be Equal    ${item}    ${exp}
+            Exit For Loop
```
pekkaklarck, line 48: Unconditionally exiting the loop loops really strange.

atest/testdata/libdoc/DataTypesLibrary.json
```diff
@@ -0,0 +1,326 @@
+{
+  "name": "DataTypesLibrary",
+  "doc": "<p>This Library has Data Types.</p>\n<p>It has some in <code>__init__</code> and others in the <a href=\"#Keywords\" class=\"name\">Keywords</a>.</p>\n<p>The DataTypes are the following that should be linked. <span class=\"name\">HttpCredentials</span> , <a href=\"#GeoLocation\" class=\"name\">GeoLocation</a> , <a href=\"#Small\" class=\"name\">Small</a> and <a href=\"#AssertionOperator\" class=\"name\">AssertionOperator</a>.</p>",
+  "version": "",
+  "generated": "2020-11-26 21:27:03",
+  "type": "LIBRARY",
+  "scope": "TEST",
+  "docFormat": "HTML",
+  "source": "/Source/robotframework/robotframework/atest/testdata/libdoc/DataTypesLibrary.py",
```
pekkaklarck, line 9: Does this work regardless where Robot's source is checked out?

doc/schema/libdoc_schema.json
```diff
@@ -247,38 +257,50 @@
...
-            "type": "string"
-          },
-          "uniqueItems": true
+        "type": {
+          "type": "string"
+        },
+        "required": {
+          "type": [
+            "boolean",
+            "null"
```
pekkaklarck, line 302: I would rather omit `required` altogether than set it `null`, but I guess using `null` has been the approach used with JSON specs in general.

pekkaklarck (changes requested):
Looks really good but please take a look at the separate line comments. Major things to do:

- Fix detecting typing extension usage in tests.

- Add documentation to the User Guide. Can be done after merging this and possibly also aftre beta 1.

- Issue about the enhancement(s) to be listed in the released notes.

atest/robot/libdoc/datatypes_json-xml.robot
Snooz82: that is just an Enum that is used by the only library that uses enums and it has funny enum Members.

So i just took it.
I can delete some if it is a problem.

atest/robot/libdoc/libdoc_resource.robot
Snooz82: it only exits if the conditional continue is not executed.

i can change it to IF

doc/schema/libdoc_schema.json
Snooz82: yes it is omitted in XML and we do not omit anything in json

src/robot/libdocpkg/model.py
Snooz82: The idea was to be kind of compatible with the interface of a set().

atest/testdata/libdoc/DataTypesLibrary.json
Snooz82: Yes because it is only used to convert to JSON again (roundtrip) and to be converted to XML
And the JSON and XML Spec is used to be read by Libdoc and written to JSON and the schema is verified.

i wanted to have every input ones to test that the output is a valid Schema
And only the dataTypes are really used.

atest/robot/libdoc/LibDocLib.py
```diff
@@ -8,7 +8,7 @@
 from xmlschema import XMLSchema
 
 from robot.api import logger
-from robot.utils import CONSOLE_ENCODING, SYSTEM_ENCODING, unicode
+from robot.utils import PY_VERSION, CONSOLE_ENCODING, SYSTEM_ENCODING, unicode
```
pekkaklarck, line 11: Unused import.

src/robot/libdocpkg/model.py
```diff
@@ -13,13 +13,39 @@
...
+    class TypedDictType(object):
+        pass
+
+try:
+    from typing_extensions import TypedDict as ExtTypedDict
+
+    ExtTypedDictType = type(ExtTypedDict('TypedDictDummy', {}))
+except ImportError:
+    class ExtTypedDictType(object):
+        pass
```
pekkaklarck, line 42: Are these imports needed anymore? If they are, couldn't they be imported from the new `datatypesdoc` module?

src/robot/libdocpkg/specbuilder.py
```diff
@@ -90,3 +93,32 @@ def _create_arguments(self, elem):
...
+        items = []
+        for item in dt.findall('items/item'):
+            required = item.get('required', None)
+            if required is not None:
+                required = True if required == 'true' else False
+            items.append({'key': item.get('key'),
+                          'type': item.get('type'),
+                          'required': required})
+        return TypedDictDoc(name=dt.get('name'),
+                            type='TypedDict',
```
pekkaklarck, line 122: Needing to pass `TypedDict` as type explicitly is somewhat strange. Couldn't that be de default with `TypedDictDoct`?

utest/libdoc/test_libdoc.py
```diff
@@ -38,6 +38,14 @@ def run_libdoc_and_validate_json(filename):
     validate(instance=json.loads(json_spec), schema=schema)
 
 
+def is_extended_typing():
+    try:
+        from typing_extensions import TypedDict
+        return True
+    except ImportError:
+        return PY_VERSION[0] == 3 and PY_VERSION[1] >= 9
```
pekkaklarck, line 46: Is this used anywhere?

utest/libdoc/test_libdoc.py
```diff
@@ -221,6 +229,18 @@ def test_TypesViaKeywordDeco(self):
...
 
+        def test_DataTypesLibrary_json(self):
+            run_libdoc_and_validate_json('DataTypesLibrary.json')
+
+        def test_DataTypesLibrary_xml(self):
+            run_libdoc_and_validate_json('DataTypesLibrary.xml')
+
+        def test_DataTypesLibrary_py(self):
+            if PY_VERSION[0] == 3 and PY_VERSION[1] >= 7:
+                run_libdoc_and_validate_json('DataTypesLibrary.py')
```
pekkaklarck, line 240: I'd prefer creating the test only when it's applicable and comparing version as a tuple:

```python
if PY_VERSION > (3, 7):
    def test_DataTypesLibrary_py(self):
        ...
```

atest/requirements-run.txt
```diff
@@ -1 +1,2 @@
 xmlschema
+typing-extensions; python_version == '3.7'
```
pekkaklarck, line 2: This should be in `atest/requirements.txt` or in `utest/requirements.txt` that's read by the former, not here.

pekkaklarck:
Fixed problems mentioned in above comments in 50788fd101dda49ce5c044830e3d3971e71d83b5.

PR #3745, opened 2020-10-29
Make Should Be Title compatible with Convert To Title

### LINKED ISSUE #3651: String: `Should Be Tittle Case` is not consistent with `Convert To Title Case`
`Convert To Title Case` was added in RF 3.2 (#2706). Unfortunately its not fully compatible with the old `Should Be Title Case` keyword. For example, the second test below fails with `'This Is Not OK!' is not titlecase.`:

```robotframework
*** Settings ***
Library        String

*** Test Cases ***
Passes
    ${title} =    Convert To Title Case    This is ok.
    Should Be Title Case    ${title}

Fails
    ${title} =    Convert To Title Case    This is not OK!
    Should Be Title Case    ${title}
```

The reason for the failure is that `Should Be Title Case` simply uses Python's [str.istitle()](https://docs.python.org/3/library/stdtypes.html#str.istitle) function while `Convert To Title Case` uses a lot more sophisticated and configurable algorithm. Most importantly, `Convert To Title Case` leaves words like `XML`, `OK` or `iPhone` as-is and but `str.istitle()` doesn't consider them to be in title case.

The best way to fix this is probably changing `Should Be Title Case` so that it it uses `Convert To Title Case` internally and compares the result with the original value. Because `Convert To Title Case` allows configuration, `Should Be Title Case` probably should too.

mikesmith:
This change makes `Should Be Title` compatible with `Convert To Title`. Fixes #3651.

- Changes `should_be_titlecase` to `should_be_title_case`
- Updates `should_be_title_case` to use `convert_to_title_case` as an internal function to help determine if a string is in title case.
- Adds `exclude` argument to `should_be_title_case`
- Adds additional positive, exclude, and exclude regex tests for `should_be_title_case`. Re-used many of the `convert_to_title_case` tests to ensure compatibility
- Updates documentation for `should_be_title_case`

Please send over any feedback and anything I can do to improve the PR. Thanks!


mikesmith:
Looks like some tests are failing in the `Python2.7` and `pypy2` Acceptance Tests due to the handling of the bytes in these changes.

More specifically, it is getting the error 'TypeError: don't know how to handle UnicodeDecodeError in error callback'. 

Looking into it.

mikesmith:
@pekkaklarck I think this should be ready for review. Please let me know if there is anything that needs to be fixed and updated. Thanks!

pekkaklarck:
Looks awesome! I really like the comprehensive tests and docs.

I'm not sure about the chosen approach to handle bytes, though. Using `str()` causes an exception if there are non-ASCII characters with Python 2 and `string.decode(errors='backslashreplace')` with Python 3 may do undesirable conversion. I actually think it's best we explicitly disallow for these reasons:

- `Convert To Title Case` doesn't handle them.
- Without knowing encoding, you cannot really convert bytes to title case in the first place. For example, `b'\xe4iti'` is `äiti` (mother in Finnish) in Latin-1 and should be converted to `Äiti` i.e. `b'\xc4iti',` but if encoding is something else then title case version would also be something else. Notice that Python ignores non-ASCII bytes and `b'\xe4iti'.title()` yield `b'\xe4Iti'`.
- As the earlier point already demonstrated, Python doesn't handle title casing well (it just can't) with non-ASCII bytes.

The only downside I see with not supporting bytes is that with Python 2 just using `'xxx'` means creating bytes. We convert allour data like `xxx` to Unicode strings, but a library may return `'xxx'` instead of `u'xxx'` and such return value wouldn't anymore work. We can argue that Python 2 is pretty much dead and this kind of change is OK in a major release like RF 4, but we could also pretty easily handle this so that non-ASCII strings are accepted with Python 2. I can do that myself after merging this PR.

mikesmith:
Thanks @pekkaklarck for the feedback and the merge! Also, thanks for sharing your thoughts on handling of bytes. I wasn't 100% sure what the right solution should be but your explanation makes a lot of sense. I learned a couple things that I will keep in mind when dealing with bytes in the future.

Issue #3750, opened 2020-11-01
Improve embedded keyword logging in output.xml


MaciejWiczk:
Embedded keywords are great improvement when comes to testcases readability, but can be cumbersome when debugging logs and processing them.

Having following keyword:
``` 
*** Test Cases ***
Add One To Two
    Add 1 To 2    

Add Two To One
    Add ${a} To ${b}

*** Variables ***
${a}    1
${b}    2

*** Keywords ***
Add ${n} To ${m}
    ${o}    Evaluate    ${n}+${m}
```
will be processed in both log.html and output.xml as either `Add 1 To 2` or `Add ${a} To ${b}` while it's actual definition, `Add ${n} To ${m}` isn't stored in logs. It would be nice improvement, to allow to store keywords definition in addition to current keyword properties. 
Such a change would allow for post processing output.xml for statistical and analytical purposes, which I'm facing [in rfhub2](https://github.com/pbylicki/rfhub2/issues/291)

I found few places, which would need to be changed:
- https://github.com/robotframework/robotframework/blob/master/src/robot/running/librarykeywordrunner.py
- https://github.com/robotframework/robotframework/blob/master/src/robot/running/userkeyword.py
but I believe, it's a tip of an iceberg. Any help and advice would be much appriciated. 
![image](https://user-images.githubusercontent.com/35737166/97808803-403fd100-1c69-11eb-9a17-9eb23b28d140.png)

pekkaklarck:
Do I get it right that you'd like to store the original name of the keyword somewhere? Would it need to be in logs or would having it only in output.xml be enough? I guess it could be added as a new attribute to the ´<kw>` element.

A problem with adding info like this to output.xml is that its size will be increased. If this is done only with keywords using embedded arguments, that shouldn't be too big a problem, though.

MaciejWiczk:
Hi @pekkaklarck, You are exactly right. This argument could be only logged when keyword is embedded keyword, otherwise keyword definition, or however this attribute is called, would be the same as name/longname and skipped. Storing it in output.xml is enough for me, but maybe, using DEBUG/TRACE level, and storing in log.html, would be useful for others.

Last, but not least, did I get to right places, where changes should be applied first?

pekkaklarck:
I think this is a good idea and assigned this issue tentatively to RF 4.0 scope. I also commented the PR with changes that should still be done.

pekkaklarck:
[...]

PR #3753, opened 2020-11-03
Storing embedded keyword sourcename in output.xml as sourcename property

MaciejWiczk:
PR for change #3750


MaciejWiczk:
Here are proposed changed, that will implement additional property logged in output.xml for embedded keywords only. 
Current output.xml will look like this:
```
<kw name="Add ${a} To ${b}">
<msg timestamp="20201104 11:01:46.362" level="TRACE">Arguments: [ ${n}='1' | ${m}='2' ]</msg>
```
And new output.xml will look like this:
```
<kw name="Add ${a} To ${b}" definition="Add ${n} To ${m}">
<msg timestamp="20201104 11:03:52.179" level="TRACE">Arguments: [ ${n}='1' | ${m}='2' ]</msg>
```
Currently one acceptance test fails, but it's also failing in master branch, and I do not see a way to fix it, and a change that introduced it.

MaciejWiczk:
@pekkaklarck, could I ask for review? I know that there is a need to change xml schema, and I can add additional changes in #3727 since they are connected or any other recommended way.

pekkaklarck:
Sorry for late review due to having other bigger tasks like IF/ELSE under development. Quick comments now:

- Adding the information about the original keyword name to output.xml sounds good to me in general.
- I'm not sure is `definition` the best name for it but I cannot immediately come up with anything much better. 
- I think this information should be added only to keywords in the result model i.e. `robot.result.model`, not to "base model" in `robot.model` where it would be got also by the execution side keywords. Could you test would that work as well?
- You are right that the schema needs to be updated as well but that's a small task.

pekkaklarck:
We can continue discussion about this on the #devel channel on our Slack as well.

MaciejWiczk:
@pekkaklarck, no worries, IF/ELSE is way hotter topic, I cannot wait for as well :)
- great! It will be only working for embedded keywords, both user defined or from libraries
- I believe I added it everywhere, because the information about `definition` was only available there, but I will check again
- nam is also not exactly perfect for me, bo other like `base_name` `orig_name` are even worse ;)

src/robot/running/model.py
```diff
@@ -53,14 +53,15 @@ class Keyword(model.Keyword):
...
     message_class = None  #: Internal usage only.
 
     def __init__(self, name='', doc='', args=(), assign=(), tags=(), timeout=None,
-                 type=model.Keyword.KEYWORD_TYPE, lineno=None, parent=None):
+                 type=model.Keyword.KEYWORD_TYPE, lineno=None, parent=None, sourcename=''):
         model.Keyword.__init__(self, name, doc, args, assign, tags, timeout, type,
-                               parent)
+                               parent, sourcename)
         self.lineno = lineno
+        self.sourcename = sourcename
```
pekkaklarck, line 64: `sourcename` shouldn't be needed here at all.

src/robot/running/userkeywordrunner.py
```diff
@@ -34,6 +34,7 @@ class UserKeywordRunner(object):
     def __init__(self, handler, name=None):
         self._handler = handler
         self.name = name or handler.name
+        self.sourcename = handler.name
```
pekkaklarck, line 37: There shouldn't be need to assign `self.sourcename` when we can access it via `self._handler`. Exactly like with library keywords, it ought to be possible to enhance only the `EmbeddedArgumentsRunner`.

src/robot/variables/tablesetter.py
```diff
@@ -71,7 +71,7 @@ def resolve(self, variables):
     @contextmanager
     def _avoid_recursion(self):
         if self._resolving:
-            raise DataError('Recursive variable definition.')
+            raise DataError('Recursive variable sourcename.')
```
pekkaklarck, line 74: This looks like an accidental renaming.

pekkaklarck (changes requested):
Good start but changes are still needed:

- `sourcename` should only be added to result model and only be set when actually using embedded arguments.

- Tests must be added.

- The new output.xml attribute should be explained in `doc/schema/README.rst`.

src/robot/output/xmllogger.py
```diff
@@ -66,6 +66,8 @@ def start_keyword(self, kw):
         attrs = {'name': kw.kwname, 'library': kw.libname}
         if kw.type != 'kw':
             attrs['type'] = kw.type
+        if kw.sourcename not in ('', kw.kwname):
+            attrs['sourcename'] = kw.sourcename
```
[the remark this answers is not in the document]
MaciejWiczk: Applying the above condition would always log sourcename to output.xml, no matter if keyword is embedded or not. Perhaps I should move condition to set sourcename only if it differs form kwname/name?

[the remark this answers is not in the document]
pekkaklarck: I think setting `sourcename` only when using embedded args is a good idea. If we only save it to output.xml in that case, we wouldn't have it with other keywords when post-processing results anyway. Looking at the code it seems this has already been done, so the condition above can also be changed to `if kw.sourcename`.

doc/schema/README.rst
```diff
@@ -90,5 +90,6 @@ kw - keyword element, name is given as an attribute. Type attribute describes th
     * ``msg`` - any number of optional keyword messages
     * ``timeout`` - optional keyword timeout. Before 3.0 this was an attribute.
     * ``status`` - keyword has to have a status
+    * ``sourcename`` - name of keyword with embedded arguments, from it's definition
```
pekkaklarck, line 93: "it's" -> "its"

Could also be changed to something like

> original name of the keyword when using embedded arguments, not set otherwise. New in 4.0.

altogether.

doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
```diff
@@ -1667,6 +1667,10 @@ __ `Setting custom name`_
 __ `Embedding arguments into keyword names`_
 __ `Specifying argument types using @keyword decorator`_
 
+.. note:: Storing original keyword name, as ``sourcename`` property
+          when it contains embedded arguments is a new feature in Robot Framework 4.0.
+          It allows for easier matching keywords definition with data from output.xml files.
+
```
pekkaklarck, line 1673: - I don't think this should be a note because this topic hasn't been discussed otherwise. Could be just a regular paragraph explaining that the original name of the keyword is stored as `sourcename` in the output file starting from RF 4.0. Or at least the note should be worded differently like "To ease finding keyword usages from output files, the original name of the keyword is stored ....".
- This should be explained in the section discussing embedded arguments, not in the general `@keyword` decorator section. This applies also to user keywords so perhaps this should be mentioned in the section discussing using embedded args, not here in the creating libraries section.
- This is documented in the schema so I'm not 100% sure is a separate doc needed here in the UG at all. A short paragraph/note certainly wouldn't hurt.

pekkaklarck (approved):
1. Code changes look very good but one "if" condition could still be simplified. See line comments for details.
2. Great tests!
3. Documentation needs some tuning still. See line comments for details.

I hope you @MaciejWiczk could still take a look at the smallish issues mentioned in the line comments. If you can get them sorted out I can merge this possibly before beta 2 that's planned for Wednesday. If not, I can merge the PR in the current form and fix these myself after the beta.

doc/schema/README.rst
MaciejWiczk: Done

MaciejWiczk:
> 1. Code changes look very good but one "if" condition could still be simplified. See line comments for details.

I don't know why I didn;t push this change, fixed now, working like a charm :)
> 2. Great tests!

Thanks! Seems like those years as a tester were not wasted :D
> 3. Documentation needs some tuning still. See line comments for details.

Yeah, I see that they are far from perfect. It is not so simple to follow the right path. Sorry bout that. 
> I hope you @MaciejWiczk could still take a look at the smallish issues mentioned in the line comments. If you can get them sorted out I can merge this possibly before beta 2 that's planned for Wednesday. If not, I can merge the PR in the current form and fix these myself after the beta.

I hope that I fixed everything that was needed in this PR, and it could make it to the beta2 :)
Thanks for great review :)

pekkaklarck:
Looks good! There was a conflict due to recent changes to result model but I fixed that.

Issue #3079, opened 2019-02-04
Support for nested control structures


pekkaklarck:
Robot Framework currently only has FOR loops but the plan is to add support for IF/ELSE (#3074) and TRY/EXCEPT (#3075) in the future as well. At the moment FOR loops cannot be nested directly, so nested loops need to be implemented so that a loop contains a keyword which has another loop. This isn't too convenient, and the problem becomes even worse when new syntax is added.

We have made a decision to add support for nested structures so that examples like this will work:

```
FOR    ${row}    IN    @{rows}
    FOR    ${cell}    IN    @{row}
        IF    "${cell}" != "IGNORE"
            Process Cell    ${cell}
        END
    END
END
```

The plan is to add support for nested FOR loops already in RF 3.2 as a feature provided by the new test data parser (#3076). Possible problem is that in RF 3.2 we still need to support FOR loops without the explicit END marker (#3078). If that adds too much complexity, the support for nested control structures can be postponed to RF 3.3 where FOR loops without END won't be supported anymore.

pekkaklarck:
No time for this in RF 3.2. We want to the new parser out ASAP so that it supports old syntax, and then we can start concentrating on enhancing syntax with IF/ELSE, TRY/EXCEPT, and nested structures. This is definitely a critical feature in RF 3.2.

Snooz82:
readability of nested control features is always a pain.
And statics code analysis count every nesting as a huge impact to code complexity.

With Keywords nesting is possible and readability is good. 

If you would ask me, i would downvote this feature.
If you need this stuff, do it in python.
If you do not understand python, you do not understand nested structures...

pekkaklarck:
In my opinion IF/ELSE makes no sense if we don't support nested structures. Especially FOR loop with IF sounds like a pretty common use case. It would be annoying to be able to use IF/ELSE elsewhere but in a loop you needed to revert to the hackish Run Keyword If.

PR #3757, opened 2020-11-04
Topic/nested for

mkorpela:
Extracted nested for loop from topic/if.
This also contains the parsing side refactoring for fileparser.py and blockparser.py.


pekkaklarck:
I'm not a big fan of factory methods and believe using concrete classes as parsers would be better. Anyway, I'll merge this now and see can the methods be turned to classes later. This new functionality also needs documentation before we can consider #3079 ready.

PR #3765, opened 2020-11-17
Fix environment variable setting

Hi-Fi:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions\#environment-files


mkorpela:
Thanks @Hi-Fi !

PR #3768, opened 2020-11-18
rename JSON keys to camelCase #3730

Snooz82:


pekkaklarck:
Looks good to me. Two questions:
- Was `default` supposed to be changed to `defaultValue` as well?
- Could we actually use `tags` and not `allTags`?

Snooz82:
@pekkaklarck 
That default actually causes a lot of issues in the atests because then they are different in names between XML and JSON and i just decided that the effort is too high.

i can change tags.
and fix the 3.5 test that has weird orders

pekkaklarck:
So apparently `default` doesn't cause too much problems in JS side? I don't care strongly should we use `tags` or `allTags`. Go with the one you think works better. I just realized XML specs don't have tags listed like that at all. Shouldn't they be added there as well? We should have same content in XML and JSON specs. That's a topic for a separate commit/PR and probably deserves its own issue as well.

Issue #3748, opened 2020-10-30
Libdoc: Support argument types with multiple possible values


pekkaklarck:
RF 4.0 alpha 2 added support for argument conversion based on multiple types (#3738) and Libodc should support that as well both in specs and in HTML. In specs it's probably best to store info for each possible type separately (i.e. instead of having just `<type>int</type>` make it valid to have `<type>int</type><type>float</type>`) and in HTML it might be a good idea to use something like `<int> or <float>` as the type info.

When getting argument info from Robot, Libdoc should support both unions and tuples (i.e. `Union[int, float]` and `(int, float)`) like the execution side does.. We need to decide should we normalize this information already at the execution side or should Libdoc just handle both cases. When reading specs and recreating models, Libdoc can read multiple types into a tuple.

mkorpela:
Union support has been added.

Issue #3724, opened 2020-10-19
Libdoc: Drop `typing.` prefix from type hints originating from the `typing` module


Snooz82:
Type hints from typing looks a bit uncomfortable ...

```python
from typing import Optional, Union, Any

def keyword(whatever: Any, stringy: str = 'some', maybe: Optional[str] = None, num: Union[int, str] = 2, )
    pass
```

```
whatever: typing.Any,
stringy: str = some,
maybe: typing.Optional[str] = None,
num: typing.Union[int, str] = 2
```

That should be something like that:
```
whatever: Any,
stringy = some,
maybe: Optional[str] = None,
num: Union[int, str] = 2
```

pekkaklarck:
#3748 proposes storing information about Unions separately in specs and showing that in HTML like `<int> or <float>`. That would solve this problem with Union and because Optional is actually a shortcut for a Union with None also with that.

An easy solution for Any is not showing that information at all. There's only a semantic difference between `def kw(arg: Any)` and `def kw(arg)` anyway and it may be best not to bother users with that.

MaciejWiczk:
@Snooz82, don't You think that `stringy: str = some` have better meaning than `stringy = some` ? I can imagine a situation where this argument would have value None by default, and would be hard o catch what kind of variable it should be.

pekkaklarck:
This has been already implemented as part of PR #3771 but needs a test still.

PR #3771, opened 2020-11-19
Added Union support to libdoc (#3748)

mkorpela:
Great work from @Snooz82 !!!!! 🏆

### DIFF
--- a/doc/schema/libdoc_schema.json
+++ b/doc/schema/libdoc_schema.json
@@ -73,6 +73,16 @@
         "$ref": "#/definitions/keyword"
       },
       "uniqueItems": true
+    },
+    "dataTypes": {
+      "type": "array",
+      "items": {
+        "oneOf": [
+          {"$ref": "#/definitions/typedDictType"},
+          {"$ref": "#/definitions/enumType"}
+        ]
+      },
+      "uniqueItems": true
     }
   },
   "required": [
@@ -143,7 +153,7 @@
       "type": "object",
       "required": [
         "name",
-        "type",
+        "types",
         "default",
         "kind",
         "required",
@@ -153,11 +163,11 @@
         "name": {
           "type": "string"
         },
-        "type": {
-          "type": [
-            "string",
-            "null"
-          ]
+        "types": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
         },
         "default": {
           "type": [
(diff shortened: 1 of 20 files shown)


Snooz82:
Resolves #3748

src/robot/libdocpkg/specbuilder.py
```diff
@@ -85,9 +85,9 @@ def _create_arguments(self, elem):
             default_elem = arg.find('default')
             if default_elem is not None:
                 spec.defaults[name] = default_elem.text or ''
-            type_elem = arg.find('type')
-            if type_elem is not None:
+            type_elems = arg.findall('type')
+            if type_elems is not None:
```
pekkaklarck, line 89: `type_elems` cannot be `None` anymore. It's always a list and may be empty.

The JSON side had exactly the same problem and the fix is the same on both side as well.

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -93,20 +93,18 @@ def _write_arguments(self, kw, writer):
                                  'repr': unicode(arg)})
             if arg.name:
                 writer.element('name', arg.name)
-            if arg.type is not arg.NOTSET:
-                writer.element('type', arg.type_repr)
+            if arg.types:
+                for type_repr in arg.types_reprs:
+                    writer.element('type', type_repr)
```
pekkaklarck, line 98: Isn't `if arg.types` unnecessary? If the arg doesn't have types, `arg.types_reprs` yields nothing and the loop isn't executed.

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -93,20 +93,18 @@ def _write_arguments(self, kw, writer):
...
     def _get_start_attrs(self, kw_type, kw, lib_source, writer):
-        if kw_type == 'init':
-            attrs = {}
-        else:
-            attrs = {'name': kw.name}
-            if kw.deprecated:
-                attrs['deprecated'] = 'true'
+        attrs = {'name': kw.name}
+        if kw_type != 'init' and kw.deprecated:
+            attrs['deprecated'] = 'true'
```
pekkaklarck, line 107: I don't think there's need to check `kw_type` when deciding should `deprecated` attribute be added. Not doing that might remove the need to pass that info to this method altogether.

src/robot/running/arguments/argumentspec.py
```diff
@@ -137,14 +164,15 @@ def required(self):
...
-        return unicode(self.type)
+    def types_reprs(self):
+        return [self._type_repr(t) for t in self.types]
+
+    def _type_repr(self, typ):
+        if typ == type(None):
+            return 'None'
+        if isclass(typ):
+            return typ.__name__
+        return re.sub(r'^typing\.(.+)', lambda m: m.group(1), unic(typ))
```
pekkaklarck, line 175: Dropping the `typing.` prefix isn't related to rest of this PR but it's so simple that having it here isn't a big problem. It would need some tests, but I guess it's fine to do that as part of #3724 as well.

I'd in general rather use `if type_repr.startswith('typing.'): return type_repr[7:]` than `re.sub`, but in this particular case that would be somewhat more code. Please replace the lambda with `r'\1'` anyway.

doc/schema/libdoc_schema.json
```diff
@@ -184,6 +194,93 @@
...
+          "uniqueItems": true
+        },
+        "optional_keys": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "uniqueItems": true
+        }
+      }
```
pekkaklarck, line 283: Seems to be related to data types and not to this PR. That enhancement is coming soon so no biggie.

pekkaklarck (approved):
Looks very good in general. Please take a look at my comments and when fixing problems you consider worth fixing go ahead and merge. At least conditions like `if types is None` where `types` never can be `None´ should be either fixed or removed.

PR #3774, opened 2020-11-23
Topic/if user guide

mkorpela:


doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -739,16 +739,52 @@ to make the syntax easier to read.
...
 
 In general, it is not recommended to have conditional logic in test
 cases, or even in user keywords, because it can make them hard to
 understand and maintain. Instead, this kind of logic should be in test
 libraries, where it can be implemented using natural programming
-language constructs. However, some conditional logic can be useful at
-times, and even though Robot Framework does not have an actual if/else
-construct, there are several ways to get the same effect.
+language constructs. However, conditional logic can be useful at
+times and from Robot Framework version 4.0 onwards IF block is included
```
pekkaklarck, line 752: "IF block" doesn't sound that good to me but I'm not sure what would be better. Perhaps "IF syntax"?

doc/userguide/src/CreatingTestData/AdvancedFeatures.rst
```diff
@@ -739,16 +739,52 @@ to make the syntax easier to read.
...
+in the syntax.
+
+Conditional execution can be done with IF block.
```
pekkaklarck, line 755: This seem unnecessary considering what the paragraph above says.

doc/userguide/src/CreatingTestData/CreatingTestCases.rst
```diff
@@ -904,6 +904,25 @@ all the looped elements even if there are failures.
            1st arg    ${index}
        END
 
+Templates with IF block
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+`if block`_ can be also used together with templates.
```
pekkaklarck, line 910: I'm still not certain is "block" good or should it be e.g. "syntax". What's certain is that "if" is wrong case in the beginning of a sentence. I think "IF" would be good because it demonstrates the actual syntax as well ,l but normal "If" could be fine as well. Should check has "FOR" or "for" been used with for loops.

pekkaklarck (changes requested):
Looks good. Few things to do:

- Consider how to call this new feature. I don't particularly like "IF block" but I'm not sure is "IF expression", "IF syntax", or what best. We can go with the block if we don't agree on something better.

- Use case consistently. Some places now use "IF" while others use "if". Need to check how for loops are handled in this regard.

- Explain the expression syntax. Users should know that it's evaluated in Python so, for example, strings need to be quoted. The syntax also has all kind of interesting features like using "real" variable values using `$var` syntax and automatically importing modules that need to be explained. I would assume the "Evaluating expressions" section from the BuiltIn library docs can be used as the base. You can also look the "Inline Python evaluation" section under the Variables section in the User Guide itself for inspiration.

PR #3775, opened 2020-11-23
Fixed problem with variable name like ${var_name} for `Variable Should Exist` and `Variable Should Not Exist`

miktuy:
If you pass variable name like `${var_name}` to msg in keywords `Variable Should Exist` and `Variable Should Not Exist` it will raise error like `Variable '${var_name}' not found.

```
*** Test Cases ***
Example
    Variable should not exist   ${var}   msg=Var ${var} is exist!!!
```
Before fix
image
```
*** Variables ***
${var}=  42
*** Test Cases ***
Example
    Variable should not exist   ${var}   msg=Var ${var} is exist!!!
```
image

___
After fix
image

```
*** Variables ***
${var}=  42
*** Test Cases ***
Example
    Variable should not exist   ${var}   msg=Var ${var} is exist!!!
```
image

@pekkaklarck Can you explain why ```self._variables.replace_string(msg)``` was used? Mb my problem is not serious than problem which was solved by this call.

### DIFF
--- a/atest/testdata/standard_libraries/builtin/variable_should_exist.robot
+++ b/atest/testdata/standard_libraries/builtin/variable_should_exist.robot
@@ -71,7 +71,7 @@ Variable Should Not Exist With Given Error Message
     Variable Should Not Exist    ${scalar}    This is the error message
 
 Variable Should Not Exist With Error Message Containing Variables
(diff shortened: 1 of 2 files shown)


miktuy:
Oh! I have seen atests for this keywords) I can use slash in msg to escape `$` symbol...

PR #3796, opened 2020-12-10
Listener v2: source DRAFT and tests

mkorpela:


mkorpela:
For some reason also keyword inside (else if or else at least) branch is not getting source.

pekkaklarck:
This functionality has been implemented elsewhere.

Issue #3362, opened 2019-11-01
Enhance performance of getting information about keywords with big remote libraries


MyMindWorld:
Hello! 
We have really big remote libraries (about 50.000  kw) and we stuck on performance issues

While researching it i found that when client asks for kw names, server sends them all at once, but documentation, arguments e.t.c. sending by one. So we have around 200-300k of requests and to run single case we need to wait around 3 minutes to actually start.

Seems like ok, but in real life when we need to debug smth or run pabot(which simply run allocated instance of robot) 3 minutes becomes 5 hours for 100 suites and our parallel setup becomes useless also

Maybe its possible to make request for all at once and send them also like that?

For now we simply removed this functions from XmlRpcRemoteClient class in Remote.py and it starting now in less than 1 sec)

We also researched dynamic libraries API, but it seems like its not what we need)

pekkaklarck:
Sending all information in one go sounds like a good idea, and is definitely possible, but would need someone to implement it. I can help with the design but don't have time for implementation in the near future. Are you interested? Once this feature would be added to the Remote API, it then needed to be added to various remote servers before actually helping.

Alternatively, or additionally, we could add a feature to the Remote library to disable querying other keyword information than names and possibly arguments. It would probably be easier to implement and wouldn't require changes to remote servers. Depending on the context, of some of this information may be needed during execution (especially argument names and types), so this wouldn't be as good solution than getting all information in one request.

MyMindWorld:
Yeah, im definately interested in this!

pekkaklarck:
We can easily pass arguments to `Remote` itself when it's imported (it currently gets the optional URI), but there's no way we can pass such initialization information to remote servers. I'm not sure why we needed to pass any such information to implement this functionality, though. This is how I think this should be implemented:

1. Add a new method to the remote API (and also the the dynamic API used by it) to get all information in one go. Two possibilities here:

    - Implement a single `get_library_information` that returns *all* information of the library (doc, version, ...) and *all* keywords (names, docs, args, ...). Basically a big dictionary in format like 

      ```
      {'name': 'ExampleLibrary', 'doc': 'Example doc', 'version': '0.0.1a1',
       'keywords': [{'name': 'Kw 1', 'doc': 'xxx', 'args': ['arg1', 'arg2=default']}, 
                    {'name': 'Kw 2', 'args': ['**kws'], 'tags': ['example']}]}
      ```

[...]

Issue #73, opened 2014-06-29
It should be possible to generate links to keywords inside keyword documentation using 'libdoc.py'


spooning:
> _Originally submitted to [Google Code](http://code.google.com/p/robotframework/issues/detail?id=73) by @yanne on 28 Aug 2008_

It would nice to able to link inside keyword documentation so that libdoc
would generate real links from some suitable syntax.

PR #3802, opened 2020-12-17
Implement get_library_information interface for Remote libraries (Rob…

JFoederer:
In preparation of resolving issue #3362 about load-time performance of large libraries, this pull request offers a forward and backward compatible interface extension as suggested in the issue.

With this change the Remote library will utilize the new `get_library_information` call to retrieve all information in a single go. This significantly reduces load times of large libraries, from minutes to a couple of seconds. When the new call is unavailable or fails, the regular interface is used to provide both backward compatibility and maintain analyzability in case of failure.

This pull request includes the change to Robot framework code (1 file), a new acceptance test (3 files) and a fix for the remote server (1 file) in the testdata. The latter could return None while None marshalling is disabled by default. There is no impact on documentation.

There appears to be some technical debt on the [remote server](https://github.com/robotframework/PythonRemoteServer) side. Pull request [#73](https://github.com/robotframework/PythonRemoteServer/pull/73) has been opened to first fix the current baseline of tests before pushing any updates to that side of the interface. Input requested.


JFoederer:
My initial statement about there being no documentation impact was false. That was only true for the user part. Documention has now been extended in the remote protocol section.

atest/testdata/standard_libraries/remote/libraryinfo.py
```diff
@@ -0,0 +1,38 @@
...
+        Individual get_keyword_* methods are not registered.
+        This removes the fall back scenario should get_library_information fail.
+        """
+        self.register_function(self.get_library_information)
+        self.register_function(self.run_keyword)
+
+    def get_library_information(self):
+        info_dict = dict()
+        for kw in self.get_keyword_names():
+            kw_dict = dict()
```
bhirsz, line 19: what about just:

```
info_dict[kw] = {
    'args': self.get_keyword_arguments(kw),
    'tags': self.get_keyword_tags(kw)
    'doc': self.get_keyword_documentation(kw)
}
```

atest/testdata/standard_libraries/remote/libraryinfo.py
```diff
@@ -0,0 +1,38 @@
...
+            kw_dict['args'] = self.get_keyword_arguments(kw)
+            kw_dict['tags'] = self.get_keyword_tags(kw)
+            kw_dict['doc'] = self.get_keyword_documentation(kw)
+            info_dict[kw] = kw_dict
+        return info_dict
+
+class The10001KeywordsLibrary(object):
```
bhirsz, line 26: It's the test library? Shouldn't it be stored in test data somehow?

src/robot/libraries/Remote.py
```diff
@@ -227,23 +228,49 @@ def _server(self):
...
+        returns False or a dict with structure:
+        { "keyword_name" : { 'args':[], 'tags':[], 'doc':"", 'types':[]},
+          "kw2"          : {...}, etc. }
+        """
+        if self.__kw_cache is None:
+            self.__kw_cache = False
+            try:
+                with self._server as server:
+                    self.__kw_cache = server.get_library_information()
+            except:
```
bhirsz, line 243: Too broad exception, what are usual expected exceptions here? (in case of not supported version etc)

src/robot/libraries/Remote.py
```diff
@@ -227,23 +228,49 @@ def _server(self):
...
+                pass # Best efort failed, fall back to regular loading
+        return self.__kw_cache
+
     def get_keyword_names(self):
+        if self._kw_cache:
```
bhirsz, line 248: Not sure what is possible with the Python version we need to support but I would try to implement decorator similiar to lru_cache instead, with optional argument. Ie:

@keyword_cache()

@keyword_cache(type='tags')
..

atest/testdata/standard_libraries/remote/libraryinfo.py
JFoederer: Hi @bhirsz and thanks for reviewing!

Good suggestion, I might even prefer:
```
info_dict[kw] = dict(args=self.get_keyword_arguments(kw),
                     tags=self.get_keyword_tags(kw),
                     doc=self.get_keyword_documentation(kw))
```
I have similar code in the [Python remote server implementation](https://github.com/robotframework/PythonRemoteServer/pull/74). I will update there as well.

src/robot/libraries/Remote.py
JFoederer: I figured there were 2 possible ways to go here.
1. Check if get_library_information is implemented. If so, use it and let it fail whenever it does.
2. Just try to call it and fall back to the _old way_ if it doesn't work, for whatever reason.

I went for the latter , because I thought it stayed closest to the existing behaviour. Not a fan of broad exception handling myself, so we could still opt for the first option.

src/robot/libraries/Remote.py
JFoederer: I have now also processed this comment.

src/robot/libraries/Remote.py
JFoederer: Introspection isn't available, so checking if it is implemented can't be done. But as it turns out the exceptions were already narrowed down on another level, that also contained a retry. Now the try on get_library_info replaces the first attempt and only the specific error is caught.

pekkaklarck:
As I just commented issue #3362, it would be great to get this into RF 4.0. I looked at this PR now and it seems ok so that ought to be doable.

The code is somewhat complex with all decorators and caching and I'm not sure is that needed. Wouldn't it be enough to call `get_library_information` in `get_keyword_names`? We know Robot calls `get_keyword_names` first and also that it is called exactly once. I'll checkout this PR locally and will test that approach.

pekkaklarck:
If you @JFoederer are around, we can discuss this further on the #devel channel on our Slack. Already started a thread about this enhancement there.

pekkaklarck:
Thanks @JFoederer for updating the implementation to call `get_library_information` directly in `get_keyword_names`. Lot simpler that way.

I think we can simplify this a bit more by replacing the decorator with a helper method. In addition to that, probably need to enhance the documentation a bit. At least a note about this being new in RF 4.0 is needed. I'll do these changes myself after merging this PR.

PR #3804, opened 2020-12-18
Fix typo `datatime` -> `datetime`

dandersson:
Also "to some extend" -> "to some extent" near one of the other fixes.


pekkaklarck:
Thanks! I've made that typo _so_ many times. 😄

PR #3808, opened 2020-12-22
Add class method from_params for creating statements

bhirsz:
Simplify creating statements - we don't need to pass token list and we can use param values instead. 

To be discussed - what statements should have from_params? All of them? Or just particular.

### DIFF
--- a/src/robot/parsing/lexer/tokens.py
+++ b/src/robot/parsing/lexer/tokens.py
@@ -143,7 +143,7 @@ class Token(object):
             value = {
                 Token.IF: 'IF', Token.ELSE_IF: 'ELSE IF', Token.ELSE: 'ELSE',
                 Token.FOR: 'FOR', Token.END: 'END', Token.CONTINUATION: '...',
-                Token.EOL: '\n'
+                Token.EOL: '\n', Token.WITH_NAME: 'WITH NAME'
             }.get(type, '')
         self.value = value
         self.lineno = lineno
--- a/src/robot/parsing/lexer/tokens.py
+++ b/src/robot/parsing/lexer/tokens.py
@@ -35,9 +35,9 @@ class Token(object):
 
     If :attr:`value` is not given when :class:`Token` is initialized and
     :attr:`type` is :attr:`IF`, :attr:`ELSE_IF`, :attr:`ELSE`, :attr:`FOR`,
-    :attr:`END` or :attr:`CONTINUATION`, the value is automatically set
-    to the correct marker value like ``'IF'`` or ``'ELSE IF'``. If :attr:`type`
-    is :attr:`EOL` in this case, the value is set to ``'\\n'``.
+    :attr:`END`, :attr:`WITH_NAME` or :attr:`CONTINUATION`, the value is
+    automatically set to the correct marker value like ``'IF'`` or ``'ELSE IF'``.
+    If :attr:`type` is :attr:`EOL` in this case, the value is set to ``'\\n'``.
     """
 
     SETTING_HEADER = 'SETTING HEADER'
--- a/src/robot/parsing/model/statements.py
+++ b/src/robot/parsing/model/statements.py
@@ -337,6 +337,13 @@ class Variable(Statement):
 class TestCaseName(Statement):
     type = Token.TESTCASE_NAME
 
+    @classmethod
+    def from_params(cls, name, eol='\n'):
+        return TestCaseName([
+            Token(Token.TESTCASE_NAME, name),
+            Token(Token.EOL, eol)
+        ])
+
     @property
     def name(self):
         return self.get_value(Token.TESTCASE_NAME)
@@ -450,6 +457,16 @@ class ForHeader(Statement):
 class IfHeader(Statement):
     type = Token.IF
 
+    @classmethod
+    def from_params(cls, condition, indent='    ', separator='    ', eol='\n'):
+        return IfHeader([
+            Token(Token.SEPARATOR, indent),
+            Token(Token.IF),
+            Token(Token.SEPARATOR, separator),
+            Token(Token.ARGUMENT, condition),
+            Token(Token.EOL, eol)
+        ])
+
     @property
     def condition(self):
         return self.get_value(Token.ARGUMENT)
@@ -466,10 +483,28 @@ class IfHeader(Statement):
 class ElseIfHeader(IfHeader):
     type = Token.ELSE_IF
 
+    @classmethod
+    def from_params(cls, condition, indent='    ', separator='    ', eol='\n'):
+        return ElseIfHeader([
+            Token(Token.SEPARATOR, indent),
+            Token(Token.ELSE_IF),
+            Token(Token.SEPARATOR, separator),
+            Token(Token.ARGUMENT, condition),
+            Token(Token.EOL, eol)
+        ])
+
 
 @Statement.register
 class ElseHeader(Statement):
     type = Token.ELSE
+
+    @classmethod
+    def from_params(cls, indent='    ', eol='\n'):
+        return ElseHeader([
+            Token(Token.SEPARATOR, indent),
+            Token(Token.ELSE),
+            Token(Token.EOL, eol)
+        ])
 
     @property
     def condition(self):
@@ -484,6 +519,14 @@ class ElseHeader(Statement):
 class End(Statement):
     type = Token.END
 
+    @classmethod
+    def from_params(cls, indent='    ', eol='\n'):
+        return End([
+            Token(Token.SEPARATOR, indent),
+            Token(Token.END),
+            Token(Token.EOL, eol)
+        ])
+
     def validate(self):
         if self.get_tokens(Token.ARGUMENT):
             self.errors += ('END does not accept arguments.',)
@@ -492,6 +535,14 @@ class End(Statement):
 @Statement.register
 class Comment(Statement):
     type = Token.COMMENT
+
+    @classmethod
+    def from_params(cls, comment, indent='    ', eol='\n'):
+        return Comment([
+            Token(Token.SEPARATOR, indent),
+            Token(Token.COMMENT, comment),
+            Token(Token.EOL, eol)
+        ])
 
 
 @Statement.register
@@ -514,5 +565,5 @@ class EmptyLine(Statement):
     type = Token.EOL
 
     @classmethod
-    def from_value(cls, value):
+    def from_params(cls, value):
         return EmptyLine([Token(Token.EOL, value)])
--- a/src/robot/parsing/model/statements.py
+++ b/src/robot/parsing/model/statements.py
@@ -395,9 +395,9 @@ class ForceTags(MultiValue):
     type = Token.FORCE_TAGS
 
     @classmethod
-    def from_params(cls, tags, separator=FOUR_SPACES, eol=EOL):
+    def from_params(cls, values, separator=FOUR_SPACES, eol=EOL):
         tokens = [Token(Token.FORCE_TAGS, 'Force Tags')]
-        for tag in tags:
+        for tag in values:
             tokens.append(Token(Token.SEPARATOR, separator))
             tokens.append(Token(Token.ARGUMENT, tag))
         tokens.append(Token(Token.EOL, eol))
@@ -409,9 +409,9 @@ class DefaultTags(MultiValue):
     type = Token.DEFAULT_TAGS
 
     @classmethod
-    def from_params(cls, tags, separator=FOUR_SPACES, eol=EOL):
+    def from_params(cls, values, separator=FOUR_SPACES, eol=EOL):
         tokens = [Token(Token.DEFAULT_TAGS, 'Default Tags')]
-        for tag in tags:
+        for tag in values:
             tokens.append(Token(Token.SEPARATOR, separator))
             tokens.append(Token(Token.ARGUMENT, tag))
         tokens.append(Token(Token.EOL, eol))
@@ -423,11 +423,11 @@ class SuiteSetup(Fixture):
     type = Token.SUITE_SETUP
 
     @classmethod
-    def from_params(cls, keyword, args=(), separator=FOUR_SPACES, eol=EOL):
+    def from_params(cls, name, args=(), separator=FOUR_SPACES, eol=EOL):
         tokens = [
             Token(Token.SUITE_SETUP, 'Suite Setup'),
             Token(Token.SEPARATOR, separator),
-            Token(Token.NAME, keyword)
+            Token(Token.NAME, name)
         ]
         for arg in args:
             tokens.append(Token(Token.SEPARATOR, separator))
@@ -441,11 +441,11 @@ class SuiteTeardown(Fixture):
     type = Token.SUITE_TEARDOWN
 
     @classmethod
-    def from_params(cls, keyword, args=(), separator=FOUR_SPACES, eol=EOL):
+    def from_params(cls, name, args=(), separator=FOUR_SPACES, eol=EOL):
         tokens = [
             Token(Token.SUITE_TEARDOWN, 'Suite Teardown'),
             Token(Token.SEPARATOR, separator),
-            Token(Token.NAME, keyword)
+            Token(Token.NAME, name)
         ]
         for arg in args:
             tokens.append(Token(Token.SEPARATOR, separator))
@@ -459,11 +459,11 @@ class TestSetup(Fixture):
     type = Token.TEST_SETUP
 
     @classmethod
-    def from_params(cls, keyword, args=(), separator=FOUR_SPACES, eol=EOL):
+    def from_params(cls, name, args=(), separator=FOUR_SPACES, eol=EOL):
         tokens = [
             Token(Token.TEST_SETUP, 'Test Setup'),
             Token(Token.SEPARATOR, separator),
-            Token(Token.NAME, keyword)
+            Token(Token.NAME, name)
         ]
         for arg in args:
             tokens.append(Token(Token.SEPARATOR, separator))
@@ -477,11 +477,11 @@ class TestTeardown(Fixture):
     type = Token.TEST_TEARDOWN
 
     @classmethod
-    def from_params(cls, keyword, args=(), separator=FOUR_SPACES, eol=EOL):
+    def from_params(cls, name, args=(), separator=FOUR_SPACES, eol=EOL):
         tokens = [
             Token(Token.TEST_TEARDOWN, 'Test Teardown'),
             Token(Token.SEPARATOR, separator),
-            Token(Token.NAME, keyword)
+            Token(Token.NAME, name)
         ]
         for arg in args:
             tokens.append(Token(Token.SEPARATOR, separator))
@@ -495,11 +495,11 @@ class TestTemplate(SingleValue):
     type = Token.TEST_TEMPLATE
 
     @classmethod
-    def from_params(cls, keyword, separator=FOUR_SPACES, eol=EOL):
+    def from_params(cls, value, separator=FOUR_SPACES, eol=EOL):
         return cls([
             Token(Token.TEST_TEMPLATE, 'Test Template'),
             Token(Token.SEPARATOR, separator),
-            Token(Token.NAME, keyword),
+            Token(Token.NAME, value),
             Token(Token.EOL, eol)
         ])
 
@@ -509,11 +509,11 @@ class TestTimeout(SingleValue):
     type = Token.TEST_TIMEOUT
 
     @classmethod
-    def from_params(cls, timeout, separator=FOUR_SPACES, eol=EOL):
+    def from_params(cls, value, separator=FOUR_SPACES, eol=EOL):
         return cls([
             Token(Token.TEST_TIMEOUT, 'Test Timeout'),
             Token(Token.SEPARATOR, separator),
-            Token(Token.ARGUMENT, timeout),
+            Token(Token.ARGUMENT, value),
             Token(Token.EOL, eol)
         ])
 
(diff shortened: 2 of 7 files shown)


src/robot/parsing/model/statements.py
```diff
@@ -450,6 +457,16 @@ def _add_error(self, error):
 class IfHeader(Statement):
     type = Token.IF
 
+    @classmethod
+    def from_params(cls, condition, indent='    ', separator='    ', eol='\n'):
```
MaciejWiczk, line 461: @bhirsz maybe default indent, separator and eol could be extracted to constants like:
```
def from_params(cls, condition, indent=INDENT, separator=SEPARATOR, eol=UNIX_EOL):
```
where INDENT, SEPARATOR and UNIX_EOL:
```
INDENT='    '
SEPARATOR='    '
UNIX_EOL='\n'
```

pekkaklarck: I'm not sure about this. Problems:
- Seeing actual values directly would have its benefits.
- Because default values are bound when functions are created, changing constant values won't have any effect anyway.
- `UNIX_EOL` is a confusing name because `\n` is a generic newline in Python and converted to OS specific newline when saving. Passing `\r\n` as the EOL would in most cases cause wrong results on Windows.

The above said, repeating `'    '` everywhere is annoying as well and avoiding that may be a good idea anyway. `UNIX_EOL` should nevertheless be just `EOL`. Instead of `SEPARATOR` and `INDENT` we could also use `FOUR_SPACES` to make the actual value explicit. It's longer than the alternatives, though.

src/robot/parsing/model/statements.py
```diff
@@ -514,5 +565,5 @@ class EmptyLine(Statement):
     type = Token.EOL
 
     @classmethod
-    def from_value(cls, value):
+    def from_params(cls, value):
```
pekkaklarck, line 568: `value` could be renamed to `error` when making changes here.

pekkaklarck (commented):
Looks good! Comments:

- I think all statements should get this factory method. They might all be needed in some use cases and all statements having them would also be good for consistency reasons.
- Perhaps `create` would be better than `from_params`.
- See also individual line comments.

utest/parsing/test_statements.py
```diff
@@ -0,0 +1,131 @@
...
+    assert_equal(len(model.tokens), len(expected.tokens))
+    for m, e in zip(model.tokens, expected.tokens):
+        assert_equal(m, e, formatter=repr)
+    assert_equal(model._attributes, ('lineno', 'col_offset', 'end_lineno',
+                                     'end_col_offset', 'errors'))
+    assert_equal(model.lineno, expected.tokens[0].lineno)
+    assert_equal(model.col_offset, expected.tokens[0].col_offset)
+    assert_equal(model.end_lineno, expected.tokens[-1].lineno)
+    assert_equal(model.end_col_offset, expected.tokens[-1].end_col_offset)
+    assert_equal(model.errors, expected.errors)
```
pekkaklarck, line 29: Should we perhaps add `__eq__` to `Statement`? Then comparing expected and actual statements would be trivial. `Token` already has it originally for same reasons. Implementation shouldn't require more than this:

```python
    def __eq__(self, other):
        return (isinstance(other, type(self))
                and self.tokens == other.tokens
                and self.errors == other.errors)

    def __ne__(self, other):
        return not self == other    
```

Better error reporting would then require adding also `__repr__`. I prototyped that a bit and this seems to produce pretty nice output:

```python
    def __repr__(self):
        errors = '' if not self.errors else ', errors=%s' % list(self.errors)
        return '%s(tokens=%s%s)' % (type(self).__name__, list(self.tokens), errors)
```

Alternatively we could use `ast.dump` in `__repr__` but I think repr that resembles how to construct the object is better.

src/robot/parsing/model/statements.py
MaciejWiczk: @pekkaklarck my idea was to provide constans as default values, unless function is invoked with other values, just to avoid repeating and possible typos when using `'    '`. I agree that `EOL` is way better. I didn't know that python was replacing that based on OS used. Good to know. `FOUR_SPACES` is descriptive, so even tough it's longer name it is better one.

pekkaklarck: Ok, let's go with `FOUR_SPACES` and `EOL`.

pekkaklarck:
Although I wrote earlier that all statements should get this treatment, they don't necessarily need to get it in RF 4 and definitely not in beta 2. Right now it's important to agree on the overall design and to add methods that the new Tidy needs.

bhirsz:
> Looks good! Comments:
> 
> * I think all statements should get this factory method. They might all be needed in some use cases and all statements having them would also be good for consistency reasons.
> * Perhaps `create` would be better than `from_params`.
> * See also individual line comments.

I will update all statements with this factory method. 

I mentioned this in Slack but if it's about `from_params` I prefer the current version. Not 100% happy with it but it's akin to 'from_tokens' method so for someone who was using parsing API already it will be easier to understand what it's doing.

utest/parsing/test_statements.py
bhirsz: Good idea. We don't need any other comparison like `assert_equal(model.col_offset, expected.tokens[0].col_offset)` etc? I copied the whole comparison method from statements unit tests. But I guess if we're checking if we have the same tokens, it's enough as tokens are used to fill all other data.

bhirsz:
Updated with all comments. Still not covered all statements - on it.

src/robot/parsing/model/statements.py
pekkaklarck: Yeah, `eol`.

utest/parsing/test_statements.py
pekkaklarck: `Statement.col_offset` and others are properties. Their values depend solely on tokens so just testing tokens ought to be enough here. These properties themselves should be tested separately, and I think they already have some tests.

bhirsz: Yes it's covered somewhere else. I used your code for __eq__

src/robot/parsing/model/statements.py
```diff
@@ -355,42 +577,152 @@ def name(self):
...
+        return cls(tokens)
+
 
 @Statement.register
 class KeywordCall(Statement):
     type = Token.KEYWORD
     handles_types = (Token.KEYWORD, Token.ASSIGN)
 
+    @classmethod
+    def from_params(cls, keyword, assign=None, args=None, indent=FOUR_SPACES, separator=FOUR_SPACES, eol=EOL):
```
pekkaklarck, line 712: `assign=(), args=()` would simplify code below a bit.

src/robot/parsing/model/statements.py
```diff
@@ -408,6 +740,16 @@ def assign(self):
 class TemplateArguments(Statement):
     type = Token.ARGUMENT
 
+    @classmethod
+    def from_params(cls, args, indent=FOUR_SPACES, separator=FOUR_SPACES, eol=EOL):
+        tokens = [Token(Token.SEPARATOR, indent)]
+        for index, arg in enumerate(args):
+            if index != 0:
+                tokens.append(Token(Token.SEPARATOR, separator))
+            tokens.append(Token(Token.ARGUMENT, arg))
```
pekkaklarck, line 749: Could be formatted like this as well:
```python
tokens = []
for index, arg in enumerate(args):
    tokens.append(Token(Token.SEPARATOR, separator if index else indent))
    tokens.append(Token(Token.ARGUMENT, arg))
```

src/robot/parsing/model/statements.py
```diff
@@ -417,6 +759,23 @@ def args(self):
 class ForHeader(Statement):
     type = Token.FOR
 
+    @classmethod
+    def from_params(cls, for_separator, values, variables, indent=FOUR_SPACES, separator=FOUR_SPACES, eol=EOL):
```
pekkaklarck, line 763: Not sure about the order of arguments. `variables, for_separator, values` would be the same order as in the data, but using `variables, values, for_separator='IN'` would allow using a default value with the separator.

pekkaklarck (changes requested):
Looks great! I added several comments but many of them were style related (i.e. nitpicking). Some things that really need to be changed as well, though. We can discuss more about them here in comments or on Slack.

src/robot/parsing/model/statements.py
```diff
@@ -196,6 +212,18 @@ def name(self):
 class LibraryImport(Statement):
     type = Token.LIBRARY
 
+    @classmethod
+    def from_params(cls, library, alias=None, separator=FOUR_SPACES, eol=EOL):
+        separator_token = Token(Token.SEPARATOR, separator)
```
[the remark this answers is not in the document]
bhirsz: I will rename it to `sep` 👍

src/robot/parsing/model/statements.py
```diff
@@ -214,6 +242,16 @@ def alias(self):
...
 
+    @classmethod
+    def from_params(cls, resource, separator=FOUR_SPACES, eol=EOL):
+        tokens = [
+            Token(Token.RESOURCE, 'Resource'),
+            Token(Token.SEPARATOR, separator),
+            Token(Token.NAME, resource),
+            Token(Token.EOL, eol)
+        ]
+        return cls(tokens)
```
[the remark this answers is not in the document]
bhirsz: Yeah I also noticed that I'm switched the way I call cls() -> I will go with cls([]) whenever I define token list directly, without any further changes so it will be consistent.

src/robot/parsing/model/statements.py
```diff
@@ -236,6 +285,44 @@ def args(self):
 class Documentation(DocumentationOrMetadata):
     type = Token.DOCUMENTATION
 
+    @classmethod
+    def from_params(
+            cls,
+            doc,
+            test_documentation=False,
+            split_on_eol=True,
```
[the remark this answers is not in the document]
bhirsz: I agree on `settings_section` (with default to True, since it's opposite of current flag meaning). 

I will use `split_multiline` since I'm using this name somewhere else too, and move to the end of args.

src/robot/parsing/model/statements.py
```diff
@@ -236,6 +285,44 @@ def args(self):
...
+            for line in doc_lines[1:]:
+                if test_documentation:
+                    tokens.append(Token(Token.SEPARATOR, indent))
+                tokens.append(Token(Token.CONTINUATION, '...'))
+                tokens.append(Token(Token.SEPARATOR, separator))
+                tokens.append(Token(Token.ARGUMENT, line))
+                tokens.append(Token(Token.EOL, eol))
+        else:
+            tokens.append(Token(Token.ARGUMENT, doc))
+            tokens.append(Token(Token.EOL, eol))
```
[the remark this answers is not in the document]
bhirsz: Yeah I also wondered about '\n' when creating test cases for documentation. I will go with always splitting multiline docs and simplify algorithm.

src/robot/parsing/model/statements.py
```diff
@@ -260,46 +368,146 @@ def value(self):
...
+        tag_tokens.append(Token(Token.EOL, eol))
+        return cls(tag_tokens)
+
 
 @Statement.register
 class SuiteSetup(Fixture):
     type = Token.SUITE_SETUP
 
+    @classmethod
+    def from_params(cls, keyword, args=None, separator=FOUR_SPACES, eol=EOL):
```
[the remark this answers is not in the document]
bhirsz: I wanted to avoid putting mutable default value in argument, but didn't thought of using unmutable tuple. Thanks for the tip

src/robot/parsing/model/statements.py
```diff
@@ -260,46 +368,146 @@ def value(self):
...
+        tokens = [Token(Token.SUITE_SETUP, 'Suite Setup')]
+        if keyword:
```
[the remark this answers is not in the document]
bhirsz: I agree, I will remove it (also for similiar cases, except statements where empty value is allowed).

src/robot/parsing/model/statements.py
```diff
@@ -246,6 +333,27 @@ def value(self):
 class Metadata(DocumentationOrMetadata):
     type = Token.METADATA
 
+    @classmethod
+    def from_params(cls, metadata, split_on_eol=True, separator=FOUR_SPACES, eol=EOL):
```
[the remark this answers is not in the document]
pekkaklarck: Probably even better to just always split the value and remove this argument altogether.

src/robot/parsing/model/statements.py
```diff
@@ -337,6 +545,13 @@ def _is_valid_dict_item(self, item):
 class TestCaseName(Statement):
     type = Token.TESTCASE_NAME
 
+    @classmethod
+    def from_params(cls, name, eol=EOL):
+        return cls([
+            Token(Token.TESTCASE_NAME, name),
+            Token(Token.EOL, eol)
+        ])
```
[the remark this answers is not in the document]
bhirsz: In this case we can support it since it's not that exotic (at least for my experience, keeping test case name and template arguments in same line).

src/robot/parsing/model/statements.py
```diff
@@ -355,42 +577,152 @@ def name(self):
...
 
+    @classmethod
+    def from_params(cls, keyword=None, indent=FOUR_SPACES, separator=FOUR_SPACES, eol=EOL):
+        tokens = [
+            Token(Token.SEPARATOR, indent),
+            Token(Token.TEMPLATE, '[Template]')
+        ]
+        if keyword:
+            tokens.append(Token(Token.SEPARATOR, separator))
+            tokens.append(Token(Token.NAME, keyword))
```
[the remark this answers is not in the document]
bhirsz: It's rarely used and probably not known (overriding suite settings by empty values) so we ought to support only non empty values in `from_params`. In case someone needs empty value there is always `from_tokens` method.

src/robot/parsing/model/statements.py
```diff
@@ -355,42 +577,152 @@ def name(self):
...
+        tokens.append(Token(Token.EOL, eol))
+        return cls(tokens)
+
 
 @Statement.register
 class Arguments(MultiValue):
     type = Token.ARGUMENTS
 
+    @classmethod
+    def from_params(cls, args, split_to_multiline=False, indent=FOUR_SPACES, separator=FOUR_SPACES, eol=EOL):
```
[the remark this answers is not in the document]
bhirsz: I'm going to add documentation (especially to every method that have some control flag). Agreed on rest. 
Splitting arguments functionality could be useful (I'm already doing it in some cases in some tools) but I'm still not 100% if I should add this to core library. I will think about it.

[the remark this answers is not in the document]
bhirsz: Decided to remove splitting. There is too many possiblities to cover (for example, split every line vs split after x arguments, first argument should be the same line or the next..). If someone needs something more advanced it will be better to use `from_tokens` instead.

src/robot/parsing/model/statements.py
bhirsz: `variables, values, for_separator='IN'` sounds handy so I will use it.

src/robot/parsing/model/statements.py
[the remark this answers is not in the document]
pekkaklarck: I was also thinking that there are multiple ways to split args. Handling them is Tidy functionality.

src/robot/parsing/model/statements.py
pekkaklarck: Just realized the name of the property accessing this value is `flavor` and that might be a good name for the argument as well. The token type is `FOR_SEPARATOR` so the current `for_separator` is fine as well.

bhirsz: When debugging the first thing people see is most likely `flavor` rather that type of the token (`FOR_SEPARATOR`). And if they check source code, they will see variables, values and flavor properties first. So I will change argument name to follow it 👍

src/robot/parsing/model/statements.py
```diff
@@ -246,6 +324,23 @@ def value(self):
...
+        tokens = [
+            Token(Token.METADATA, 'Metadata'),
+            Token(Token.SEPARATOR, separator)
+        ]
+        metadata_lines = metadata.splitlines()
+        if metadata_lines:
+            tokens.append(Token(Token.ARGUMENT, metadata_lines[0]))
+            tokens.append(Token(Token.EOL, eol))
+        for line in metadata_lines[1:]:
+            tokens.append(Token(Token.CONTINUATION, '...'))
```
pekkaklarck, line 338: Value isn't actually needed because `CONTINUATION` tokens get `'...'` automatically. Having the value visible perhaps make the code easier to understand, though, but then again e.g. `IF` code doesn't have it.

If this is changed, then possibly should also change `WITH_NAME` to use same approach. That type doesn't get the value automatically, but this could be easily changed in `Tken.__init__`. Feel free to do that as part of this PR if you think it makes sense.

bhirsz: Right, and I actually used Token(Token.Continuation) in other parts of code. I will update WITH_NAME token as well

src/robot/parsing/lexer/tokens.py
```diff
@@ -143,7 +143,7 @@ def __init__(self, type=None, value=None, lineno=-1, col_offset=-1, error=None):
             value = {
                 Token.IF: 'IF', Token.ELSE_IF: 'ELSE IF', Token.ELSE: 'ELSE',
                 Token.FOR: 'FOR', Token.END: 'END', Token.CONTINUATION: '...',
-                Token.EOL: '\n'
+                Token.EOL: '\n', Token.WITH_NAME: 'WITH NAME'
```
pekkaklarck, line 146: Tokens automatically getting the value are listed in the class doc string and this should be added there as well. Could also consider adding a class attribute like `AUTO_VALUE = {...}` that could be referenced from the docs instead.

src/robot/parsing/model/statements.py
```diff
@@ -64,6 +68,22 @@ def from_tokens(cls, tokens):
...
+        There is one notable difference for `Documentation` statement where
+        ``settings_header`` flag is used to determine if statement belongs to settings header
+        or test/keyword.
+
+        Most implementations support following general properties:
+        - `separator` whitespace inserted between each token. Default is ``'    '``.
+        - ``indent`` whitespace inserted before first token. Default is ``'    '``.
+        - ``eol`` end of line sign. Default is ``'\n'``.
+        """
+        pass
```
pekkaklarck, line 85: Should use `raise NotImplementedError` here.

src/robot/parsing/model/statements.py
```diff
@@ -64,6 +68,22 @@ def from_tokens(cls, tokens):
...
+        should match class properties. Values are used to create matching tokens.
+
+        There is one notable difference for `Documentation` statement where
+        ``settings_header`` flag is used to determine if statement belongs to settings header
+        or test/keyword.
+
+        Most implementations support following general properties:
+        - `separator` whitespace inserted between each token. Default is ``'    '``.
+        - ``indent`` whitespace inserted before first token. Default is ``'    '``.
+        - ``eol`` end of line sign. Default is ``'\n'``.
```
pekkaklarck, line 83: - Instead of `Default is '    '.` it would probably be better to say `Default is four spaces.`.
- `\n` in the docstring is a real newline. It should be changed to `\\n` or the docstring prefixed with `r`.

src/robot/parsing/model/statements.py
```diff
@@ -64,6 +68,22 @@ def from_tokens(cls, tokens):
                 return handlers[token.type](tokens)
         return EmptyLine(tokens)
 
+    @classmethod
+    def from_params(cls, *args, **kwargs):
+        """Create statement from passed parameters. Required and optional arguments
+        should match class properties. Values are used to create matching tokens.
```
pekkaklarck, line 74: This doesn't follow the [general docstring conventions](https://www.python.org/dev/peps/pep-0257/) fully. Just using the first sentence as the short doc i.e. formatting like

```
"""Create statement from passed parameters.

Required and optional arguments should match class properties. Values are ...
```

would be a bit better.

More importantly, docs themselves are great!

src/robot/parsing/model/statements.py
```diff
@@ -181,6 +213,13 @@ class SectionHeader(Statement):
                      Token.TESTCASE_HEADER, Token.KEYWORD_HEADER,
                      Token.COMMENT_HEADER)
 
+    @classmethod
+    def from_params(cls, type, name, separator=FOUR_SPACES, eol=EOL):
+        return cls([
+            Token(type, name),
+            Token('EOL', '\n')
+        ])
```
pekkaklarck, line 221: `separator` isn't used for anything. Is it in the signature for consistency with other `from_param` methods or by accident? I don't think consistency is a good enough reason to have it there and it should be removed.

pekkaklarck (approved):
Great work! I added few comments but nothing urgent. If needed, I can fix them myself after merging this as well.

src/robot/parsing/model/statements.py
bhirsz: It's by accident, thanks for noticing

src/robot/parsing/lexer/tokens.py
bhirsz: I added missing token to docs (manually). `AUTO_VALUE` class attribute can be added separetly. If we're going to add it, should it be also used in `__init__`? ie:
```
class Token:
    """
    :attr:`AUTO_VALUE`
    """
    AUTO_VALUE = {
                Token.IF: 'IF', Token.ELSE_IF: 'ELSE IF', Token.ELSE: 'ELSE',
                Token.FOR: 'FOR', Token.END: 'END', Token.CONTINUATION: '...',
                Token.EOL: '\n', Token.WITH_NAME: 'WITH NAME'
    }
    
    def __init__(self, value=None, ...):
        if value is None:
            value = self.AUTO_VALUE.get(type, '')

Comment (pekkaklarck) [reply src/robot/parsing/lexer/tokens.py:146]: Yeah, if we'd add AUTO_VALUE it should work like that. Probably in `__init__` like
```
if value is None and type in self.AUTO_VALUE:
    value = self.AUTO_VALUE[type]
```
Probably not worth the effort at least now.

pekkaklarck:
Thanks a lot for a great PR @bhirsz!

PR #3812, opened 2020-12-27
Trim collapse spaces 3240

vokiput:


vokiput:
Created it acidentally.

PR #3832, opened 2021-01-20
Revert adding __eq__ methods to Statement class

### LINKED ISSUE #3830: AST classes no longer hashable.
It seems that AST classes now implement `__eq__` and `__ne__` (added in https://github.com/robotframework/robotframework/commit/fd641abe368fa5eb02708225e0fc64b63d33d373), but don't provide a `__hash__` and thus the AST classes are no longer hashable.

Is it possible to make them hashable again? (this is breaking the Robot Framework Language Server)

My suggestion would be to not even add `__eq__` and `__ne__` and leave the `__eq__` / `__hash__` by `object id` as it doesn't really seem that useful as cases probably vary quite a bit on what needs to be used to actually consider that one AST equals another -- considering only the structure or actual lines/columns comes to mind (so, if someone has a special case for comparing ASTs I think that in general that should be done by who is doing that compare -- possibly providing one helper function for the current use-case if needed).

bhirsz:
This should fix #3830

### DIFF
--- a/utest/parsing/test_statements.py
+++ b/utest/parsing/test_statements.py
@@ -37,7 +37,8 @@ from robot.parsing.model.statements import (
     Comment,
     EmptyLine
 )
-from robot.utils.asserts import assert_equal
+from robot.utils.asserts import assert_equal, assert_true
+from robot.utils.robottypes3 import type_name
 
 
 def assert_created_statement(tokens, base_class, **params):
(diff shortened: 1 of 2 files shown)


utest/parsing/test_statements.py
```diff
@@ -56,11 +57,21 @@ def assert_created_statement(tokens, base_class, **params):
...
+
+
+def assert_statements(st1, st2):
+    for t1, t2 in zip(st1, st2):
         assert_equal(t1, t2, formatter=repr)
-    assert_equal(first, second)
-    assert_equal(len(first), len(second))
+    assert_true(
+        compare_statements(st1, st2),
+        'Statements are not equal. %s (%s) != %s (%s)' % (st1, type_name(st1),
```
bhirsz, line 71: I could also import report_inequality from robot.asserts OR create new method inside robot asserts that will check if given condition is true.

utest/parsing/test_statements.py
```diff
@@ -37,7 +37,8 @@
     Comment,
     EmptyLine
 )
-from robot.utils.asserts import assert_equal
+from robot.utils.asserts import assert_equal, assert_true
+from robot.utils.robottypes3 import type_name
```
pekkaklarck, line 41: Use `from robot.utils import type_name` instead. We generally import from package level when possible and the above isn't Python 2 compatible.

pekkaklarck (changes requested):

fabioz:
Is anything blocking this merge? -- I'd really like to have tests passing on the language server when running with the robotframework `master` again ;)

pekkaklarck:
The thing blocking merge was me forgetting about this. Great that @bhirsz reminded me on Slack and this made it to beta 2.

Issue #3240, opened 2019-07-12
BuiltIn: Support stripping whitespace with `Should Be Equal` and other comparison keywords


lucielavickova:
Would it be possible to add "**trim**" parameter to keyword "**Should Be Equal As Strings**" (+related keywords)? 
Many times I need to compare two strings, but there are some trailing or leading whitespaces in the actual value, which I don't care about, so I always have to trim them before. Having trim as a parameter in these keywords would help me a lot :)

pekkaklarck:
This is a good idea. Should also have an option to collapse internal spaces and to normalize spaces so that newlines, tabs, and non-ASCII spaces are all converted to spaces. Also, should support trimming/stripping from both ends and only from left/right.

Implementing something like this is very easy by using Python's string methods (and regexps when normalizing whitespace). The hard part is designing how to configure all this, but perhaps it would be fine to have separate options like `strip_spaces=False`, `collapse_spaces=False`, and so on. Exact argument names (`strip` vs `strip_spaces` vs `strip_whitespace`) are also subject to some discussion.

Would you @lucietriskova be interested to implement this enhancement? As I already wrote, implementation ought to be pretty easy and I can help if needed.

lucietriskova:
ok, I think next week I could start working on it

pekkaklarck:
Great! Let me know if I can help somehow. We can discuss about the design here or on the `#devel` channel on our Slack.

sohum:
Hey guys! I'd be happy to take a stab at this, @lucielavickova have you had a chance to get started on it?

lucielavickova:
hey @sohum, feel free to do it, I did not find time to have a look yet. Thank you!!!

sohum:
> This is a good idea. Should also have an option to collapse internal spaces and to normalize spaces so that newlines, tabs, and non-ASCII spaces are all converted to spaces. Also, should support trimming/stripping from both ends and only from left/right.
> 
> Implementing this is pretty easy by using Python's string methods (and regexps when normalizing whitespace). The hard part is designing how to configure all this, but perhaps it would be fine to have separate options like `strip_spaces=False`, `collapse_spaces=False`, and so on. Exact argument names (`trim` vs `strip` vs `strip_spaces` vs `strip_whitespace`) are also subject to some discussion.
> 
[...]

PR #3843, opened 2021-02-03
PR for the issue 3240

vokiput:
https://github.com/robotframework/robotframework/issues/3240


src/robot/libraries/BuiltIn.py
```diff
@@ -618,6 +618,13 @@ def should_be_equal(self, first, second, msg=None, values=True,
         If both arguments are multiline strings, this keyword uses
         `multiline string comparison`.
 
+        If ``strip_spaces`` is given a true value (see `Boolean arguments`)
+        and both arguments are strings, the comparison is done without leading
+        and trailing spaces. If ``strip_spaces`` is given a string value
+        ``leading`` or ``trailing`` and both arguments are strings, the
+        comparison is done without leading or trailing spaces respectively.
+        The default value is ``False``.
```
pekkaklarck, line 626: Good documentation! Few things can be still enhanced, though:

- Perhaps `leading or trailing` could be changed to `LEADING or TRAILING (case-insensitively)`. 
- The second `both arguments are strings` can be removed.
- Mentioning the actual default _value_ isn't necessary when you can see it in arguments yourself. Something like `By default spaces are not stripped.` would explain the default _behavior_ but that ought to be clear also otherwise.
- Need a note that this is new in Robot Framework 4.0 (hoping we still get this in!).

All these comments obviously apply to same doc elsewhere as well.

src/robot/libraries/BuiltIn.py
```diff
@@ -663,21 +673,44 @@ def _raise_multi_diff(self, first, second, formatter):
...
         return is_truthy(values) and str(values).upper() != 'NO VALUES'
 
+    def _strip_spaces(self, string_value, strip_spaces):
+        if is_string(strip_spaces):
+            if strip_spaces.upper() == 'LEADING':
+                string_value = string_value.lstrip()
+            elif strip_spaces.upper() == 'TRAILING':
+                string_value = string_value.rstrip()
+            elif is_truthy(strip_spaces):
+                string_value = string_value.strip()
```
pekkaklarck, line 683: Last `elif` seems redundant considering the code below.

src/robot/libraries/BuiltIn.py
```diff
@@ -663,21 +673,44 @@ def _raise_multi_diff(self, first, second, formatter):
...
+        comparison is done without leading or trailing spaces respectively.
+        The default value is ``False``.
+
         New option in Robot Framework 3.0.1.
         """
         self._log_types_at_info_if_different(first, second)
         if is_truthy(ignore_case) and is_string(first) and is_string(second):
             first = first.lower()
             second = second.lower()
+        if strip_spaces:
```
pekkaklarck, line 711: Need to check `first` and `second` are strings here as well.

I actually think it should be an error to use non-string values when `strip_spaces` isn't `False`. We don't do that with `ignore_case` so probably better to not do that here either now, but this could be changed in the future.

src/robot/libraries/BuiltIn.py
```diff
@@ -1050,7 +1158,16 @@ def should_contain_x_times(self, container, item, count, msg=None,
...
+        case-insensitive.
+
+        If ``strip_spaces`` is given a true value (see `Boolean arguments`)
+        and both arguments are strings, the comparison is done without leading
+        and trailing spaces. If ``strip_spaces`` is given a string value
+        ``leading`` or ``trailing`` and both arguments are strings, the
+        comparison is done without leading or trailing spaces respectively.
+        The default value is ``False``.
+
+        If the ``container`` is a list-like object, string
```
pekkaklarck, line 1170: Docs messed up a bit here.

src/robot/libraries/BuiltIn.py
```diff
@@ -1100,6 +1223,10 @@ def should_not_match(self, string, pattern, msg=None, values=True,
         ``*``, ``?`` and ``[chars]`` acting as wildcards. See the
         `Glob patterns` section for more information.
 
+        If ``ignore_case`` is given a true value (see `Boolean arguments`) and
+        compared items are strings, it indicates that comparison should be
+        case-insensitive.
```
pekkaklarck, line 1228: Great catch that this documentation was missing!

With this keyword arguments are always strings, so the `and compared items are strings.` can be removed. Same applies to the other place where this documentation was added.

pekkaklarck (approved):
Great work! Few things can still we enhanced but I can take care of them myself. I want to get RF 4.0 rc 1 out today and because this PR is in such a good shape I want it in.

PR #3849, opened 2021-02-10
Add Codecov badge

thomasrockhu:
Tom from Codecov here. I thought I would go ahead and add the Codecov badge here in case it's useful!


pekkaklarck:
We aren't actively following coverage. The main reason is that unit tests alone don't tell much because most of our tests are acceptance tests.

Issue #3857, opened 2021-02-17
Parsing API error handling does not detect fatal errors properly


d-biehl:
Hello,

Just playing around with the parsing Model of Robotframework.
I've noticed that the Error statement does not read Token.FATAL_ERROR in errors property, but it is defined to handle both Token.Error and Token.FATAL_ERROR.

The following code:

```python
import io

from robot.api.parsing import ModelVisitor, get_resource_model


class ErrorReporter(ModelVisitor):
    # Implement `generic_visit` to visit all nodes.
    def generic_visit(self, node):
        if node.errors:
            print(f'Error on line {node.lineno}:')
            for error in node.errors:
                print(f'- {error}')
        ModelVisitor.generic_visit(self, node)


text = "*** Test Cases ***"

with io.StringIO(text) as data:
    model = get_resource_model(data)

    ErrorReporter().visit(model)
```

should return like:

    Error on line 1:
    - Resource file with 'Test Cases' section is invalid.

but it doesn't.

See code from Master Branch:

https://github.com/robotframework/robotframework/blob/fb355fbc99bd8ced392017a4317ea9a9dff4b659/src/robot/parsing/model/statements.py#L914

pekkaklarck:
First of all, it's great to see these APIs investigated and used by others! Secondly, great catch! I can fix this myself but if you are interested to provide a PR that obviously works too. I try to get RF 4.0 rc 1 out early next week and this definitely needs to included.

pekkaklarck:
This error reporting was added in 2127623386f43e5a536b55ae1f27d8be1a6b0800 with some tests in `test_model.py`. It seems that the bug was introduced in 0194b6271b8512464c0149cb604df3dea2a7c4f2 when handling token types was changed and the `Error` statement started to handle both `ERROR` and `FATAL_ERROR` tokens. Ought to be trivial to fix with enhanced/new tests in `test_model.py`.

d-biehl:
just add added the PR #3861 to correct this issue, also added/modified the unittests for it.

pekkaklarck:
Thanks for the PR @d-biehl!

PR #3861, opened 2021-02-19
PR for #3857: correct Error statement to correctly handle FATAL_ERRORS

d-biehl:
this corrects issue: #3857

also added some simple unit tests

### DIFF
--- a/src/robot/parsing/model/statements.py
+++ b/src/robot/parsing/model/statements.py
@@ -911,7 +911,7 @@ class Error(Statement):
     @property
     def errors(self):
         """Errors got from the underlying ``ERROR`` tokens."""
-        return tuple(t.error for t in self.get_tokens(Token.ERROR)) + self._errors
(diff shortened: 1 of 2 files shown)


src/robot/parsing/model/statements.py
```diff
@@ -911,7 +911,7 @@ class Error(Statement):
     @property
     def errors(self):
         """Errors got from the underlying ``ERROR`` tokens."""
-        return tuple(t.error for t in self.get_tokens(Token.ERROR)) + self._errors
+        return tuple(t.error for t in self.get_tokens(Token.ERROR, Token.FATAL_ERROR)) + self._errors
```
pekkaklarck, line 914: This line looks pretty long. CONTRIBUTING.rst states that the max line length with out Python code is 88 characters. Not a big problem can can be changed later.

pekkaklarck (approved):
Looks good! Very good tests!

Issue #3625, opened 2020-07-31
Support custom statuses


pekkaklarck:
It would be convenient to be able to configure Robot Framework to have custom statuses in addition to the current PASS and FAIL and the forthcoming SKIP (#3622). Currently tests can have tags to differentiate them, but explicit statuses would make the distinction more clear.

This functionality could be implemented so that new statuses are defined based on old statuses and tags. The specification could be given from the command like with a command line option like `--addstatus` with an argument in format such as `NEWSTATUS:OLDSTATUS:TAGPATTERN:COLOR`. In this spec `NEWSTATUS` would be the name of the new status to add, `OLDSTATUS` would be the old status to base this status on, `TAGPATTERN` would be the tag or tag pattern (e.g. `tag`, `tag:*` or `tag1ORtag2`) matching test must have, and `COLOR` would be the color associated with the status (possibly optional). Example:

    --addstatus KNOWN_ISSUE:FAIL:bug-id-*:purple

One use case for this feature would be separating tests skipped explicitly during execution and tests skipped by the forthcoming `--skiponfailure` feature. This would be convenient for those who oppose removing the criticality concept (#3624):

    --addstatus NON-CRITICAL:SKIP:non-critical:pink --skiponfailure non-critical

pekkaklarck:
This would be nice already in RF 4.1.

pekkaklarck:
Too big task for RF 4.1. I moved this tentatively to RF 5.0 scope but we haven't yet started more detailed planning of that release and it's not clear will this one be actually included or not. If you find this important, comment here and explain your use case!

pekkaklarck:
Oooops, there's PR #3869 adding this functionality. If the PR implements what's needed and doesn't have problems, this certainly can be included into RF 4.1 I don't have time to review the PR right now, but I moved this issue back to RF 4.1 to make sure I don't forget about it.

pekkaklarck:
PR #3869 is a good start. There is, however, so much work still to be done both internally to handle custom statuses and in log report to show them nicely in statistics and elsewhere that I move this issue to RF 5.0 scope.

Tattoo:
[...]

PR #3869, opened 2021-03-02
Support custom statuses

mihaiparvu:
PR for #3625 

Added the possibility to set custom status from command line option `--addstatus`. The new status is set based on tags and displayed both in console output and log.html. The color can be customized but only in the log file, in console it maintains the old colors.


pekkaklarck (changes requested):
This is a good start but the core logic how custom statuses are set to objects need to be changed to something less fragile. Let's discuss about better alternative separately.

There were also some other issues explained in individual comments. In addition to them, I noticed these problems when testing the code locally:

- Statistics aren't collected for the custom statuses. There probably should be new status columns in both log and report statics, but needed to test how that works if there are multiple custom statuses. Should also show custom statuses in status graphs.

  Right now also collecting "normal" statistics is broken. I tested creating a new status based on PASS, and those tests were considered failed in all statistics.

- Dark status colors make reading the black status text hard in the log file. We preferably should automatically decide should the status text be black or white depending on the specified status color. Alternatively we could require giving that as well but it would make the usage a bit more complicated.

pekkaklarck:
The hard part with this issue is designing how to store information about custom statuses internally. For example, should the `status` attribute tests have get these custom statuses or should it always be one of the base statuses `PASS/FAIL/SKIP`? The former initially feels better, but it would require changes in all code looking for statuses. In the latter case test could/should have a separate attribute that returns possible custom status based on configs and test base status and tags. We could consider renaming the current `status` to e.g. `base_status` and using `status` with the new attribute, though.

We also need to design how to pass the custom status information to tests. Most likely it could be given to the top level result suite where tests could query it. We used similar approach with criticality information earlier.

We can discuss all this further here or on Slack. It seems pretty clear that there's so much work still to be done that it's nevertheless best to move this issue from RF 4.1 to RF 5.0 scope.

PR #3876, opened 2021-03-08
differentiate test and suite statuses from other statuses

bitcoder:
Differentiate test and suite statuses from other statuses, as "NOT RUN" is only available for keywords, FORs and IFs.
Related to #3726.

### DIFF
--- a/doc/schema/robot.02.xsd
+++ b/doc/schema/robot.02.xsd
@@ -38,7 +38,7 @@
             <xs:element name="test" type="Test" minOccurs="0" maxOccurs="unbounded" />
             <xs:element name="doc" type="xs:string" minOccurs="0" />
             <xs:element name="meta" type="Metadata" minOccurs="0" maxOccurs="unbounded" />
-            <xs:element name="status" type="SuiteAndTestStatus" />
+            <xs:element name="status" type="Status" />
         </xs:choice>
         <xs:attribute name="name" type="xs:string" />
         <xs:attribute name="source" type="xs:string" />
@@ -60,7 +60,7 @@
             <xs:element name="doc" type="xs:string" minOccurs="0" />
             <xs:element name="tag" type="xs:string" minOccurs="0" maxOccurs="unbounded" />
             <xs:element name="timeout" type="Timeout" minOccurs="0" />
-            <xs:element name="status" type="SuiteAndTestStatus" />
+            <xs:element name="status" type="Status" />
         </xs:choice>
         <xs:attribute name="name" type="xs:string" />
         <xs:attribute name="id" type="xs:string" />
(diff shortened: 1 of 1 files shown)


doc/schema/robot.02.xsd
```diff
@@ -60,13 +60,15 @@
...
-            <xs:element name="status" type="Status" />
+            <xs:element name="status" type="SuiteAndTestStatus" />
         </xs:choice>
         <xs:attribute name="name" type="xs:string" />
         <xs:attribute name="id" type="xs:string" />
     </xs:complexType>
     <xs:complexType name="Keyword">
         <xs:choice maxOccurs="unbounded">
+            <!-- These keywords can be regular keywords or test setup and teardown.
+                 If the later, they must have the `type` attribute set accordingly. -->
```
pekkaklarck, line 71: What does this comment refer to? Does it refer to the `<xs:complexType name="Keyword">` or keywords it can contain? In the latter case the comment would be pretty strange because keywords inside a keyword cannot be test setups or teardowns. In the former case the comment would be places a bit strangely. Mentioning only test setup/teadown would be also a bit strange when also suites have them and keywords can have teardown. If the comment would be moved a few lines and it would talk only about setups and teardowns (i.e. no word "test") it would make more sense to me.

pekkaklarck:
Good idea to restrict status values explicitly instead of just having a comment like I did. Two questions:

1. Did you try structuring this so that we'd have a base `StatusValue` without `NOT RUN` and then another status like `BodyItemStatus` that extends it adds `NOT RUN`? I haven't done that myself but AFAIK XSD supports it. Not sure is this worth the effort but ought to remove few lines from the schema when common statuses didn't need to be repeated. Should then also rename the current `Status` to `BodyItemStatus` and the `TestAndSuiteStatus` you added to `Status` to be consistent.
2. The added comment about setups and teardowns was a bit strange. I added a separate comment about it.

doc/schema/robot.02.xsd
bitcoder: you're right @pekkaklarck  I've updated it

bitcoder:
> 1. Did you try structuring this so that we'd have a base `StatusValue` without `NOT RUN` and then another status like `BodyItemStatus` that extends it adds `NOT RUN`? I haven't done that myself but AFAIK XSD supports it. Not sure is this worth the effort but ought to remove few lines from the schema when common statuses didn't need to be repeated. Should then also rename the current `Status` to `BodyItemStatus` and the `TestAndSuiteStatus` you added to `Status` to be consistent.

I think I can do that.

doc/schema/robot.02.xsd
```diff
@@ -65,9 +65,11 @@
         <xs:attribute name="name" type="xs:string" />
         <xs:attribute name="id" type="xs:string" />
     </xs:complexType>
+    <!-- Keywords can be regular keywords or setup and teardown keywords.
+         If the later, they must have the `type` attribute set accordingly. -->
     <xs:complexType name="Keyword">
         <xs:choice maxOccurs="unbounded">
-            <xs:element name="kw" type="Keyword" minOccurs="0" maxOccurs="unbounded" />
+           <xs:element name="kw" type="Keyword" minOccurs="0" maxOccurs="unbounded" />
```
pekkaklarck, line 72: Seems like an accidental change.

pekkaklarck:
Merged. I'll fix the accidental whitespace change myself. Thanks for the PR @bitcoder!

Issue #3867, opened 2021-02-26
Argument conversion is different between 3.2.2. and 4.0


aaltat:
When I have library like this:
```
from typing import Union


def keyword(arg: Union[str, dict, None] = None):
    print(arg)
    print(type(arg))
```
And tests like this
```
*** Setting ***
Library    MyLib.py

*** Variable ***
&{A}    key=value

*** Test Cases ***
Test A
    Keyword    ${A}

Test B
    Keyword    None

Test C
    Keyword    ${None}
```
Then the Robot Framework 4.0 the logging shows that arg type is always a string: `<class 'str'>`. But with Robot Framework 3.2.2, Test A logs `<class 'robot.utils.dotdict.DotDict'>`, the Test B logs: `<class 'str'>` and Test C logs `<class 'NoneType'>`.

I would thing that the conversion of the 3.2.2 to be more correct although it is not perfect.

pekkaklarck:
This is due the new Union conversion support (#3738) and to some extend string conversion support (#3736). The first type in the `Union` is `str` and Robot somewhat blindly converts all arguments to strings. The exact problem is discussed in the description of the former issue as the first design decision to be made:

> Should we first check does the given argument match any of the possible types before doing conversion? If, for example, we have `arg: Union[str, int]` and the given argument is an integer, should we pass it as is (it matches one of the possible types) or should we test conversion with possible types one-by-one? In the latter case we'd convert the integer to a string because the first possible type is `str`. In this case checking does the argument match any of the types feels better.
>
> On the other hand, if we have `arg: Union[int, str]` and call with a string `2`, checking does the given argument match any of the type would mean that we don't do any conversion because one of the types is `str`. If we'd go through types one-by-one, we'd convert it to an integer because `int` is before `str`. In this case testing conversion one-by-one works better.
>
> We could also make it so that we test conversions one-by-one if the given argument is a string and otherwise first check does the argument match any of the types. That feels a bit inconsistent and hard to explain, though. I got a feeling that going through types one-by-one works best. The key thing to remember is that `str` should be the last type to avoid converting all arguments to strings.

As explained above, the alternative to check does the argument match some of the types in the `Union` would be even worse than the current situation and handling `str` in `Union` somehow specially would be pretty complicated. For the user a simple solution is having `str` last in the `Union` i.e. using `Union[dict, None, str]` in the original example. If that's not considered adequate, we can still consider special `str` handling in this context.

pekkaklarck:
[...]

Issue #3738, opened 2020-10-25
Support type conversion with multiple possible types


pekkaklarck:
Currently our argument conversion logic allows specifying only one possible type for each argument. There are, however, use cases where declaring that, for example, either a list or a string is accepted and Robot should then try out different conversions. With type hints the natural syntax to use is  [Union](https://docs.python.org/3/library/typing.html#typing.Union) like `arg: Union[list, str]`.

The basic logic with multiple possible types would be going through all types one-by-one, trying to convert using each type, and going to the next one if conversion fails. There are, however, plenty of design decisions to be made about different details:

1. Should we first check does the given argument match any of the possible types before doing conversion? If, for example, we have `arg: Union[str, int]` and the given argument is an integer, should we pass it as is (it matches one of the possible types) or should we test conversion with possible types one-by-one? In the latter case we'd convert the integer to a string because the first possible type is `str`. In this case checking does the argument match any of the types feels better.

   On the other hand, if we have `arg: Union[int, str]` and call with a string `2`, checking does the given argument match any of the type would mean that we don't do any conversion because one of the types is `str`. If we'd go through types one-by-one, we'd convert it to an integer because `int` is before `str`. In this case testing conversion one-by-one works better.

   We could also make it so that we test conversions one-by-one if the given argument is a string and otherwise first check does the argument match any of the types. That feels a bit inconsistent and hard to explain, though. I got a feeling that going through types one-by-one works best. The key thing to remember is that `str` should be the last type to avoid converting all arguments to strings.

2. What to do if we don't recognize one of the possible types? I think we should test conversion with all recognized types that are before it but return the value as-is if no conversion succeeds. For example, `arg: Union[int, MyThing]` would convert string `2` to an integer but string `xxx` would be left as-is. This would be consistent with how we handle `arg: MyThing`. 

3. How to support multiple types with the `@keyword` decorator? We could require using Union also with it, but then this syntax wouldn't be supported with Python 2 and needing to use `typing` may not always be desirable with Python 3 either. I think we could allow using a tuple to types like `@keyword(types={'arg': (list, str)}` as an alias for Union.

4. If we support a tuple of types with `@keyword`, should we also support it with type hints like `arg: (list, str)`? Type checkers probably won't like that but I see no harm us supporting it assuming it doesn't require extra work.

[...]

PR #3895, opened 2021-03-18
Don't convert value if it is already a valid type.

KotlinIsland:
#3867
#3738

If a value is already a valid type from a type union then there is no need to convert it.

### DIFF
--- a/src/robot/running/arguments/typeconverters.py
+++ b/src/robot/running/arguments/typeconverters.py
@@ -451,7 +451,9 @@ class CombinedConverter(TypeConverter):
         return True
 
     def _no_conversion_needed(self, value):
-        return False
+        NoneType = type(None)
+        types = tuple(NoneType if it is None else it for it in self.args)
+        return isinstance(value, types)
 
     def _convert(self, value, explicit_type=True):
         for typ in self.args:
--- a/atest/robot/keywords/type_conversion/unions.robot
+++ b/atest/robot/keywords/type_conversion/unions.robot
@@ -28,3 +28,6 @@ Optional argument
 
 Optional argument with default
      Check Test Case    ${TESTNAME}
+
+Avoid unnecessary conversion
+     Check Test Case    ${TESTNAME}
--- a/atest/testdata/keywords/type_conversion/keyword_decorator.robot
+++ b/atest/testdata/keywords/type_conversion/keyword_decorator.robot
@@ -513,7 +513,7 @@ Multiple types using Union
     NONE          None
     ${1}          1
     ${1.2}        1.2
-    ${None}       None
+    ${None}       ${None}
 
 Argument not matching Union tupes
     [Tags]        require-py3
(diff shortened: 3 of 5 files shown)


KotlinIsland:
Or would this be better in `ArgumentConverter._convert`?

pekkaklarck:
The current functionality is there by design. The motivation is that we want conversion to be done if you have type like `Union[int, str]` and use argument like `42`. I don't want that to be changed, but I don't like too much that currently if you have `Union[int, float]` and use it with `${42.9}` that gets converted to an integer `42`. Possibly we should always convert strings but with others with check is the type valid already. Anyway, this change needs a separate issue and cannot be done before RF 4.1.

KotlinIsland:
The current functionality breaks optional strings `Optional[str]` completely.

If the purpose of this is to allow int/float/enum literals to override `str` then can we skip forced conversion when value is not a string

```py
    def _no_conversion_needed(self, value):
        if isinstance(value, str):
            return False
        # None is not a valid type, convert to NoneType
        valid_instance_tuple = tuple(type(it) if it is None else it for it in self.args)
        return isinstance(value, valid_instance_tuple)
```

Or just always put `str` last in the conversion? I don't like this option as it's adding more complexity for the user.
```
        if type(str) in self.args:
            args = tuple(arg for arg in self.args if arg != type(str)) + type(str)
        else:
            args = self.args
        for typ in args:
            converter = TypeConverter.converter_for(typ)
            if not converter:
                return value
            try:
                return converter.convert('', value, explicit_type)
            except ValueError:
                pass
        raise ValueError
```

pekkaklarck:
How does the current functionality break `Optional[str]`? This anyway needs an issue to discuss the design.

KotlinIsland:
`Optional[str]` becomes `Union[str, None]` making it impossible to pass `None` from robot.

KotlinIsland:
issue: #3897

pekkaklarck:
I agree `${None}` being converted to a string is not good if you use `Optional[str]`, but I wouldn't say that totally breaks it. There's also a simple workaround that you omit `Optional` and just use `arg=None`. Anyway, let's discuss this more in #3897.

pekkaklarck:
After thinking this more, I believe the approach this PR originally used is the best one. See [my comment](https://github.com/robotframework/robotframework/issues/3897#issuecomment-808336570) to issue #3897 for more details. I also submitted a separate issue about the problem with `arg: Optional[str] = None` (#3908).

src/robot/running/arguments/typeconverters.py
```diff
@@ -451,7 +451,11 @@ def _handles_value(self, value):
         return True
 
     def _no_conversion_needed(self, value):
-        return False
+        if isinstance(value, str):
+            return False
+        # None is not a valid type, convert to NoneType
+        valid_instance_tuple = tuple(type(it) if it is None else it for it in self.args)
```
pekkaklarck, line 457: This could be written like
```python
NoneType = type(None)
types = tuple(typ if typ is not None else NoneType for typ in self.args)
```
to avoid the need for the comment.

pekkaklarck:
Looks great. Docs related to Union conversion need to still be updated but I can do that myself. I'll also make sure we have tests covering the `arg: Optional[str] = None` case (#3908).

pekkaklarck:
One of the tests was actually failing. I should've checked CI status more before merging, but I was anyway planning to enhance tests a bit so I'll fix the test at the same time.

PR #3899, opened 2021-03-19
Support using pathlib.Path with markupwriters

mhwaage:
This fixes a simple omission for supporting Path objects

### DIFF
--- a/src/robot/utils/markupwriters.py
+++ b/src/robot/utils/markupwriters.py
@@ -14,7 +14,7 @@
 #  limitations under the License.
 
 from .markuputils import attribute_escape, html_escape, xml_escape
-from .robottypes import is_string
+from .robottypes import is_string, is_pathlike
 from .robotio import file_writer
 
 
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Looks ok. Could you nevertheless tell the use case for this change?

mhwaage:
e.g. when doing post-processing of execution results:
```python
from pathlib import Path

from robot.api import ExecutionResult
from robot.result.visitor import ResultVisitor

my_old_result = Path("location/to/old/report.xml")
my_postprocessed_result = Path("location/to/new/result.xml")

result = ExecutionResult(my_old_result) # Already works
result.visit(SomeResultVisitorClass())
result.save(my_postprocessed_result) # Internally creates and uses an xml-markupwriter, currently fails because a Path is not a string (workaround is to wrap the path in str)
```

pekkaklarck:
Sounds good and safe enough to be included in 4.0.1 as well. Could you please submit an issue aa well so that this change gets listed in the release notes? Creating a unit test that `save()` works with `Path` would be good too to make sure we don't accidentally break this later. I believe there are tests already for creating `ExecutionResult` using `Path`.

mhwaage:
For the unit tests, I notice that none of the tests appear to actually print the output (presumably to avoid cleanup). Since the goal is simply to test taht the save function doesn't throw, would it be acceptable to target `os.devnull`?

pekkaklarck:
Looks good to me. The new test could have possibly been included into the existing suite using `Path` but then at least the class name needs to be changed. I could to that myself after merging. Thanks for the PR!

PR #3916, opened 2021-04-01
Update LICENSE.txt

Snooz82:


pekkaklarck:
My understanding is that this LICENSE.txt file should be exactly the one got from Apache. We have separate COPYRIGHT.txt with the actual copyright information. For example, the instructions above the modified place contains this note that is pretty strange after updates:

> To apply the Apache License to your work, attach the following  boilerplate notice, with the fields enclosed by brackets "[]" replaced with your own identifying information.

I might be wrong with this so if you have documentation explaining this otherwise we certainly can change this. In that case we need to add Nokia Networks there as well, though.

pekkaklarck:
Sorry, didn't mean to close this. Annoying that "Close with comment" and "Comment" buttons are next to each others with the former being bigger.

PR #3925, opened 2021-04-07
Enable continuable failures via robot:continue-on-failure test or kw tag

oboehmer:
This PR implements one of the features in #2285 , the ability to control continuable failures via test or keyword tags.
We introduce `robot:continue-on-failure` as a test or user-keyword tag to denote all failures within the test (and within all user keywords executed within) or within a user-keyword to be continuable.


oboehmer:
Finished with the code, please review, @pekkaklarck

oboehmer:
@pekkaklarck , I understand you're now looking at 4.1 features. Would you consider this one to be part of 4.1? Please let me know what you think of the approach and the implementation. 

main discussion point is how to treat continuable keywords in such a scenario:

```
*** test Cases ***
test
    [Tags]    robot:no-continue-on-failure
    Run Keyword and Continue on Failure    Should be Equal   3   4
    Log to console     continued
```
    
I feel we should still continue, i.e. not change the semantic of continuable keywords.

pekkaklarck:
Yeah, this definitely fits well for RF 4.1. Tuesday we'll have one week from RF 4.0.3 and I believe that's enough time to see are there new critical regressions. That means we can start RF 4.1 work for real. I have a training course on Tuesday, though, so most likely it'll be Wednesday before I really get started.

src/robot/running/userkeywordrunner.py
```diff
@@ -144,6 +144,19 @@ def _format_trace_log_args_message(self, args, variables):
...
+        # (unless robot:no-continue-on-failure is set in keyword)
+        try:
+            tc_tags = context.test.tags
+        except AttributeError:
+            return kw_tags
+        result = list(kw_tags)
+        if ('robot:continue-on-failure' in tc_tags
+           and 'robot:no-continue-on-failure' not in kw_tags):
+            result.append('robot:continue-on-failure')
+        return result
```
pekkaklarck, line 158: Do I get it right that the idea of this function is to get the default value for the continuation functionality from the test case? That sounds good, but shouldn't we also look at possible parent user keywords? I just looked at the issue #2285 and it seems that we haven't yet agreed on [this part of the design](https://github.com/robotframework/robotframework/issues/2285#issuecomment-472770743). Should we discuss that further on Slack?

If we want to look at the parent keywords, we have a problem they may not be easily accessible at the moment. It ought to be pretty easy to change the current context object to preserve them, though.

This function returns tags as a list, not as a `Tags` object, and thus `tag in tags` check isn't anymore case and space insensitive. Instead of returning tags in any container, perhaps this method should just return `True` or `False` depending on is continue-on-failure enabled or not? That would then require changes everywhere else as well, but the logic would probably get simpler. Case and space insensitivity with `robot:continue-on-failure` should also be tested.

Perhaps we should actually move checking is continue-on-failure enabled fully to the context object? It already has access to the current test and, if needed, could also get access to parent user keywords. It could thus just have a property like `continue_on_failure` or `continue_on_failure_mode` that returns `True/False`. Then we could simply pass that value to `ExecutionStatus.can_continue` instead of tags.

src/robot/running/userkeywordrunner.py
```diff
@@ -152,7 +165,9 @@ def _execute(self, context):
             return None, None
         error = return_ = pass_ = None
         try:
-            BodyRunner(context).run(handler.body)
+            BodyRunner(context,
+                       tags=self._inherit_tc_continue_on_failure(handler.tags, context)
+                       ).run(handler.body)
```
pekkaklarck, line 170: See my comment related to similar code in `suiterunner.py`.

pekkaklarck (changes requested):
I only looked at code at this point and it mostly looks pretty good. For things that should still be changed see individual comments. 

The main problem now is that we apparently haven't yet agreed on how to handle nested `robot:continue-on-failure` usage (see [my comment](https://github.com/robotframework/robotframework/issues/2285#issuecomment-472770743) to issue #2285). The current implementation gets the default value from the test case, but doesn't look at possible parent user keywords. I think that's a bit odd and we either should either look at all parents or none. This might be best to discuss further on Slack and once we have agreed on the design then document it here and in comments of #2285.

src/robot/errors.py
```diff
@@ -151,6 +151,11 @@ def can_continue(self, teardown=False, templated=False, dry_run=False):
             return False
         if teardown:
             return True
+        if 'robot:no-continue-on-failure' in tags:
+            # continuable keywords can still continue
+            return self.continue_on_failure
+        if 'robot:continue-on-failure' in tags:
+            return True
```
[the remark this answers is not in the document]
oboehmer: without the inheritance, I no longer see the need for `robot:no-continue-on-failure`, wouldn't you agree? then the above wouldn't apply anymore..

src/robot/running/bodyrunner.py
```diff
@@ -30,10 +30,11 @@
 
 class BodyRunner(object):
 
-    def __init__(self, context, run=True, templated=False):
+    def __init__(self, context, run=True, templated=False, tags=None):
         self._context = context
         self._run = run
         self._templated = templated
+        self._tags = tags or []
```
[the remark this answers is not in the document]
oboehmer: ack, removed

src/robot/running/bodyrunner.py
```diff
@@ -136,16 +138,17 @@ def ForRunner(context, flavor='IN', run=True, templated=False):
...
     runner = runners[flavor or 'IN']
-    return runner(context, run, templated)
+    return runner(context, run, templated, context.test.tags)
 
 
 class ForInRunner(object):
     flavor = 'IN'
 
-    def __init__(self, context, run=True, templated=False):
+    def __init__(self, context, run=True, templated=False, tags=None):
```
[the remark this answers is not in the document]
oboehmer: I feel ForInRunner needs to honour continue-on-failure tags, which is actually one common use case (at least based on what I see our users do). I also added it to IfRunner now as well, even if this is not such a common case, but needed for consistency.

src/robot/running/suiterunner.py
```diff
@@ -142,7 +142,10 @@ def visit_test(self, test):
         self._run_setup(test.setup, status, result)
         try:
             if not status.failed:
-                BodyRunner(self._context, templated=bool(test.template)).run(test.body)
+                BodyRunner(self._context,
+                           templated=bool(test.template),
+                           tags=test.tags
+                           ).run(test.body)
```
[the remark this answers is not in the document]
oboehmer: ack, fixed

src/robot/running/userkeywordrunner.py
```diff
@@ -144,6 +144,19 @@ def _format_trace_log_args_message(self, args, variables):
         args = ['%s=%s' % (name, prepr(variables[name])) for name in args]
         return 'Arguments: [ %s ]' % ' | '.join(args)
 
+    def _inherit_tc_continue_on_failure(self, kw_tags, context):
+        # apply test-case's continue-on-failure on keywords
+        # (unless robot:no-continue-on-failure is set in keyword)
+        try:
+            tc_tags = context.test.tags
+        except AttributeError:
+            return kw_tags
```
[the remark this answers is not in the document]
oboehmer: removed this code.. but it was indeed to catch some scenarious I encountered where context.test or context.test.tags were not set, so just caught AttributeError to address both cases..

src/robot/running/userkeywordrunner.py
```diff
@@ -144,6 +144,19 @@ def _format_trace_log_args_message(self, args, variables):
...
+        result = list(kw_tags)
+        if ('robot:continue-on-failure' in tc_tags
+           and 'robot:no-continue-on-failure' not in kw_tags):
```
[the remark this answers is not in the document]
oboehmer: removed this code..

src/robot/running/userkeywordrunner.py
oboehmer: removed this code

src/robot/running/bodyrunner.py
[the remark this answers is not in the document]
oboehmer: done, just pushed a WIP commit which works, but still requires a bit polishing.. would appreciate overall feedback..

src/robot/model/tagstatistics.py
```diff
@@ -71,6 +71,13 @@ def _add_to_combined_statistics(self, test):
             if stat.match(test.tags):
                 stat.add_test(test)
 
+    def _suppress_reserved(self, tag):
+        # don't suppress reserved tags if the user explicitly included them
+        if tag.startswith('robot:') and not (self._included and self._included.match(tag)):
+            return True
+        else:
+            return False
```
pekkaklarck, line 79: This can be simplified to just
```python
return tag.startswith('robot:') and not self._included.match(tag)
```
Not 100% sure can that `self._included and` part be removed but I think it can.

src/robot/running/context.py
```diff
@@ -124,6 +126,16 @@ def in_teardown(self):
...
 
+    @property
+    def continue_on_failure(self):
+        kw_or_test = self.user_keywords[-1] if self.user_keywords else self.test
+        if not kw_or_test:
+            return False
+        if 'robot:continue-on-failure' in kw_or_test.tags:
+            return True
+        return self.test and 'robot:continue-on-failure-recursive' in self.test.tags \
+            or any([kw for kw in self.user_keywords if 'robot:continue-on-failure-recursive' in kw.tags])
```
pekkaklarck, line 137: This could be simplified a bit. Something like this (untested):
```python
parents = ([self.test] if self.test or []) + self.user_keywords
if not parents:
    return False
if 'robot:continue-on-failure' in parents[-1].tags:
    return True
return any('robot:continue-on-failure-recursive' in p.tags for p in parents)
```

doc/userguide/src/ExecutingTestCases/TestExecution.rst
```diff
@@ -364,6 +364,85 @@ converting any failure into a continuable failure. These failures are
 handled by the framework exactly the same way as continuable failures
 originating from library keywords.
 
+Controlling Continue on Failure Using Reserved Tags
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
pekkaklarck, line 368: We haven't used title case with other titles so this should also be spelled like "Controlling continue on failure ...".

As a non-native English speaker I'm not sure is "continue on failure" better than "continue-on-failure" but for some reason the latter "feels" better for me. Should check what style has been used elsewhere and use one style consistently (which can mean updating other docs).

doc/userguide/src/ExecutingTestCases/TestExecution.rst
```diff
@@ -364,6 +364,85 @@ converting any failure into a continuable failure. These failures are
...
+
+All Keywords executed as part of test cases or a user keywords which are
```
pekkaklarck, line 370: "All Keywords" -> "All keywords"

doc/userguide/src/ExecutingTestCases/TestExecution.rst
```diff
@@ -364,6 +364,85 @@ converting any failure into a continuable failure. These failures are
...
+tagged with the reserved tag `robot:continue-on-failure` are considered continuable
+by default.
+
+Thus, the following two test cases `Test 1` and `Test 2` behave identically:
```
pekkaklarck, line 374: We've used `name` role with keyword and test names. Should thus change
```
`Test 1`
```
to
```
:name:`Test 1`
```

doc/userguide/src/ExecutingTestCases/TestExecution.rst
```diff
@@ -364,6 +364,85 @@ converting any failure into a continuable failure. These failures are
...
+       Should be Equal   3   4
+       Log   log from keyword 1
+       User Keyword 2
+
+   User Keyword 2
+       Should be Equal   5   6
+       Log   log from keyword 2
+
+
+The `robot:continue-on-failure*` tags are new in Robot Framework 4.1.
```
pekkaklarck, line 444: Perhaps list both tags explicitly here instead of using `*` as a wild card.

atest/robot/running/continue_on_failure_tag.robot
```diff
@@ -0,0 +1,70 @@
+*** Settings ***
+Suite Setup       Run Tests    ${EMPTY}    running/continue_on_failure_tag.robot
+Resource          atest_resource.robot
+
+*** Test Cases ***
+Continue in test with tag
+    ${tc}=    Check Test Case    ${TESTNAME}
```
pekkaklarck, line 7: `${tc}=    ` part is unnecessary when no further validation is done for these tests.

atest/robot/tags/tag_stat_include_and_exclude.robot
```diff
@@ -5,6 +5,7 @@ Resource          atest_resource.robot
 *** Variables ***
 ${DATA SOURCE}    tags/include_and_exclude.robot
 ${F}              force
+${INT}            robot:foobar
```
pekkaklarck, line 8: `${INT}` is a bit strange variable name because it sounds like `int` i.e. integer. It seems to be used as an example of an internal tag so perhaps `${INTERNAL}` would be better.

The value could also be changed to `robot:just-an-example` to make it a bit more explicit as well. I first though this was an example of an invalid value but it seems to be just an example.

atest/robot/tags/tag_stat_include_and_exclude_with_rebot.robot
```diff
@@ -8,6 +8,7 @@ Resource          rebot_resource.robot
 ${DATA SOURCE}    tags/include_and_exclude.robot
 ${INPUT FILE}     %{TEMPDIR}${/}robot-test-tagstat.xml
 ${F}              force
+${INT}            robot:foobar
```
pekkaklarck, line 11: Same comments here as with `${INT}` and `robot:foobar` above.

atest/testdata/running/continue_on_failure_tag.robot
```diff
@@ -0,0 +1,265 @@
...
+    FOR    ${val}    IN    1    2    3
+        Set Tags   robot:continue-on-failure
+        Fail   loop-${val}
+    END
+
+Continue in for loop without tag
+    [Documentation]    FAIL loop-1
+    FOR    ${val}    IN    1    2    3
+        Fail   loop-${val}
+    END
```
pekkaklarck, line 83: This seems to be testing the normal behavior that ought to be tested thoroughly elsewhere. Perhaps having an example here is fine, but at least the name should be changed to something like "No continue ...".

atest/testdata/running/continue_on_failure_tag.robot
```diff
@@ -0,0 +1,265 @@
...
+    ...    1) kw1a\n\n
+    ...    2) kw1b\n\n
+    ...    3) kw1a\n\n
+    ...    4) kw1b\n\n
+    ...    5) This should be executed
+    [Tags]   robot:continue-on-failure
+    Failure in user keyword with tag     run_kw=Failure in user keyword with tag
+    Fail   This should be executed
+
+Continue in for loop with tag
```
pekkaklarck, line 59: "for loop" could be changed to "FOR" in test case names to be consistent with how "IF" is used with other tests.

atest/testdata/running/continue_on_failure_tag.robot
```diff
@@ -0,0 +1,265 @@
...
+    [Documentation]    FAIL kw-loop-1
+    For Loop in in user keyword without tag
+
+Continue in IF with tag
+    [Documentation]    FAIL ${HEADER}\n\n
+    ...    1) 1\n\n
+    ...    2) 2\n\n
+    ...    3) 3\n\n
+    ...    4) 4
+
```
pekkaklarck, line 102: Unnecessary empty line.

atest/testdata/running/continue_on_failure_tag.robot
```diff
@@ -0,0 +1,265 @@
...
+        Fail    3
+        Fail    4
+    END
+
+Continue in IF with set and remove tag
+    [Documentation]    FAIL ${HEADER}\n\n
+    ...    1) 1\n\n
+    ...    2) 2\n\n
+    ...    3) 3
+
```
pekkaklarck, line 120: Unnecessary empty line.

atest/testdata/running/continue_on_failure_tag.robot
```diff
@@ -0,0 +1,265 @@
...
+    END
+    Remove Tags   robot:continue-on-failure
+    IF   1==2
+        No Operation
+    ELSE
+        Fail    3
+        Fail    this is not executed
+    END
+
+Continue in IF without tag
```
pekkaklarck, line 134: Like similar test with FOR, this could be removed (ought to be tested elsewhere) or possibly renamed.

atest/testdata/tags/include_and_exclude.robot
```diff
@@ -1,5 +1,5 @@
 *** Settings ***
-Force Tags      force
+Force Tags      force    robot:foobar
```
pekkaklarck, line 2: Not sure where this is used, but `robot:just-an-example` would be better value.

atest/testdata/running/continue_on_failure_tag.robot
```diff
@@ -0,0 +1,265 @@
...
+Recursive Continue in test without tag and two nested UK without and with recursive tag
+    [Documentation]    FAIL kw2a
+    Failure in user keyword without tag     run_kw=Failure in user keyword with recursive tag
+    Fail   This should not be executed
+
+*** Keywords ***
+
+Failure in user keyword with tag
+    [Tags]   robot:continue-on-failure
+    [Arguments]    ${run_kw}=No Operation
```
pekkaklarck, line 211: We've typically had `[Arguments]` as the first setting.

atest/testdata/running/continue_on_failure_tag.robot
```diff
@@ -0,0 +1,265 @@
...
+    ...    1) kw1a\n\n
+    ...    2) kw1b\n\n
+    ...    3) kw2a\n\n
+    ...    4) kw2b\n\n
+    ...    5) This should be executed
+    [Tags]   robot:continue-on-failure-recursive
+    Failure in user keyword with tag     run_kw=Failure in user keyword without tag
+    Fail   This should be executed
+
+Recursive Continue in test without tag and two nested UK with and without recursive tag
```
pekkaklarck, line 193: Starting the name with "Recursive continue in test" is a bit strange when the test itself doesn't specify anything about continue-on-failure. Something like "Recursive continue in user keyword" ought to be clearer.

atest/testdata/running/continue_on_failure_tag.robot
```diff
@@ -0,0 +1,265 @@
...
+    [Documentation]    FAIL ${HEADER}\n\n
+    ...    1) kw1a\n\n
+    ...    2) kw1b\n\n
+    ...    3) kw2a\n\n
+    ...    4) kw2b
+    Failure in user keyword with recursive tag     run_kw=Failure in user keyword without tag
+    Fail   This should not be executed
+
+Recursive Continue in test without tag and two nested UK without and with recursive tag
```
pekkaklarck, line 202: This name is a bit strange for same reason as the above. Something like "Recursive continue in nested user keyword" could be fine. You probably have a better idea what exactly these two tests are testing and can come up with even better names.

atest/testdata/running/continue_on_failure_tag.robot
```diff
@@ -0,0 +1,265 @@
...
+    If in user keyword without tag
+
+Run Keywords with tag
+    [Documentation]    FAIL ${HEADER}\n\n
+    ...    1) 1\n\n
+    ...    2) 2
+    [Tags]   robot:continue-on-failure
+    Run Keywords    Fail   1   AND   Fail   2
+
+Recursive Continue in test with tag and two nested UK without tag
```
pekkaklarck, line 160: "Recursive Continue" -> "Recursive continue" here and in other names

pekkaklarck (changes requested):
Awesome work! I added several comments about various smallish things that can still be enhanced but this now looks really good.

src/robot/model/tagstatistics.py
oboehmer: I am not sure either.. I took this pattern from 
```
    def _is_included(self, tag):
        if self._included and not self._included.match(tag):
            return False
        return not self._excluded.match(tag)
```
and when I change the above to 
```
        if not self._included.match(tag):
            return False
```
some acceptance tests fail (have not checked why)

I can simplify my code as suggested by you (and a _doh!_ for the stupid if), the tests will pass..  I'll change as suggested..

src/robot/running/context.py
oboehmer: I didn't like `parents` as a var name as the list includes the current context too, but I am fine if you are fine.. thanks for the suggestion, much nicer!

doc/userguide/src/ExecutingTestCases/TestExecution.rst
oboehmer: We do call it `Continue on failure` mode, including a section title.. I checked with a native english speaker, and he also recommend to stick without dashes. so would suggest we keep it as is.

atest/testdata/running/continue_on_failure_tag.robot
```diff
@@ -199,16 +197,16 @@ Recursive Continue in test without tag and two nested UK with and without recurs
     Failure in user keyword with recursive tag     run_kw=Failure in user keyword without tag
     Fail   This should not be executed
 
-Recursive Continue in test without tag and two nested UK without and with recursive tag
+No recursive continue in user keyword
```
pekkaklarck, line 200: Is this name accurate? Doesn't the test validate recursive continue works also with nested keywords?

pekkaklarck (approved):

Issue #3799, opened 2020-12-14
Dedicated search functionality in log.html


mihaiparvu:
Searched a bit on the issue tracker but didn't find a similar request. A colleague in my organization who wanted to start using Robot Framework asked me if there is the possibility to search the logs with regular expressions. This question immediately triggered an enhancement idea for Robot Framework, because there were a few times in which I would have found it useful to have a dedicated search functionality directly in the log.html file. The search box provided by the web browser is pretty limited in this regard as it only searches elements that are already displayed in the page. For example, if a test case is collapsed all the containing keywords are hidden and not available for search.

Having to search a huge file for a specific keyword call (that may have failed or not) can be sometimes really tedious and a dedicated search box in the bottom of the page would help a lot. I know that we already have a search button on the keyword documentation html file that helps finding keywords faster, so I think it should fit logs well.

If this is considered a valid enhancement request, I am also interested in trying to implement it.

pekkaklarck:
I agree this would be handy. Probably wouldn't be too complicated to implement but coming up with a nice UI requires some thinking. Anyone interested to prototype this?

stdedos:
Recap (complete quote) from the duplicate ^^^:

> I would like to be able to search for a keyword (that I know it has executed) on an execution, but I don't want to find out under which keyword it is hiding.
> 
> More ideas. Search box with:
> 
> - Search only keyword names
> - Search on Fully Qualified Keyword Names
> - Search on visible log level
> - Search on all log levels

Additionally, it would be nice if the anchor-quote did not change as part of said search; sometime it is troubling to me that an existing anchor in the URL would block from the failed keywords being expanded.

anhtuta:
Can we merge this PR? It's very difficult for us to search for anything, we have to expand all steps in order to search! That's so annoying

PR #3933, opened 2021-04-12
Search functionality in log files

mihaiparvu:
I created a prototype search in log.html for #3799.
The interface was inspired from the old search function in libdoc and it looks like this:

Capture0

Capture

In the search dialog, all the results are links to the keywords in the log file and inside the log the results are highlighted in orange.

A limitation I've encountered is that not all keywords are available from the start in DOM and are not accessible to search through them. A proposal I have is to add a checkbox to `Expand all keywords and search`, but maybe there is a better way to do it that I cannot see at the moment.

@pekkaklarck any comments and suggestions from you are welcomed :)

### DIFF
--- a/src/robot/htmldata/rebot/log.css
+++ b/src/robot/htmldata/rebot/log.css
@@ -181,3 +181,64 @@
 #total-stats tr:hover, #tag-stats tr:hover {
     cursor: default;
 }
+
+.highlighted {
+    background: orange;
+}
+
+#search, #open-search {
+    position: fixed;
+    bottom: 5px;
+    right: 5px;
+    z-index: 1000;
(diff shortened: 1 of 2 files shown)


LucianCrainic:
Hi @mihaiparvu does it only search for keywords ? Since it would be great to search the whole contents of the log.html. I will make an example:

let's say we have a keyword that returns a dictionary data and we make a comparison with an expected dictionary. If they do not match the keyword would be repeated for an amout of time (eg. 2 minutes) untill it fails if they do not match. Searching through all the repeated keywords and looking at the dictionary that i am looking for takes a lot of time, while with a search function it would save a lot of time.

Issue #2717, opened 2017-11-17
`Dictionaries Should Be Equal` should support ignoring keys


bbpatel2001:
Problem:  
Current implementation of "Dictionaries Should Be Equal" returns true  if two dictionaries  are  identical. Sometimes we need to compare two dictionaries by ignoring some keys. i.e.  If we have large dictionaries and one dictionary does not have some keys at nth level but still keyword should return True for all other keys if matches . Take example of following two dictionaries.  expected_dict is my expected dictionary and Actual dictionary is created from UI data.  Actual dictionary does not contain   "Forged Source" key.  If key is 1st or 2nd level we can pop key and compare it but if we want to ignore any key which is at the nth level it is very difficult. 

```
expected_dict = {
  "alert_main_list": {
  "test":"test1",
  "IP": {
      "1.1.1.16": {
        "#_of_CLUSTERS": "1",
        "#_of_FINGERPRINTS": "1",
        "IP_ADDRESS": "1.1.1.1",
        "HIGHEST_CONFIDENCE": "49",
        "EVENT_COUNT": "1",
        "abcd" : ['I','Love','Robot', 'Framework'],
        "numbers" : [ 1, 2, 3, 4, 5,"aa"]
      }
    },
  },
  "User": {
      "user@test.com": {
        "ALERT_TRIGGERS": {
          "Suspicious Behavior": {
            "Username is an email address": "1"
          },
          "Suspicious Source": {
            "Forum Spam Blacklist": "1"
          },
          "Forged Source": {
            "Web Browser : Watson Lookup": "1"
          }
        },        
      }
    },
}
Actual_dict = {
  "alert_main_list": {
  "test":"test1",
  "IP": {
      "1.1.1.16": {
        "#_of_CLUSTERS": "1",
        "#_of_FINGERPRINTS": "1",
        "IP_ADDRESS": "1.1.1.1",
        "HIGHEST_CONFIDENCE": "49",
        "EVENT_COUNT": "1",
        "abcd" : ['I','Love','Robot', 'Framework'],
        "numbers" : [ 1, 2, 3, 4, 5,"aa"]
      }
    },
  },
  "User": {
      "user@test.com": {
        "ALERT_TRIGGERS": {
          "Suspicious Behavior": {
            "Username is an email address": "1"
          },
          "Suspicious Source": {
            "Forum Spam Blacklist": "1"
          },
        },        
      }
    },
}
```

I am proposing
[...]

PR #3942, opened 2021-04-21
`Dictionaries Should Be Equal` added support for ignoring keys 

yuriverweij:
Issue #2717 
Added additional optional parameter to provide a list of keys to ignore in the comparison 

- [x] updated keyword documentation
- [x] added acceptance tests


src/robot/libraries/Collections.py
```diff
@@ -737,19 +737,28 @@ def dictionary_should_not_contain_value(self, dictionary, value, msg=None):
...
+    def dictionaries_should_be_equal(self, dict1, dict2, msg=None, values=True, ignore_keys=None):
         """Fails if the given dictionaries are not equal.
 
         First the equality of dictionaries' keys is checked and after that all
         the key value pairs. If there are differences between the values, those
         are listed in the error message. The types of the dictionaries do not
         need to be same.
 
+        ``ignore_keys`` can be used to provide a list of keys to ignore in the
+        comparison.
```
pekkaklarck, line 749: Need a note when this option has been added.

src/robot/libraries/Collections.py
```diff
@@ -737,19 +737,28 @@ def dictionary_should_not_contain_value(self, dictionary, value, msg=None):
...
         See `Lists Should Be Equal` for more information about configuring
         the error message with ``msg`` and ``values`` arguments.
         """
+
+        if ignore_keys is None:
+            ignore_keys = []
         self._validate_dictionary(dict1)
         self._validate_dictionary(dict2, 2)
+        dict1 = {k: v for k, v in dict1.items() if k not in ignore_keys}
+        dict2 = {k: v for k, v in dict2.items() if k not in ignore_keys}
```
pekkaklarck, line 760: This should be done only when `ignore_keys` is set. Now dictionaries are always recreated which wastes resources and in with custom dict subsclasses can also change semantics.

src/robot/libraries/Collections.py
```diff
@@ -737,19 +737,28 @@ def dictionary_should_not_contain_value(self, dictionary, value, msg=None):
...
         are listed in the error message. The types of the dictionaries do not
         need to be same.
 
+        ``ignore_keys`` can be used to provide a list of keys to ignore in the
+        comparison.
+
         See `Lists Should Be Equal` for more information about configuring
         the error message with ``msg`` and ``values`` arguments.
         """
+
```
pekkaklarck, line 754: We don't generally use empty lines inside functions or methods.

pekkaklarck (approved):
First of all, sorry for a very late review. I had missed this PR earlier and now noticed it when finalizing RF 6.1. This looks so good that I'll merge it and fix the small issues mentioned in separate comments.

I also plan to change this so that the list of ignored keys can be given as a Python list literal. That would be easiest to implement by having `list` as a type hint for `ignore_keys`, but because this library doesn't use type hints otherwise (it should!) I'll probably implement it using `ast.literal_eval` separately.

yuriverweij:
No problem. Thanks for the feedback!

Issue #3884, opened 2021-03-13
BuiltIn: Support collapsing whitespaces with `Should Be Equal` and other comparison keywords


vokiput:
Based on requirements in #3240

collapse_spaces
- False - do not collapse whitespaces
- True - Collapse consecutive whitespaces into one space

pekkaklarck:
Sounds good. Should be added to the same keywords that got `strip_spaces` in RF 4.0 as part of #3240. I'll add this to RF 4.1 scope but it still needs someone to implement it.

vokiput:
Ok, I can do it. Should be trivial after strip_spaces

pekkaklarck:
Great!

vokiput:
Ready for review

pekkaklarck:
Commented the great PR!

pekkaklarck:
PR #3949 merged after updates. Thanks for contribution @vokiput!

PR #3949, opened 2021-04-26
Feature/support collapsing whitespaces

vokiput:
Implements code for #3884


pekkaklarck (changes requested):
Looks great! The only problem is that keywords should get a not that this is a new feature in RF 4.1. Could you please still add them?

I noticed that the PR splits several long list comprehensions and generator expressions in related code to multiple lines. That's nice in general as there are some overly long lines, but it is perhaps done a bit too aggressively. We have 88 character max line length in Python code (inspired by [Black](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length)) and many expressions would fit into it. Too short lines are better that too long lines so you can leave them as-is, but merging not-too-long lines back together is fine too.

vokiput:
1. added notes that this is a new feature in RF 4.1
2. Restored long list comprehensions and generator expressions

src/robot/libraries/BuiltIn.py
```diff
@@ -636,7 +636,7 @@ def should_be_equal(self, first, second, msg=None, values=True,
         | Should Be Equal | ${x} | expected | ignore_case=True | formatter=repr |
 
         ``formatter`` is new in Robot Framework 3.1.2 and ``strip_spaces`` is new
```
pekkaklarck, line 638: "and" should be replaced by a comma here, but we could also remove the reference to old RF 3.1.2 version and the whole problems goes away. I'll do that myself so no changes needed.

pekkaklarck (approved):

pekkaklarck:
When looking at the code a bit more closely, I noticed that `should_not_contain_any` calls a wrong helper method. I'll fix that and also enhance tests so that the bug is caught.

I also noticed that the long list comprehensions discussed above can be shortened by enhancing `_strip_spaces` and `_collapse_spaces` helpers so that they check is the value a string or not. This makes it possible to change `[helper(x) if is_string(x) else x for x in container]` to `[helper(x) for x in container]`. I'll do this change as well.

PR #3953, opened 2021-04-29
ISSUE-3941 fix: "Rebot removes sourcename attribute from kw tag"

miktuy:
@pekkaklarck @rasjani Guys, I hope that I have fixed this problem.
image

### DIFF
--- a/atest/robot/rebot/output_file.robot
+++ b/atest/robot/rebot/output_file.robot
@@ -14,6 +14,7 @@ Generate output with Robot
     ...    misc/for_loops.robot
     ...    misc/if_else.robot
     ...    misc/warnings_and_errors.robot
+    ...    keywords/embedded_arguments.robot
     Run tests    -L TRACE    ${inputs}
     Run keyword and return    Parse output file
 
--- a/src/robot/result/xmlelementhandlers.py
+++ b/src/robot/result/xmlelementhandlers.py
@@ -131,7 +131,8 @@ class KeywordHandler(ElementHandler):
         except AttributeError:
             body = self._get_body_for_suite_level_keyword(result)
         return body.create_keyword(kwname=elem.get('name', ''),
-                                   libname=elem.get('library'))
+                                   libname=elem.get('library'),
+                                   sourcename=elem.get('sourcename'))
 
     def _get_body_for_suite_level_keyword(self, result):
         # Someone, most likely a listener, has created a `<kw>` element on suite level.


pekkaklarck:
Fix looks good but a test making sure we don't break this again would still be needed.

miktuy:
> Fix looks good but a test making sure we don't break this again would still be needed.

Tests have done

pekkaklarck:
I only see code changes in this PR. An acceptance test should be added or an existing edited to cover the change.

miktuy:
> I only see code changes in this PR. An acceptance test should be added or an existing edited to cover the change.

Sorry, my English is not good. I thought that you told about running of existed tests. 
Now I have added simple tests for checking.

miktuy:
@pekkaklarck Hi! Do you have any remarks?

pekkaklarck:
Thanks for adding tests. It would, however, be better to validate the attribute using XML tooling, not just by looking what the output file contains. You found the right suite where to add the test and it would actually be very easy to enhance the existing test there so that it validates the fix. I just verified that changing it as below fails the test locally, and I'm fairly certain your fix would then fix it. If that's the case no new tests need to be added. 

```diff
diff --git a/atest/robot/rebot/output_file.robot b/atest/robot/rebot/output_file.robot
index 512afdad7..293461f9a 100644
--- a/atest/robot/rebot/output_file.robot
+++ b/atest/robot/rebot/output_file.robot
@@ -14,6 +14,7 @@ Generate output with Robot
     ...    misc/for_loops.robot
     ...    misc/if_else.robot
     ...    misc/warnings_and_errors.robot
+    ...    keywords/embedded_arguments.robot
     Run tests    -L TRACE    ${inputs}
     Run keyword and return    Parse output file
```

miktuy:
> Thanks for adding tests. It would, however, be better to validate the attribute using XML tooling, not just by looking what the output file contains. You found the right suite where to add the test and it would actually be very easy to enhance the existing test there so that it validates the fix. I just verified that changing it as below fails the test locally, and I'm fairly certain your fix would then fix it. If that's the case no new tests need to be added.
> 
> ```diff
> diff --git a/atest/robot/rebot/output_file.robot b/atest/robot/rebot/output_file.robot
> index 512afdad7..293461f9a 100644
> --- a/atest/robot/rebot/output_file.robot
> +++ b/atest/robot/rebot/output_file.robot
> @@ -14,6 +14,7 @@ Generate output with Robot
>      ...    misc/for_loops.robot
>      ...    misc/if_else.robot
>      ...    misc/warnings_and_errors.robot
> +    ...    keywords/embedded_arguments.robot
>      Run tests    -L TRACE    ${inputs}
>      Run keyword and return    Parse output file
> ```

Thanks!!! Run test after adding existed test with `sourcename` and remove fix and catch
image

pekkaklarck:
Yeah, that was the error I got as well. I assume the error went away with the fix.

pekkaklarck:
Thanks for the PR @miktuy!

PR #3979, opened 2021-05-20
Set Test Variable doc update to clarify error conditions

chriscallan:
Simple update to clarify that "Set Test Variable" will throw a fatal exception if called outside the scope of a "Test";


pekkaklarck:
Should this kind of documentation rather go to the documentation of `Set Test Variable` itself?

chriscallan:
Added the same warning about calling Set Test Variable outside of a test scope to the Set Test Variable documentation section.

doc/userguide/src/CreatingTestData/Variables.rst
```diff
@@ -774,7 +774,8 @@ everywhere within the scope of the currently executed test case. For
 example, if you set a variable in a user keyword, it is available both
 in the test case level and also in all other user keywords used in the
 current test. Other test cases will not see variables set with this
-keyword.
+keyword. It is a fatal exception to call :name:`Set Test Variable`
+outside the scope of a test (e.g. in a Suite Setup or Teardown).
```
pekkaklarck, line 778: Two small problems here:
- In our terminology "fatal exception" means an exception stopping the whole execution and that's not what happens here.
- There is not other scope outside test than suite setup and teardown.

Better wording would thus be something like this:
> It is an error to call :name:`Set Test Variable` in suite setup or teardown.

doc/userguide/src/CreatingTestData/Variables.rst
```diff
@@ -1171,6 +1172,8 @@ Variables with the test case scope are visible in a test case and in
 all user keywords the test uses. Initially there are no variables in
 this scope, but it is possible to create them by using the BuiltIn_
 keyword :name:`Set Test Variable` anywhere in a test case.
+It is a fatal exception to call :name:`Set Test Variable` outside the
+scope of a test (e.g. in a Suite Setup or Teardown).
```
pekkaklarck, line 1176: Wording should be changed here as well. Either same style as above or something like this to avoid repetition:
> Calling :name:`Set Test Variable` in suite setup or teardown is an error.

chriscallan:
I think I've got all the right places updated and using your suggested terminology.  Let me know if I missed something.

PR #3988, opened 2021-05-28
Duplicate stdout and stderr to robot log and original location

### LINKED ISSUE #3976: Show std(out/err) when debugging
When using PyCharm to debug robot keywords written in python, robot will capture all the console output.

This is really annoying, and am forced to use `logger.console()` when in the console, but even that isn't good enough because if there's an exception in the expression it won't print anything so I'm basically forced to use:

```py
try: logger.console(expr)
except Exception as e: logger.console(e)
```

image
# 😲

KotlinIsland:
Closes #3976

### DIFF
--- a/src/robot/running/outputcapture.py
+++ b/src/robot/running/outputcapture.py
@@ -64,13 +64,13 @@ class PythonCapturer(object):
             self._original = sys.stderr
             self._set_stream = self._set_stderr
         self._stream = StringIO()
-        self._set_stream(self._stream)
+        self._set_stream(self._stream, True)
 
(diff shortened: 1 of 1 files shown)


pekkaklarck:
This PR has multiple problems:

- I don't think writing all output to the console by default is a good idea. Having a command line option to enable something like that could be considered, though. We can discuss that further in #3976.
- This feature has no tests.
- Plenty of existing tests are failing.

djdespot (approved):

pekkaklarck:
No updates since June and I'm not sure is a good idea in general. Closing.

PR #4010, opened 2021-06-10
#ISSUE-4009 try to fix problem with doc tag

miktuy:
@pekkaklarck @mkorpela Hi, guys! I try to fix this problem.
**Tests before fix:**
image
**After fix:**
image

### DIFF
--- a/atest/robot/cli/rebot/remove_keywords/all_passed_tag_and_name.robot
+++ b/atest/robot/cli/rebot/remove_keywords/all_passed_tag_and_name.robot
@@ -132,7 +132,7 @@ Errors Are Not Removed In Tag Mode
 
 *** Keywords ***
 Run Some Tests
-    Create Output With Robot    ${INPUTFILE}    ${EMPTY}    misc/pass_and_fail.robot misc/warnings_and_errors.robot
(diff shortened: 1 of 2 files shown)


pekkaklarck:
The fix looks good bug the test should rather go here:
https://github.com/robotframework/robotframework/blob/master/atest/robot/cli/rebot/remove_keywords/all_passed_tag_and_name.robot

It actually ought to be possible to just add `misc/if_else.robot` to the list of executed suites here:
https://github.com/robotframework/robotframework/blob/master/atest/robot/cli/rebot/remove_keywords/all_passed_tag_and_name.robot#L135

miktuy:
> Thanks for the PR! The fix looks good but the test should rather go here:
> https://github.com/robotframework/robotframework/blob/master/atest/robot/cli/rebot/remove_keywords/all_passed_tag_and_name.robot
> 
> It actually ought to be possible to just add `misc/if_else.robot` to the list of executed suites here:
> https://github.com/robotframework/robotframework/blob/master/atest/robot/cli/rebot/remove_keywords/all_passed_tag_and_name.robot#L135

I have done.
**Before fix**
image
**After fix**
image

pekkaklarck:
Looks good! The updated test makes sure there's no crash, but it would be good to also actually validate that `IF` works correctly. I can add a test for that separately. I also want to anyway check how this looks in the log file.

PR #4019, opened 2021-06-21
Disable collecting unit tests code coverage data for PRs

tysonite:
These data don't make much sense without collecting coverage
data also for acceptance tests. Decided to disable it.


pekkaklarck:
Thanks @tysonite!

PR #4043, opened 2021-07-21
Add pyroject.toml with mypy section and type annotations

KotlinIsland:
Now that development has started on rf 5 which is dropping support for python 2.7-3.5 adding type annotations and mypy integration is a good idea (imo).

Annotations have been added such that `mypy src/robot/api` runs successfully.

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -1,0 +1,42 @@
+name: Unit tests
+
+on:
+  push:
+    branches:
+      - main
+      - master
+    paths:
+      - '.github/workflows/**'
+      - 'src/**'
+      - 'utest/**'
+      - '!**/*.rst'
+
+
+jobs:
+  test_using_builtin_python:
+
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ 'ubuntu-latest', 'windows-latest' ]
+        python-version: [ '3.6', '3.7', '3.8', '3.9', 'pypy3' ]
+        exclude:
+          - os: windows-latest
+            python-version: 'pypy3'
+
+    runs-on: ${{ matrix.os }}
+
+    name: Python ${{ matrix.python-version }} on ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2.2.2
+        with:
+          python-version: ${{ matrix.python-version }}
+          architecture: 'x64'
+
+      - name: Run mypy
+        run: |
+          python -m pip install mypy
+          python -m mypy --install-types --non-interactive src
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,21 @@
+[tool.mypy]
+allow_redefinition = true
+show_error_codes = true
+warn_redundant_casts = true
+
+[[tool.mypy.overrides]]
+module = "robot.*"
+
+[[tool.mypy.overrides]]
+module = "robot.model.testcase"
+strict_equality = true
+warn_unused_ignores = true
+disallow_any_generics = true
+disallow_untyped_defs = true
+
+[[tool.mypy.overrides]]
+module = "robot.api.*"
+strict_equality = true
+warn_unused_ignores = true
+disallow_any_generics = true
+disallow_untyped_defs = true
--- a/src/robot/__main__.py
+++ b/src/robot/__main__.py
@@ -20,7 +20,7 @@ import sys
 # Allows running as a script. __name__ check needed with multiprocessing:
 # https://github.com/robotframework/robotframework/issues/1137
 if 'robot' not in sys.modules and __name__ == '__main__':
-    import pythonpathsetter
+    import pythonpathsetter  # type: ignore[import]
 
 from robot import run_cli
 
--- a/src/robot/api/deco.py
+++ b/src/robot/api/deco.py
@@ -15,8 +15,17 @@
 
 import inspect
 
+from typing import overload, cast, Callable, TypeVar, Union, List, Any, TYPE_CHECKING, Mapping
 
-def not_keyword(func):
+if TYPE_CHECKING:
+    from typing_extensions import Literal
+
+Fn = TypeVar("Fn", bound=Callable[..., Any])
+Cls = TypeVar("Cls", bound=type)
+_Types = Union[Mapping[str, type], List[type], None]
+
+
+def not_keyword(func: Fn) -> Fn:
     """Decorator to disable exposing functions or methods as keywords.
 
     Examples::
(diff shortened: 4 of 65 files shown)


pekkaklarck:
There are lot of APIs where adding type information helps, but I don't think adding them everywhere and making MyPy happy is worth the effort. My biggest problem is the added noise in cases like these:
- Magic comments like `# type: ignore[attr-defined]` to make MyPy happy.
- Redundant return value annotation `-> None` in general and especially with `__init__` that just cannot return anything.
- Argument information when it's generic (e.g. `Any`) or redundant (e.g. `arg: bool = True`).
- Code like `ExecutionResult(*sources: str | TextIO | bytes | Path, merge: bool = ..., rpa: Literal[True] = ...) -> Result: ...` that requires deep typing knowledge to even understand.

There's also the problem that RF 5.0 will still support Python 3.6 and we cannot use more modern typing features. For example, we couldn't use `list[int|float]` and needed to use a lot more verbose `from typing import List, Union; List[Union[int,float]]` instead. We cannot even use `from __future__ import annotations` to solve that since Python 3.6 doesn't support that either. That particular future import and how to evaluate annotations is also subject to change in future Python versions and I'd rather just avoid that thing altogether until there are some decisions related to that.

This PR also has a problem that it was started before we removed Python 2 support and there are thus a lot of conflicts.

Before spending more time for this, it would be a good idea to discuss this more either here or on Slack. As I already wrote, I don't think being MyPy compatible is worth the effort, but if you can list good concrete benefits we'd get I may be convinced otherwise. Adding type information to public APIs is definitely a good idea, but should probably be done so that there's an issue/PR for each API.

One API that would especially benefit from type information is visitor. Changing current
```python
class Visitor:
    ...
    def start_test(self, test):
        ...
 ```
to
```python
from robot.model import TestCase

class Visitor:
    ...
    def start_test(self, test: TestCase):
        ...
 ```
would mean that users of this API would get support for auto-completion. That's a concrete benefit and there's no extra noise either.

**Comment by KotlinIsland** [comment]
> Magic comments like `# type: ignore[attr-defined]`.

Type ignore comments aren't magic, they are part of [PEP 484](https://www.python.org/dev/peps/pep-0484/#toc-entry-31)

> Redundant return value annotation `-> None`
I completely agree (https://github.com/python/mypy/issues/9413), but is currently a requirement.

> Argument information when it's generic (e.g. `Any`) or redundant (e.g. `arg: bool = True`)

`Any` is not generic, it's dynamic, not sure what you mean there.

The redundant annotation when there's a default argument is questionable, but I think it's better to be explicit.

> Code like `ExecutionResult(*sources: str | TextIO | bytes | Path, merge: bool = ..., rpa: Literal[True] = ...) -> Result: ...` that requires deep typing knowledge to even understand.

Not sure what you mean here, all these features are standard typing constructs:
- `...` eludes an expression when it's not meaningful (ie a type context)
- ` | ` is the type union operator
- `Literal[True]` is a literal type: it can only be the value `True`.

> I don't think being MyPy compatible is worth the effort

Don't you think that by adding type information to the internal api it would make development / refactoring a much smoother process?

**Comment by DetachHead** [comment]
i think typing even the internal API would be highly beneficial especially for new contributors. in my experience, projects with types are so much easier to jump right in and start making changes with little to no experience with the codebase, because it allows your IDE to help out a lot when looking for a property you need or when you're trying to figure out how to call a function. another huge benefit is that the type checker informs me of other parts of the code i may have broken

**Comment by pekkaklarck** [comment]
> Don't you think that by adding static type information to the internal api it would make development / refactoring a much smoother process?

No I don't. For me most changes this PR just make the code harder to understand because there's so much extra noise. I don't like how the code looks and I'd even less like to write code like this. If I'd like static typing, I'd use Java or C#. I see benefits from type hints in selected APIs, especially in public APIs, but I really see more problems than benefits with full MyPy support.

**Comment by pekkaklarck** [comment]
> i think typing even the internal API would be highly beneficial especially for new contributors.

This is a valid point, I'm not the only one working with this code base. I'm not sure is this true with all new contributors, though. Typing in Python, especially if you want full MyPy compatibility, is somewhat advanced topic and lot of our contributors are novice programmers. I'm afraid this could be a barrier of entry for people who aren't already familiar with typing.

**Comment by KotlinIsland** [comment]
Community Ng into this project is it seems very overwhelming, with multiple classes with the same name and functions with very ambiguous types. Even trying to figure out the types for this change has been extremely confusing.

**Comment by pekkaklarck** [comment]
I'm sure we can add types to various places to make the code easier to understand and to ease navigation around. Adding more documentation and creating some kind of an architecture diagram could help as well.

It's the full MyPy compatibility that bothers me. A very good example of noise it adds without benefits is the [robot.api.deco](https://github.com/robotframework/robotframework/blob/1593a82085edc8416a9c26a2a77fbf5878cda1d1/src/robot/api/deco.py) module. The original code is pretty trivial (assuming you know how decorators work), but the added `# type: ignore` comments, `@overload`s and changing signatures like
```python
def keyword(name=None, tags=(), types=()):
```
to
```python
def keyword(name: Union[Fn, str, None] = None, tags: Union[List[str], Tuple[None, ...]] = (), types: Union[Dict[str, type], List[type], Tuple[None, ...], None] =()) -> Union[Callable[[Fn], Fn], Fn]:
```
make it a lot more complicated. As far as I can see, none of the changes in that file actually has any concrete benefits.

KotlinIsland:
Mypy is very flexible around what needs to have annotations / what is checked. Everything is entirely optional and confurable down to the individual module.

Tbh in my projects that use robot framework the `@keyword` decorator was the first thing I typed with a local stub because otherwise it erases the entire original signature.

Maybe there is a way to simplify this type annotation.

pekkaklarck:
> Tbh in my projects that use robot framework the @keyword decorator was the first thing I typed with a local stub because otherwise it erases the entire original signature.

Could you clarify this? Does the variant in this PR fix the problem?

KotlinIsland:
When a decorator isn't typed the signature of the original function becomes dynamic(basically completely uncheck d), there is a mypy rule relating to this https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_any_decorated.

```py
@untyped_decorator
def foo(a: int) -> str: ...

reveal_type(foo) # Any
```
`Any` is the dynamic type, meaning that any usage is entirely unchecked.

So when `keyword` isn't typed it disables and type checks on any function it's used on. And a naïve implementation of typing wouldn't allow the optional arts etc.

I do have an idea about how to drastically simplify this implementation but will need to get to it tomorrow.

KotlinIsland:
```py
@overload
def keyword(name: Fn) -> Fn: ...

@overload
def keyword(name: str = None, tags: List[str] = ..., types: _Types = ...) -> Callable[[Fn], Fn]: ...

@not_keyword
def keyword(name: object = None, tags: object = (), types: object = ()) -> object:
```
Simplified the defs to this

KotlinIsland:
@pekkaklarck If you think type annotations look gross they could be implemented as separate `pyi` stubs, I think sellib does it that way. Or they could be added to TypeShed.

Issue #3410, opened 2019-12-09
Add `--maxassignlength` to control how much to automatically log when assigning variables


CharterMitch:
Similar to #2576 I have a use case that requires increasing the _MAX_ASSIGN_LENGTH variable in Utils/text.py when running robot tests.

The implementation and CLI flag --maxassignlength would mimic --maxerrorlines.

--maxassignlength documentation would show the default MAX_ASSIGN_LENGTH of 200 words and allow the value to be increased when running robot.

It would also be nice to set the default behavior of format_assign_message(cut_long=True) to cut_long=False. Perhaps as a --assignlogall flag?

pekkaklarck:
Quick comments:

1. I'd prefer only increasing the value of `_MAX_ASSIGN_LENGTH` from 200 characters to something like 400. Trivial to implement and updating tests and docs shouldn't be a big task either. Could still make it to RF 3.2 if someone would create a pull request.

2. Adding `--maxassignlength` (or possibly `--maxassignmessage` or even `--maxassignmessagelength`) that works similarly as `--maxerrorlines` is possible but a lot bigger task than 1. If someone finds this important enough to create a pull request, I'd be happy to review it. I won't have time for that before RF 3.2, though.

3. I absolutely don't want to have more than one setting to control this behavior. That one option could have way to tell everything to be logged, for example, like `--maxassignlength -1`. I think `--maxerrorlines` had such a feature and this one should be consistent.

4. In general keywords should log information that is important themselves. Relaying on automatic logging when variables are created feels a bit strange for me, especially when no such logging occurs if there's no variable assignment.

wuwentao:
Hi Robot support team.

we are also want to get full log in robot log file.
For example, robot script will run in linux server, and do an tcp socket to windows pc, then execute python script in windows remote PC for GUI automation, it's out of the linux/robot/windows command line control.
this python script will return all the execute log and response to robot, and robot can check/confirm all the steps/error/exception.

without robot, we can manual run this python script, but it's a part of robot script, we want to run all these script in robot test suite.

is there any option can be add to robot script content and enable the full log?  command line is not a good solution, it's a hard code solution and need manual add action when we run all scripts at the same time. 


Thanks

MitchellMahan:
+1 for this feature. I am using Robot Framework for a lot of network validation cases where the log.html becomes an artifact.
I have had to forke the repo internally and set trim_long=False in utils/text.py but not diverging would be more ideal.

[...]

PR #4051, opened 2021-08-04
ISSUE-3410 Added `--maxassignlength` option

miktuy:
Here I try to implement #ISSUE-3410
Now it seems like
image

@pekkaklarck If this approach is Ok I will add test for this feature.


miktuy:
@pekkaklarck Hi! How about this approach? Is it ok or do you see another way for this task?

pekkaklarck:
Sorry, I had missed this PR. I have been looking at #3410 this wasn't linked to it (but after referring it here it will be) and somehow had missed this otherwise.

Why did you @miktuy close the PR? On a quick look it looks fine except that it needs tests.

Issue #3255, opened 2019-08-03
Feature Request:  Access to --include and --exclude Tags from ListenerV3/PreRunModifier


Snooz82:
Hi Peke,

I need access to the list of excluded and included tags from ListenerV3.

UseCase:
DataDriver loads data from external Source and creates TestCases. 
Tags can also be defined in this external Source and will be added to TestCases.
Because DataDriver is initialized on Start Suite the excluded tags has no effect.

I want DataDriver Listener to be able to exclude data from source that has a tag that is excluded by command line option --exclude. same for --include.

Any Possibility with robot.api or with Suite data model would be OK.

There are actual usecases that require this feature.

Thanks.

pekkaklarck:
Being able to access execution configs would definitely be useful. For reasons explained below, it is somewhat big task to implement and unfortunately not likely to happen in the near future.

A workaround is specifying important configs using variables like `--include foo --variable include:foo`. That's not too convenient if you use the standard `robot` command, but if you have a custom start-up script it can accept options itself and then call `robot` accordingly.

Reasons why this is rather big task to implement:

1. Current system for processing command line options is pretty old and bad. Most importantly, there isn't a good context object having all configs in one place, but there should be also new features like support for config files. I don't want to use the current (and bad) `RobotSettings` class in any public API because it will be changed sooner or later. Rewriting the whole config system isn't an absolutely mandatory precondition for this issue, but at least we should define the API of the object containing configs.

2. Listeners could potentially get a new `initialize` method like #722 proposed or a separate `config` method, but same approach wouldn't work with model modifiers, libraries, or in the test data. One simple possibility would be adding a new built-in variable like `${CONFIG}` (or possibly `&{CONFIG}` as a dict) that could be directly used in test data and programmatically accessed via `BuiltIn.get_variable_value()`. It probably wouldn't work with model modifiers, though, so a separate public API somewhere in `robot.api` might be needed anyway. Designing and implementing all this is some work. Variable containing configs would probably also need its own issue.

3. Robot's internal execution context handling is not very good. Most importantly, you cannot execute tests in different threads because the framework isn't thread safe. We've been slowly enhancing this situation and hopefully can introduce proper context handling sometime in the future. We definitely don't want to make this situation worse, so this needs to be taken into account if we decide to add something execution configs related under `robot.api`.

Everything above is doable but there's just quite a lot to do and also to design. I won't have time to do that myself in the near future, but can help if someone else is interested.

[...]

PR #4052, opened 2021-08-04
Access to --include and --exclude Tags from Automatic Variables (#3255)

nicobucher:
This PR adds two new automatic/built-in variables: @{INCLUDE_TAGS} and @{EXCLUDE_TAGS} containing the tags set by the respective command line options.

As requested in feature request #3255

### DIFF
--- a/atest/robot/variables/automatic_variables.robot
+++ b/atest/robot/variables/automatic_variables.robot
@@ -14,10 +14,10 @@ Test Documentation
     Should Be Equal    ${tc.doc}    My doc.\nIn 2 lines! And with variable value!!
 
 Test Tags
-    Check Test Tags    ${TEST NAME}    Force 1    Hello, world!    id-42    variable value
+    Check Test Tags    ${TEST NAME}    Force 1    Hello, world!    id-42    include this test    variable value
 
 Modifying \${TEST TAGS} does not affect actual tags test has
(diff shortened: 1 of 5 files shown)


pekkaklarck:
First of all, sorry for missing this PR earlier. The implementation looks good and it's great that it contains also tests and docs. A problem is that I'm not sure is it a good idea to expose different configuration parameters as individual built-in variables. There are lot of such options and adding new variable for each doesn't feel that good.

It would, on the other hand, be fine to have one variable like `&{SETTINGS}`that would be a dictionary containing all options we want to expose. Then, for example, `--exclude` information could be accessed like `${SETTINGS}[exclude]` and like `${SETTINGS.exclude}`.

I'm not sure would it make sense to expose all our command line options via such variable. In the beginning it could even contain only include and exclude info and others could be added later based on needs. Adding new entries to the dictionary would be so safe from backwards compatibility point of view, contrary to adding new variables, that additions could be done even in patch releases if needed.

Are you @nicobucher are still interested in this and how the above sounds for you? If you think it's a good idea, it would probably be best to submit a new issue about this enhancement because the original #3255 proposes something a bit different. Implementation of this new variable ought to be pretty much as easy as what's done in this PR already.

nicobucher:
I created a new issue describing the more general use case. #4229

Issue #4047, opened 2021-07-27
Variables in unexecuted FOR loops overwrite local variables


MoreFamed:
```
*** Test Cases ***
Test
    ${a}  Set Variable  abc
    IF  '${a}' == ''
        FOR  ${a}  IN  def  ghi
            Log  ${a}
        END
    ELSE
        Log To Console  ${a}
    END
```
This test logs ```${a}``` to the console instead of ```abc``` as I would expect. The problem is in reusing variable ```a``` in the FOR-cycle. When I use another variable instead, like ```FOR  ${x} IN ...```, then ```abc``` is written correctly. Seems as a bug to me.

RF v. 4.0 and 4.0.3, Python 3.7.4

miktuy:
@MoreFamed @pekkaklarck Guys! I think I have fixed this problem in https://github.com/robotframework/robotframework/pull/4055

miktuy:
uups. Another tests are broken((((

pekkaklarck:
The reason for the problem is this:
1. We go through keywords and control structures in unexecuted IF branches to get them listed in the log file.
2. When going through FOR loops, we assign variables used in the loop to a dummy value.
3. In the above case variable `${a}` gets overwritten by the dummy value.

Fixing this shouldn't be overly complicated. Alternatives:
- Create a new temporary variable scope for when going through unexecuted FOR loops.
- Don't set variables when going through unexecuted FOR loops. This probably has problems with the log file.
- Restore original variable values after going through unexecuted FOR loops. This is probably the easiest solution but I haven't tested it myself yet.

pekkaklarck:
Turned out creating a temporary variable context was the easiest solution to fix this.

PR #4055, opened 2021-08-08
#4047 Variable in normal syntax is not replaced with its value in some cases;

miktuy:
Improved test ``For loop inside if`` in  ``atest/testdata/running/if/complex_if.robot`` is failed before fix
image
After fix it is OK:
image

### DIFF
--- a/atest/testdata/running/if/complex_if.robot
+++ b/atest/testdata/running/if/complex_if.robot
@@ -45,14 +45,19 @@ Setting after if
 
 For loop inside if
     ${value}    Set Variable    0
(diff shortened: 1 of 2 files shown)


pekkaklarck:
This approach didn't work because the problem wasn't in IF but in FOR. We explicitly want to go through FOR loops also in unexecuted IF branches to get their keywords listed in the log file, but we just shouldn't set variables to the real execution scope when doing that. I fixed that in 60f6ca42cd3657d5854b178eec1ce091fcbe3672. Thanks for the PR anyway!

PR #4085, opened 2021-09-13
Narrow down the scope of ReST parsing.

### LINKED ISSUE #4086: All irrelevant errors are not silenced when parsing reStructuredText data
#### Bug description

When processing an ReST file that bears code blocks and `csv-table` directives, ReST parsing fails if their `:file:` paths cannot be found by `robot`.

#### Steps to reproduce

1. Paste the following to a `test.rst` file:
    ```rst
    .. code-block:: robotframework

        *** Test Cases ***
        Dummy Test
             No Operation

    .. csv-table:: will cause an error
         :file: not/a/real/path/yet.csv 
    ```
2. Run it:
    ```sh
    robot test.rst
     ```

##### Expected Result

```
Test                                                                          
Dummy Test                                                            | PASS |
Test                                                                  | PASS |
1 test, 1 passed, 0 failed
```

##### Actual Result

```
[ ERROR ] Unexpected error: SystemMessage: /home/michel/VirtualEnvs/test.rst:7: (SEVERE/4) Problems with "csv-table" directive path:
[Errno 2] No such file or directory: 'not/a/real/path.csv'.
```

hidmic:
Closes #4086. This patch decorates `docutils` functionality so as to ignore every directive and role that is not directly relevant to RobotFramework. Alternative to error suppression (through `report_level` configuration, see https://github.com/robotframework/robotframework/issues/2093).

### DIFF
--- a/atest/testdata/parsing/data_formats/rest/sample.rst
+++ b/atest/testdata/parsing/data_formats/rest/sample.rst
@@ -1,3 +1,6 @@
+..
+   When parsing ReST files, only robotframework code blocks
+   and includes need to be parsed.
 .. include:: empty.rest
 .. include:: include.rst
 
@@ -21,10 +24,13 @@ We have a devious plan to rule the world with robots.
    | Variables  | ../resources/variables.py
    | Library    | OperatingSystem | | | | | | | | | | | | | | | |
 
-The following are non-standard docutils directives, and we should ignore
-errors when parsing this.
+.. csv-table:: cannot and should not be parsed
+   :file: not/a/real/path.csv
 
-Testing also a :term:`test` as it should generate an error.
+The following are non-standard docutils directives and no errors
+should arise when parsing this.
+
+Testing also a :term:`test` as it should not generate an error.
 
 .. highlight:: robotframework
 
--- a/src/robot/utils/restreader.py
+++ b/src/robot/utils/restreader.py
@@ -16,6 +16,7 @@
 import functools
 
 from robot.errors import DataError
+from robot.utils import PY2
 
 try:
     from docutils.core import publish_doctree
@@ -72,9 +73,6 @@ def directive(*args, **kwargs):
     return directive_class, messages
 
 
-directives.directive = directive
-
-
 @functools.wraps(roles.role)
 def role(*args, **kwargs):
     role_function = role.__wrapped__(*args, **kwargs)
@@ -83,6 +81,12 @@ def role(*args, **kwargs):
     return role_function
 
 
+if PY2:
+    directive.__wrapped__ = directives.directive
+    role.__wrapped__ = roles.role
+
+
+directives.directive = directive
 roles.role = role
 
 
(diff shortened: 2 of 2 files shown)


hidmic:
To put things in context, I'm using RobotFramework for RPA and attempting to use the same ReST file to describe the automation and to layout its output in a document. Thus, not all files that'd be required to render the ReST file are present when at the time it is parsed by RobotFramework. 

This patch allows RobotFramework to pick up what it needs from the ReST and nothing else.

pekkaklarck:
Could you clarify why this change is needed? More precisely, how is this better than what was done in #2093? It might be best to submit a separate issue explaining that. We anyway want to have issues for tracking purposes.

On a quick look the code looks ok but lacks tests. If we decide to add this, I'll review the PR more closely.

hidmic:
> Could you clarify why this change is needed? More precisely, how is this better than what was done in #2093?

In the situation I described [above](https://github.com/robotframework/robotframework/pull/4085#issuecomment-918178455), #2093 won't do. Parsing a `.. csv-file::` directive with a `:file:` uri that points to a nonexistent file raises. But beyond my specific use case, with this patch there's less to process and there's no need to silently ignore any warnings or errors.

> We anyway want to have issues for tracking purposes.

Got it. I took a look at the CONTRIBUTING guidelines and it wasn't clear to me whether you always had a tracking issue or not. I'll open one ASAP.

> On a quick look the code looks ok but lacks tests. If we decide to add this, I'll review the PR more closely.

I did extend one of the acceptance tests. Would you like to have a unit test for  `read_rest_data()`? There's none at the moment.

pekkaklarck:
So this fix is more general than what we have? Sounds good to me. I missed that you had modified test data in this regard. Didn't know `csv-file` import caused problems.

pekkaklarck:
As I commented #4086, this PR needs to be Python 2 compatible if we want it in RF 4.1.2. Looking at the code the only problem is `__wrapped__` that's not set by `functools.wraps` in Python 2. That ought to be pretty easy to handle:

```python
from robot.utils import PY2

...

@functools.wraps(directives.directive)
def directive(*args, **kwargs):
    ...

@functools.wraps(roles.role)
def role(*args, **kwargs):
    ...

if PY2:
    directive.__wrapped__ = directives.directive
    role.__wrapped__ = roles.role

directives.directive = directive
roles.role = role
```

hidmic:
@pekkaklarck suggestion applied in https://github.com/robotframework/robotframework/pull/4085/commits/35393bcd23ee315152807177ecb11c62547c7a17.

hidmic:
@pekkaklarck acceptance tests for ReST format parsing pass locally for Python 2.7.18 and Python 3.8.10 using `docutils` 0.9 and `docutils` 0.17.1.

Could we fire up the workflows in this PR?

pekkaklarck:
Thanks for the PR @hidmic!

Issue #3125, opened 2019-03-18
Make resolving connectioncache alias into public api.


rasjani:
Root of this issue here: https://github.com/robotframework/SeleniumLibrary/issues/1319

SeleniumLibrary has WebDriverCache which inherits ConnectionCache from robot utils.  In case someone adds a new connection/driver by calling with `register()`, a new connection is added into cache pool and index of that connection is returned.   In scenario where user only then as "alias" but no index, a connection can be still fetched via `get_connection()` but there is no public way to fetch the index of existing connection.  

Internal implementation in ConnectionCache does however have functionality to resolve alias into index via `_resolve_alias_or_index()`.

Im suggesting that `_resolve_alias_or_index()` to be made into public api `resolve_alias_or_index()` and old one is deprecated.

PR #4108, opened 2021-10-04
Process: Return created process object

### LINKED ISSUE #4104: Process: Change `Start Process` to return created process object instead of generic handle
Currently if you want to interact with the process started using the `Start Process` keyword you need to do it like this:
```
Start Process    command
${process} =    Get Process Object
# Do something with the process
```
The extra step needed to get the process object is somewhat annoying and it would be much more convenient to just use:
```
${process} =    Start Process    command
```
The problem with this is that `Start Process` currently returns a process handle that can be used with other keywords:
```
${handle} =    Start Process    command
# Other keywords possibly starting more processes
${result} =    Wait For Process    ${handle}
```
The handle is actually just an integer, and it ought to be possible to use the started process object as a handle instead. In practice the above would still work but `${handle}` would be a process object instead of an integer. This would nevertheless be a backwards incompatible change, because someone could have used integers as handles directly:
``` 
Start Process    command
# Other keywords possibly starting more processes
${result} =    Wait For Process    1
```
The above usage hasn't been recommended, though, and aliases should be used instead:
```
Start Process    command    alias=example
# Other keywords possibly starting more processes
${result} =    Wait For Process    example
```

onurcelep:
Modifies create_process method as it returns a process object.
Iterable for process objects  goes through connection indexes
so we can directly call get_connection for generated index from
register method.

Fixes #4104

### DIFF
--- a/atest/robot/standard_libraries/process/process_library.robot
+++ b/atest/robot/standard_libraries/process/process_library.robot
@@ -23,5 +23,8 @@ Running a process in a shell
 Input things to process
     Check Test Case    ${TESTNAME}
 
+Assign process object to variable
+    Check Test Case    ${TESTNAME}
+
 Get process id
     Check Test Case    ${TESTNAME}
--- a/atest/testdata/standard_libraries/process/process_library.robot
+++ b/atest/testdata/standard_libraries/process/process_library.robot
@@ -41,6 +41,12 @@ Input things to process
     ${result}=    Wait For Process
     Should Match    ${result.stdout}    *inp 42*
 
+Assign process object to variable
+    ${process} =    Start Process  python  -c  print('Hello, world!')
+    ${result} =    Run Process  python  -c  import sys; print(sys.stdin.read().upper().strip())  stdin=${process.stdout}
+    Wait For Process    ${process}
+    Should Be Equal As Strings    ${result.stdout}  HELLO, WORLD!
+
 Get process id
     ${handle}=    Some process
     ${pid}=    Get Process Id    ${handle}
--- a/atest/testdata/standard_libraries/process/terminate_process.robot
+++ b/atest/testdata/standard_libraries/process/terminate_process.robot
@@ -96,10 +96,12 @@ Terminate all processes
     END
 
 Terminating all empties cache
+    [Documentation]    FAIL Non-existing index or alias '1'.
     Some process
+    Some process
+    Switch process    ${2}
     Terminate All Processes    kill=True
-    ${handle} =    Some Process
-    Should Be Equal    ${handle}    ${1}
+    Switch process    ${1}
 
 *** Keywords ***
 Terminate should stop countdown
--- a/src/robot/libraries/Process.py
+++ b/src/robot/libraries/Process.py
@@ -375,19 +375,32 @@ class Process(object):
         for more information about the arguments, and `Run Process` keyword
         for related examples.
 
-        Makes the started process new `active process`. Returns an identifier
-        that can be used as a handle to activate the started process if needed.
+        Makes the started process new `active process`. Returns the created 
+        process object [https://docs.python.org/3/library/subprocess.html#popen-constructor]
+        which can be assigned to a variable if needed.
+
+        Note: Returning an actual process object introduced in RF 5.0, previous versions
+        would only return a generic handler.
 
         Processes are started so that they create a new process group. This
         allows sending signals to and terminating also possible child
         processes. This is not supported on Jython.
+
+        Examples:
+        | Start Process | ${command} | alias=example |
+        | ${result} = | Wait For Process | example |
+        | ${process} = | Start Process | python | -c | print('Hello, world!') |
+        | ${result} = | Run Process | python | -c | import sys; print(sys.stdin.read().upper().strip()) | stdin=${process.stdout} |
+        | Wait For Process | ${process} |
+        | Should Be Equal | ${result.stdout} | HELLO, WORLD! |
         """
         conf = ProcessConfiguration(**configuration)
         command = conf.get_command(command, list(arguments))
         self._log_start(command, conf)
         process = subprocess.Popen(command, **conf.popen_config)
         self._results[process] = ExecutionResult(process, **conf.result_config)
-        return self._processes.register(process, alias=conf.alias)
+        self._processes.register(process, alias=conf.alias)
+        return self._processes.current
 
     def _log_start(self, command, config):
         if is_list_like(command):
--- a/src/robot/libraries/Process.py
+++ b/src/robot/libraries/Process.py
@@ -380,7 +380,7 @@ class Process(object):
         which can be assigned to a variable if needed.
 
         Note: Returning an actual process object introduced in RF 5.0, previous versions
-        would only return a generic handler.
+        would only return a generic process handle.
 
         Processes are started so that they create a new process group. This
         allows sending signals to and terminating also possible child
--- a/src/robot/utils/connectioncache.py
+++ b/src/robot/utils/connectioncache.py
@@ -138,12 +138,12 @@ class ConnectionCache(object):
         return self.current is not self._no_current
 
     def resolve_alias_or_index(self, alias_or_index):
-        for resolver in self._resolve_alias, self._resolve_index:
+        for resolver in self._resolve_alias, self._resolve_index, self._is_connection:
             try:
                 return resolver(alias_or_index)
             except ValueError:
                 pass
-        raise ValueError("Non-existing index or alias '%s'." % alias_or_index)
+        raise ValueError(f"Non-existing index or alias '{alias_or_index}'.")
 
     def _resolve_alias(self, alias):
         if is_string(alias) and alias in self._aliases:
@@ -158,6 +158,9 @@ class ConnectionCache(object):
         if not 0 < index <= len(self._connections):
             raise ValueError
         return index
+
+    def _is_connection(self, conn):
+        return self._connections.index(conn) + 1
 
 
 @py3to2


pekkaklarck:
Changing `Start Process` isn't enough, all keywords accepting `handle` need to be updated as well. For example,

```
${process} =    Start Process
${result} =    Wait For Process    ${process}
```
should work but this change alone breaks it. I'll add a comment how this can handled to the original issue.

This change would also need its own tests.

atest/testdata/standard_libraries/process/terminate_process.robot
```diff
@@ -98,8 +98,9 @@ Terminate all processes
 Terminating all empties cache
     Some process
     Terminate All Processes    kill=True
-    ${handle} =    Some Process
-    Should Be Equal    ${handle}    ${1}
+    ${process_object}=    Some Process
+    ${index}=    Get Process Object Index     ${process_object}
+    Should Be Equal    ${index}    ${1}
```
pekkaklarck, line 103: I don't think it's a good idea to add separate `Get Process Object Index` keyword for just this usage. Other alternatives to make sure the cache is empty:
1. Use `Switch Process    1` and handle resulting error by adding `[Documentation]` that specifies what error should occur. Although `Start Process` now returns Popen objects, `Switch Process` should work with indices. But not when there are no processes.
2. Use `BuiltIn.Get Library Instance` to get access to the library itself and check that `lib._processes` is empty. Accessing an attribute starting with `_` is a bit questionable but ought to be fine in tests.

src/robot/libraries/Process.py
```diff
@@ -375,19 +375,32 @@ def start_process(self, command, *arguments, **configuration):
...
         for related examples.
 
-        Makes the started process new `active process`. Returns an identifier
-        that can be used as a handle to activate the started process if needed.
+        Makes the started process new `active process`. Returns the created 
+        process object [https://docs.python.org/3/library/subprocess.html#popen-constructor]
+        which can be assigned to a variable if needed.
+
+        Note: Returning an actual process object introduced in RF 5.0, previous versions
+        would only return a generic handler.
```
pekkaklarck, line 383: "handler" -> "process handle"

src/robot/libraries/Process.py
```diff
@@ -375,19 +375,26 @@ def start_process(self, command, *arguments, **configuration):
...
 
         Processes are started so that they create a new process group. This
         allows sending signals to and terminating also possible child
         processes. This is not supported on Jython.
+
+        Examples:
+        | Start Process | ${command} | alias=example |
+        | ${result} = | Wait For Process | example |
+        | ${process} = | Start Process | python | -c | print('Hello, world!') |
+        | Should Be Equal | ${process.stdout} | Hello, world! |
```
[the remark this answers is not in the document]
pekkaklarck: Validation at the end should have `HELLO, WORLD!` instead of `Hello, world!`. Would be good to actually test the example as well!

src/robot/libraries/Process.py
```diff
@@ -667,6 +680,19 @@ def get_process_object(self, handle=None):
...
+
+        If ``process`` is not given, uses the current `active process`.
+
+        Examples:
+        | Terminate All Processes |
+        | ${process_object}=      | Start Process | ${command} |
+        | ${process_index}=       | Get Process Object Index   | ${process_object} |
+        | Should Be Equal         | ${process_index}           | ${1}
+        """
+        return self._processes.get_connection_index(process)
```
pekkaklarck, line 694: I don't think anyone needs the old process handle (i.e. index) for anything. Apparently this keyword was added to make it fix a test that was broken when `Start Process` doesn't anymore return indices, but I already commented that there are other ways to validate that the internal cache is empty.

In other words, I believe this keyword should be removed. Can be added later if someone actually needs it.

src/robot/utils/connectioncache.py
```diff
@@ -100,13 +100,28 @@ def get_connection(self, alias_or_index=None):
...
+        """
+        if popen is None:
+            if not self:
+                self.current.raise_error()
+            return self._connections.index(self.current)+1
+        try:
+            return self._connections.index(popen)+1
+        except ValueError:
+            raise ValueError(f"Provided input: '{popen}' is not a process object (popen).")
+
```
pekkaklarck, line 124: This code seems to be only used by the keyword that I believe isn't needed. Thus this code can be removed as well. If it is preserved, it should be made independent also on terminology level from Popen.

src/robot/utils/connectioncache.py
```diff
@@ -137,13 +152,14 @@ def __len__(self):
     def __bool__(self):
         return self.current is not self._no_current
 
-    def resolve_alias_or_index(self, alias_or_index):
-        for resolver in self._resolve_alias, self._resolve_index:
+
+    def _resolve_connection(self, alias_or_index):
```
pekkaklarck, line 156: This method is part of the public API (see #3125) and thus cannot be renamed. Now that we support also connections themselves, `resolve_alias_or_index` is not a totally accurate name, but renaming it would require first a deprecation period. I personally consider this name good enough and would perhaps only update documentation. If you want to change the name, let's do that as a separate PR.

src/robot/utils/connectioncache.py
```diff
@@ -137,13 +152,14 @@ def __len__(self):
...
+        for resolver in self._resolve_alias, self._resolve_index, self._is_connection:
             try:
                 return resolver(alias_or_index)
             except ValueError:
                 pass
-        raise ValueError("Non-existing index or alias '%s'." % alias_or_index)
+        raise ValueError(f"Non-existing index,alias or connection '{alias_or_index}'.")
```
pekkaklarck, line 162: Space needed after a comma (i.e. `index, alias`, not `index,alias`). I'd be actually fine considering connections as aliases which would mean we could preserve the old error message.

pekkaklarck (changes requested):
Looks good but some changes needed still. Please take a look at the individual comments for more information.

pekkaklarck:
You mentioned that one acceptance test fails when you run tests locally and CI confirms that:
```
FAIL: Robot.Libdoc.Python Library.KwArgs and VarArgs
Process Should Be Stopped != Run Process
```
The reason is that this particular tests generates documentation for the Process library using Libdoc and validates that `Run Process` has expected argument information. Now that you have added a new keyword to the library, finding right keyword by an index fails. This could be fixed by incrementing the index, but I believe that this new keyword is unnecessary and it should be removed instead.

src/robot/utils/connectioncache.py
onurcelep: reverted to old error msg

src/robot/utils/connectioncache.py
onurcelep: reverted method name change , we can figure that in another PR

src/robot/utils/connectioncache.py
onurcelep: removed index getter method

src/robot/libraries/Process.py
onurcelep: got rid of `get_connection_index` method

src/robot/libraries/Process.py
onurcelep: fixed typo

atest/testdata/standard_libraries/process/process_library.robot
```diff
@@ -41,6 +41,12 @@ Input things to process
     ${result}=    Wait For Process
     Should Match    ${result.stdout}    *inp 42*
 
+Assign process object to variable
+    ${process} =  Start Process  python  -c  print('Hello, world!')
+    ${result} =  Run Process  python  -c  import sys; print(sys.stdin.read().upper().strip())  stdin=${process.stdout}
+    Wait For Process  ${process}
+    Should Be Equal As Strings  ${result.stdout}  HELLO, WORLD!
```
pekkaklarck, line 48: Our convention is to use four spaces (or more if needed) between arguments and keywords.

atest/robot/standard_libraries/process/process_library.robot
```diff
@@ -25,3 +25,6 @@ Input things to process
 
 Get process id
     Check Test Case    ${TESTNAME}
+
+Assign process object to variable
+    Check Test Case    ${TESTNAME}
```
pekkaklarck, line 30: Newline missing at the end of file. Good idea to configure your IDE to add them automatically.

This test should actually be before Get process id to keep them in same order as in the test data side.

atest/testdata/standard_libraries/process/terminate_process.robot
```diff
@@ -96,10 +96,11 @@ Terminate all processes
     END
 
 Terminating all empties cache
+    [Documentation]    FAIL Non-existing index or alias '1'.
     Some process
+    Switch process  ${1}
     Terminate All Processes    kill=True
-    ${handle} =    Some Process
-    Should Be Equal    ${handle}    ${1}
+    Switch process  ${1}
```
pekkaklarck, line 103: Good idea to validate that `Switch process    1` (`1` should work in addition to `${1}`) works before cache is cleared, but now it's possible that the test fails already there and test status validation doesn't detect that. To make sure it fails only at the end, we'd either need to validate that on the `atest/robot` side or change the test a bit. The latter is probably easier. Could, for example, start two processes and then use `Switch process    2` before clearing the cache.

pekkaklarck (approved):
Looks great! You still may want to take a look at my final and pretty nitpickey comments, but I consider this ready to be merged also without them.

atest/testdata/standard_libraries/process/terminate_process.robot
onurcelep: updated the test case as it creates 2 processes then clears cache

atest/robot/standard_libraries/process/process_library.robot
onurcelep: moved the test case

atest/testdata/standard_libraries/process/process_library.robot
onurcelep: fixed

pekkaklarck:
Looks great! Awesome how little changes were needed in the end.

PR #4111, opened 2021-10-05
No _start/_end_keyword when in FOR loop

Harri:
This PR aims to fix missing html output when using start_keyword or end_keyword listener and failure is in FOR loop.

Given following test suite:

    *** Test Cases ***
    List
        ${list}    Create List    foo    bar
        FOR    ${element}    IN    @{list}
            Log    ${element}
            Fail    Error
        END

    *** Keywords ***
    Log Error Situation
        Log    Error Situation
        
and following listener.py

    from robot.libraries.BuiltIn import BuiltIn

    ROBOT_LISTENER_API_VERSION = 2

    def end_keyword(name, attributes):
        if attributes['status'] == 'FAIL':
            BuiltIn().run_keyword('Log Error Situation')

When executing `robot --listener .\listener.py .` We get this output

    > robot --listener .\listener.py .
    Rf-For
    Rf-For.Test
    List                                                                  | FAIL |
    Error
    Rf-For.Test                                                           | FAIL |
    1 test, 0 passed, 1 failed
    Rf-For                                                                | FAIL |
    1 test, 0 passed, 1 failed
    Output:  C:\Users\hpaavola\Development\sandbox\rf-for\output.xml
    [ ERROR ] Reading XML source 'C:\Users\hpaavola\Development\sandbox\rf-for\output.xml' failed: Incompatible child element 'kw' for 'for'.

    Try --help for usage information.
    
Using Robot Framework 4.1.1 After this fix, the output is as expected; html files are created and listener is working as expected.

### DIFF
--- a/src/robot/output/listeners.py
+++ b/src/robot/output/listeners.py
@@ -46,11 +46,11 @@ class Listeners(object):
         self._is_logged.set_level(level)
 
     def start_keyword(self, kw):
-        if kw.type != kw.IF_ELSE_ROOT:
+        if kw.type != kw.IF_ELSE_ROOT and kw.type != kw.FOR:
             self._start_keyword(kw)
 
     def end_keyword(self, kw):
-        if kw.type != kw.IF_ELSE_ROOT:
+        if kw.type != kw.IF_ELSE_ROOT and kw.type != kw.FOR:
             self._end_keyword(kw)
 
     def log_message(self, msg):


pekkaklarck:
Thanks for the PR. Could you please also submit an issue about the underlying bug that `BuiltIn.run_keyword` causes in this case?

Although this PR fixes the problem, I'm not absolutely sure is this the best way. Instead if not calling these listener methods we could also fix output.xml parsing to support `<kw>` elements inside `<for>`. That ought to be easy, but it could also be that the result model doesn't support such constructs. Enhancing the model could be an option, but simply using the current approach in this PR could be better.

Regardless how the problem is fixed, it needs acceptance tests to make sure we don't introduce the bug again.

Harri:
Sure thing, issue is here https://github.com/robotframework/robotframework/issues/4112

pekkaklarck:
Thanks for submitting #4112 and for providing a simple example to reproduce the problem. I investigated this quickly and was able to fix output.xml parsing pretty easily. More details in the submitted issue.

pekkaklarck:
This has been already fixed in d64f3c25fec38c7778780c619e3e4ad7b60b4926.

PR #4126, opened 2021-10-22
Add len and type log formatter options to log keyword

### LINKED ISSUE #4095: Add type formatter to Log keyword
Add a formatter=type to Log keyword for verification of variable type

Wolfe1:
Resolves #4095

- Adds the len (suggested) and type formatter options to _get_formatters used by Log.
- Acceptance tests added for each.
- Doc and error message modifications.

Made an attempt at the additional suggestion in #4095 to support any builtin but couldn't get it down. After looking at the [list of builtins](https://docs.python.org/3/library/functions.html) as well I am not confident that I am not opening some can of worms in the process there. :sweat_smile:

### DIFF
--- a/atest/robot/standard_libraries/builtin/log.robot
+++ b/atest/robot/standard_libraries/builtin/log.robot
@@ -105,6 +105,22 @@ formatter=repr pretty prints
     Stdout Should Contain    ${small dict}
     Stdout Should Contain    ${small list}
 
+formatter=len
+    ${tc} =    Check Test Case    ${TEST NAME}
+    Check Log Message    ${tc.kws[0].msgs[0]}    20
+    Check Log Message    ${tc.kws[1].msgs[0]}    13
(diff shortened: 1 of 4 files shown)


pekkaklarck:
Looks great! Only missing thing is adding a note to docs that `len` and `type` are new in RF 5.0 but I can do that myself after merging this PR.

Supporting any builtin would be pretty easy as we could just do something like `formatter = getattr(builtin, name)`, but I agree it's very unlikely others would be useful. Being able to use other Python functions like `math.ceil()` could be more useful, but that's also more work. Users needing that can already use inline Python evaluation like `${{math.cail($value)}}` and if this turns out to be a common need such functionality can be added later.

A more important possible enhancement would be supporting multiple formatters like `formatter=type, repr`. Users obviously can use `Log` multiple times, but getting all information printed in one go would have benefits. We can still discuss should that be added as part of  #4095, but this PR can be merged now anyway.

Issue #4079, opened 2021-09-06
New `BREAK` and `CONTINUE` statements for controlling `FOR` and `WHILE` loop execution


pekkaklarck:
We currently have `Exit For Loop` (#502) and `Continue For Loop` (#1125) keywords for this purpose. Having explicit statements would make the usage more clear and closer to how "real" programming languages handle loops. This is especially important when we add `WHILE` loops (#4084).

When `Exit For Loop` was added (#502), we decided to use that name instead of `Break` or `Break For Loop` because the exiting sounds more natural for non-programmers. That's still a valid point, but I believe these new statements should be named the same way as they commonly are named in other languages. That way they are more familiar for programmers and also people learning basics of programming using Robot's syntax will have it easier to learn other languages later.

See also #4078 about adding somewhat similar `RETURN` statement.

pekkaklarck:
One obvious thing to decide is what we do with `Exit For Loop` and `Continue For Loop` keywords after `BREAK` and `CONTINUE` are implemented. I think its better to deprecate and remove them. It's better to have just one way to accomplish one thing and making BuiltIn library smaller is good as well. We should, however, wait at least until RF 5.1 before emitting deprecation warnings.

A related question is what to do with conditional variants `Exit For Loop If` and `Continue For Loop If`. They make shorter constructs than `IF` with `BREAK` or `CONTINUE`, but I think having just one way to break/continue is more important. We could consider some kind of inline `IF` structure later if shorter structures are considered important. It would then work also with `RETURN`.

ericbjones:
Single line IF option would be great. The new IF syntax is soooo much better for conditional blocks than the prior `Run Keyword If      Run Keywords  …   AND    …    AND    …    AND` silliness we had to do. But when you start using the new IF everywhere instead of Run Keyword If , you end up with awkwardness of `IF`  -  `END`’s around single lines all over the place.

I would never choose having to put  `…` and `AND` stuff for blocks like it was for `run keywords` Nor would I want anything that only lets you put a keyword that cannot have arguments. But it would be great if `IF` had two syntaxes:

```
IF    <condition>    <keyword>   [*args]
``` 

``` 
IF    <condition>
    <whatever>
    <you>
    <want>
END
```
So if the `IF` keyword has 3+ arguments, the first syntax is automatically chosen and you can not continue to new lines without using the `…` continuation keyword, and you may not have more than one keyword (please no AND syntax! :) ) and you may not nest.

[...]

PR #4136, opened 2021-11-04
#4079 Continue and  Break

bbpatel2001:


src/robot/parsing/model/statements.py
```diff
@@ -955,6 +955,40 @@ class End(NoArgumentHeader):
...
+    def from_params(cls, indent=FOUR_SPACES, eol=EOL):
+        return cls([
+            Token(Token.SEPARATOR, indent),
+            Token(Token.BREAK),
+            Token(Token.EOL, eol)
+        ])
+
+    def validate(self):
+        if self.get_tokens(Token.ARGUMENT):
+            self.errors += ('BREAK does not accept arguments.',)
```
pekkaklarck, line 989: There's new `NoArgumentHeader` base class that both `Continue` and `Break` could inherit. It would provide both `from_params` and `validate` methods.

src/robot/parsing/model/statements.py
```diff
@@ -974,6 +1008,74 @@ def from_params(cls, values=(), indent=FOUR_SPACES, separator=FOUR_SPACES, eol=E
...
```
pekkaklarck, line 1076: No need to to have `Break` and `Continue` three times!

Not sure what would be the best place for them in this rather long module. Probably after `Return`.

src/robot/running/model.py
```diff
@@ -175,6 +176,34 @@ def run(self, context, run=True, templated=False):
                 raise ReturnFromKeyword(self.values)
 
 
+@Body.register
+class Continue(model.Return):
```
pekkaklarck, line 180: There should be `Continue` and `Break` classes in the `model` package that are extended here. They are also needed in the `result` package.

src/robot/running/model.py
```diff
@@ -175,6 +176,34 @@ def run(self, context, run=True, templated=False):
...
+    __slots__ = ['lineno']
+
+    def __init__(self, values=(), parent=None, lineno=None):
+        model.Return.__init__(self, values, parent)
```
pekkaklarck, line 184: Neither `CONTINUE` nor `BREAK` accepts values in data (`RETURN` does) so model objects shouldn't have `values` arguments either.

utest/parsing/test_lexer.py
```diff
@@ -1816,5 +1816,149 @@ def _verify(self, data, expected, test=False):
...
+class TestContinue(unittest.TestCase):
+
+    def test_in_keyword(self):
+        data = '    CONTINUE'
+        expected = [(T.CONTINUE, 'CONTINUE', 3, 4),
+                    (T.EOS, '', 3, 12)]
+        self._verify(data, expected)
+
+    def test_in_test(self):
+        # This is not valid usage but that's not recognized during lexing.
```
pekkaklarck, line 1828: It's true using `CONTINUE` and `BREAK` directly under a test isn't valid, but using them directly inside a user keyword isn't valid either. It should be possible to use them only inside `FOR` either directly or inside `IF` and `TRY` structures. This is different to `RETURN` that can be used anywhere inside a user keyword, also directly under them. In other words, this comment applies also to the above test.

Notice that it's possible to use `Continue For Loop` and `Exit For Loop` keywords directly inside user keywords as long as those user keywords are used inside a loop. That's bad design and not something I want to repeat with `CONTINUE` and `BREAK`. Although those keywords will be eventually deprecated and removed, I wouldn't like to change their behavior to avoid backwards incompatible change. We need to see can that be handled nicely when we get further with `CONTINUE/BREAK`.

Notice also that I have an idea how these error situations could be detected already during lexing/parsing. I can look at that after this PR has been merged and update tests then.

utest/parsing/test_lexer.py
```diff
@@ -1816,5 +1816,149 @@ def _verify(self, data, expected, test=False):
...
+            name_type = T.TESTCASE_NAME
+        data = f'{header}\nName\n{data}'
+        expected = [(header_type, header, 1, 0),
+                    (T.EOS, '', 1, len(header)),
+                    (name_type, 'Name', 2, 0),
+                    (T.EOS, '', 2, 4)] + expected
+        assert_tokens(data, expected, data_only=True)
+
+
+class TestBreak(unittest.TestCase):
```
pekkaklarck, line 1890: Parsing tests for `CONTINUE` and `BREAK` are so similar that I think it would be best to have a single class for them. Each test in that class could then test same functionality with both of them.

pekkaklarck (changes requested):
Looks good! See individual comments for things to improve.

src/robot/parsing/model/statements.py
bbpatel2001: Removed.

src/robot/parsing/model/statements.py
bbpatel2001: Resolved. Modified code as per comment.

src/robot/running/model.py
bbpatel2001: Removed following code blocked and directly  raised...  Is this correct or should I just remove values .... ?
with StatusReporter(self, ReturnResult(self.values), context, run):
            if run:
                raise ExitForLoop(self.values)

utest/parsing/test_lexer.py
bbpatel2001: I also thought same thing but word length is different so I wrote it in separate class. Need your suggestion here.

utest/parsing/test_lexer.py
bbpatel2001: I am not sure I am following all your points mentioned here however I have modified test little.

src/robot/running/model.py
```diff
@@ -175,6 +176,30 @@ def run(self, context, run=True, templated=False):
                 raise ReturnFromKeyword(self.values)
 
 
+@Body.register
+class Continue(model.Continue):
+    __slots__ = ['lineno']
+
+    def __init__(self, values=(), parent=None, lineno=None):
+        model.Return.__init__(self)
```
pekkaklarck, line 184: Should be `model.Continue`, not `model.Return` here! Even better would be using `super()` that we have now started to use when we don't need to care about Python 2 (`super()` syntax was different with it). Also need to pass `parent` to the parent class. In other words, this should be `super().__init__(parent)`.

src/robot/running/model.py
```diff
@@ -175,6 +176,30 @@ def run(self, context, run=True, templated=False):
...
+    def run(self, context, run=True, templated=False):
+        raise ContinueForLoop()
+
+
+@Body.register
+class Break(model.Break):
+    __slots__ = ['lineno']
+
+    def __init__(self, values=(), parent=None, lineno=None):
+        model.Return.__init__(self)
```
pekkaklarck, line 196: Same problems as with `Continue` here as well.

src/robot/running/model.py
bbpatel2001: added in result pkg but need you guidance a not able to understand this pkg fully.

utest/parsing/test_lexer.py
pekkaklarck: Do you mean that `BREAK` and `CONTINUE` have different length and thus offset of the following tokens are different? There's ought to be only an `EOS` on the same line and its offset ought to be pretty easy to set correctly based on the token type. I imagine something like this could work:
```python
class TestBreakAndContinue(unittest.TestCase):
    def test_in_keyword(self):
        for value in 'BREAK', 'CONTINUE':
            data = f'   {value}'
            expected = [(value, value, 3, 4),
                        (T.EOS, '', 3, 4 + len(value)]
            self._verify(data, expected)
```

This can potentially add some extra complexity so having separate test classes is fine as well. They can possibly be combined later if needed.

utest/parsing/test_lexer.py
```diff
@@ -1816,5 +1816,199 @@ def _verify(self, data, expected, test=False):
...
+    END
+'''
+            expected = [(T.FOR, 'FOR', 3, 4),
+                    (T.VARIABLE, '${x}', 3, 11),
+                    (T.FOR_SEPARATOR, 'IN', 3, 19),
+                    (T.ARGUMENT, '@{STUFF}', 3, 25),
+                    (T.EOS, '', 3, 33),
+                    (T.END, 'END', 7, 4),
+                    (T.EOS, '', 6, 7)]
+            self._verify(data, expected)
```
pekkaklarck, line 1896: There's no `CONTINUE` in expected. I assume this doesn't succeed yet.

utest/parsing/test_lexer.py
```diff
@@ -1816,5 +1816,199 @@ def _verify(self, data, expected, test=False):
...
+                    (name_type, 'Name', 2, 0),
+                    (T.EOS, '', 2, 4)] + expected
+        assert_tokens(data, expected, data_only=True)
+
+
+class TestBreak(unittest.TestCase):
+
+    def test_in_keyword(self):
+        data = '    BREAK'
+        expected = [(T.CONTINUE, 'BREAK', 3, 4),
```
pekkaklarck, line 1919: Should be `T.BREAK` here and in tests below.

utest/parsing/test_model.py
```diff
@@ -832,6 +832,60 @@ def test_return(self):
...
+            ),
+            body=[
+
+                Keyword(
+                    header=KeywordName(
+                        tokens=[Token(Token.KEYWORD_NAME, 'Name', 2, 0)]
+                    ),
+                    body=[
+                        KeywordCall([Token(Token.KEYWORD, 'Break', 3, 4)]),
+                        ReturnStatement([Token(Token.BREAK, 'BREAK', 4, 4)])
```
pekkaklarck, line 881: Should be `ContinueStatement` here!

utest/parsing/test_lexer.py
pekkaklarck: This ought to be fine. Can be changed later if parser is enhanced to detect `RETURN` and others when used in a wrong place. That's not something this PR needs to care.

pekkaklarck (changes requested):
Lexing part of the parser is not in good shape, but parser still needs to be enhanced to create appropriate parsing model. Also transformer needs to be enhanced to convert that parsing model into the executable model.

utest/parsing/test_model.py
bbpatel2001: corrected.

utest/parsing/test_lexer.py
bbpatel2001: corrected

utest/parsing/test_model.py
bbpatel2001: Corrected.

utest/parsing/test_lexer.py
bbpatel2001: Corrected.

src/robot/parsing/lexer/blocklexers.py
```diff
@@ -26,8 +26,7 @@
                               ForHeaderLexer, InlineIfHeaderLexer,
                               IfHeaderLexer, ElseIfHeaderLexer, ElseHeaderLexer,
                               TryHeaderLexer, ExceptHeaderLexer, FinallyHeaderLexer,
-                              EndLexer, ReturnLexer)
-
+                              EndLexer, ReturnLexer, ContinueLexer, BreakLexer)
 
 class BlockLexer(Lexer):
```
pekkaklarck, line 31: An empty line has been accidentally removed. PEP-8 says there should be two such lines between imports and the class declaration.

src/robot/parsing/lexer/blocklexers.py
```diff
@@ -233,7 +232,7 @@ def handles(self, statement):
 
     def lexer_classes(self):
         return (InlineIfLexer, IfHeaderLexer, ElseIfHeaderLexer, ElseHeaderLexer,
-                ForLexer, TryLexer, EndLexer, ReturnLexer, KeywordCallLexer)
+                ForLexer, TryLexer, EndLexer, ReturnLexer, KeywordCallLexer, ContinueLexer, BreakLexer)
```
pekkaklarck, line 235: I'm not absolutely sure about this, but I believe these new lexers should be before `KeywordCallLexer` here and in other `lexer_classes`. Otherwise `BREAK` and `CONTINUE` are likely to be lexed as keyword calls.

src/robot/result/model.py
```diff
@@ -264,6 +264,29 @@ def doc(self):
         return ''
 
 
+@Body.register
+class Continue(model.Continue, StatusMixin, DeprecatedAttributesMixin):
+    __slots__ = ['status', 'starttime', 'endtime']
+
+    def __init__(self, values=(), status='FAIL', starttime=None, endtime=None, parent=None):
+        super().__init__(values, parent)
```
pekkaklarck, line 272: `Continue` shouldn't accept `values`. Same also with `Break`.

src/robot/running/builder/transformers.py
```diff
@@ -229,6 +229,11 @@ def visit_ReturnStatement(self, node):
         self.test.body.create_keyword(name='RETURN', args=node.values,
                                       lineno=node.lineno)
 
+    def visit_Continue(self, node):
+        self.test.body.create_keyword(name='CONTINUE',lineno=node.lineno)
+
+    def visit_Break(self, node):
+        self.test.body.create_keyword(name='BREAK',lineno=node.lineno)
 
 class KeywordBuilder(NodeVisitor):
```
pekkaklarck, line 238: 1. When mapping CONTINUE and BREAK used outside a loop to a keyword call, we should pass also their possible arguments. That's done with RETURN above, but using `node.values` isn't possible here because CONTINUE and BREAK nodes don't currently have such attribute. Probably `NoArgumentsHeader` should get this property to allow that:
   ```python
    @property
    def values(self):
        return self.get_values(Token.ARGUMENT)
    ```
2. There should be two empty lines between classes.

src/robot/running/model.py
```diff
@@ -38,6 +38,7 @@
 from robot import model
 from robot.conf import RobotSettings
 from robot.errors import ReturnFromKeyword
+from robot.errors import ContinueForLoop, ExitForLoop
```
pekkaklarck, line 41: Combine the above two import lines together.

src/robot/running/model.py
```diff
@@ -175,6 +176,30 @@ def run(self, context, run=True, templated=False):
                 raise ReturnFromKeyword(self.values)
 
 
+@Body.register
+class Continue(model.Continue):
+    __slots__ = ['lineno']
+
+    def __init__(self, values=(), parent=None, lineno=None):
+        super().__init__(parent)
```
pekkaklarck, line 184: `Continue`, and `Break`, shouldn't accept `values`.

pekkaklarck (changes requested):
Great progress!

src/robot/parsing/lexer/blocklexers.py
bpateldg: Corrected.

src/robot/parsing/lexer/blocklexers.py
bpateldg: Corrected.

src/robot/result/model.py
bpateldg: I had checked this, and I was thinking that it will be a empty, so did not remove. 
Removing now.

src/robot/running/builder/transformers.py
bpateldg: In this case, I am adding back args=node.values to the method. 

def visit_Continue(self, node):
        self.test.body.create_keyword(name='CONTINUE', args=node.values,
                                      lineno=node.lineno)

src/robot/running/model.py
bpateldg: Corrected. Thought it will be empty so had kept. 

Removed

src/robot/model/control.py
```diff
@@ -242,3 +242,29 @@ def __init__(self, values=(), parent=None):
...
+
+
+@Body.register
+class Continue(BodyItem):  
+    type = BodyItem.CONTINUE
+    __slots__ = []
+
+    def __init__(self, parent=None):
+        self.parent = parent
+        #self.body = None
```
pekkaklarck, line 254: Remove the commented code here and in `Break`. These structures don't have body.

src/robot/parsing/lexer/blocklexers.py
```diff
@@ -11,7 +11,7 @@
 #  distributed under the License is distributed on an "AS IS" BASIS,
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
-#  limitations under the License.
+#  limitations under the License.
```
pekkaklarck, line 14: Remove accidentally added whitespace.

src/robot/result/xmlelementhandlers.py
```diff
@@ -240,13 +240,31 @@ def start(self, elem, result):
...
+
+
+@ElementHandler.register
+class BreakHandler(ElementHandler):
+    tag = 'break'
+    children = frozenset(('status', 'msg'))
+
+    def start(self, elem, result):
+        return result.body.create_break()
+
```
pekkaklarck, line 259: In addition to creating these handlers for `<break>` and `<continue>`, they need to be listed in `children` for `IterationHandler` and `BranchHandler`. After this change `BREAK` and `CONTINUE` should be visible in log as well.

`<return>` is also used with `KeywordHandler`, but these statements aren't usable directly under keywords so that shouldn't be needed.

pekkaklarck (changes requested):
Good progress again! Small change still needed to get BREAK and CONTINUE visible in log.

Issue #4133, opened 2021-11-04
Support variable length in `Generate Random String`


asimell:
`Generate Random String` only supports a static length for a random string. It would be nice if if you could give a range of values how long the random string should be.

E.g. something like this
```
# Assume the following variable has been set: @{values}    5    10
${str}=    Generate Random String    length=${values}
${str}=    Generate Random String    length=[5, 10]
```

pekkaklarck:
Sounds good to me in general. Not sure is an actual list or list-literal syntax best way to give ranges, though. Couldn't it be just `5-10` or `5..10`?

asimell:
I thought about this as well. `5-10` would be the most intuitive IMO, but it could possibly support a list as well, instead of a list-like string.

So

```
Generate Random String    length=5
Generate Random String    length=${values}
Generate Random String    length=5-10
```

I have a feeling supporting a list and a string in format `<int>-<int>` is fairly straightforward to implement.

pekkaklarck:
Supporting actual lists is easy but I'm not sure is it worth the extra work needed for testing and documentation. Is there some use case where it would be better than simple `5-10`?

asimell:
No direct use cases come to mind. It can probably be ignored for now and added later on if there is bigger need?

pekkaklarck:
Yeah, adding later if needed is always easy.

pekkaklarck:
This was implemented in PR #4138 by @asimell. Thanks for a contribution!

PR #4138, opened 2021-11-05
Support generating random strings with random length

asimell:
Implementation for #4133.

Allows the use of `Generate Random String` with a range of values, instead of a single value.

### DIFF
--- a/atest/testdata/standard_libraries/string/generate_random_string.robot
+++ b/atest/testdata/standard_libraries/string/generate_random_string.robot
@@ -22,6 +22,19 @@ Generate Random String With Random Length
         String Length Should Be Within    ${result}    5    10
     END
 
+Generate Random String With Invalid Ranges
+    # Cannot convert 'length' to integer
+    Run Keyword And Expect Error    ValueError: *    Generate Random String    5-
+    Run Keyword And Expect Error    ValueError: *    Generate Random String    foo-10
+    Run Keyword And Expect Error    ValueError: *    Generate Random String    5-bar
+    Run Keyword And Expect Error    ValueError: *    Generate Random String    foo-bar
+    Run Keyword And Expect Error    ValueError: *    Generate Random String    -
+    # Length is not a valid range of integers
+    Run Keyword And Expect Error    ValueError: Length is not a valid range of integers: *
+    ...    Generate Random String    --
+    Run Keyword And Expect Error    ValueError: Length is not a valid range of integers: *
+    ...    Generate Random String    1-2-3
+
 Generate Random String From Non Default Characters
     Test Random String With    %=}$+^~*äö#    %=}$+^~*äö#
 
--- a/atest/testdata/standard_libraries/string/generate_random_string.robot
+++ b/atest/testdata/standard_libraries/string/generate_random_string.robot
@@ -24,15 +24,10 @@ Generate Random String With Random Length
 
 Generate Random String With Invalid Ranges
     # Cannot convert 'length' to integer
-    Run Keyword And Expect Error    ValueError: *    Generate Random String    5-
-    Run Keyword And Expect Error    ValueError: *    Generate Random String    foo-10
-    Run Keyword And Expect Error    ValueError: *    Generate Random String    5-bar
-    Run Keyword And Expect Error    ValueError: *    Generate Random String    foo-bar
-    Run Keyword And Expect Error    ValueError: *    Generate Random String    -
-    # Length is not a valid range of integers
-    ${range error} =    Set Variable    ValueError: Length is not a valid range of integers: *
-    Run Keyword And Expect Error    ${range error}    Generate Random String    --
-    Run Keyword And Expect Error    ${range error}    Generate Random String    1-2-3
+    FOR    ${range}    IN     5-    foo-10    5-bar    foo-bar    -    5--10    1-2-3
+        Run Keyword And Expect Error    ValueError: Cannot convert 'length' argument '${range}' to an integer.
+        ...    Generate Random String    ${range}
+    END
 
 Generate Random String From Non Default Characters
     Test Random String With    %=}$+^~*äö#    %=}$+^~*äö#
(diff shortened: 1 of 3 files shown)


src/robot/libraries/String.py
```diff
@@ -587,10 +590,16 @@ def generate_random_string(self, length=8, chars='[LETTERS][NUMBERS]'):
         | ${low} = | Generate Random String | 12 | [LOWER]         |
         | ${bin} = | Generate Random String | 8  | 01              |
         | ${hex} = | Generate Random String | 4  | [NUMBERS]abcdef |
+        | ${rnd} = | Generate Random String | 5-10 | # Generates a string 5 to 10 characters long |
         """
         if length == '':
             length = 8
-        length = self._convert_to_integer(length, 'length')
+        if '-' in str(length):
+            min_length, max_length = length.split('-')
```
pekkaklarck, line 598: Error checking needed in case the value contains more than one `-`. Probably try/except around this assignment and raising an exception with a descriptive error message is easiest.

Should also add tests for cases like `foo-bar` and `-10`. Most likely `_convert_to_integer` already handles them fine, but explicit tests make that sure.

We could actually make `-10` same as `0-10`, but since it looks like a negative number, explicitly adding `0` is easy, and range like `10-` doesn't make any sense I doubt that's a good idea.

src/robot/libraries/String.py
```diff
@@ -571,6 +571,9 @@ def fetch_from_right(self, string, marker):
     def generate_random_string(self, length=8, chars='[LETTERS][NUMBERS]'):
         """Generates a string with a desired ``length`` from the given ``chars``.
 
+        ``length`` can be given as a number, a string representation of a number,
+        or as a range of numbers, such as ``5-10``.
```
pekkaklarck, line 575: Need a note here or at the end of the documentation that this is new in Robot Framework 5.0.

pekkaklarck (changes requested):
Looks good but few small changes needed.

src/robot/libraries/String.py
pekkaklarck: Perhaps should also more explicitly explain that when the value is given as a range, actual length is selected randomly from that range.

src/robot/libraries/String.py
asimell: `-10` is actually parsed currently normally by `_convert_to_integer`. In that case, the keyword doesn't do anything as string length cannot be negative.

This behaviour can be persisted by adding a try-except for an `AttributeError` to `length.split('-')`. Although a negative length string doesn't make any sense, but that's the current behaviour anyway.

src/robot/libraries/String.py
asimell: Improved documentation.

src/robot/libraries/String.py
```diff
@@ -587,10 +591,23 @@ def generate_random_string(self, length=8, chars='[LETTERS][NUMBERS]'):
...
         if length == '':
             length = 8
-        length = self._convert_to_integer(length, 'length')
+        if '-' in str(length):
+            try:
+                min_length, max_length = length.split('-')
+                length = randint(self._convert_to_integer(min_length, "length"),
+                                 self._convert_to_integer(max_length, "length"))
+            except ValueError:
+                raise ValueError("Length is not a valid range of integers: {0}".format(length))
```
pekkaklarck, line 606: We are Python 3.6+ nowadays so you can use f-sting here. I'd also format it e.g. like `f"Length '{length}' is not ...`.

src/robot/libraries/String.py
```diff
@@ -587,10 +591,23 @@ def generate_random_string(self, length=8, chars='[LETTERS][NUMBERS]'):
...
+            except AttributeError:
+                pass    # To keep behaviour as it is with negative numbers
```
pekkaklarck, line 608: Catching and ignoring AttributeError is too magic even with a comment.

atest/testdata/standard_libraries/string/generate_random_string.robot
```diff
@@ -10,11 +10,29 @@ ${NUMBERS}        1234567890
...
+    # Cannot convert 'length' to integer
+    Run Keyword And Expect Error    ValueError: *    Generate Random String    5-
+    Run Keyword And Expect Error    ValueError: *    Generate Random String    foo-10
+    Run Keyword And Expect Error    ValueError: *    Generate Random String    5-bar
+    Run Keyword And Expect Error    ValueError: *    Generate Random String    foo-bar
+    Run Keyword And Expect Error    ValueError: *    Generate Random String    -
+    # Length is not a valid range of integers
+    ${range error} =    Set Variable    ValueError: Length is not a valid range of integers: *
+    Run Keyword And Expect Error    ${range error}    Generate Random String    --
+    Run Keyword And Expect Error    ${range error}    Generate Random String    1-2-3
```
pekkaklarck, line 35: This could possibly be handled using a FOR loop. Something like this:
```
FOR    ${range}    IN    5-    foo-10    # ...
    Run Keyword And Expect Error
    ...    ValueError: Cannot convert 'length' argument '${range}' to an integer.
    ...    Generate Random String    ${range}
END
```
The above assumes checking the range has been changed as I propose in another comment and invalid ranges are validated in `_convert_to_integer`.

src/robot/libraries/String.py
pekkaklarck: Perhaps it would be easier to check that the range is in correct format so that no error checking would be needed here in general. Something like this:
```python
if isinstance(length, str) and re.match(r'^\d+-\d+$', length):
    min_length, max_length = length.split('-')
    ...
else:
    length = ...
```

pekkaklarck (changes requested):
Looks mostly really good but error checking is a bit complicated and magic. Checking that the range is valid beforehand, like I propose in one comment, could make it easier.

src/robot/libraries/String.py
asimell: Moved all error handling to `_convert_to_integer` as suggested.

src/robot/libraries/String.py
pekkaklarck: It's actually a bit questionable that negative numbers work, we could make it an error and require 0 when an empty string is ok, but that's not related to this issue.

pekkaklarck (approved):

pekkaklarck:
Thanks for a great PR!

PR #4144, opened 2021-11-13
Resolves  #4074 Add timestamp attribute testsuite element in xunit output

### LINKED ISSUE #4074: xUnit outputs: Add `timestamp` attribute for `testsuite` elements
Python [junitparser](https://pypi.org/project/junitparser/) expects that there is  timestamp attribute testsuite element. 

Also Pytest puts timestamp attribute testsuite element. Example Pytest has: 
`<testsuite name="pytest" errors="0" failures="0" skipped="2" tests="46" time="18.169" timestamp="2021-09-01T04:43:16.364650" hostname="something">` 

In Robot Framework there is: `<testsuite name="Test" tests="435" errors="0" failures="4" skipped="0" time="244.974">`

Is somewhat related to: #2982 and #3944

rikerfi:
Resolves issue #4074 by adding the timestamp attribute to testsuite element in xunit output. The timestamp is generated from the testsuite ```starttime``` which seems to be procedure in other tools as well. Furthermore, strong assumption is done that RF starttime format is always ```YYYYMMDD hh:mm:ss.mil```. Starttime with empty string ```''``` or ```None``` yields xunit timestamp attribute to empty value as well: ```timestamp=""```.

Valid RF starttime yields e.g. ```2021-11-10T21:34:30.006000```

### DIFF
--- a/atest/robot/output/xunit.robot
+++ b/atest/robot/output/xunit.robot
@@ -1,15 +1,21 @@
 *** Settings ***
 Documentation       Tests for xunit-compatible xml-output.
 Resource            atest_resource.robot
+Resource            rebot_resource.robot
 Variables           unicode_vars.py
-Suite Setup         Run Tests    -x xunit.xml -l log.html --skiponfailure täg    ${TESTDATA}
+Suite Setup         Run Keywords
+...    Create Output With Robot    ${OUT ONE}    ${EMPTY}    ${PASS AND FAIL}
+...    AND
+...    Create Output With Robot    ${OUT TWO}    ${EMPTY}    ${TESTDATA}
+...    AND
+...    Run Tests    -x xunit.xml -l log.html --skiponfailure täg    ${TESTDATA}
 
 *** Variables ***
 ${TESTDATA}         misc/non_ascii.robot
 ${PASS AND FAIL}    misc/pass_and_fail.robot
 ${INVALID}          %{TEMPDIR}${/}ïnvälïd-xünït.xml
-${MERGE ONE}        %{TEMPDIR}${/}merge1.xml
-${MERGE TWO}        %{TEMPDIR}${/}merge2.xml
+${OUT ONE}          %{TEMPDIR}${/}out1.xml
+${OUT TWO}          %{TEMPDIR}${/}out2.xml
 
 *** Test Cases ***
 XUnit File Is Created
@@ -78,17 +84,11 @@ Skipping non-critical tests is deprecated
     Stderr Should Contain   Command line option --xunitskipnoncritical has been deprecated and has no effect.
 
 Merge outputs
-    Run Tests Without Processing Output    --output ${MERGE ONE}    ${PASS AND FAIL}
-    Run Tests Without Processing Output    --output ${MERGE TWO}    ${TESTDATA}
-    Execute On Existing Execution Environment
-    ...    ${INTERPRETER.rebot}    --xUnit xunit.xml -l log.html    ${MERGE TWO} ${MERGE ONE}    ${COMMON DEFAULTS}
+    Run Rebot    -x xunit.xml -l log.html    ${OUT ONE} ${OUT TWO}
     Verify XUnit Timestamp    ${EMPTY}
 
 Rebot with start and end time
-    Run Tests    -l log.html    ${PASS AND FAIL}
-    Execute On Existing Execution Environment
-    ...    ${INTERPRETER.rebot}    -x xunit.xml -l log.html --starttime 20211215-12:11:10.456 --endtime 20211215-12:13:14.789
-    ...    ${OUTFILE}    ${COMMON DEFAULTS}
+    Run Rebot    -x xunit.xml -l log.html --starttime 20211215-12:11:10.456 --endtime 20211215-12:13:14.789    ${OUT ONE}
     Verify XUnit Timestamp    2021-12-15T12:11:10.456000
 
 *** Keywords ***
@@ -116,11 +116,3 @@ Verify XUnit Timestamp
     File Should Exist    ${OUTDIR}/xunit.xml
     ${suite} =    Get XUnit Node
     Element Attribute Should Match    ${suite}    timestamp    ${pattern}
-
-Execute On Existing Execution Environment
-    [Arguments]    ${executor}    ${options}    ${sources}    ${default options}=
-    @{arguments} =    Get Execution Arguments    ${options}    ${sources}    ${default options}
-    ${result} =    Run Process    @{executor}    @{arguments}
-    ...    stdout=${STDOUTFILE}    stderr=${STDERRFILE}    output_encoding=SYSTEM
-    ...    timeout=5min    on_timeout=terminate
-    [Return]    ${result}
--- a/atest/robot/output/xunit.robot
+++ b/atest/robot/output/xunit.robot
@@ -8,6 +8,8 @@ Suite Setup         Run Tests    -x xunit.xml -l log.html --skiponfailure täg  
 ${TESTDATA}         misc/non_ascii.robot
 ${PASS AND FAIL}    misc/pass_and_fail.robot
 ${INVALID}          %{TEMPDIR}${/}ïnvälïd-xünït.xml
+${MERGE ONE}        %{TEMPDIR}${/}merge1.xml
+${MERGE TWO}        %{TEMPDIR}${/}merge2.xml
 
 *** Test Cases ***
 XUnit File Is Created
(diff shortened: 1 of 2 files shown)


src/robot/reporting/xunitwriter.py
```diff
@@ -90,3 +91,9 @@ def visit_errors(self, errors):
 
     def end_result(self, result):
         self._writer.close()
+
+    def _starttime_to_isoformat(self, stime):
+        isoformat = None
+        if stime:
+            isoformat = f'{stime[:4]}-{stime[4:6]}-{stime[6:8]}T{stime[9:22]}000'
+        return isoformat
```
pekkaklarck, line 99: What happens if `None` is returned here? Will we get `timestamp='None'`, not `timestamp` at all, or a crash? I hope it's the middle one but test should be added for it.

I'd also write this code like below to simplify it a bit:
```
if not stime:
    return None
return f'...'
```

pekkaklarck (changes requested):
Looks good but a test for suite not having a start time is needed.

src/robot/reporting/xunitwriter.py
rikerfi: ```return None``` yields ```timestamp=""``` currently. I made that part unit testing by hand, but will write the test as well. Also, I'll modify the routine more simply as proposed.

atest/robot/output/xunit.robot
```diff
@@ -71,6 +84,43 @@ Skipping non-critical tests is deprecated
...
+    Execute On Existing Execution Environment
+    ...    ${INTERPRETER.rebot}    --xUnit xunit.xml -l log.html    ${OUTFILE}    ${COMMON DEFAULTS}
+    Verify Xfile Timestamp    ${EMPTY}
+
+Missing suite status
+    Run Tests Without Processing Output    -l log.html    ${PASS AND FAIL}
+    Remove Suite Status Element    ${OUTFILE}
+    Execute On Existing Execution Environment
+    ...    ${INTERPRETER.rebot}    --xUnit xunit.xml -l log.html    ${OUTFILE}    ${COMMON DEFAULTS}
+    Verify Xfile Timestamp    ${EMPTY}
```
pekkaklarck, line 102: These two tests seem to mainly validate that Rebot can parse output.xml files with missing information. I don't think it has anything to do with xunit files and it would probably be best to just remove these tests.

atest/robot/output/xunit.robot
```diff
@@ -71,6 +84,43 @@ Skipping non-critical tests is deprecated
...
+    Run Tests Without Processing Output    --output ${MERGE TWO}    ${TESTDATA}
+    Execute On Existing Execution Environment
+    ...    ${INTERPRETER.rebot}    ${OPTIONS WITH TIMES}    ${MERGE TWO} ${MERGE ONE}    ${COMMON DEFAULTS}
+    Verify Xfile Timestamp    ${XTIMESTAMP}
+
+Rebot with start and end time
+    Run Tests Without Processing Output    -l log.html    ${PASS AND FAIL}
+    Execute On Existing Execution Environment
+    ...    ${INTERPRETER.rebot}    ${OPTIONS WITH TIMES}    ${OUTFILE}    ${COMMON DEFAULTS}
+    Verify Xfile Timestamp    ${XTIMESTAMP}
```
pekkaklarck, line 122: These tests look fine but I don't understand the need to use the new `Execute On Existing Execution Environment` keyword. As far as I can the normal `Run Rebot` would work.

With the last test it probably would be possible to use also `Run Tests`, but then we needed to create the expected xunit timestamp based on suite start time ourselves. Something like this could possibly work:

```
Run Tests    --xunit xunit.xml    ${PASS AND FAIL}
Verify XUnit Timestamp    ${{datetime.datetime.strptime($SUITE.starttime, '%Y%m%d %H:%M:%S.%f').strftime('%Y-%m-%dT%H:%M:%S.%f')}}
```
Yeah, that timestamp conversion is quite complicated. With tests using Rebot specifying the start time using ´--starttime´ is a very good idea.

pekkaklarck (changes requested):
Code is fine but tests are pretty complicated and there are some tests that don't seem to really test xunit outputs at ll. Please take a look at the individual comments and ping me on Slack if you need any help.

atest/resources/atest_resource.robot
```diff
@@ -74,6 +74,14 @@ Execute
...
     [Return]    ${result}
 
+Execute On Existing Execution Environment
+    [Arguments]    ${executor}    ${options}    ${sources}    ${default options}=
+    @{arguments} =    Get Execution Arguments    ${options}    ${sources}    ${default options}
+    ${result} =    Run Process    @{executor}    @{arguments}
+    ...    stdout=${STDOUTFILE}    stderr=${STDERRFILE}    output_encoding=SYSTEM
+    ...    timeout=5min    on_timeout=terminate
+    [Return]    ${result}
+
```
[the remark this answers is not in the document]
rikerfi: We need this keyword because of missing option(s) in between test execution and post-processing. Example of this kind of option is ```--starttime```. Thus we need to execute test without processing the output. Then we need this new keyword as legacy ```Execute``` will set the execution environment causing the test output flushed. This is the rationale behind the new keyword. How generally this is needed? Probably not so much. I'll move it into the suite keywords.

atest/robot/output/xunit.robot
rikerfi: True. These are more like a play around and learn the system. Will remove.

atest/robot/output/xunit.robot
```diff
@@ -71,6 +84,43 @@ Skipping non-critical tests is deprecated
     Run tests    --xUnit xunit.xml --xUnitSkipNonCritical     ${PASS AND FAIL}
     Stderr Should Contain   Command line option --xunitskipnoncritical has been deprecated and has no effect.
 
+Nullish timestamp prerebotmodifier
+    Nullish Timestamp Should Be    ${EMPTY}
```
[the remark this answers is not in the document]
rikerfi: Re-thinking this one as well. Not much added value and will keep tests more streamlined if I just remove it.

atest/robot/output/xunit.robot
rikerfi: I feel ```Merge outputs with times``` and ```Rebot with start and end time``` are falling to very same test in the end. Removing the merge one.

atest/robot/output/xunit.robot
```diff
@@ -8,6 +8,13 @@ Suite Setup         Run Tests    -x xunit.xml -l log.html --skiponfailure täg
 ${TESTDATA}         misc/non_ascii.robot
 ${PASS AND FAIL}    misc/pass_and_fail.robot
 ${INVALID}          %{TEMPDIR}${/}ïnvälïd-xünït.xml
+${MERGE ONE}        %{TEMPDIR}${/}merge1.xml
+${MERGE TWO}        %{TEMPDIR}${/}merge2.xml
+${STIME}            20211215-12:11:10.456
+${XTIMESTAMP}       2021-12-15T12:11:10.456000
+${ETIME}            20211215-12:13:14.789
+${OPTIONS WITH TIMES}
+...                 -x xunit.xml -l log.html --starttime ${STIME} --endtime ${ETIME}
```
pekkaklarck, line 17: Most if not all of these variables seem to be used only once. Would be easier to understand if values were used directly.

pekkaklarck, line 17: Most, if not all, of these variables are used in tests only once. Tests would probably be easier to understand if actual values were used directly.

atest/robot/output/xunit.robot
```diff
@@ -89,3 +113,18 @@ Suite Stats Should Be
     Element Attribute Should Be       ${elem}    skipped     ${skipped}
     Element Attribute Should Match    ${elem}    time        ?.???
     Element Attribute Should Be       ${elem}    errors      0
+    Element Attribute Should Match    ${elem}    timestamp   ????-??-??T??:??:??.???000
+
+Verify XUnit Timestamp
+    [Arguments]    ${pattern}
+    File Should Exist    ${OUTDIR}/xunit.xml
+    ${suite} =    Get XUnit Node
+    Should match    ${suite.attrib['timestamp']}    ${pattern}
```
pekkaklarck, line 122: This could probably be `Should Be Equal` instead of `Should Match`. For consistence reasons and also otherwise it would be better to use `Element Attribute Should Be` like in the keyword above.

atest/robot/output/xunit.robot
```diff
@@ -89,3 +113,18 @@ Suite Stats Should Be
...
+
+Execute On Existing Execution Environment
+    [Arguments]    ${executor}    ${options}    ${sources}    ${default options}=
+    @{arguments} =    Get Execution Arguments    ${options}    ${sources}    ${default options}
+    ${result} =    Run Process    @{executor}    @{arguments}
+    ...    stdout=${STDOUTFILE}    stderr=${STDERRFILE}    output_encoding=SYSTEM
+    ...    timeout=5min    on_timeout=terminate
+    [Return]    ${result}
```
pekkaklarck, line 130: I still don't understand the need for this special keyword. Why couldn't test just use `Run Rebot`?

pekkaklarck (approved):
Looks good now. New tests using Rebot could validate also other things than timestamps but I can fine tune these tests myself after merging this.

pekkaklarck:
Rebot actually has a separate suite for xunit file. I'll move Rebot related tests there.

PR #3210, opened 2019-06-18
Issue #3208: Fix for py logging formatter

Ka55i0peia:
Please refer to https://github.com/robotframework/robotframework/issues/3208

* apply logging formatter as stated here: https://dzone.com/articles/python-custom-logging-handler-example

PR #4147, opened 2021-11-18
[Fix] [3208] RobotHandler not formatting messages correctly

Robtom5:
Fixed issue with RobotHandler not applying and configured logging.Formatters
Added missing unit tests for pylogging.conf
Changed RobotHandler to inject library logger to decouple


utest/run.py
```diff
@@ -48,40 +58,19 @@ def get_tests(directory=None):
         elif testfile.match(name):
             modname = os.path.splitext(name)[0]
             if modname in imported:
-                print("Test module '%s' imported both as '%s' and '%s'. "
-                      "Rename one or fix test discovery."
-                      % (modname, imported[modname],
-                         os.path.join(directory, name)), file=sys.stderr)
+                print(
```
Robtom5, line 61: this formatting was done automatically by python black

pekkaklarck: Both old and new formatting look a bit odd for me. Perhaps using f-string here would make it better but since this isn't in actual production code I don't care too much.

Could also consider using `pathlib` instead of `os.path`, but then there would be even more changes and it would probably be better to have all these `utest/run.py` changes in a separate PR.

utest/run.py
```diff
@@ -93,10 +82,24 @@ def usage_exit(msg=None):
...
-    docs, vrbst = parse_args(sys.argv[1:])
-    tests = get_tests()
+    parser = argparse.ArgumentParser(add_help=False, description=__doc__)
+    parser.add_argument('-I', '--interpreter', default=sys.executable)
+    parser.add_argument('-h', '--help', action='store_true')
+    parser.add_argument('-q', dest='vrbst', action='store_const', const=0)
+    parser.add_argument('-v', dest='vrbst', action='store_const', const=2)
+    parser.add_argument('-d', dest='docs', action='store_true')
+    parser.add_argument(dest='directory', nargs='?', action='store', default=None)
+    parser.set_defaults(vrbst=1)
```
pekkaklarck, line 92: :+1: for `argparse`. These changes are unrelated to the actual fix and would be better to have them in a separate PR, but no need to separate them now.

src/robot/output/pyloggingconf.py
```diff
@@ -65,11 +68,11 @@ def emit(self, record):
         method = self._get_logger_method(record.levelno)
         method(message)
         if error:
-            librarylogger.debug(error)
+            self.library_logger.debug(error)
 
     def _get_message(self, record):
         try:
-            return record.getMessage(), None
+            return self.format(record), None
```
pekkaklarck, line 75: This is the actual fix, right?

utest/output/test_pylogging.py
```diff
@@ -0,0 +1,141 @@
...
+        for handler in root.handlers:
+            root.removeHandler(handler)
+        root.addHandler(self.test_handler)
+
+    def tearDown(self):
+        root = logging.getLogger()
+        root.removeHandler(self.test_handler)
+
+    def test_python_logging_debug(self):
+        TESTMESSAGE = "This is a test message"
```
pekkaklarck, line 52: Using `ALLCAPS` with local variables is a bit odd and against PEP-8. Could you change them? This particular message is used multiple times so perhaps there could be `test_message = 'This is a test message'` on the class level so that tests could just use `self.test_message`.

utest/output/test_pylogging.py
```diff
@@ -0,0 +1,141 @@
...
+        logging.debug(TESTMESSAGE)
+        message, level = self.library_logger.last_message
+        assert_equal(message, TESTMESSAGE)
+        assert_equal(level, logging.DEBUG)
```
pekkaklarck, line 56: These last three steps seem to be repeated with most if not all tests. Perhaps it would be a good idea to add a helper like this to avoid duplication:
```python
def assert_message(self, message, level):
    assert_equal(self.library_logger.last_message, message)
    assert_equal(self.library_logger.last_level, level)
```
Notice that the above would require changing `MockLibraryLogger` to store `last_message` and `last_level` separately.

pekkaklarck: Another alternative would be enhancing `MockLibraryLogger` so that it would get something like
```python
def assert_last(self, message, level):
    assert_equal(self.last_message, message)
    assert_equal(self.last_level, level)
```
that could then me used like `self.library_logger.assert_last(...)` in tests.

src/robot/output/pyloggingconf.py
```diff
@@ -56,6 +56,9 @@ def set_level(level):
 
 
 class RobotHandler(logging.Handler):
+    def __init__(self, level=logging.NOTSET, library_logger=librarylogger):
```
pekkaklarck, line 59: Our convention has been adding an empty row above `__init__`. I know Black doesn't do that but arguable having it is in line with PEP-8. I don't care super strongly about that, but I do care about consistency and would prefer that empty line here. Preferably they should be added to tests as well.

src/robot/output/pyloggingconf.py
Robtom5: Yup, that's the fix that causes messages to now be passed through the set formatter

pekkaklarck:
The fix seems to be the same as in #3210 that never got tests and thus wasn't merged. It looks good but few things need to be changed:

1. One acceptance tests is failing after the change. That seems to be caused by `self.format(record)` automatically adding possible traceback information to the message. We add that info ourselves so most likely it's there now twice. Probably this can be fixed by simply removing our traceback handling code.
2. It would be good to add an acceptance test for this particular feature as well. It belongs to the same suite where the failing test discussed above is. See my [comment to #3210](https://github.com/robotframework/robotframework/pull/3210#issuecomment-521955231) for more info about adding acceptance tests.
3. It's great that there are comprehensive unit tests. If got it right, injecting `library_logger` is done mostly to ease unit testing which is fine. I added some individual comments about style issues related to tests that I hope you can look at.
4. Our convention is having an empty row above `__init__` methods. Hope you can add them.
5. Changes to `utest/run.py` look really good. Having unrelated changes in a single PR isn't great, but changes are simple enough that there's no need to split this PR now.

Robtom5:
Pushed fixes following proposed changes. Have seen an issue with the error message validation for Log Exception as it doesn't seem to expand the * correctly on my computer, but I'm hoping it expands it correctly on the CI system

utest/run.py
```diff
@@ -21,12 +28,12 @@
...
 if not sys.warnoptions:
-    warnings.simplefilter('always')
+    warnings.simplefilter("always")
 
 
 base = os.path.abspath(os.path.normpath(os.path.split(sys.argv[0])[0]))
-for path in ['../src', '../atest/testresources/testlibs']:
-    path = os.path.join(base, path.replace('/', os.sep))
+for path in ["../src", "../atest/testresources/testlibs"]:
+    path = os.path.join(base, path.replace("/", os.sep))
```
pekkaklarck, line 36: Changing quotes is a bit problematic for these reasons:
- We prefer `'` over `"` in general.
- This change has nothing to do with the actual fix or even with `utest/run.py` enhancements.

Not a huge problem and assuming there are no other things to change in this PR we can leave these.

atest/testdata/test_libraries/logging_with_logging.robot
```diff
@@ -37,3 +37,6 @@ Timestamps are accurate
 Logging when timeout is in use
     [Timeout]    5 seconds
     Log something
+
+Log with format
+    Log with format
```
pekkaklarck, line 42: Newline missing at the end of the file. Good idea to configure your IDE to add them automatically. Not a biggie but will create unnecessary diff later if/when this file is modified.

utest/run.py
Robtom5: Annoyingly the latest version of black is responsible for these changes

utest/run.py
```diff
@@ -48,55 +57,47 @@ def get_tests(directory=None):
...
-if __name__ == '__main__':
-    docs, vrbst = parse_args(sys.argv[1:])
-    tests = get_tests()
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(add_help=False, description=__doc__)
+    parser.add_argument("-I", "--interpreter", default=sys.executable)
+    parser.add_argument("-h", "--help", action="store_true")
+    parser.add_argument("-q", dest="vrbst", action="store_const", const=0)
+    parser.add_argument("-v", dest="vrbst", action="store_const", const=2)
+    parser.add_argument("-d", dest="docs", action="store_true")
```
pekkaklarck, line 88: `-q`, `-v` and `-d` should get long options like they had earlier. `--quiet` being missing is the reason running unit tests as part of acceptance tests fails.

pekkaklarck (changes requested):
Code changes and also tests look good. Two acceptance tests are still failing:
- `Robot.Test Libraries.Logging With Logging.Log exception` fails because line number where the exception occurs has changes.
- Running unit tests as part of acceptance tests fails because unit tests are run with `--quiet` and this PR changes `utest/run.py` so that it isn't anymore supported. This is actually a good example why such unrelated changes should rather be in a separate PR. They can be preserved now, but `--quiet` and other long options that used to be supported need to be added back.

utest/run.py
pekkaklarck: Yeah, Black prefers `"` instead of `'` but that can be configured. I wouldn't in general use code formatters when creating a PR but you can also select what changes you commit. Anyway, not a big issue this. Although we generally use `'`, there's code using `"` as well.

Robtom5: Would it be worth adding a pyproject.toml file with this setting to the project so that the issue isnt hit going forward?

Robtom5: I've reverted the change but haven't included a pyproject file in this pr

pekkaklarck:
Changes look good. One acceptance test is failing but it seems be flakey. The failing test is actually related to logging it's also related to timeouts and they cause this kind of problems now and then.

pekkaklarck:
Same test failed again on CI and this time also another test in the same suite related to `logging` and timeouts failed. My assumption is that this enhanced formatting is a bit slower than old formatting. In my opinion correct behavior is more important than ultimate speed here and it also otherwise seems that the test itself could be improved to avoid this flakeynes.

PR #4153, opened 2021-11-28
Resolves #2982 Add separate <testsuite> entries for each suite - No Tests included

### LINKED ISSUE #2982: xUnit outputs: Add separate `<testsuite>` entries for each suite
Currently Robot's xUnit output only contains one `<testsuite>` element and all executed tests as its direct child elements. This is true regardless does the executed top-level test suite have child suites or not. The underlying problem with the xUnit format is that there is not official specification for it, but based on the discussion in #2292 it seems that tools consuming these outputs ought to support nested suite structures and that reporting suites like that would be better than the current flat approach. The problem is that although tools that have been tested seem to support nested structures, there may be other tools, especially custom scripts, that don't.

This kind of potentially backwards-incompatible change can only be done in a major release. We are somewhat late in RF 3.1 release cycle, but if the enhancement is considered useful and possible problems not severe, the feature can still get it. For that we'd need feedback from the user community.

rikerfi:
Please, note that this draft PR is based on https://github.com/robotframework/robotframework/pull/4144, but also addressing fix on #2982. However, absolutely no test cases included yet.


pekkaklarck:
Implementation looks good. This PR should only have changes to adding `<testsuite>` for each suite, not it also contains changes related to adding `timestamp`.

PR #4156, opened 2021-11-30
 Resolves #2982 xUnit outputs: Add separate <testsuite> entries for each suite

### LINKED ISSUE #2982: xUnit outputs: Add separate `<testsuite>` entries for each suite

rikerfi:
Introduces xunit output with separate <testsuite> element for each of suite. Allows nested suites targeting to design topics mentioned in #2982.

### DIFF
--- a/atest/robot/output/xunit.robot
+++ b/atest/robot/output/xunit.robot
@@ -8,9 +8,8 @@ Suite Setup         Run Tests    -x xunit.xml -l log.html --skiponfailure täg  
 ${TESTDATA}         misc/non_ascii.robot
 ${PASS AND FAIL}    misc/pass_and_fail.robot
 ${INVALID}          %{TEMPDIR}${/}ïnvälïd-xünït.xml
-${OUT ONE}          %{TEMPDIR}${/}out1.xml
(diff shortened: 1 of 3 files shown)


pekkaklarck:
On a very quick look on mobile this looks good. Should also update tests in `robot/output/xunit.robot`, though.

rikerfi:
I'll keep working with ```robot/output/xunit.robot``` :)

pekkaklarck:
Merged. Tests could possibly be cleaned up still, but they were a mess already before. Thanks for contribution!

PR #4157, opened 2021-12-01
Topic/grep file regex syntax

JockeJarre:
here is my pull request for #4132

### DIFF
--- a/atest/testdata/standard_libraries/operating_system/get_file.robot
+++ b/atest/testdata/standard_libraries/operating_system/get_file.robot
@@ -113,25 +113,28 @@ Get Binary File returns bytes as-is
 Grep File
     [Template]    Grep And Check File
     ${EMPTY}    foo\nbar\nfoo bar\n\nA Foo
-    ${EMPTY}    foo\nbar\nfoo bar\n\nA Foo    regexp=${True}
     foo         foo\nfoo bar
-    foo         foo\nfoo bar    regexp=${True}
     foo?        foo bar
-    foo.        foo bar    regexp=${True}
     ?foo        ${EMPTY}
-    .foo        ${EMPTY}    regexp=${True}
     ?oo         foo\nfoo bar\nA Foo
-    .+oo         foo\nfoo bar\nA Foo    regexp=${True}
     [Ff]oo      foo\nfoo bar\nA Foo
-    .*[Ff]oo      foo\nfoo bar\nA Foo    regexp=${True}
     f*a         foo bar
-    f.*a         foo bar    regexp=${True}
     ?           foo\nbar\nfoo bar\nA Foo
-    .           foo\nbar\nfoo bar\nA Foo    regexp=${True}
     ????        foo bar\nA Foo
-    ....        foo bar\nA Foo    regexp=${True}
     foo bar     foo bar
-    foo\\sbar     foo bar    regexp=${True}
+
+Grep File with regexp
+    [Template]    Grep And Check File
+    ${EMPTY}    foo\nbar\nfoo bar\n\nA Foo    regexp=True
+    f\\wo         foo\nfoo bar    regexp=True
+    foo.        foo bar    regexp=True
+    .foo        ${EMPTY}    regexp=True
+    .oo         foo\nfoo bar\nA Foo    regexp=True
+    [Ff]oo      foo\nfoo bar\nA Foo    regexp=True
+    f.*a         foo bar    regexp=True
+    .           foo\nbar\nfoo bar\nA Foo    regexp=True
+    ....        foo bar\nA Foo    regexp=True
+    foo\\sbar     foo bar    regexp=True
 
 Grep File with empty file
     Create File    ${TESTFILE}    ${EMPTY}
@@ -141,10 +144,14 @@ Grep File non Ascii
     [Setup]    Create File    ${TESTFILE}    fää\nbär\nföö bär\n\nA Fåå
     [Template]    Grep And Check File
     fää     fää        ${TESTFILE}
-    fää     fää        ${TESTFILE}    regexp=${True}
     ö       föö bär    ${TESTFILE}
-    .*ö     föö bär    ${TESTFILE}    regexp=${True}
     A       A Fåå      ${TESTFILE}
+
+Grep File non Ascii with regexp
+    [Setup]    Create File    ${TESTFILE}    fää\nbär\nföö bär\n\nA Fåå
+    [Template]    Grep And Check File
+    f\\wä     fää        ${TESTFILE}    regexp=True
+    ö       föö bär    ${TESTFILE}    regexp=yes
     A       A Fåå      ${TESTFILE}    regexp=${True}
 
 Grep File with UTF-16 files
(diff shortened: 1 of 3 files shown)


src/robot/libraries/OperatingSystem.py
```diff
@@ -286,34 +287,36 @@ def grep_file(self, path, pattern, encoding='UTF-8', encoding_errors='strict'):
         Possible trailing newline is never returned.
 
         A line matches if it contains the ``pattern`` anywhere in it and
-        it *does not need to match the pattern fully*. The pattern
-        matching syntax is explained in `introduction`, and in this
+        it *does not need to match the pattern fully when not in regex mode*. 
+        The pattern matching syntax is explained in `introduction`, and in this
         case matching is case-sensitive.
```
pekkaklarck, line 292: 1. Shouldn't partial matching work also in the regexp mode? Pattern type affecting this would be somewhat inconsistent and AFAIK `grep` on the command line supports partial matching as well.
2. A bit more information is needed about regexp matching here. Should probably explain both patterns separately so that glob explanation refers to the docs in the introduction and regexp is briefly explained here. I can fine tune this later if needed as well.
3. A note either here or at the end of the documentation explaining that regexp support is new in RF 5.0 is needed.

atest/testdata/standard_libraries/operating_system/get_file.robot
```diff
@@ -113,15 +113,25 @@ Get Binary File returns bytes as-is
...
     [Ff]oo      foo\nfoo bar\nA Foo
+    .*[Ff]oo      foo\nfoo bar\nA Foo    regexp=${True}
     f*a         foo bar
+    f.*a         foo bar    regexp=${True}
     ?           foo\nbar\nfoo bar\nA Foo
+    .           foo\nbar\nfoo bar\nA Foo    regexp=${True}
     ????        foo bar\nA Foo
+    ....        foo bar\nA Foo    regexp=${True}
     foo bar     foo bar
+    foo\\sbar     foo bar    regexp=${True}
```
pekkaklarck, line 134: 1. I think it would be better to add separate `Get File with regexp` test. Then it needs to be added to matching file on the `atest/robot` side as well.
2. No need to use `${True}`, just `True` will work. I generally use both and also variants like `true` and even `yes` if such a flag is needed multiple times. No need for separate tests, though.

atest/testdata/standard_libraries/operating_system/get_file.robot
```diff
@@ -131,8 +141,11 @@ Grep File non Ascii
     [Setup]    Create File    ${TESTFILE}    fää\nbär\nföö bär\n\nA Fåå
     [Template]    Grep And Check File
     fää     fää        ${TESTFILE}
+    fää     fää        ${TESTFILE}    regexp=${True}
     ö       föö bär    ${TESTFILE}
+    .*ö     föö bär    ${TESTFILE}    regexp=${True}
     A       A Fåå      ${TESTFILE}
+    A       A Fåå      ${TESTFILE}    regexp=${True}
```
pekkaklarck, line 148: 1. Separate regexp test might be a good idea here as well.
2. Should test that `\w` matches non-ASCII characters. Possibly `fää` could be changed to `f\\wä` (note double `\\` needed by Robot's escaping rules).

atest/testdata/standard_libraries/operating_system/get_file.robot
```diff
@@ -169,8 +183,8 @@ Get And Check File
     Should Be Equal    ${content}    ${expected}
 
 Grep And Check File
-    [Arguments]    ${pattern}    ${expected}    ${test FILE}=${UTF-8 LONG FILE}
-    ${content} =    Grep File    ${test FILE}    ${pattern}
+    [Arguments]    ${pattern}    ${expected}    ${test FILE}=${UTF-8 LONG FILE}    ${regexp}=${False}
+    ${content} =    Grep File    ${test FILE}    ${pattern}    regexp=${regexp}
```
pekkaklarck, line 187: Could also use `&{config}` both in the signature and when calling the keyword to support any kind of configuration. Now that there's only `regexp` option, this approach works as well.

pekkaklarck (changes requested):
Looks good and I especially like the idea to translate glob patterns to regexps as well. I added several comments but most of them are related to code/test style and documentation.

The only open question is that should regexp patterns match lines fully or should they support partial match. I somewhat strongly believe that latter is better approach for consistency reasons and otherwise and that requires code change. Actually matching fully would also need code change because the current implementation requires lines only to start with the pattern, not to match it fully.

src/robot/libraries/OperatingSystem.py
```diff
@@ -286,34 +287,36 @@ def grep_file(self, path, pattern, encoding='UTF-8', encoding_errors='strict'):
...
         This keyword supports special ``SYSTEM`` and ``CONSOLE`` encodings that
         `Get File` supports only with Robot Framework 4.0 and newer. When using
         Python 3, it is possible to use ``${NONE}`` instead of ``SYSTEM`` with
         earlier versions.
         """
         path = self._absnorm(path)
-        pattern = '*%s*' % pattern
+        if is_falsy(regexp):
+            pattern = fnmatch.translate('*%s*' % pattern)
+        reobj = re.compile(pattern)
```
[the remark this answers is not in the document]
JockeJarre: The * around the pattern for the re.match pattern is still needed, otherwise it does not trigger when switching to search.
The hint about re.search, was nice though! 
The comment about is_falsy noted.

src/robot/libraries/OperatingSystem.py
```diff
@@ -286,34 +287,36 @@ def grep_file(self, path, pattern, encoding='UTF-8', encoding_errors='strict'):
...
         encoding = self._map_encoding(encoding)
         lines = []
         total_lines = 0
         self._link("Reading file '%s'.", path)
         with open(path, encoding=encoding, errors=encoding_errors) as f:
-            for line in f.readlines():
+            for line in f:
```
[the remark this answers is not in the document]
JockeJarre: Ok, understood and changed.

src/robot/libraries/OperatingSystem.py
```diff
@@ -286,34 +287,36 @@ def grep_file(self, path, pattern, encoding='UTF-8', encoding_errors='strict'):
...
                 total_lines += 1
                 line = line.rstrip('\r\n')
-                if fnmatch.fnmatchcase(line, pattern):
+                if reobj.match(line):
```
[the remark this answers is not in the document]
JockeJarre: I still need to check if fnmatchcase translates that good to the regular expression?

src/robot/libraries/OperatingSystem.py
JockeJarre: Did some updates to the documentation, should be clearer now

atest/testdata/standard_libraries/operating_system/get_file.robot
JockeJarre: The tests are now split up.

I do not know if I do these comments in the right way though. How do I get my changes into the PR?

src/robot/libraries/OperatingSystem.py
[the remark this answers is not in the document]
JockeJarre: The reason is because `fnmatch.translate` adds a \Z to the regular expression, matching only end of string.

`>>> import fnmatch, re
>>>
>>> regex = fnmatch.translate('*.txt')
>>> regex
'(?s:.*\\.txt)**\\Z**'
`

src/robot/libraries/OperatingSystem.py
```diff
@@ -286,34 +287,36 @@ def grep_file(self, path, pattern, encoding='UTF-8', encoding_errors='strict'):
...
 
         This keyword supports special ``SYSTEM`` and ``CONSOLE`` encodings that
         `Get File` supports only with Robot Framework 4.0 and newer. When using
         Python 3, it is possible to use ``${NONE}`` instead of ``SYSTEM`` with
         earlier versions.
         """
         path = self._absnorm(path)
-        pattern = '*%s*' % pattern
+        if not regexp:
+            pattern = fnmatch.translate(f'{pattern}*')
```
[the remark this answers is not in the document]
pekkaklarck: I was wrong, the trailing `*` is added because `translate` adds `\Z` denoting the end of the string at the end of the pattern automatically.

Issue #4115, opened 2021-10-08
New `format` option to `Log To Console` to control alignment, fill characters, and so on


IlfirinPL:
In some cases its look nice to print msg on center of screen with some padding

default values for new arguments
align=left    
padding=' '

example use after changes
Log To Console    ${msg}    align=center    padding=*

current workaround but it`s not easy to get actual screen width
Log To Console    ${msg.center(78, '*')}

links
https://www.programiz.com/python-programming/methods/string/center
https://docs.python.org/3/library/stdtypes.html#str.center

pekkaklarck:
We cannot expose only console width but exposing all command line options somehow is possible. Not sure should there be a built-in variable like `${SETTINGS}` or some API to get this info. We may already have an issue about that and if not, one can be submitted. This issue can be about possibly adding formatting option to `Log To Console`. Getting either of them implemented in foreseeable future will require external contribution. I have other issues to look at for RF 5.0 that have a lot higher priority.

pekkaklarck:
Anyone interested to add `format` option to `Log To Console`? As explained in the [docs](https://docs.python.org/3/library/string.html#formatspec), it would support alignment, fill characters, and so on. Implementation ought to be very easy so this is a good first issue for anyone interested to contribute!

makeevolution:
@pekkaklarck I'm interested to implement this, do you mind assigning me?

pekkaklarck:
Go ahead @makeevolution! Just create a PR and I'll review it.

nick-grout1:
Hi @makeevolution , I was wondering what the status of this was - I would be happy to take this issue on it sounds really interesting.

Just let me know if you don't have time for this issue/would like me to take it on. 

Thank you!

makeevolution:
Hey @nick-grout1 thanks for the offer but I'm still actively working on it, already made pull requests and working on the feedback now

pekkaklarck:
Implemented in PR #4162 by @makeevolution. Thanks for contribution!

PR #4162, opened 2021-12-05
Format option for "log to console" issue #4115

makeevolution:
Hi,

This is my pull request to solve issue #4115. Please let me know what to improve/change. If you like it I can edit the doc string on the function to reflect the new change


src/robot/libraries/BuiltIn.py
```diff
@@ -2964,6 +2964,8 @@ def log_to_console(self, message, stream='STDOUT', no_newline=False):
         This keyword does not log the message to the normal log file. Use
         `Log` keyword, possibly with argument ``console``, if that is desired.
         """
+        format = "{:" + format + "}"
+        message = format.format(message)
```
pekkaklarck, line 2968: 1. It would be a bit safer to add `if format:` before these two lines. It ought to be always safe to call `{:}.format(message)`, but because `message` can be anything, not only a string, there's a small risk it could blow up.
2. This new functionality needs documentation. The docs should contain a link to https://docs.python.org/3/library/string.html#formatspec and a new example should be added.
3. Docs also need a note that `format` is new in Robot Framework 5.0.

src/robot/libraries/BuiltIn.py
```diff
@@ -2944,7 +2944,7 @@ def _yield_logged_messages(self, messages):
             else:
                 yield value
 
-    def log_to_console(self, message, stream='STDOUT', no_newline=False):
+    def log_to_console(self, message, stream='STDOUT', no_newline=False, format = ""):
```
pekkaklarck, line 2947: Please change this to `format=''`. Spaces around `=` are against PEP-8 and we have generally used `'` instead of `"` in our code base. The latter isn't too big a deal, but using different quotes in arguments (there's `'STDOUT'` already) would be somewhat inconsistent.

atest/robot/output/format_log_to_console.robot
```diff
@@ -0,0 +1,15 @@
...
+    Log to console  hello console  format=*^78
+Right align log to console with hastag padding
+    Log to console  test123$#@!$  format=#>32
+Right align log to console with space padding
+    Log To Console  +000000120\./asdf  format=>30
+Test if formatting works with no_newline argument
+    Log To Console  message starts here,  format=>44  no_newline=true
+    Log To Console  this sentence should be on the same sentence as "message starts here"
+Test if formatting works with stderr argument
+    Log to console  log to stderr  format=>44  stream=STDERR
```
pekkaklarck, line 15: These tests belong here:
https://github.com/robotframework/robotframework/blob/master/atest/testdata/standard_libraries/builtin/log.robot

Then you also need to add matching test to the `atest/robot` side to validating the logged output. See [atest/README.rst](https://github.com/robotframework/robotframework/blob/master/atest/README.rst) for more information about our acceptance system. If you need help with it, don't hesitate to ask here or on Slack.

pekkaklarck (changes requested):
Great work but docs and tests still need enhancements.

makeevolution:
Hi @pekkaklarck I have implemented the changes you requested. Please let me know if there's more I need to do.

makeevolution:
Hi @pekkaklarck any news on this PR perhaps?

atest/testdata/standard_libraries/builtin/log.robot
```diff
@@ -163,3 +163,10 @@ Log To Console
     Log To Console    stderr äö w/ newline    stdERR
     Log To Console    ...line äö   stdout    continue without newlines
     Log To Console    ${42}
+    Log to console    test middle align with star padding    format=*^60
+    Log to console    test right align with hash padding    format=#>32
+    Log To Console    test-with-spacepad-and-weird-characters+%?,_\>~}./asdf    format=>30
+    Log To Console    message starts here,    format=>44  no_newline=true
+    Log To Console    this sentence should be on the same sentence as "message starts here"
+    Log to console    test log to stderr    format=>44  stream=stdERR
+
```
pekkaklarck, line 172: Perhaps a separate test like "Log To Console with formatting" should be added with these changes. There could be such a test also on the `atest/robot` side.

There's also some strange whitespace at the end that should be removed.

src/robot/libraries/BuiltIn.py
```diff
@@ -2944,26 +2944,38 @@ def _yield_logged_messages(self, messages):
...
         By default appends a newline to the logged message. This can be
         disabled by giving the ``no_newline`` argument a true value (see
         `Boolean arguments`).
 
+        By default adds no alignment formatting. The format argument can be 
+        used to set the alignment of the log message, along with adding 
+        custom padding. Please see 
+        https://docs.python.org/3/library/string.html#formatspec for
+        detailed alignment possibilities. This argument is new in Robot
+        Framework 5.0.
```
pekkaklarck, line 2963: - A link, not url, would be a bit nicer.
- Argument name should be in double backtics.

Something like this perhaps:

    The ``format`` argument can be used to format the logged message according to the
    [https://docs.python.org/3/library/string.html#formatspec|format specification]. It allows,
    for example, alignment and padding. This argument is new in Robot Framework 5.0.

You can see how the generated documentation looks like by installing `invoke` and then running

    invoke library-docs builtin

src/robot/libraries/BuiltIn.py
```diff
@@ -2944,26 +2944,38 @@ def _yield_logged_messages(self, messages):
...
+
         Examples:
         | Log To Console | Hello, console!             |                 |
         | Log To Console | Hello, stderr!              | STDERR          |
         | Log To Console | Message starts here and is  | no_newline=true |
         | Log To Console | continued without newline.  |                 |
+        | Log To Console | center message with * pad   | format=*^60     |
+        | Log To Console | 30 spaces before msg starts | format=>30      |
```
pekkaklarck, line 2971: Good examples!

atest/robot/standard_libraries/builtin/log.robot
```diff
@@ -193,10 +193,16 @@ Log Many with dict variable containing non-dict
...
     END
     Stdout Should Contain    stdout äö w/ newline\n
     Stdout Should Contain    stdout äö w/o new......line äö
     Stderr Should Contain    stderr äö w/ newline\n
     Stdout Should Contain    42
+    Stdout Should Contain    ************test middle align with star padding*************
+    Stdout Should Contain    ####################test right align with hash padding
+    Stdout Should Contain                  test-with-spacepad-and-weird-characters+%?,_\>~}./asdf
+    Stdout Should Contain                            message starts here,this sentence should be on the same sentence as "message starts here"
+    Stderr Should Contain                                   test log to stderr
```
pekkaklarck, line 207: Using spaces here doesn't work, there's just a long separator between the Stdout Should Contain keyword and its arguments. To verity there's actually spaces in output, use e.g. `${SPACE * 44}`.

pekkaklarck (commented):
Looks good. Could probably be merged already as-is, but there are few things that could still be enhanced.

makeevolution:
@pekkaklarck I've updated the branch with your comments, please let me know what you think :)

pekkaklarck:
Thanks for contribution @makeevolution!

Issue #3944, opened 2021-04-22
Request for `log_property` functionality for xunit writer


MrKevinWeiss:
I would like to request a log property capability in order to story key values in the xunit output.

### Reasoning

The output.xml is very verbose and makes it a bit of a challenge when wanting use just some test results with benchmarks.
xunit (or junit) provides a nice standard way of viewing only the test result (pass, fail, skip) and does allow for key values to be stored in properties.

### Background

It seems like pytest supports [`record_property`](https://docs.pytest.org/en/stable/reference.html#record-property).
Many [junit/xunit schemas](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd#L56) include this in.
We currently are doing [post processing](https://github.com/RIOT-OS/RobotFW-tests/blob/e21fc5113320ba17aa110e353e233875f5559b9f/dist/tools/output_to_xunit/output_to_xunit.py) on the output.xml file and have wrapper keyword using [`LOG`](https://github.com/RIOT-OS/RobotFW-tests/blob/e21fc5113320ba17aa110e353e233875f5559b9f/dist/robotframework/res/util.keywords.txt#L12)

<details><summary>Example of xunit with properties</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
   <testsuite name="JUnitXmlReporter" errors="0" tests="0" failures="0" time="0" timestamp="2013-05-24T10:23:58" />
   <testsuite name="JUnitXmlReporter.constructor" errors="0" skipped="1" tests="3" failures="1" time="0.006" timestamp="2013-05-24T10:23:58">
      <properties>
         <property name="java.vendor" value="Sun Microsystems Inc." />
         <property name="compiler.debug" value="on" />
         <property name="project.jdk.classpath" value="jdk.classpath.1.6" />
      </properties>
      <testcase classname="JUnitXmlReporter.constructor" name="should default path to an empty string" time="0.006">
         <failure message="test failure">Assertion failed</failure>
      </testcase>
      <testcase classname="JUnitXmlReporter.constructor" name="should default consolidate to true" time="0">
         <skipped />
      </testcase>
      <testcase classname="JUnitXmlReporter.constructor" name="should default useDotNotation to true" time="0" />
   </testsuite>
</testsuites>
```

Thanks to https://stackoverflow.com/questions/4922867/what-is-the-junit-xml-format-specification-that-hudson-supports
</details>

[...]

PR #4165, opened 2021-12-08
Draft only: Various properties to xunit output

rikerfi:
This is a draft PR for solving issue #3944 Request for log_property functionality for xunit writer.
Now also tests implemented.

### DIFF
--- a/atest/robot/output/xunit.robot
+++ b/atest/robot/output/xunit.robot
@@ -1,13 +1,21 @@
 *** Settings ***
 Documentation       Tests for xunit-compatible xml-output.
 Resource            atest_resource.robot
+Resource            rebot_resource.robot
 Variables           unicode_vars.py
-Suite Setup         Run Tests    -x xunit.xml -l log.html --skiponfailure täg    ${TESTDATA}
+Suite Setup         Run Keywords
+...    Create Output With Robot    ${OUT ONE}    ${EMPTY}    ${PASS AND FAIL}
+...    AND
+...    Create Output With Robot    ${OUT TWO}    ${EMPTY}    ${TESTDATA}
+...    AND
+...    Run Tests    -x xunit.xml -l log.html --skiponfailure täg    ${TESTDATA}
 
 *** Variables ***
 ${TESTDATA}         misc/non_ascii.robot
 ${PASS AND FAIL}    misc/pass_and_fail.robot
 ${INVALID}          %{TEMPDIR}${/}ïnvälïd-xünït.xml
+${OUT ONE}          %{TEMPDIR}${/}out1.xml
+${OUT TWO}          %{TEMPDIR}${/}out2.xml
 
 *** Test Cases ***
 XUnit File Is Created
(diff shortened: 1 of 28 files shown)


src/robot/reporting/xunitwriter.py
```diff
@@ -57,6 +60,14 @@ def _get_stats(self, statistics):
         )
 
     def end_suite(self, suite):
+        if suite.metadata or suite.doc:
+            self._writer.start('properties')
+            if suite.doc:
+                self._writer.element('property', attrs={'name': 'Suite Documentation', 'value': suite.doc})
```
pekkaklarck, line 66: Just `Documentation` would probably be better than `Suite Documentation`. We use that in the data as well.

src/robot/reporting/xunitwriter.py
```diff
@@ -57,6 +60,14 @@ def _get_stats(self, statistics):
...
+            if suite.metadata:
+                for n, v in suite.metadata.items():
+                    self._writer.element('property', attrs={'name': n, 'value': v})
+            self._writer.end('properties')
```
pekkaklarck, line 70: No need for `if` here, it would be fine to just loop over possible metadata items. Using `name` and `value´ instead of `n` and `w` would be a bit better as well. We've generally used one character variable names only in very tight scopes such as in list comprehensions.

atest/robot/output/xunit.robot
```diff
@@ -33,6 +31,8 @@ File Structure Is Correct
     Element Attribute Should Be    ${skips}[0]    message
     ...    Test failed but its tags matched '--SkipOnFailure' and it was marked skipped.\n\nOriginal failure:\n${MESSAGES}
     Element Attribute Should Be    ${skips}[0]    type    SkipExecution
+    Run Keyword And Expect Error    No element matching 'testsuite/properties' found.
+    ...    Get XUnit Node    testsuite/properties
```
pekkaklarck, line 35: Using `Element Should Not Exist` would be better than checking error message.

atest/robot/output/xunit.robot
```diff
@@ -112,3 +157,20 @@ Suite Stats Should Be
...
+Get Root Node
+    ${root} =    Get XUnit Node
+    Should Be Equal    ${root.tag}    testsuite
+    [Return]    ${root}
+
+Get Properties Node
+    [Arguments]    ${source}
+    ${properties} =    Get Elements    ${source}    properties
+    Length Should Be    ${properties}    1
+    [Return]    ${properties}
```
pekkaklarck, line 176: You can use new `RETURN` here instead of ´[Return]`.

src/robot/reporting/xunitwriter.py
```diff
@@ -37,6 +37,7 @@ class XUnitFileWriter(ResultVisitor):
 
     def __init__(self, xml_writer):
         self._writer = xml_writer
+        self.root = None
```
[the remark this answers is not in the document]
rikerfi: Good point. Not Anymore

pekkaklarck:
Apparently #4193 has replaced this PR.

Issue #4161, opened 2021-12-04
Add new builtin tags `robot:exclude`, `robot:skip` and `robot:skip-on-failure`


Snooz82:
Having a permanent skipping/exclusion of the tags `robot:skip` and `robot:exclude` would ease the marking of testcases that are temporarily excluded from execution.

For example tests, that has to be maintained, due to changed test object or in a concrete request of a user, to mark DataDriven Tests as skip for a certain use-case.

That would add the comfort, that users do not have to add `--skip robot:skip` and `--exclude robot:exclude` to the command line.

Would be nice for 5.0, because there is a real request for this. see [DataDriver Issue 68](https://github.com/Snooz82/robotframework-datadriver/issues/68)

[Slack discussion](https://robotframework.slack.com/archives/CJSNFRZMK/p1638581281028000)

pekkaklarck:
Sounds good to me. Should also add `robot:skip-on-failure` at the same time.

makeevolution:
@pekkaklarck would you mind if I try this out? I also would like to join the slack channel, but it requires an talousverkko.fi email address. Do you know how can I make one?

pekkaklarck:
Go ahead @makeevolution! No special email should be needed to join the Slack community. You can get an invite at https://rf-invite.herokuapp.com/.

nick-grout1:
Hi @makeevolution , I was wondering what the status of this was - I would be happy to pick this up if you're unable to work on this at the moment.

Just let me know if I can take this on, thank you!

makeevolution:
Hey @nick-grout1 thanks for the offer but I'm still actively working on it, already made pull requests and working on the feedback now

tw39124-1:
Upvote for a `robot:skip-on-failure` tag. I'm using `--skiponfailure` at the moment but it's hard to only apply to a subset of test cases. Being able to use a tag to do it would be great.

pekkaklarck:
This was implemented in PR #4172 by @makeevolution. Thanks for contribution!

pekkaklarck:
I just noticed that if you all tests in a suite are excluded with `robot:exclude`, the suite itself is still executed without any tests. This is different compared to when using `--exclude` because then empty suites are pruned automatically. The problem is due to `robot:exclude` being inspected only at execution time when the suite has already been started. Filtering using `--exclude` and other such options is done already before the whole execution starts and thus empty suites can be removed.

[...]

PR #4172, opened 2021-12-17
Add new builtin tags issue #4161

makeevolution:
Hi @pekkaklarck this is my pull request for issue #4161. Please let me know what you think.

I still am stuck with trying to implement the one for robot:exclude, since if the test is already in the stack that runs suiterunner.py, it means the test is already included in the suite and thus shall be in the report itself. But exclude means the test is as if it doesn't exist. I still am struggling with finding a way to read the tags of the test cases in a test suite without invoking the suiterunner. But it seems I need to invoke the suiterunner to get access to tags. I hope you get my conundrum/dillemna :) . Maybe you have a suggestion on this?


src/robot/running/status.py
```diff
@@ -202,6 +201,7 @@ def _skip_on_failure(self):
         critical = not critical_pattern or critical_pattern.match(tags)
         skip_on_fail_pattern = TagPatterns(self._skip_on_failure_tags)
         skip_on_fail = skip_on_fail_pattern and skip_on_fail_pattern.match(tags)
+        skip_on_fail = TagPatterns("robot:skip-on-failure").match(tags)
```
pekkaklarck, line 204: Overwriting earlier `skip_on_fail` variable certainly cannot be right. I'm surprised if there aren't tests that fail due to that. It would probably be easiest to just add `robot:skip-on-failure` at the end of `skip_on_failure_tags` either here or already in `__init__`.

src/robot/running/suiterunner.py
```diff
@@ -125,6 +125,11 @@ def visit_test(self, test):
         if status.exit:
             self._add_exit_combine()
             result.tags.add('robot:exit')
+        if TagPatterns("robot:skip").match(test.tags):
+            status.test_skipped(
+                test_or_task(
+                    "{Test} skipped since test is tagged with robot:skip.",
+                    self._settings.rpa))
```
pekkaklarck, line 132: Cannot use word `test` as-is in the message because we want to use `task` in RPA mode. That `test_or_task` helper converts placeholders like `{Test}` to right format automatically, but I wouldn't repeat test/task in one sentence anyway. Perhaps something like `{Test} skipped because it had 'robot:skip' tag.` would be fine.

atest/robot/running/skip.robot
```diff
@@ -118,6 +118,12 @@ Skipped with --SkipOnFailure when Failure in Test Teardown
 Skipped with --SkipOnFailure when Set Tags Used in Teardown
     Check Test Case    Skipped with --SkipOnFailure when Set Tags Used in Teardown
 
+Skipped when test is tagged with robot:skip
+    Check Test Case    ${TEST NAME}    status=SKIP
+
+Skipped although test fails since test is tagged with robot:skip-on-failure
+    Check Test Case    ${TEST NAME}    SKIP
+
```
pekkaklarck, line 126: - No need to pass status to `Check Test Case` because tests in the `atest/tesdata` have it in their documentation.
- First test could be right after `Skipped with --skip` test because they are highly related. The latter could be here after various `--skiponfailure` tests. Test on the data side should be organized the same way.

atest/testdata/running/skip/skip.robot
```diff
@@ -248,6 +248,20 @@ Skipped with --SkipOnFailure when Set Tags Used in Teardown
     Fail    Ooops, we fail!
     [Teardown]    Set Tags    skip-on-failure
 
+Skipped when test is tagged with robot:skip
+    [Documentation]    SKIP
+    ...    Test skipped since test is tagged with robot:skip.
+    [Tags]    robot:skip
+
```
pekkaklarck, line 255: Would be good to add something like `Fail    Not executed` to the body of the test. Empty test is an error situation and although apparently now test is skipped before that error occurs, this behavior could easily change in the future.

pekkaklarck (changes requested):
Looks pretty good. See separate comments about things that should be changed in code and could be changed in tests.

pekkaklarck:
`robot:exclude` is easiest to implement like this:
```diff
diff --git a/src/robot/model/filter.py b/src/robot/model/filter.py
index b5487f33c..58998d78c 100644
--- a/src/robot/model/filter.py
+++ b/src/robot/model/filter.py
@@ -44,7 +44,7 @@ class Filter(EmptySuiteRemover):
         self.include_suites = include_suites
         self.include_tests = include_tests
         self.include_tags = include_tags
-        self.exclude_tags = exclude_tags
+        self.exclude_tags = tuple(exclude_tags or ()) + ('robot:exclude',)
```

pekkaklarck:
This functionality also needs documentation in the User Guide. Let me know if you need help getting started with it.

makeevolution:
Hi @pekkaklarck I updated the files with the new changes and also added the functionality to the documentation, all the tests passed now in my local machine but except the include-and-exclude atest for robot:exclude. I uploaded my attempt anyway; could you help me with how to write this test?

atest/robot/tags/include_and_exclude.robot
```diff
@@ -73,6 +73,9 @@ Include with NOT
 Exclude with NOT
     --exclude excl1NOTexcl2ANDexcl3    Excl-123    @{INCL_ALL}
 
+Exclude with robot:exclude tag
+    --tag no-use    Robot-exclude
+
```
pekkaklarck, line 78: I assume this is the test that's failing? Tests in this suite validate what tests were actually run. The new test with `robot:exclude` should never be executed so testing that it's run isn't possible. That it's not run is actually tested already by other tests in this suite. For example, the first fest validates that all tests are run when `--include/--exclude` aren't used and the new test isn't run even in that case.

It would thus be enough to just add the new test with `robot:exclude` to the `atest/testdata` side, but it would nevertheless be nice to somehow make it clear on the `atest/robot` side that this functionality is tested. Possibly it would be enough to just add a documentation to the first test that the test with `robot:exclude` is always excluded automatically.

atest/testdata/tags/include_and_exclude.robot
```diff
@@ -14,6 +14,10 @@ Incl-123
     [Tags]  incl_1  incl_2  incl_3
     No Operation
 
+Robot-exclude
+    [Tags]  robot:exclude
+    Fail  test not executed
+
```
pekkaklarck, line 20: This could be at the end of the suite. There could also be two tests using the tag in different variations like `robot:exclude` and `ROBOT: EXCLUDE`.

doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
```diff
@@ -130,6 +130,14 @@ combining individual tags or patterns together::
...
    --include fooNOTbar
 
+It is also possible to use the reserved tag `robot:exclude` to achieve
+the same effect as with using the `--exclude` option::
+
+    *** TEST CASES ***
+    Test Case
+        [Tags]    robot:exclude
+        Fail      this test will be excluded
+
```
pekkaklarck, line 140: 1. Should mention this functionality is new in RF 5.0. Could add an explicit note about that, but perhaps here the sections could just start like "Staring from Robot Framework 5.0, it is also possible ...".
2. Robot code should be formatted. You can use `.. sourcecode:: robotframework` directive to enable that.
3. We have used `*** Test Cases ***` style elsewhere and for consistency it would be good here as well.
4. Perhaps the test could be named "Example".

doc/userguide/src/ExecutingTestCases/TestExecution.rst
```diff
@@ -215,6 +215,14 @@ specified tags or tag patterns are skipped::
...
     --skip python2.* --skip python3.[0-6]
 
+In addition, a test case can also be skipped by tagging the test with the 
+reserved tag `robot:skip`::
+
+    *** TEST CASES ***
+    Test Case
+        [Tags]    robot:skip
+        Log       this message is not logged
+
```
pekkaklarck, line 225: See comments to `robot:exclude` documentation.

doc/userguide/src/ExecutingTestCases/TestExecution.rst
```diff
@@ -254,6 +262,14 @@ the :option:`--skip` option discussed above::
...
     --skiponfailure experimentalANDmobile
 
+In addition, the reserved tag `robot:skip-on-failure` can alternatively be used to
+achieve the same effect as above::
+
+    *** TEST CASES ***
+    Test Case
+        [Tags]    robot:skip-on-failure
+        Fail      this test will be marked as skipped instead of failed
+
```
pekkaklarck, line 272: See comments to `robot:exclude` documentation.

src/robot/running/status.py
```diff
@@ -168,6 +168,7 @@ def __init__(self, parent, test, skip_on_failure=None, critical_tags=None,
         super().__init__(parent)
         self._test = test
         self._skip_on_failure_tags = skip_on_failure
+        self._skip_on_failure_tags.append('robot:skip-on-failure')
```
pekkaklarck, line 171: `skip_on_failure` can be `None` (that's the default!) or a tuple so using `append` isn't safe. Two lines above could be replaced with this:
```python
self._skip_on_failure_tags = tuple(skip_on_failure or ()) + ('robot:skip-on-failure',)
```

src/robot/running/suiterunner.py
pekkaklarck: Above comment already handled. Two nitpicky things to still change:
- Add `.` at the end of the sentence.
- You can use `"` around the string to be able to use `'` in the string without escaping like `"{Test} ... 'robot:skip' tag."`

pekkaklarck (changes requested):
Looks really good! Some smallish changes still needed, though.

makeevolution:
@pekkaklarck I've implemented all the fixes!

pekkaklarck (approved):
Looks great! Assuming tests pass on CI (I just enabled running them) I'll merge this as-is.

pekkaklarck (changes requested):
Running tests unfortunately uncovered multiple problems:
1. It isn't possible to simply add `robot:exclude´to the list of tags given using `--exclude`. The reason is that using `--exclude` changes filtering logic with Rebot and also otherwise and with that change it effectively was always set. This broke several acceptance tests.
2. When starting execution programmatically, it is possible to give a single excluded tags using `exclude='tag'` in addition to `exclude=['tag']`. Thus blindly converting given tags to a tuple like `tuple(excluded_tags or ())` doesn't work. This broke one unit test.
3. Acceptance test for `robot:skip` fails due to the actual message not exactly matching the expected message.

The first problem is probably easiest to fix by reverting changes related to `robot:exclude` and just dropping tests if they have that tag using same approach they are skipped if they have ´robot:skip`. In practice adding the following to the beginning of `visit_test` in `suiterunner.py` ought to be enough:
```python
if TagPatterns("robot:exclude").match(test.tags):
    return
```

The above change fixes programmatic usage with `exclude` as well. The failing `robot:skip` tests would need to be fixed as well, though.

After going these changes, please run all acceptance tests to make sure they are passing. See [atest/README.rst](https://github.com/robotframework/robotframework/tree/master/atest#readme) for more details if needed.

doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
```diff
@@ -130,6 +130,17 @@ combining individual tags or patterns together::
...
+Starting from RF 5.0, it is also possible to use the reserved 
+tag `robot:exclude` to achieve
+the same effect as with using the `--exclude` option:
+
+.. sourcecode:: robotframework
+
+   *** Test Cases ***
+   Example
+      [Tags]    robot:exclude
+      Fail      this test will be excluded
```
pekkaklarck, line 142: This message is pretty confusing considering the test is _not_ executed. Adding the word "not" to the message or changing the whole `Fail    this test ...` to `This is not executed` would be good.

atest/testdata/tags/include_and_exclude.robot
```diff
@@ -25,3 +25,7 @@ Excl-12
 Excl-123
     [Tags]  excl_1  excl_2  excl_3
     No Operation
+
+Robot-exclude
+    [Tags]  robot:exclude  ROBOT:EXCLUDE
+    Log  this test will never be run
```
pekkaklarck, line 31: - Could use `Fail`, not `Log` to make it more explicit this isn't run.
- Newline missing at the end of the file.

src/robot/model/filter.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, include_suites=None, include_tests=None,
         self.include_tests = include_tests
         self.include_tags = include_tags
         self.exclude_tags = exclude_tags
-
+
```
pekkaklarck, line 48: Accidentally added whitespace. Please remove.

PR #4188, opened 2022-01-17
fix #4181

### LINKED ISSUE #4181: Parsing Argument Setting of Keywords with line continuation and no argments, throws an IndexError
When I run this piece of code:

```robotframework
*** Keywords ***
test
    [Arguments]  
    ...    
```

robot crashes with an:

```
[ ERROR ] Unexpected error: IndexError: string index out of range
 File "c:\develop\robot\robotframework\src\robot\utils\application.py", line 83, in _execute
    rc = self.main(arguments, **options)

...
  File "c:\develop\robot\robotframework\src\robot\running\arguments\argumentparser.py", line 118, in parse
    elif self._is_kwargs(arg):
  File "c:\develop\robot\robotframework\src\robot\running\arguments\argumentparser.py", line 215, in _is_kwargs
    return arg[0] == '&'
```

this is happening with RF 4.1.3 and with the current main branch

Tried to investigate what happens and stumbled upon this line:

https://github.com/robotframework/robotframework/blob/ac4803649702ada9d18c5357ad7fcbbebcd0fb89/src/robot/parsing/lexer/tokenizer.py#L80

If I comment out this line, everything works fine, some testcases are broken, those that wanted to have these extra token in it. But there is no testcase that explicitly test this extra token.

The problem is, that `_is_kwargs` Method (see traceback) get's an empty string as argument, and throws an error, because there this extra empty token is added by `_ensure_data_after_continuation` Method and later then, it is converted to an ARGUMENT token.

Is adding this extra token after a line continuation really nessessary?

I can provide a PR, but wanted to ask first for other opinions.

d-biehl:
try to fix #4181


pekkaklarck:
Fix looks good but tests are needed.

d-biehl:
I'm working on it, but first family weekend 😉

pekkaklarck:
Something strange has happened with this PR most likely when rebasing it. According to the tab on the top, there are changed to 162 files which obviously cannot be true.

d-biehl:
created a new PR #4206 with only the changes for the fix
please reject this PR

pekkaklarck:
New PR merged!

PR #4148, opened 2021-11-18
improvements for UserGuide for #4149

Snooz82:
Updated css.

Features:
- Nicer Font
- Sidebar with TOC in 3 levels (this is depth 3 in rst file)
- If Browser to small TOC with 2 levels at the beginning
- DarkMode
- limited width for better readability
- looks good :-) 

Issue #4149

PR #4190, opened 2022-01-18
Dark mode for report and log

psaikkonen:
### Draft PR for adding dark mode for report and log files

I have replaced most of the color definitions with CSS variables, and added a dark mode color palette based on the user guide. While I was working on this I noticed there is an open pull request https://github.com/robotframework/robotframework/pull/4148 for adjusting the looks of the user guide, but the palette is easy to replace as it is just a relatively small bunch of variables defined in common.css.

---

Colors of icons are [inverted](https://github.com/robotframework/robotframework/compare/master...psaikkonen:master#diff-6515006c7bb31bab0f022fec942192c8269520b9499ad15f14ca4e0a0af423b3R11-R12) with `filter: invert(1)` when dark mode is enabled. Active sorting icons are the only exception - they have [modified versions of the icons for dark mode](https://github.com/robotframework/robotframework/compare/master...psaikkonen:master#diff-c69b4c09f9de76eff72c595f34c4abe737208bde6b3ed7418f8717b503232b05R25-R26) where the toggled direction is highlighted using white instead of black.

---

There are small changes for dropdown lists and input fields & buttons that affect the light mode as well. Earlier these were using browser defaults but now their styles are defined in the stylesheets. See the screenshots below.

---

Background colors of the report page are a bit tricky and I am not sure what to do with them. Haven't done any changes for them yet, and at least the default yellow color (skipped) makes texts unreadable and the other two are painful to look at as well. 😄 

As these colors are user configurable I was wondering whether the defaults should be adjusted so that they would possibly work on both light and dark themes, or should there be a separate setting for dark mode background color which the user can change as well? 

Or maybe even the whole feature could be redesigned: is it necessary that the whole page is screaming in red, maybe just the header containing title and timestamp would be enough to quickly visualize the results?

---

Here are some screenshots of how it is looking currently, comments & ideas are welcome:

Screenshot 2022-01-18 at 21-32-50 Test Log
Screenshot 2022-01-18 at 21-33-00 Test Log

Screenshot 2022-01-18 at 21-34-06 Test Report
Screenshot 2022-01-18 at 21-34-13 Test Report

Screenshot 2022-01-18 at 21-34-39 Test Report
Screenshot 2022-01-18 at 21-34-50 Test Report


pekkaklarck:
Great, I'm more than happy with log and report styles being enhanced. Some important things to take into account:
- We cannot support only the latest browser versions. Some users are using RF on internal networks where browsers are ancient. I cannot say exactly where to draw the line, though.
- Color blindness needs to be taken into account. We've tried to do that with the current colors as well.
- Resulting code/styles need to be easy to maintain going forward.

It might be best to discuss these enhancements with a wider audience and the #devel channel on our Slack is probably the best place. I'm sure at least @Snooz82 and @samipe are interested on this.

pekkaklarck:
FWIW, I'm fine with report background not being red/green/yellow depending on the status as long as there's some other clear visual indication about the overall status.

kirananto:
To your point @pekkaklarck We should also take into consideration the accessibility, 

Right now the background being red --> Doesn't really help alone, We could do lot of enhancements on the visual side to make it visually appealing to customers, while not breaking any of the accessibility & Backward compatibility for the older browsers. 

We could also drop support for old browsers with a flag, that way if the flag is turned on, users will get more modern, polished Reports with better UX and modern javascript, and if not, by default they get the current one. That way we aren't breaking any functionality for ancient browsers, and users will have the freedom to switch to newer ones.

alepinio:
Really waiting for this to become true!

I'm currently using Dark Reader in my browser to keep my eyes healthy. Here's a quick look of how it renders `log.html` from https://github.com/robotframework/RobotDemo/blob/master/keyword_driven.robot in my laptop (I've introduced an error in first test case):

image

psaikkonen:
It has been quite a long time since I last worked on this, I haven't been using RF myself lately so I completely forgot this draft. However the comments popped on my GitHub feed so I thought I could re-visit this again.

I tried modifying the report page based on @TurekBot comment and added a vertical status bar which runs on the left side of the page. Here are some screenshots of how that could look like:
Screenshot 2023-09-03 at 21 03 07
Screenshot 2023-09-03 at 17 20 44
Screenshot 2023-09-03 at 22 02 43

I didn't make changes to the original default background colors yet, but as the last screenshot with light theme shows they could perhaps be more saturated on light mode. Colors can still be [configured the same way as with report background earlier](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#setting-background-colors). And since other colors are now defined in css variables it could be possible to let the user configure the whole log/report page theme by themselves in the future, but that's out of scope for this one.

As of now the status bar is pretty slim, but then again I wouldn't grow it very much either as it eats screen real estate. This might not be clear enough for visually impaired folks though. I configured the left-margin required by status bar in common.css, so that contents of Log and Report pages align nicely when moving between them. However this means that widening the bar would eat space from log file as well. 

Considering backwards compatibility, these are the features used.
[filter](https://caniuse.com/?search=filter) 
Used to invert colors of icons instead of adding new versions in the code. This is used for all icons except active sorting icons, as those required a more fine-grained tuning. This was done solely for ease-of-developments sake and could be replaced by adding a dark mode version for each icon used. I'm not sure if using filter could have an effect on performance as well when there are gazillion tests.

[prefers-color-scheme](https://caniuse.com/?search=prefers-color-scheme) 
[css-variables](https://caniuse.com/css-variables)
Fallback values still need to be added for browsers that don't support CSS variables. Since these are pure CSS features older browsers should just ignore them and use fallbacks instead, but it's something that I haven't tested yet.

I'll make a post about this on RF Slack too as @pekkaklarck suggested, to discuss about these changes with a wider audience.

pekkaklarck:
Sorry for no activity from my side lately. Quick comments now:

1. We are going to rewrite log and report in the future, hopefully sooner than later, and bigger changes to them doesn't make much sense. Adding dark mode sounds good to me, though, but I believe changes to styles otherwise should be avoided.
2. To some extend related to the above, I believe the light/normal mode should stay the same as nowadays.
3. I'm not a good person to review these changes because I don't know the used CSS well enough nor do I use the dark mode myself. I just asked could @Snooz82 or someone else take a look at this on Slack.
4. If we get @Snooz82 to review this, I'm happy to merge this once he's fine with changes. Or he can merge this himself.
5. Because there are no real Python code changes, it is possible to get this into RF 7.0 also after release candidates are out. The target date for the final RF 7.0 release is January 8.

pekkaklarck:
Thanks @Snooz82 for taking a look at this PR, fixing issues, and merging it. If you @psaikkonen have any questions related to changes he did, you can ask here or on Slack. Huge thanks for a great contribution as well!

PR #4193, opened 2022-01-19
Properties and property elements to xunit output

rikerfi:
This PR is for solving issue #3944 Request for log_property functionality for xunit writer.

Based on previous draft PR #4165


pekkaklarck:
This PR looks good and the added functionality is useful.

Few things to think about still:
1. Should we add properties also for tests? For example, pytest has [record_property](https://docs.pytest.org/en/stable/usage.html#record-property-example) that allows adding them. Our tests have documentation and tags that we could include. A problem with tags is that they are just a list and properties are name-value pairs, but I guess we could just leave the value empty.
2. The issue that initiated this PR (#3944) talks about a callable, similar to the aforementioned `record_property`, to allow logging properties. This PR doesn't do that but instead writes normal suite documentation and metadata as properties. I personally consider that a lot better than having some special callable (i.e. keyword) that would only affect xunit outputs. This PR and the issue not being in sync is nevertheless the problem. Probably best to submit a new issue about adding properties and close the old one.

PR #4197, opened 2022-01-21
fixes #4195: wrong node was expanded in case of ELSE-IF or TRY-EXCEPT branches

### LINKED ISSUE #4195: Log: Linking to warnings and `--expandkeywords` broken with IF/ELSE structures
Running 
> robot --expandkeywords name:*.Comment ExpandBug.robot

on this ExpandBug.robot:
```
*** Test Cases ***
ExpandBug
	IF  ${True}
		Log  true-ranch
	ELSE IF  ${False}
		Log  dead code
	END
	Comment  Hello
```
results in *wrong* expanded keyword *Log dead code* instead of expected keyword *Comment*:
grafik

**Reason:** Rebot correctly identifies the IF and ELSE IF lines as single keyword and thus gives the expand-directive path for the log.html as s1-t1-k2 in order to expand the 2nd keyword. Unfortunately in log space, the ELSE IF branch is now a single keyword-node, thus s1-t1-k3 would be correct.

DirkRichter:
Just root-if node was counted when generating id. All sub-nodes of IF-root was ignored. I moved the flatten method from _Builder to BodyItem in order to apply flatten here, too. Now for computation of keyword position the correct amount of body-steps is used.

This should also solve similar problems with TRY-EXCEPT branches.

### DIFF
--- a/src/robot/model/body.py
+++ b/src/robot/model/body.py
@@ -56,10 +56,22 @@ class BodyItem(ModelObject):
             return 'k1'
         setup = getattr(self.parent, 'setup', None)
         body = getattr(self.parent, 'body', ())
-        teardown = getattr(self.parent, 'teardown', None)
-        steps = [step for step in [setup] + list(body) + [teardown]
+        flatbodylist = self.flatten(list(body))
+        teardown = getattr(self.parent, 'teardown', None)        
+        steps = [step for step in [setup] + flatbodylist + [teardown]
                  if step and step.type != step.MESSAGE]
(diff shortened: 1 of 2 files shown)


DirkRichter:
**new issue found:** i wonder about crashing browser with outOfMem from time to time (using this pull-request)
now i found the reason: expanding works very well now, thus also not-executed/skipped mathing keywords will be expanded. But this is a problem for hughe test-suites, because alot of the log will be expanded then. 
We need to exclude skipped keywords. I'm working now on a refined pull-request.

Issue #3661, opened 2020-09-11
Run process hangs on large outputs


anton264:
RF-version 3.2
When doing "Run process" keyword that produces a large amount of output the keyword will hang.
This seems to be related to this issue in python 
https://stackoverflow.com/questions/39477003/python-subprocess-popen-hanging
This used to work in rf 3.1

There is a workaround to the issue:
`Run process     myVeryLargeOutputCommand    stdout=False    stderr=False`

pekkaklarck:
Quick comments:
- Process library documentation explains that `Run Process` can hang if there's lot of output and the output is not redirected to a file.
- Redirecting is done using `stdout` and `stderr` arguments but I don't understand why you set them to `False`.
- Are you certain this worked with RF 3.1? I don't remember any changes that ought to be able to affect this.

anton264:
Setting `stdout` and `stderr` to False only writes it to a file so I guess it's a bit weird to do this.

We have been running the same tests for a very long time (years) and never had a problem with `run process` before but I guess the fault is on my side for not reading the documentation. It was however after we upgraded that we started to see problems with this keyword. It might be related to other libraries though and not robots fault, hard to say.

pekkaklarck:
Any new information related to this? Can you confirm the exact same tests works in exact same environment with RF 3.1 but not with 3.2? Can you provide a simple example I could run locally to reproduce and debug the issue?

csphoenix1:
Hi, I have the same problem if I do a `cat /tmp/foo` with foo as a text file wich contain a `dpkg -l` stdout command (ie. 2000 lines).

I code a workaround bellow. What do you think about it?

[modified_file_with_workaround.txt](https://github.com/robotframework/robotframework/files/7649807/modified_file_with_workaround.txt)
[original_file_with_problem.txt](https://github.com/robotframework/robotframework/files/7649808/original_file_with_problem.txt)
.

pekkaklarck:
Could you @csphoenix1 clarify your problem? How did you use `Run Process`?

Could you also make your proposed changes available so that it would be easier to see them? If you there aren't much changes, you could just include them into a comment here with formatting. If there are more changes, you could create a branch with changes. It could then later be turned into a pull request if needed.

bhorkarg:
Can confirm that this happens with Robot Framework 4.1.2 (Python 3.8.10 on linux). We gather Kubernetes pod logs using the following keyword.

[...]

PR #4205, opened 2022-01-27
Solution to solve #3661

csphoenix1:
Hi,

It work with this .robot

```*** Settings ***
Library           Process

*** Test Cases ***
Dpkg
	Log To Console	Test with a few stdout lines
    ${result}	Process.Run Process	cat foo.txt | head	shell=true
	Log To Console	${result.stdout}
	Log To Console	Test with a lot stdout lines
    ${result}	Process.Run Process	cat foo.txt	shell=true
	Log To Console	${result.stdout}
	Log To Console	${result.stderr}
```

And a file "foo.txt" which contain around 1000 lines.

RobotFramework hung just after display "Test with a lot stdout lines".


pekkaklarck:
Did you test running acceptance tests with these changes? This change itself needs tests as well.

pekkaklarck:
This fix causes other problems, existing tests fail, and doesn't have tests for the fix itself. No updates since January. Closing.

Issue #4210, opened 2022-01-30
Enhance error detection at parsing time


pekkaklarck:
Robot's parser detects many errors already at the parsing time, but there are also many syntax errors that are only detected at the execution time. A problem with detecting errors only at the execution time is that they aren't reported to tools like editor plugins that use Robot's parser. We should thus move error detection to parser as much as possible. This includes at least these cases:

- Empty test and keyword.
- `END`, `ELSE`, `ELSE IF`, `EXCEPT` or `FINALLY` when there is no matching parent element like `IF` or `TRY`.
- ~`RETURN` used in test.~ Fixed already in RF 5.0.
- ~`BREAK/CONTINUE` outside loops.~ Fixed already in RF 5.0.
- ~`RETURN/BREAK/CONTINUE` in `FINALLY`.~ Fixed already in RF 5.0.

We also have a similar problem that some of the error situations detected by the parser are not detected as early as they could be. For example, having `IF` with more than one condition is detected by the parser when the parsing model is validated, but the underlying tokens do not get `ERROR` type. This is a problem if only using tokens (i.e. when using `get_tokens`, not `get_model`) and also when using the model it would be more convenient if invalid tokens could be easily detected. This problem occurs at least in these cases:

- `IF`, `ELSE IF` and `WHILE` with more than one condition.
- `ELSE`, `END`, `BREAK`, `CONTINUE`, `TRY` and `FINALLY` with values.
- `AS` with invalid variable or with more than one variable.

Because these latter errors are detected by the parser, only a bit late, I don't consider this limitation as bad as the problem with the earlier errors.

yanne:
These three points are covered already in 5.0:

- RETURN used in test.
- BREAK/CONTINUE outside loops.
- RETURN/BREAK/CONTINUE in FINALLY.

pekkaklarck:
After ffa0d4e65795d2392bf3166b33eb7ff9800e27c9 syntax errors are reported during execution using `raise DataError(msg, syntax=True)`. All these could be detected already at parsing time.

pekkaklarck:
Need to take a closer look at this and submit concrete issues about errors that could be detected earlier.

yanne:
Fixed empty name and keyword validation #4376e59

yanne:
We decided to leave this part for a later stage:

[...]

Issue #3075, opened 2019-02-04
Native support for `TRY/EXCEPT` functionality


pekkaklarck:
The plan is to add support for native `IF/ELSE` syntax in RF ~3.3~ 4.0. See the discussion in #3074 for background information why we think adding this kind of programmatic features is a good idea. Another similar and commonly needed feature is `TRY/EXCEPT` functionality to ease handling errors. Currently it is possible to use `Run Keyword And Expect Error` and `Run Keyword And Ignore` error but they are awkward similarly as `Run Keyword If` is an awkward replacement for `IF/ELSE` syntax.

Exact semantics how `TRY/EXCEPT` will work are not yet clear. They will be decided later before implementation starts. A precondition is a new parser, similarly as for `IF/ELSE`, and we also consider `IF/ELSE` more important and will start implementing it first.

Snooz82:
Just an idea:

At the moment we have already TRY/FINALLY with teststeps and Teardown.
Teardown can be on every level.

What is missing is only a [On-Error] setting.
If we would have this [On-Error] we could also do stuff like capture screenshots there and closing connections an so on.

I personally would not like to see a real TRY/EXCEPT like IF/ELSE, because the we start re-implementing Python.
[On-Error] is also understandable for non programmers.

```
*** Test Cases ***
Test
    [Setup]     Open Browser.   http://example.com
    Do Keyword
    Some Other Test
    [On-Error]    Save error log and capture Screenshot
    [Teardown]    Close All Browsers
```

Same could work on Suite and on Keyword Level.
```
*** Settings *** 
Suite Setup   Open Browser
Error Handler    Save error log and capture Screenshot
Suite Teardown    Close All Browsers
```

pekkaklarck:
I've been thinking about adding an on-error handler but nowadays I think TRY/EXCEPT would probably be better. Some of the problems I see with on-error handler:

- Unless on-error can also ignore the error, it's not really different from teardown with `Run Keyword If Test Failed`.
- If on-error should be able to ignore errors, it needed to get exception information as argument to be able to do decisions based on that.
- One test (or keyword) could only have one on-error handler. If there were multiple possible errors to handle, the handler would get complicated.

I got a feeling TRY/EXCEPT would be a lot more generic solution. In simple cases on-error might be easier to understand, but in more complex scenarios TRY/EXCEPT would be a lot more flexible.

[...]

PR #4211, opened 2022-01-31
Alternative error handling for try/except fix #4208 #4210 #3075

### LINKED ISSUE #4208: Parsing differences in try/except statements with get_model and data_only parameter

d-biehl:
Referring to your comment https://github.com/robotframework/robotframework/pull/4209#discussion_r795202824

I tried to move the error handling for TRY/EXCEPT/ELSE/FINALLY/END into the lexer, so that errors are detected and reported as early as possible.

At the moment it is implemented in a way that the invalid tokens are reported but have no effect at runtime, similar for example if a library is not found. And I haven't corrected or created any tests yet.

In a similar way IF/WHILE/FOR could be adapted, but this should be done in separate PR's.

### DIFF
--- a/atest/testdata/running/try_except/invalid_try_except.robot
+++ b/atest/testdata/running/try_except/invalid_try_except.robot
@@ -77,7 +77,8 @@ Multiple default EXCEPTs
     END
 
 AS not the second last token
-    [Documentation]    FAIL    EXCEPT's AS marker must be second to last.
+    [Documentation]    FAIL    EXCEPT's AS can only have one variable.
+    
     TRY
         Fail   Should not be executed
     EXCEPT    AS    foo    ${foo}
--- a/src/robot/parsing/lexer/statementlexers.py
+++ b/src/robot/parsing/lexer/statementlexers.py
@@ -14,7 +14,7 @@
 #  limitations under the License.
 
 from robot.utils import normalize_whitespace
-from robot.variables import is_assign
+from robot.variables import is_assign, is_scalar_assign
 
 from .tokens import Token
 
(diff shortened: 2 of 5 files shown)


pekkaklarck:
Using invalid number of parameters with `AS`, `IF`, etc. definitely should be a syntax error that fails execution, only reporting such errors at parsing time is not an option. To keep the usage of the parsing model easy and backwards compatible, all these errors should get available via `errors` attribute of the nodes in the model. Possibly it could be a property or `validate()` could add errors from tokens there. Either way, it must not have adverse effect for parsing performance. 

Because it's unlikely these are common errors and because they _are_ already now detected by the parser, I don't consider it too high priority to detect them earlier. I understand it's a bit better for editors to be able to highlight only the latter value in cases like
```
IF    $foo    $bar
```
instead of highlighting the whole `IF` node, but editors anyway need to use the latter approach in cases like
```
IF
```
where the problem is that a token is missing. I would assume it would be easier for editors to just check does the `IF` node have errors instead of going through each tokens. At least for Robot itself the current validation approach is simple and I'm not at all certain making it more complicated is worth the benefits.

PR #4215, opened 2022-02-02
fix  #4214: just expand passed keywords 

### LINKED ISSUE #4214: Do not expland `NOT RUN` keywords even if they match `--expandkeywords`
Using RF 5.0a1 skipped keywords gets expanded. This is a problem for huge logs with many matching skipped keywords and results in browser crash.

grafik

DirkRichter:
- do not expand skipped/NotRun to avoid browser crash on large logs
- failed kws expanded per default, thus no need to handle failed kws in --expandkeyword logic

### DIFF
--- a/src/robot/reporting/jsbuildingcontext.py
+++ b/src/robot/reporting/jsbuildingcontext.py
@@ -73,7 +73,8 @@ class JsBuildingContext:
 
     def check_expansion(self, kw):
         if self._expand_matcher is not None:
-            self._expand_matcher.match(kw)
+            if kw.passed:
+                self._expand_matcher.match(kw)
 
     @property
     def expand_keywords(self):


src/robot/reporting/jsbuildingcontext.py
```diff
@@ -73,7 +73,8 @@ def create_link_target(self, msg):
 
     def check_expansion(self, kw):
         if self._expand_matcher is not None:
-            self._expand_matcher.match(kw)
+            if kw.passed:
+                self._expand_matcher.match(kw)
```
pekkaklarck, line 77: 1. I don't think not expanding keywords with SKIP status is a good idea. Keywords only get that status if they explicitly skip a test. That's pretty rare, so shouldn't cause extra overhead, and it could also be useful to be able to expand them. Should thus use `not kw.not_run` instead of `kw.passed`.
2. I think `ExpandKeywordMatcher` itself should make the decision about expanding. This `if` would thus rather belong there.

pekkaklarck (changes requested):
Would need a test as well as a note in the User Guide that starting from RF 5.0 not run keywords aren't expanded.

pekkaklarck:
As I commented above, I think it should be possible to expand keywords with the SKIP status. It's a good point that failed tests are automatically expanded so they could be excluded here. Perhaps the condition should be `if kw.passed or kw.skipped` to explicitly list statuses we want to expand.

src/robot/reporting/jsbuildingcontext.py
DirkRichter: 1.+2.: ok, i agree. 
I updated the pull request and tested: Expanding also skipped keywords is safe.

pekkaklarck:
As I commented earlier, this still needs a test. Could also add a note about the changed functionality to the User Guide, but I'm not sure is that really needed. Probably not.

pekkaklarck:
I fixed this with tests in f22b78e118f7280535e7f7a41b0343b202bf4e07.

PR #4217, opened 2022-02-04
fix incorrect spelling of "SKIPPED"

DetachHead:
image

### DIFF
--- a/src/robot/output/console/dotted.py
+++ b/src/robot/output/console/dotted.py
@@ -79,7 +79,7 @@ class StatusReporter(SuiteVisitor):
                               'test' if not suite.rpa else 'task',
                               plural_or_not(stats.total),
                               secs_to_timestr(suite.elapsedtime/1000.0)))
-        self._stream.highlight(suite.status + ('P' if suite.status == 'SKIP' else '') + 'ED', suite.status)
+        self._stream.highlight(suite.status + 'PED' if suite.status == 'SKIP' else 'ED', suite.status)
         self._stream.write('\n%s\n' % stats.message)
 
     def visit_test(self, test):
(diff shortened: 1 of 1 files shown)


src/robot/output/console/dotted.py
```diff
@@ -79,7 +79,7 @@ def report(self, suite):
                               'test' if not suite.rpa else 'task',
                               plural_or_not(stats.total),
                               secs_to_timestr(suite.elapsedtime/1000.0)))
-        self._stream.highlight(suite.status + 'ED', suite.status)
+        self._stream.highlight(suite.status + ('P' if suite.status == 'SKIP' else '') + 'ED', suite.status)
```
pekkaklarck, line 82: `'PED' if suite.status == 'SKIP' else 'ED'` could be a bit simpler construct here.

pekkaklarck (commented):
Good catch! There ought to be tests for this output and they need to be updated as well. I enabled GitHub Actions for this PR so we ought to soon see did anything break.

DetachHead:
it looks like all the actions were cancelled?

pekkaklarck:
Don't know what happened. Possibly GHA didn't like new commits being added when it was running. Activated again.

DetachHead:
https://github.com/robotframework/robotframework/runs/5117910325?check_suite_focus=true

no idea what's going on here

pekkaklarck:
One run failed and I believe others were cancelled after that. There were failures like this:
```
FAIL: Robot.Cli.Console.Console Type.Dotted with width
Lines differ.
Expected: FAILED
Actual:   ED
```
That error seems to be caused by a bug in this PR. I didn't see any test with `SKIPED` failing so it's possible that status isn't tested. Updating tests to cover also that would be good. If it's not easy to see how that can be done, I can do it myself.

DetachHead:
my bad didn't notice that stage failed, i was referring to this:
image

i'll try to fix the other failures

pekkaklarck:
Merged. If you consider this important enough to be added to release notes, please submit an issue as well. I'll then add it to RF 5.0 scope.

Issue #4229, opened 2022-02-13
Add `&{OPTIONS}` to automatic variables to expose command line options


nicobucher:
It would be beneficial to be able to access command line options like "exclude tags" and "include tags" from within test cases.

This enhancement was sort of already requested in #3255 but this is a more general request.

As described also by @pekkaklarck in PR #4052 , a disctionary called "SETTINGS" containing these options would be best. Starting off with only exclude and include tags, the number of exposed options can be extended in the future.

pekkaklarck:
Sounds good to me. Starting only with include and exclude info is ok, but I think it would be good to ask the community what else they'd like to be exposed. Not everyone is following this tracker so closely. Could you @nicobucher, or someone else, start a thread about this on the #devel channel in our Slack? Could obviously ask on other forums as well.

pekkaklarck:
There seems to be PR #4230 implementing this. It refers this issue but apparently mentioning it only in the title doesn't create an automatic link.

pekkaklarck:
I looked at PR #4230 quickly and it looks pretty good. I added few comments to it, though, and I hope they can be addressed soon.

I add this into RF 5.0 scope, but there's bit of hurry to get it actually included. We try to get the first RF 5.0 release candidate out this week or latest next week, and all features need to be in before that.

pekkaklarck:
I started a discussion about other things to possible expose via this variable on the #devel channel on our Slack. No other requests at least yet, but there was a _very_ good point that `SETTINGS` can be easily associated with the `*** Settings ***` section which isn't good. There was discussion about alternative names, and my current thinking is that `OPTIONS` is best. It doesn't have direct association with the Settings section and will work also if other than command line options are exposed via this variable. The name is also consistent with other built-in variable names.

pekkaklarck:
RF 5.0 rc 1 is now scheduled for Monday, February 21. That's also the deadline for getting this feature in to have it in RF 5.0 final.

nicobucher:
@pekkaklarck I udpated the PR #4230 since the deadline is approaching. I think all the comments should be adressed including renaing the dictionary to OPTIONS and using robot.model.Tags instead of a plain list. Maybe it will be possible to merge before tomorrow?

pekkaklarck:
The release candidate is a bit late for various reasons. This is now anyway in and will be included in the rc once its out.

Thanks for contribution @nicobucher!

nicobucher:
Perfect, my pleasure!

PR #4230, opened 2022-02-13
Add SETTINGS dictionary to automatic variables to expose command line options (Issue #4229)

nicobucher:
This PR adds a new disctionary "SETTINGS" to the automatic variables to expose exclude and include tags command line options (for now).

Based on previous discussions in #4052

### DIFF
--- a/atest/testdata/variables/automatic_variables/auto1.robot
+++ b/atest/testdata/variables/automatic_variables/auto1.robot
@@ -6,7 +6,7 @@ Suite Setup       Check Variables In Suite Setup    ${EXP_SUITE_NAME}
 ...               ${EXP_SUITE_DOC}    ${EXP_SUITE_META}
 Suite Teardown    Check Variables In Suite Teardown    ${EXP_SUITE_NAME}
 ...               FAIL    ${EXP_SUITE_STATS}    @{LAST_TEST}
-Force Tags        Force 1
+Force Tags        Force 1    include this test
 Resource          resource.robot
 Library           Collections
 Library           HelperLib.py    ${SUITE SOURCE}    ${SUITE NAME}
@@ -17,7 +17,7 @@ ${VARIABLE}          variable value
 ${EXP_SUITE_NAME}    Automatic Variables.Auto1
 ${EXP_SUITE_DOC}     This is suite documentation. With ${VARIABLE}.
 ${EXP_SUITE_META}    {'MeTa1': 'Value', 'meta2': '${VARIABLE}'}
-${EXP_SUITE_STATS}   16 tests, 14 passed, 2 failed
+${EXP_SUITE_STATS}   18 tests, 16 passed, 2 failed
 @{LAST_TEST}         Previous Test Variables Should Have Correct Values When That Test Fails    PASS
 
 *** Test Case ***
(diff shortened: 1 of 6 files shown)


src/robot/variables/scopes.py
```diff
@@ -185,6 +185,8 @@ def _set_cli_variables(self, settings):
     def _set_built_in_variables(self, settings):
         for name, value in [('${TEMPDIR}', abspath(tempfile.gettempdir())),
                             ('${EXECDIR}', abspath('.')),
+                            ('&{SETTINGS}',    {"EXCLUDE_TAGS": settings.suite_config['exclude_tags'], 
+                                                "INCLUDE_TAGS": settings.suite_config['include_tags']}),
```
pekkaklarck, line 189: Instead of exposing exclude and include as lists, it would probably be better to expose them as `robot.model.Tags` objects. `Tags` is a list-like object with these benefits in this context:
- Tags are matched in normalized manner. For example, `'e' in Tags(['E']) ` is true . This behavior would then also work in RF date when using e.g. `Should Contain` or `IF` expressions with include and exclude.
- Also patterns work in matching. For example, `'e*' in Tags(['example'])` and `'a AND b' in Tags(['a', 'b'])` are both true.
- Duplicate tags are automatically removed and tags are sorted.

I consider especially the first one pretty important, because otherwise users would need to implement normalization themselves to support tags similarly as Robot itself does.

The only downside of using `Tags` I can think of is that they are always unequal with lists. For example, both `Tags(['example']) ==['example']` and `Tags([]) == []` would both be false. It probably would be fine to change `Tags.__eq__` to handle this situation, though.

doc/userguide/src/CreatingTestData/Variables.rst
```diff
@@ -1007,6 +1007,13 @@ can be changed dynamically using keywords from the `BuiltIn`_ library.
    +------------------------+-------------------------------------------------------+------------+
    | ${PREV TEST MESSAGE}   | The possible error message of the previous test case. | Everywhere |
    +------------------------+-------------------------------------------------------+------------+
+   | ${SETTINGS}            | A dictionary exposing some command line options used  | Everywhere |
+   |                        | for running the current test:                         |            |
+   |                        | - EXCLUDE_TAGS                                        |            |
+   |                        |   A list of tags provided by the -exclude (-e) option |            |
+   |                        | - INCLUDE_TAGS                                        |            |
+   |                        |   A list of tags provided by the -include (-i) option |            |
+   +------------------------+-------------------------------------------------------+------------+
```
pekkaklarck, line 1016: - `--exclude` and `--include` have two hyphens.
- Should use format `:option:`--exclude` to get consistent styles
- I don't think short options need to be mentioned.
- Needs `New in RF 5.0.` note at the ends.

src/robot/variables/scopes.py
pekkaklarck: Need to decide what format to use with key names. I think it would be best to use same names we use on command line i.e. `exclude` instead of `EXCLUDE_TAGS`. That makes the names easiest for users to remember.

pekkaklarck:
This PR looks pretty good already. If my comments above are addressed soon, we can still get this merged before the first RF 5.0 release candidate.  Notice that although I commented issue #4229 that it would probably be good expose also other settings than include/exclude, that can be done after this PR has been merged.

pekkaklarck:
Failed test isn't related to this change.

oboehmer:
Introducing a new special/ builtin variable with a generic name like SETTINGS will likely create conflicts existing suites already using this name? Pekka, you might have been there when other new builtin vars have been introduced, so is this a valid concern or accepted by the community?

atest/testdata/variables/automatic_variables/auto1.robot
```diff
@@ -38,15 +38,25 @@ Test Documentation
...
+
+Include-tags Available As Automatic Variables
+    Should Contain    ${OPTIONS.include}    include this test
+
+Exclude-tags Available As Automatic Variables
+    Should Contain    ${OPTIONS.exclude}    exclude this test
+
+Skip-tags Available As Automatic Variables
+    Should Contain    ${OPTIONS.skip}    Skip_Me
+    Should Contain    ${OPTIONS.skiponfailure}    me_too
```
pekkaklarck, line 59: It would be good to test that access like `${OPTIONS}[include]` works as well.

doc/userguide/src/CreatingTestData/Variables.rst
```diff
@@ -1007,6 +1007,15 @@ can be changed dynamically using keywords from the `BuiltIn`_ library.
    +------------------------+-------------------------------------------------------+------------+
    | ${PREV TEST MESSAGE}   | The possible error message of the previous test case. | Everywhere |
    +------------------------+-------------------------------------------------------+------------+
+   | ${OPTIONS}             | A dictionary exposing command line options used for   | Everywhere |
```
pekkaklarck, line 1010: Should use `&{OPTIONS}` to match existing `&{SUITE METADATA}`.

pekkaklarck (approved):
Looks great! Two minor comments but I can handle them myself.

atest/testdata/variables/automatic_variables/auto1.robot
pekkaklarck: Noticed now that these tests aren't actually validated by tests on `atest/robot` side. I'll fix that.

PR #4246, opened 2022-03-01
feat(while): add limit to break infinite loops

yanne:


src/robot/result/model.py
```diff
@@ -236,7 +236,7 @@ def body(self, iterations):
     @property
     @deprecated
     def name(self):
-        return self.condition
+        return self.condition + f' | {self.limit}' if self.limit else ''
```
pekkaklarck, line 239: Another operator precedence problem.

src/robot/running/bodyrunner.py
```diff
@@ -334,21 +335,29 @@ def __init__(self, context, run=True, templated=False):
...
+                        self._run_iteration(data, result, run)
                 except BreakLoop:
                     break
                 except ContinueLoop:
                     continue
+                except WhileLimitExceeded:
+                    raise ExecutionFailed(
+                        "WHILE loop was aborted because it did not finish within the "
+                        f"limit of {limit}. Use the 'limit' argument to increase or "
+                        "remove the limit if needed."
```
pekkaklarck, line 359: It seems to be common in Python to prefix all lines with `f` in cases like this.

pekkaklarck: Could WhileLimit possibly raise ExecutionFailed directly? Catching WhileLimitExceeded just to raise another exception feels a bit odd.

src/robot/running/bodyrunner.py
```diff
@@ -556,3 +565,81 @@ def _run_finally(self, data, run):
...
+    @staticmethod
+    def parse_iteration_count(limit):
+        try:
+            limit = int(limit)
+            if limit <= 0:
+                raise ValueError
+            return limit
+        except ValueError:
+            raise ValueError("Iteration limit must be positive integer when using "
+                             f"'x' or 'times', got: '{limit}'")
```
pekkaklarck, line 585: This isn't part of the class API, right? Should be prefixed with `_` and be after the public `create`.

src/robot/running/bodyrunner.py
```diff
@@ -556,3 +565,81 @@ def _run_finally(self, data, run):
...
+
+    @classmethod
+    def create(cls, limit, variables):
+        try:
+            if not limit:
+                return IterationCountLimit(100)
+            value = limit.replace('limit=', '')
```
pekkaklarck, line 592: Why does the value contain `limit=` prefix here? Shouldn't only the value be returned by the parsing model? I'll add a separate comment there.

src/robot/parsing/model/statements.py
```diff
@@ -971,26 +971,38 @@ class WhileHeader(Statement):
...
+
+    @property
+    def limit(self):
         values = self.get_values(Token.ARGUMENT)
-        if len(values) != 1:
-            return ', '.join(values) if values else None
-        return values[0]
+        if len(values) < 2:
+            return None
+        return values[1]
```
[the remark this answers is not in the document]
pekkaklarck: `limit` property should return only the actual limit, not the `limit=` prefix.

src/robot/running/bodyrunner.py
```diff
@@ -556,3 +565,81 @@ def _run_finally(self, data, run):
...
+            if value.upper() == 'NONE':
+                return NoLimit()
+            value = variables.replace_scalar(value)
+            if value.endswith('x'):
```
pekkaklarck, line 596: `value` isn't necessarily a string here so `endswith` isn't safe. Changing `replace_scalar` to `replace_string` ought to fix that but it would be good to add a test as well.

src/robot/utils/robottime.py
```diff
@@ -39,10 +39,13 @@ def _float_secs_to_secs_and_millis(secs):
...
 
 
-def timestr_to_secs(timestr, round_to=3):
+def timestr_to_secs(timestr, round_to=3, accept_plain_values=True):
     """Parses time like '1h 10s', '01:00:10' or '42' and returns seconds."""
     if is_string(timestr) or is_number(timestr):
-        for converter in _number_to_secs, _timer_to_secs, _time_string_to_secs:
+        converters = [_timer_to_secs, _time_string_to_secs]
+        if accept_plain_values:
+            converters.insert(0, _number_to_secs)
```
pekkaklarck, line 47: This util may be called pretty often and `insert` is slow. This would be faster and also simpler: 
```python
if accept_plain_values:
    converters = [_number_to_secs, _timer_to_secs, _time_string_to_secs]
else:
    converters = [_timer_to_secs, _time_string_to_secs]
```
Should also add unit test for this new functionality.

pekkaklarck (changes requested):
Looks good. We already agreed that limit should be tokenized to that it gets a different token type. For few other things to fix/enhance see the separate comments.

src/robot/running/bodyrunner.py
yanne: Yes it can. Have to get rid of the f-string then :(

src/robot/parsing/model/statements.py
```diff
@@ -971,26 +971,35 @@ class WhileHeader(Statement):
...
+        tokens.append(Token(Token.EOL, eol))
         return cls(tokens)
 
     @property
     def condition(self):
-        values = self.get_values(Token.ARGUMENT)
-        if len(values) != 1:
-            return ', '.join(values) if values else None
-        return values[0]
+        return ', '.join(self.get_values(Token.ARGUMENT))
```
pekkaklarck, line 988: This changes old behavior so that if there's no condition, an empty string is returned instead of `None`. I guess that's fine because `validate` detects the error situation anyway.

src/robot/parsing/model/statements.py
```diff
@@ -971,26 +971,35 @@ class WhileHeader(Statement):
...
+
+    @property
+    def limit(self):
+        value = self.get_value(Token.OPTION)
+        return value.replace('limit=', '') if value else None
 
     def validate(self, context):
-        conditions = len(self.get_tokens(Token.ARGUMENT))
-        if conditions == 0:
+        values = self.tokens[1:]
```
pekkaklarck, line 996: I think tokens contain also separators when when parsing with `data_only=False` and this doesn't works. It seems the old logic to validate the number of conditions would work just fine. Unit test should also be added to make sure parsing works with `data_only=False`.

src/robot/parsing/model/statements.py
```diff
@@ -971,26 +971,35 @@ class WhileHeader(Statement):
...
+        if len(values) == 0:
             self.errors += ('WHILE must have a condition.',)
-        if conditions > 1:
+        if len(values) == 2 and values[1].type != Token.OPTION:
+            self.errors += (
+                f"Second WHILE loop argument must be 'limit', got {values[1]}.",)
```
pekkaklarck, line 1001: I don't think we anymore need to validate limit at all. Lexing works so that it can only be at the end and must always start with `limit=` anyway.

pekkaklarck (changes requested):
Looks great apart from a problem when parsing with `data_only=False`.

PR #4248, opened 2022-03-02
add GitHub URL for PyPi

andriyor:
Warehouse now uses the project_urls provided to display links in the sidebar on [this screen](https://pypi.org/project/requests/), as well as including them in API responses to help the automation tool find the source code for Requests.

Docs: [packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,9 @@ setup(
     author       = 'Pekka Kl\xe4rck',
     author_email = 'peke@eliga.fi',
     url          = 'https://robotframework.org/',
+    project_urls = {
+        'Source': 'https://github.com/robotframework/robotframework',
+    },
     download_url = 'https://pypi.org/project/robotframework/',
     license      = 'Apache License 2.0',
     description  = DESCRIPTION,


pekkaklarck:
Thanks for the PR, adding more info to PyPI is certainly a good idea. Some questions:

1. Would it make sense to add also link to issues? It can obviously be easily found via the GitHub project main page. Do you know how other projects are using this?
2. If we just add link to the GitHub project main page, perhaps the text should be "GitHub" instead of "Source". Again, knowing how others do this would help.
3. Which setuptools versions support `project_urls`? Do older emit warnings/errors if they see it? We can obviously make sure we use recent setuptools version when creating packages, but we cannot control what users have when they install them.

pekkaklarck:
Did little searching and via this [StackOverflow question](https://stackoverflow.com/questions/61156290/how-to-set-project-links-in-pypi) found a good example project: https://pypi.org/project/PyScaffold/. They have various links like this and we could at least use source, issues and Twitter.

It's cool that PyPI also automatically adds icons for links it recognizes. Based on the aforementioned SO question, icons are set according to rules [here](https://github.com/pypa/warehouse/blob/main/warehouse/templates/packaging/detail.html). On a quick look at the rules it seems we could also add Slack.

Possibly we could also add Changelog/What's New, but then that URL needed to be updated with each release to point to the correct release notes. We have invoke task that updates version info in setup.py and it could handle this as well.

pekkaklarck:
I'm in general very much in favor of doing this, but I'm still worried about setuptools compatibility with `project_urls`. Although these links are really nice, warnings/errors when users are installing Robot Framework are not.

bollwyvl:
> setuptools compatibility

It looks like `project_urls` was [added](https://github.com/pypa/setuptools/blob/347b847cf04d0480df240dff1b0255b596d6aefc/docs/userguide/declarative_config.rst#metadata) in [`38.3.0`](https://pypi.org/project/setuptools/38.3.0/), in early 2018. Installing e.g. `pyscaffold` from a `.tar.gz` (latest) with setuptools as old as `32` doesn't say anything about `project_urls`, so this is probably safe... a release like 5.0 which drops _python major versions_ is probably as good a time as any to actually declare a mininum setuptools... `debian/buster`, usually a good indicator of "ain't broke,"  ships [40.0.8](https://packages.debian.org/buster/python3-setuptools), and might be a reasonable candidate.

andriyor:
> 2\. If we just add link to the GitHub project main page, perhaps the text should be "GitHub" instead of "Source". Again, knowing how others do this would help.

https://github.com/pypa/warehouse/issues/5947

pekkaklarck:
Quick update:

- Let's do this!
- I added issue #4312 to track this. We should discuss there what URLs to add.
- According to the [setuptools docs](https://github.com/pypa/setuptools/blob/main/docs/userguide/declarative_config.rst#metadata) linked above by @bollwyvl, `project_urls` is new in setuptools 38.3 and `long_description_type` that we are already using is new in 38.6. Using `project_urls` thus ought to be pretty safe. We can add it at least in RF 5.1. Possibly already in RF 5.0.1, but then I'd like this to be tested with old setuptools versions.

PR #4250, opened 2022-03-04
correct handling of double underscore at end of suite names fixes #4249

### LINKED ISSUE #4249: Suites with double underscore at end of filename results in a suite with no name
A suite folder or suite file with double underscore at the end of the name, like 'test__.robot' or 'folder__', results in a suite with no name. 

Steps to reproduce:

- create a new RobotProject
- create a folder with name 'suite__'
- create a file with name 'first__.robot' in folder 'suite__'
- add a simple testcase to file 'first__.robot_ like:
```robotframework
*** Test Cases ***
simple
    Log    hello
```
- execute the suite with 'robot .'

the resulting output is:
```
Rt1
Rt1.
Rt1..
first                                                                 | PASS |
Rt1..                                                                 | PASS |
1 test, 1 passed, 0 failed
Rt1.                                                                  | PASS |
1 test, 1 passed, 0 failed
Rt1                                                                   | PASS |
1 test, 1 passed, 0 failed
```

I expect that the correct suite names are in there like:
Rt1.Suite.First

Also the log and report doesn't show the correct suite name.

I assume that robot suitebuilder thinks, that the suite name is a prefix, see [here] (https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#test-suite-name-and-documentation) for explanations. But I think, if the name after the prefix is empty, it should be handled as a normal name, without a prefix.

Used Versions:
RobotFramework 4.1.3 and 5.0 beta 1

d-biehl:
this are the fixes for #4249

do you have an idea, where I should put the testcases for it?

### DIFF
--- a/src/robot/parsing/suitestructure.py
+++ b/src/robot/parsing/suitestructure.py
@@ -149,7 +149,11 @@ class SuiteStructureBuilder:
         return incl_suites.match(self._split_prefix(name))
 
     def _split_prefix(self, name):
-        return name.split('__', 1)[-1]
+        result = name.split('__', 1)[-1]
+        if result:
+            return result
+        return name
+
 
 
 class SuiteStructureVisitor:
--- a/src/robot/parsing/suitestructure.py
+++ b/src/robot/parsing/suitestructure.py
@@ -155,7 +155,6 @@ class SuiteStructureBuilder:
         return name
 
 
-
 class SuiteStructureVisitor:
 
     def visit_file(self, structure):
--- a/atest/robot/cli/console/expected_output/dotted_exitonfailure.txt
+++ b/atest/robot/cli/console/expected_output/dotted_exitonfailure.txt
@@ -1,13 +1,13 @@
-Running suite 'Suites' with 11 tests.
+Running suite 'Suites' with 13 tests.
 ==============================================================================
-Fxxxxxxxxxx
+.Fxxxxxxxxxxx
 ------------------------------------------------------------------------------
 FAIL: Suites.Fourth.Suite4 First
 Expected
 ==============================================================================
-Run suite 'Suites' with 11 tests in *.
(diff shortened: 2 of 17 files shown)


src/robot/parsing/suitestructure.py
```diff
@@ -149,7 +149,11 @@ def _is_in_included_suites(self, name, incl_suites):
         return incl_suites.match(self._split_prefix(name))
 
     def _split_prefix(self, name):
-        return name.split('__', 1)[-1]
+        result = name.split('__', 1)[-1]
+        if result:
+            return result
+        return name
+
```
pekkaklarck, line 156: Unnecessary empty line.

pekkaklarck (changes requested):
Would be good to have an acceptance test for this as well. A suite was added already under `atest/testdata`, but it should be validated with tests under `atest/robot`. Probably the best place would be `atest/robot/parsing/suite_settings.robot`. There's already test for suite name, and another could be added for suite name when file/dir has trailing double underscore.

Probably would be good to add test for valid prefix to the same place as well so that all name related features would be tested in the same place. Name with prefix ought to be tested also elsewhere with the prefix functionality, but I don't think that having same/similar tests in two places is that bad. Alternatively the test for trailing underscores could be added to the suite testing prefixes.

src/robot/parsing/suitestructure.py
d-biehl: Working on correcting the atests. If I understand it correctly, then we have also acceptance tests for this after correction. But I have to correct 33 Tests after, this takes a little while...

pekkaklarck (approved):
Adding suites under `atest/testdata/misc/suites` certainly makes sure this is tested thoroughly. A downside is that lot of other tests needed to be updated, but because tests pass you seem to have caught them all. :smiley:

Issue #37, opened 2014-06-29
Submit package information to PyPI


spooning:
> _Originally submitted to [Google Code](http://code.google.com/p/robotframework/issues/detail?id=37) by @pekkaklarck on 25 Jun 2008_

PyPI = Python Package Index = <a href="http://pypi.python.org" rel="nofollow">http://pypi.python.org</a>

It seems to be easiest to just submit PKG-INFO file which can be created
using setup.py. Before that we need to update the metadata in setup.py,
since currently it contains only minimal information.

PR #4255, opened 2022-03-06
Suite-wide warning when tests with the same name/longname are executed, fixes #4253

### LINKED ISSUE #4253: Suites with an underscore at the end can lead to ambiguous suite names
Assuming that i have 2 files `first.robot` and `first_.robot`  and both have identical content, means a test case with the same name.

because undescores where converted to spaces and the name is stripped, both files leads to the same suite name "test"
if i want to execute the test "first.simple" with from command line like:
`robot -x xunit.xml -t "*.first.simple" .` always both tests are executed:

output is like this:
```
Rt1
Rt1.First
simple                                                                | PASS |
Rt1.First                                                             | PASS |
1 test, 1 passed, 0 failed
Rt1.First
simple                                                                | PASS |
Rt1.First                                                             | PASS |
1 test, 1 passed, 0 failed
Rt1                                                                   | PASS |
2 tests, 2 passed, 0 failed
```

report looks the same and the xunit output is like this:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuite name="Rt1" tests="2" errors="0" failures="0" skipped="0" time="0.032" timestamp="2022-03-04T21:56:07.923000">
    <testsuite name="First" tests="1" errors="0" failures="0" skipped="0" time="0.004" timestamp="2022-03-04T21:56:07.945000">
        <testcase classname="Rt1.First" name="simple" time="0.001"></testcase>
    </testsuite>
    <testsuite name="First" tests="1" errors="0" failures="0" skipped="0" time="0.004" timestamp="2022-03-04T21:56:07.950000">
        <testcase classname="Rt1.First" name="simple" time="0.001"></testcase>
    </testsuite>
</testsuite>
```

After looking at the report, I dont know why 2 testcases are executed and also statics are incorrect if the xunit file is analysed and so on...

I think, if there is another suite with the same name allready added to the suite structure, then a warning should be shown at execution, like if there are multiple testcases with the same name.

d-biehl:
I think, reporting of multiple test executions should be enough. 

What do you think? If it's ok, then I add some more testcases, if needed.


pekkaklarck:
I thought the problem in #4253 was that a suite can have multiple child suites with same name. That would be easiest to detect what suites are run similarly as how we check does a suite have multiple tests with same name. The current implementation has several problems:
- If there are two suites with same names but different tests, there's no warning at all.
- If there are two exactly same suites, each executed causes a warning causing lot of noise. Just one warning would be a lot better.

pekkaklarck:
There's also a possible problem that people may be running tests like `robot tests.robot tests.robot` to get same suite run multiple times. Showing a warning in cases like this may not be that nice, especially when there's no easy way to avoid it. Possibly the warning should be omitted altogether in cases like this. I think this situation could be detected at the runtime by checking is source of the suite `None` or not.

d-biehl:
Implement reset executed tests if we run a combined/virtual suite like `robot tests.robot tests.robot` 
Now no warnings are reported, except that these suite contains tests with the same longname...

Now we need more tests for this behaviour.

pekkaklarck:
I would assume having suites with same names also creates reporting problems, not only tests with same names. I got a feeling it's best to leave this to RF 5.1 and have more time for decisions. Or do you have some urgent use case where this features would be required?

Notice also that a better solution for differentiating tests and suites could be generating UUID for them.

d-biehl:
I have an issue in RobotCode [#37](https://github.com/d-biehl/robotcode/issues/37) where I must find a way to show/mark that there are test with same long name in the whole suite. So that VSCode's test explorer can display the tests and run correctly and also show the correct test state/result after running.
My idea is, let the user execute the test, even there are multiple tests with the same long name, but show a warning or information at execution time that there are more tests executed, then expected.

pekkaklarck:
Asking community would certainly make this RF 5.1 feature. I want the rc out latest tomorrow morning, preferably already today. I also doubt users in general have any opinion about this.

d-biehl:
It's not that I want to have any files or even directories with leading or trailing underscores. Although I have already described a use case above that I could imagine.

It is also not about the fact that I cannot execute duplicate test cases with `--test`. 

My point is that I should warn the user that something is wrong. However the user got to the point that the report says that 2 test cases/suites have been executed, although he only specified one. Maybe just because he created a backup with an underscore at the end or inserted a space somewhere too much by mistake or whatever.... And it's not only the experienced users like you and me, who immediately see that something is wrong, who use RobotFramework.

RobotCode should support the user in dealing with RobotFramework, the keyword is unknown, the parameters are not correct, the library can not be found and why not that tests exist, which have the same name and possibly influence the execution result. For this I was looking for a solution.

And since Robot already gives a warning when it finds test cases with the same name inside a .robot file, it should be possible for Robot to do that suite wide. 

If it doesn't fit into RC1 at the moment, that's ok with me. Maybe there will be another RC2 ;-) if it comes in the 5.1, it's just as ok for me.

pekkaklarck:
I agree there should be a warning. I guess we mainly disagree when the warning should be reported. In my opinion just having two suites with same name inside a directory should cause the warning. That way running two exactly same suites would only give you one warning, not warnings from each test in the suite. I also see value in warning about same suite name if suites would actually contain different tests.

I did a quick prototype and the following code worked pretty well. Would that be fine for you? If yes, I can add some tests and call it done as the last thing before RC.
```difff
diff --git a/src/robot/running/suiterunner.py b/src/robot/running/suiterunner.py
index d9769bbef..bc24402bc 100644
--- a/src/robot/running/suiterunner.py
+++ b/src/robot/running/suiterunner.py
@@ -36,7 +36,7 @@ class SuiteRunner(SuiteVisitor):
         self._variables = VariableScopes(settings)
         self._suite = None
         self._suite_status = None
-        self._executed_tests = None
+        self._executed = [NormalizedDict(ignore='_')]
         self._skipped_tags = TagPatterns(settings.skip)
 
     @property
@@ -44,6 +44,11 @@ class SuiteRunner(SuiteVisitor):
         return EXECUTION_CONTEXTS.current
 
     def start_suite(self, suite):
+        if suite.name in self._executed[-1] and suite.parent.source:
+            self._output.warn(f"Multiple suites with name '{suite.name}' executed in "
+                              f"suite '{suite.parent.longname}'.")
+        self._executed[-1][suite.name] = True
+        self._executed.append(NormalizedDict(ignore='_'))
         self._output.library_listeners.new_suite_scope()
         result = TestSuite(source=suite.source,
                            name=suite.name,
@@ -81,7 +86,6 @@ class SuiteRunner(SuiteVisitor):
                                                test_count=suite.test_count))
         self._output.register_error_listener(self._suite_status.error_occurred)
         self._run_setup(suite.setup, self._suite_status)
-        self._executed_tests = NormalizedDict(ignore='_')
 
     def _resolve_setting(self, value):
         if is_list_like(value):
@@ -102,6 +106,7 @@ class SuiteRunner(SuiteVisitor):
         self._suite.endtime = get_timestamp()
         self._suite.message = self._suite_status.message
         self._context.end_suite(ModelCombiner(suite, self._suite))
+        self._executed.pop()
         self._suite = self._suite.parent
         self._suite_status = self._suite_status.parent
         self._output.library_listeners.discard_suite_scope()
@@ -110,11 +115,11 @@ class SuiteRunner(SuiteVisitor):
         settings = self._settings
         if TagPatterns("robot:exclude").match(test.tags):
             return
-        if test.name in self._executed_tests:
+        if test.name in self._executed[-1]:
             self._output.warn(
                 test_or_task(f"Multiple {{test}}s with name '{test.name}' executed in "
-                             f"suite '{self._suite.longname}'.", settings.rpa))
-        self._executed_tests[test.name] = True
+                             f"suite '{test.parent.longname}'.", settings.rpa))
+        self._executed[-1][test.name] = True
         result = self._suite.tests.create(self._resolve_setting(test.name),
                                           self._resolve_setting(test.doc),
                                           self._resolve_setting(test.tags),
```

d-biehl:
Ok, tried a little bit. Looks good to me too. Similar to what I tried very first, but then discarded because I thought we should focus on ambiguous test cases. My PR only reports when it finds ambiguous test cases.

But thinking about it again, at the moment the easiest solution is to report ambiguous suites, because the "report/log.html" doesn't combine suites with the same name. Maybe this would be something for the 5.1 version.

So, if this is the solution, I'm happy with it.

pekkaklarck:
I consider it important in this context to report ambiguous suites. This solution also indirectly reports ambiguous tests  because you can have them only if there's an ambiguous suite. I'll add some test for this and commit it shortly.

pekkaklarck:
This is now covered by https://github.com/robotframework/robotframework/issues/4268.

PR #4256, opened 2022-03-07
Minor execution speed improvements for 5.0

xylix:
Memory usage needs to be tested. Adding various comments to highlight my findings.


src/robot/utils/robottime.py
```diff
@@ -380,6 +380,9 @@ def _timestamp_to_millis(timestamp, seps=None):
     if seps:
         timestamp = _normalize_timestamp(timestamp, seps)
     Y, M, D, h, m, s, millis = _split_timestamp(timestamp)
+    # PERFORMANCE: according to the docs https://docs.python.org/3/library/datetime.html#datetime.date.timetuple
+    # Every call of .timetuple() performs `yday = d.toordinal() - date(d.year, 1, 1).toordinal() + 1` which
+    # is a relatively expensive computation
     secs = time.mktime(datetime.datetime(Y, M, D, h, m, s).timetuple())
```
xylix, line 386: @pekkaklarck I guess it would be bad for readability to write a complete custom implementation, but if we could find some way to cache the processing of datetime `dates` (since in a single robot log those should be quite constant / small amount of different dates) that should give some performance boost.

src/robot/reporting/jsbuildingcontext.py
```diff
@@ -47,6 +48,7 @@ def string(self, string, escape=True, attr=False):
             string = (html_escape if not attr else attribute_escape)(string)
         return self._strings.add(string)
 
+    @functools.lru_cache
```
xylix, line 51: The things to keep in mind here:
- Using these automatic cachings means that the function should be side-effect free or things will go awry.
- The cache size can be tuned. My test data being just a big dumb with very few separate keywords called makes it very easy to cache.
  - Should investigate memory impact.
  - Investigate performance impact of smaller / lower cache value (the cache is dict, but even being logarithmic if it grows large it can have negative performance impact.)

xylix: But on a general level just this code change made `rebot` runs on my machine 10-15% faster (from 18 seconds to 16 without flattenkeywords and from 9 to 7-8 with flattenkeywords).

pekkaklarck: Plr

src/robot/utils/robottime.py
pekkaklarck: I think we already have a cache for timestamps. I'm on mobile and hard check that.

The whole timestamp and elapsed time generation/calculation could be rewritten. Instead of model objects having `starttime` and `endtime` as separate strings, they should internally have start time (as datetime or epoch second) and increment (timeselta or integer) and `starttime/endtime/elapsedtime` should be (cached?) propertys. Too big change fror 5.0 but possible in 5.1.

xylix:
Notes about my flamegraph analysis of running `rebot` commands:
profile

Use grain of salt when looking at exact percentages, since this is only one run on a single machine using a large while benchmark as the test data generation suite. Aim is to validate with other data.

General notices:
- Since RF codebase uses high amounts of inheritance based and overridden method abstraction patterns, the call stacks are pretty high. However there are significant percentages (10% or so) of execution spent at several hot zones. One of them being the HTML conversion of keyword objects that the PR code already looks to fix with LRU_cache
  - There are timestamp related hot paths where caching probably won't work as well, since timestamps are pretty unique for every KW call in the output.xml 

Hot code paths (% of rebot execution time):
- timestamp_to_secs 8.95% + 13.10% + 3.41% = 25.46%
- `write_over_threshold` `robot/reporting/jswriter:87` 39.94%
  - does various calls, centering around:
    - recursing itself: `robot/reporting/jswriter:87` 
    - `dump` `robot/htmldata/jsonwriter:120`
    - `dump` `robot/htmldata/jsonwriter:52`

src/robot/utils/robottime.py
pekkaklarck: Now on laptop so easier to look at the code. The cache I mentioned is ´TIMESTAMP_CACHE` and it affects getting start/end time. The code above is used for getting elapsed time and probably could be enhanced. Elapsed time is needed so often that even pretty small performance enhancement can have real measurable effects.

pekkaklarck: I was surprised the code above uses `datetime().timetuple()` and instead doesn't create a timetuple based on the values already at hand. I looked at the history, and the code used to be like this:

```python
secs = time.mktime((Y, M, D, h, m, s, 0, 0, time.daylight))
```

and it was changed to fix a DST bug in d684b6127b8d58bc72bd73f130d11246a8ecdbf5. The above code is clearly wrong, because `time.daylight` indicates is DST active at _now_ and we may be generating time for some time in the past. Based on the [time.mktime docs](https://docs.python.org/3/library/time.html#time.mktime) we could just use `-1` instead, though, and avoid the need to use `datatime` altogether.

I tested changing the code to 

```python
secs = time.mktime((Y, M, D, h, m, s, 0, 0, -1))
```
and all unit and acceptance tests pass. Unfortunately the commit where the DST bug was fixed didn't add tests to validate the fix, so I'm not 100% sure this doesn't reintroduce the issue. Should probably add some more tests and check they work the same way with the current implementation and with the code above. Before that it would obviously be a good idea to test does the above code enhance performance at all. Could you @xylix test that with your data sets?

pekkaklarck: Promising micro benchmark:
```python
>>> Y, M, D, h, m, s, *_ = time.localtime() 
>>> timeit(lambda: time.mktime(datetime(Y, M, D, h, m, s).timetuple()))
2.010632199999236
>>> timeit(lambda: time.mktime((Y, M, D, h, m, s, 0, 0, -1)))
1.285105854003632
```

xylix: Yeah I will test this change.

I'm now testing with 
- My 1 million loop iterations output.xml (80 MB)
- the Robot Framework test suite (around 32 MB) output.xml
- Robot Framework browser  output.xml (just 228 kb)

xylix: The code does seem to get around 0.5 seconds faster with the new implementation. The next bottleneck for the timestamps seems to be the `roundup` function line 47 in misc.py, which calls pythons standard library `round()`.
https://github.com/robotframework/robotframework/blob/3728874a215fa27d7f80a93488b6e68fe336b135/src/robot/utils/misc.py#L23

My opinion is that `secs = time.mktime((Y, M, D, h, m, s, 0, 0, -1))` seems to be the simpler code (I guess it does have a couple of magic numbers, but it doesn't have 2 extra datetime library function / method calls), so as long as it gives same output and performs slightly better I would consider it the nicer implementation.

pekkaklarck: I tested that the proposed implementation and the current implementation around a DST boundary. They return the same result so the change ought to be safe. Because [datetime.timetuple()](https://docs.python.org/3/library/datetime.html#datetime.datetime.timetuple) sets `tm_isdst` to `-1` by default, this isn't a big surprise.

A strange thing is that I also tested subtracting `datetime` objects around a DST boundary and what `total_seconds()` of the returned `timedelta` returned was different than what our `get_elapsed_time` returns. Either we have bug or subtracting `datetime` objects doesn't take DST into account. The latter could be the case because `datetime` objects are naive by default and know nothing about timezones. At this point of the RF 5.0 release cycle, performance enhancement that preserves the old behavior sounds anyway safer.

pekkaklarck: That `roundup` could probably be removed altogether or at least all its usages from `robottime.py` could be. Its usages are:
- Don't convert integers to floats. This use case isn't valid anymore because with Python 3 e.g. `round(10)` is `10` (but it was `10.0` with Python 2.
- Consistent rounding of "half values" between Python versions. For example, `round(0.5)` is `1` with Python 2 but `0` with Python 3. Now that we only support Python 3, we could just use Python 3 rounding.
- More familiar rounding behavior. It can be strange for users if `0.5` is rounded to `0` when in school they were told it should round to `1`. There are reasons for Python 3 using "bankers rounding" but in some contexts rounding "as thought in school" could still be useful.

I'm not is the last point important enough to preserve `roundup` and using it in some places, but I would say that when rounding times we can just go with the standard `round`. I assume it to be implemented in C and being fast.

src/robot/reporting/jsbuildingcontext.py
xylix:
```
--- 5.0b2.dev1 master ---

(.venv)  ~/C/t/rf-5.0-iteration-longrun (master)> python benchmark.py
Time for simple-million-while-output.xml
146.158634542
Time for ./rfbrowser-output.xml
0.18507958300000382
Time for ./rf-full-suite-output.xml
44.82903858399999


--- Switch branch to faster-jsbuildingcontext ---

Time for simple-million-while-output.xml
134.029772417
Time for ./rfbrowser-output.xml
0.164100042
Time for ./rf-full-suite-output.xml
38.828153250000014
```

These are summed (maybe averages with listed deviation would be better) times of 10 runs of each dataset with `timeit`.

Increase seems to be around 10% both with small and larger output.xml:s. I did not do any memory profiling yet.

src/robot/utils/robottime.py
xylix: I'm neutral on which rounding behaviour makes more sense, slightly for keeping `utils` stable at this point of release cycle.

But fair point that we could just `round` times to simplify and possibly make them faster. But that difference may require another synthetic benchmark to measure if it is too small to notice in real runs.

src/robot/reporting/jsbuildingcontext.py
```diff
@@ -47,6 +48,7 @@ def string(self, string, escape=True, attr=False):
             string = (html_escape if not attr else attribute_escape)(string)
         return self._strings.add(string)
 
+    @functools.lru_cache(maxsize=128)
     def html(self, string):
         return self.string(html_format(string), escape=False)
```
pekkaklarck, line 53: I assume time is spent in `html_format`. Another way to avoid calling it with all strings would be just passing the string to `self.string` with an indication that it needs to be HTML formatted. That formatting could then be done when strings are read from `self.strings`.

pekkaklarck: That `self.strings` is kind of a cache already. We just do work before we put stuff there.

xylix: Do you want to handle this in `self.strings`? I could close the PR if that is a better solution.

pekkaklarck: That was done in 8c08fe7cc47bc2bae3ba51b26c3af0391ff117b1. You could still check does `lru_cache` give additional benefits, but I believe difference is small.

xylix: Was unmeasurably small. Closing.

pekkaklarck:
Thanks for measurements @xylix! Pinpointing the bottlenecks was really helpful.

Issue #4262, opened 2022-03-08
Honor `SOURCE_DATE_EPOCH` environment variable when generating library documentation


Apteryks:
This leads to nondeterminism in the build artifacts (the documentation) of RobotFramework itself or its libraries.  The issue was discovered at the time RobotFramework was packaged for GNU Guix.  For more context and the associated [pull request](https://github.com/robotframework/robotframework/pull/3140).

pekkaklarck:
The PR referenced above was closed in favor of #4263. I hopefully have time to look at it later today. If it looks fine, we can still get this into RF 5.0.

pekkaklarck:
My understanding is that the idea of `SOURCE_DATE_EPOCH` is being able to generate reproducible builds. When packaging Robot Framework itself, we don't include date/time information anywhere, we only include the version number. As explained in BUILD.rst, our packages are created using setup.py:

```
python setup.py sdist --formats zip bdist_wheel
```

Version information is in `setup.py` itself and in `src/robot/version.py`, and they are both maintained using `invoke` as also explained in BUILD.rst. Version information has been set and committed before creating a tag, so you can recreate same package by checking out the same version control tag and using `setup.py` yourself.

Based on the discussion in #4263, the main motivation for `SOURCE_DATE_EPOCH` in our context is having consistent time information in generated library documentation. Libdoc outputs contain both the version information and generation date, so you cannot regenerate the exact same outputs that have been created earlier. I don't really understand why the generation time out HTML files really matters, but if it helps third-party packaging I'm fine making it configurable.

The initial implementation in #4263 had a problem that all calls to get the current timestamp required checking is the `SOURCE_DATE_EPOCH` environment variable available. We need start and end timestamp for each test, keyword, loop iteration, etc. during execution, so this code is _very_ performance critical. Checking environment variables is fast but not free, and doing that only to be able to regenerate exactly same Libdoc HTML outputs is not an important enough reason to slow down execution. The implementation needs to change and we already discussed about alternatives in the PR.

`SOURCE_DATE_EPOCH` affecting `robot.utils.get_timestamp` would also mean that if that environment variable is set, all start and end times in log and other outputs would be same. That would then mean that all elapsed times would be zero. We could argue that it allows having reproducible times, but I see absolutely no real use case for such configuration in this context.

[...]

PR #4263, opened 2022-03-08
robottime: Honor the SOURCE_DATE_EPOCH environment variable.

Apteryks:
* src/robot/utils/robottime.py:
(TimestampCache._get_epoch): Retrieve date from SOURCE_DATE_EPOCH if
it is defined, otherwise from time.time().
* utest/output/test_logger.py (TestLogger.test_write_to_one_logger):
Verify the timestamp via a regular expression.
(TestLogger.test_write_to_logger_with_source_date_epoch_set): New test.
* doc/userguide/src/Appendices/ReproducibleBuilds.rst: New file.
* doc/userguide/src/RobotFrameworkUserGuide.rst: Register it.


src/robot/utils/robottime.py
```diff
@@ -420,6 +421,8 @@ def get_timestamp(self, daysep='', daytimesep=' ', timesep=':', millissep='.'):
 
     # Seam for mocking
     def _get_epoch(self):
+        if os.getenv('SOURCE_DATE_EPOCH'):
+            return float(os.getenv('SOURCE_DATE_EPOCH'))
         return time.time()
```
pekkaklarck, line 426: This is pretty performance critical code, timestamps are generated for each keyword, loop, etc. both when they start and end, and thus always using `os.getenv` isn't great. It would be better to have the `if` outside the `_get_epoch` like this:

```python
if 'SOURCE_DATE_EPOCH' in os.environ:
    def _get_epoch(self):
        return float(os.getenv('SOURCE_DATE_EPOCH'))
else:
    def _get_epoch(self):
        return time.time()

File: doc/userguide/src/Appendices/ReproducibleBuilds.rst [hunk 2]
```diff
@@ -0,0 +1,9 @@
+Reproducible Builds
+===================
+
+Robot Framework honors the `SOURCE_DATE_EPOCH
+<https://reproducible-builds.org/docs/source-date-epoch/>` environment
+variable so that it can be built in a reproducible fashion.  When the
+`SOURCE_DATE_EPOCH` environment variable is set, timestamps embedded
+in generated documentation and log messages are set to its fixed
+value.
```
pekkaklarck, line 9: Would BUILD.rst be better place for this information than the User Guide?

pekkaklarck:
Could you @Apteryks still clarify the use case for this change a bit? I understand the general idea behind the `SOURCE_DATE_EPOCH` environment variable, but I don't fully understand how it would be used with Robot. All timestamps reading this environment value instead of using `time.time()` means that all tests, keywords, etc. in log have that time both as their start and end time. That doesn't sound too useful, and I was wondering should the environment variable be checked in some other place that where this PR now checks it.

I try to get RF 5.0 rc 1 finally out on Thursday so there's a bit hurry in getting this in. Feel free to ping me on Slack, preferably on the #devel channel so that others can also participate the discussion, to have more interactive discussion.

doc/userguide/src/Appendices/ReproducibleBuilds.rst
Apteryks: BUILD.rst seems to be specific to creating Robot Framework releases.  The `SOURCE_DATE_EPOCH` environment variable is most useful for the (automated) build tooling of 3rd parties packaging RobotFramework such as GNU Guix, NixOS or Debian that strive for reproducible builds.  So it doesn't really need to be user-facing, but it's also not at home in BUILD.rst.  I'm not sure of the value to document it, but if we do, an appendix seems the best place to me (it's there if someone looks for it, but it's in the background as most users probably do not care).

src/robot/utils/robottime.py
Apteryks: > Could you @Apteryks still clarify the use case for this change a bit?

My specific use case was just fixing the timestamps embedded in the generated documentation so that RobotFramework (and its libraries) could be built reproducibly in GNU Guix. Another, more contrived example, could be the desire that two RobotFramework test runs produce exactly the same output.

Apteryks: I've made the suggested change, and it expectedly broke the new `test_write_to_logger_with_source_date_epoch_set`; which I tried fixing by mocking the `TIMESTAMP_CACHE` global object:
```diff
modified   utest/output/test_logger.py
@@ -4,6 +4,7 @@ import time
 import unittest
 import unittest.mock
 
+import robot.utils.robottime
 from robot.utils.asserts import assert_equal, assert_true, assert_false
 from robot.utils.robottime import format_time
 
@@ -59,6 +60,8 @@ class TestLogger(unittest.TestCase):
                              logger.msg.timestamp))
 
     @unittest.mock.patch.dict(os.environ, {'SOURCE_DATE_EPOCH': "0"})
+    @unittest.mock.patch('robot.utils.robottime.TIMESTAMP_CACHE',
+                         robot.utils.robottime.TimestampCache())
     def test_write_to_logger_with_source_date_epoch_set(self):
         logger = LoggerMock(('Hello, SOURCE_DATE_EPOCH!', 'INFO'))
         self.logger.register_logger(logger)
```
But it didn't work.  Any suggestion?

doc/userguide/src/Appendices/ReproducibleBuilds.rst
pekkaklarck: Isn't `SOURCE_DATE_EPOCH` exactly about creating releases? I don't see problems enhancing BUILD.rst to cover also third party packaging.

src/robot/utils/robottime.py
pekkaklarck: Yeah, `if 'SOURCE_DATE_EPOCH' in os.environ:` in my proposal is executed when the `robottime` module is initially imported and thus patching it isn't easy. The code is so trivial that perhaps we wouldn't need a test at all. There could be just a comment explaining why `SOURCE_DATE_EPOCH` is used with an URL pointing to the original issue.

pekkaklarck:
I'm starting to believe `robottime` isn't the right place to look at `SOURCE_DATE_EPOCH` at all. See my recent comments to issue #4262 for details.

Apteryks:
This was closed by mistake, due to recreating the branch (sorry!).  I believe I've addressed the comments thus far in the newly opened https://github.com/robotframework/robotframework/pull/4286.

PR #4265, opened 2022-03-09
doc: ug2html.py: Do not depend on robotframeworklexer.

### LINKED ISSUE #4264: ug2html.py script requires obsolete robotframeworklexer
Hello,

The last release of the PyPI hosted robotframeworklexer was made on the 12th of January 2019, and hints itself that "[...] it is included in Pygments 1.6 and newer."  Pygments meanwhile has seen some fixes toward the RF lexer.

I'd suggest to remove this package from PyPI to avoid confusion and use Pygments exclusively from now on.

Thank you.

Apteryks:
Fixes #4264.

The RobotFramework lexer has long been included in Pygments now, and
the robotframeworklexer copy on PyPI wast last updated earlier 2019.

* doc/userguide/ug2html.py: Do not import robotframeworklexer, check
its version nor use it.  Bump Pygments requirement to 2.8, as that's
old enough and is what was used to test.
* requirements-build.txt: Drop robotframeworklexer requirement.
Update pygments required version.

### DIFF
--- a/doc/userguide/ug2html.py
+++ b/doc/userguide/ug2html.py
@@ -93,34 +93,26 @@ from docutils.parsers.rst import directives
 from pygments import highlight, __version__ as pygments_version
 from pygments.lexers import get_lexer_by_name
 
-# Use latest version, not version bundled with Pygments
-import robotframeworklexer
-
 
 def too_old(version_string, minimum):
     version = tuple(int(v) for v in version_string.split('.')[:2])
     return version < minimum
 
 
-if too_old(getattr(robotframeworklexer, '__version__', '1.0'), (1, 1)):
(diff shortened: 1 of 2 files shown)


pekkaklarck:
It's true that Pygments has supported RF for long time, but I'm not sure does it contain the latest version of the lexer or not. Did you check that?

A related problem is that the external lexer is badly outdated and doesn't support latest RF features. The external lexer is from the time RF's own parser was different than now. Nowadays It should just use Robot's own lexer, but I'm not sure what would be the best way to handle that dependency. It would be probably best to talk more about this on Slack.

pekkaklarck:
You seem to be right, I just tested that using Pygments on its own produces ok results.

Issue #4149, opened 2021-11-18
New User Guide styles


Snooz82:
Would be nice, if the UserGuide would look a little less 90th...

This Issue is not for changing everything to Sphinx. Just tuning the CSS a bit.
Could be nice for 4.1.3 already.

pekkaklarck:
Changing just CSS to make the UG look better is a good idea and I agree we can back port them to RF 4.1.3 that we anyway need to create at some point.

Bigger changes are probably best done by converting the UG to use [Sphinx](https://www.sphinx-doc.org/) that then makes it easy to host at [Read the Docs](https://readthedocs.org/).

pekkaklarck:
There were some reports related to problems on OSX using Chrome. Could be that problems were caused by environment configuration, but it's safer to not change anything anymore in RF 4.1.x maintenance branch. Changes can land to master and then be part of RF 5.0. Online version of the User Guide can be patched inplace separately.

pekkaklarck:
This was done in #4283. Thanks @Snooz82 for contribution!

PR #4283, opened 2022-03-17
updated-ug-style

Snooz82:
Updated css.

Features:

- Nicer Font
- Sidebar with TOC in 3 levels (this is depth 3 in rst file)
- If Browser to small TOC with 2 levels at the beginning
- DarkMode
- limited width for better readability
- looks good :-)
- Issue https://github.com/robotframework/robotframework/issues/4149

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ ext-lib
 .DS_Store
 doc/api/_build
 doc/libraries/*.html
+doc/libraries/*.json
 doc/userguide/RobotFrameworkUserGuide.html
 src/robotframework.egg-info
 log.html
--- a/doc/userguide/src/RobotFrameworkUserGuide.rst
+++ b/doc/userguide/src/RobotFrameworkUserGuide.rst
@@ -15,7 +15,7 @@
 | Licensed under the `Creative Commons Attribution 3.0 Unported`_ license
 
 .. contents:: Table of Contents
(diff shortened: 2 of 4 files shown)


Snooz82:
@pekkaklarck 

This PR replaces #4148 because i wanted to make these changes based on the current code base.
Still looks good, and should be in before Release of 5.0

pekkaklarck:
Style enhancements look good. Could you clarify the reason to generate JSON spec files?

Snooz82:
@pekkaklarck 
The json spec is for the code playground to have code completion available for standard libraries.

Snooz82:
See https://github.com/robotframework/robotframework/tree/gh-pages/latest/libdoc

pekkaklarck:
Ok, sounds reasonable. You may have mentioned that at some point but there was nothing about that in this PR.

Are changes in this PR enough to also include specs under gh-pages or does tasks.py there need to be updated? Is it a problem that this PR generates specs under `doc/libraries` but your example has `.../gh-pages/latest/libdoc/`?

pekkaklarck:
If there's still work needed related to spec generation and getting them under gh-pages, perhaps it would be best to remove that part from this PR and just merge the actual UG changes.

pekkaklarck:
The spec stuff is needed anyway so I guess it's ok to merge this.

PR #4285, opened 2022-03-20
Don't retry libdoc with :: and show all error messages.

fabioz:
This fixes a case where the user trying to build a library
(say `test_lib`) with arguments such as:

`test_lib::d:/testRepo/functions2.txt`

fails.

Given that `::` is present, it's no longer retried as a resource.

Also, the message shown to the user will now mention all errors
and not just the message saying that the resource file does not exist.

p.s.: see: https://github.com/robocorp/robotframework-lsp/issues/608#issuecomment-1073241116 for a case where this was misleading.

### DIFF
--- a/atest/robot/libdoc/cli.robot
+++ b/atest/robot/libdoc/cli.robot
@@ -61,7 +61,7 @@ Resource file in PYTHONPATH
 Non-existing resource
     [Template]    NONE
     ${stdout} =    Run Libdoc    nonexisting.resource whatever.xml
-    Should Be Equal    ${stdout}     Resource file 'nonexisting.resource' does not exist.${USAGE TIP}\n
+    Should contain    ${stdout}     Resource file 'nonexisting.resource' does not exist.${USAGE TIP}\n
 
 *** Keywords ***
 Run Libdoc And Verify Created Output File
--- a/src/robot/libdocpkg/builder.py
+++ b/src/robot/libdocpkg/builder.py
@@ -48,9 +48,15 @@ def _build(builder, source):
         # did not exist has been considered to be a library earlier, now we try to
         # parse it as a resource file.
         if (isinstance(builder, LibraryDocBuilder)
+                and '::' not in source
                 and not os.path.exists(source)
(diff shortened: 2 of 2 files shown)


fabioz:
p.s.: I see that there's 1 test failing (I can take a look at that if the general approach to show all failures is ok -- otherwise I can also just leave the last error and add the handling to not retrying if i `::` is shown).

pekkaklarck:
1. Do I get it right that the problem occurs if you use a Libdoc with a library name and argumetns like `MyLib::args.txt` and `MyLib` cannot be imported? Then Libdoc sees the `.txt` extension in arguments and retries imports as a resource file? If yes, then a bit better fix would probably be changing the `_get_extension` helper like this:
   ```python
   def _get_extension(source):
       path, args = source.split('::', 1)
       return os.path.splitext(path)[1][1:].lower()
   ```

2. Are you certain that the original error needs to be preserved if we retry import as a resource? I think it makes the error confusing in the common case. In this case where our resource file detection failed it would have helped, but I hope we don't have such problems anymore.

3. The fix needs tests. Because this is a bug fix, I can add them myself as well. Possible change to preserve original error needs a test validating both errors too.

4. This bug would need an issue for tracking purposes.

fabioz:
Well, actually in the case for the language server, this is only used when dealing with library imports, so, the attempt to import as a resource (and thus the related message is always wrong...).

I created https://github.com/robotframework/robotframework/issues/4351 to track this and gave a bit more info, so, maybe we can move the discussion there ;)

I can also add a test (after we decide on the proper approach to fix).

pekkaklarck:
I fixed the error message problem in 04db11948955c215c003af24bdf3c2aa6631cd26. The fix was slightly different than in this PR because I didn't preserve both errors if importing something both as a library and as a resource fails. I don't think that's really a good idea now that resource file detection ought to work correctly. If you disagree, please submit a separate issue and we can discuss that further. Because the main issue is now fixed, I close this PR. Thanks for reporting the error and providing the initial fix!

PR #4286, opened 2022-03-21
doc: Honor SOURCE_DATE_EPOCH for documentation timestamps.

### LINKED ISSUE #4262: Honor `SOURCE_DATE_EPOCH` environment variable when generating library documentation

Apteryks:
Fixes #4262.

* doc/userguide/ug2html.py (create_userguide): Do not embed
timestamps.
* src/robot/utils/robottime.py (get_timestamp_for_doc): New procedure.
* src/robot/utils/__init__.py: Export it.
* src/robot/libdocpkg/model.py (LibraryDoc.to_dictionary): Use it.
* src/robot/libdocpkg/xmlwriter.py (LibdocXmlWriter._write_start):
Likewise.
* BUILD.rst: Document it.

### DIFF
--- a/BUILD.rst
+++ b/BUILD.rst
@@ -204,6 +204,12 @@ Creating distributions
 
 7. Documentation
 
+   - For a reproducible build, set the ``SOURCE_DATE_EPOCH``
+     environment variable to a constant value, corresponding to the
+     date in seconds since the Epoch (also known as Epoch time).  For
+     more information regarding this environment variable, see
+     https://reproducible-builds.org/docs/source-date-epoch/.
+
    - Generate library documentation::
 
        invoke library-docs all
--- a/atest/robot/libdoc/spec_library.robot
+++ b/atest/robot/libdoc/spec_library.robot
@@ -1,4 +1,5 @@
 *** Settings ***
+Library           OperatingSystem
 Suite Setup       Run Libdoc And Parse Output    ${TESTDATADIR}/ExampleSpec.xml
 Resource          libdoc_resource.robot
 
@@ -97,6 +98,13 @@ Keyword Source Info
     Copy File    ${TESTDATADIR}/ExampleSpec.xml    %{TEMPDIR}/Example.libspec
     Run Libdoc And Parse Output    %{TEMPDIR}/Example.libspec
     Test Everything
+
+SOURCE_DATE_EPOCH is honored in Libdoc output
+    [Setup]    Set Environment Variable    SOURCE_DATE_EPOCH    0
+    Copy File    ${TESTDATADIR}/ExampleSpec.xml    %{TEMPDIR}/Example.libspec
+    Run Libdoc And Parse Output    %{TEMPDIR}/Example.libspec
+    Generated Should Be    1970-01-01T00:00:00+00:00
+    [Teardown]    Remove Environment Variable    SOURCE_DATE_EPOCH
 
 *** Keywords ***
 Test Everything
--- a/atest/robot/standard_libraries/operating_system/modified_time.robot
+++ b/atest/robot/standard_libraries/operating_system/modified_time.robot
@@ -8,7 +8,7 @@ ${TESTFILE}       %{TEMPDIR}${/}robot-os-tests${/}f1.txt
 *** Test Cases ***
 Get Modified Time As Timestamp
     ${tc} =    Check Test Case    ${TESTNAME}
-    Should Match Regexp    ${tc.kws[0].msgs[0].message}    Last modified time of '<a href=.*</a>' is 20\\d\\d-\\d\\d-\\d\\d \\d\\d:\\d\\d:\\d\\d
+    Should Match Regexp    ${tc.kws[0].msgs[0].message}    Last modified time of '<a href=.*</a>' is \\d\\d\\d\\d-\\d\\d-\\d\\d \\d\\d:\\d\\d:\\d\\d
 
 Get Modified Time As Seconds After Epoch
     ${tc} =    Check Test Case    ${TESTNAME}
--- a/src/robot/libdocpkg/model.py
+++ b/src/robot/libdocpkg/model.py
@@ -19,11 +19,11 @@ from itertools import chain
 
 from robot.model import Tags
 from robot.running import ArgumentSpec
-from robot.utils import getshortdoc, get_timestamp, Sortable, setter
+from robot.utils import getshortdoc, Sortable, setter
 
 from .htmlutils import DocFormatter, DocToHtml, HtmlToText
 from .writer import LibdocWriter
-from .output import LibdocOutput
+from .output import LibdocOutput, get_generation_time
 
 
 class LibraryDoc:
@@ -114,7 +114,7 @@ class LibraryDoc:
             'name': self.name,
             'doc': self.doc,
             'version': self.version,
-            'generated': get_timestamp(daysep='-', millissep=None),
+            'generated': get_generation_time(),
             'type': self.type,
             'scope': self.scope,
             'docFormat': self.doc_format,
--- a/src/robot/libdocpkg/output.py
+++ b/src/robot/libdocpkg/output.py
@@ -13,7 +13,9 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import datetime
 import os
+import time
 
 from robot.utils import file_writer
 
@@ -40,3 +42,14 @@ class LibdocOutput:
                 os.remove(self._output_path)
             except OSError:
                 pass
+
+
+def get_generation_time():
+    """Return a timestamp that honors `SOURCE_DATE_EPOCH`.
+
+    This timestamp is to be used for embedding in output files, so
+    that builds can be made reproducible.
+    """
+    ts = float(os.getenv('SOURCE_DATE_EPOCH', time.time()))
+    dt = datetime.datetime.fromtimestamp(round(ts), datetime.timezone.utc)
+    return dt.isoformat()
--- a/src/robot/libdocpkg/xmlwriter.py
+++ b/src/robot/libdocpkg/xmlwriter.py
@@ -13,9 +13,9 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from datetime import datetime
-
 from robot.utils import XmlWriter
+
+from .output import get_generation_time
 
 
 class LibdocXmlWriter:
@@ -32,12 +32,11 @@ class LibdocXmlWriter:
         self._write_end(writer)
 
     def _write_start(self, libdoc, writer):
-        generated = datetime.utcnow().replace(microsecond=0).isoformat() + 'Z'
         attrs = {'name': libdoc.name,
                  'type': libdoc.type,
                  'format': libdoc.doc_format,
                  'scope': libdoc.scope,
-                 'generated': generated,
+                 'generated': get_generation_time(),
                  'specversion': '4'}
         self._add_source_info(attrs, libdoc)
         writer.start('keywordspec', attrs)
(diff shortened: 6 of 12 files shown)


Apteryks:
OK, so this seems to have broken acceptance tests, due to slightly changing the timestamp format used in XML generated files.  I'll try looking at it later.

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -32,12 +30,11 @@ def write(self, libdoc, outfile):
         self._write_end(writer)
 
     def _write_start(self, libdoc, writer):
-        generated = datetime.utcnow().replace(microsecond=0).isoformat() + 'Z'
         attrs = {'name': libdoc.name,
                  'type': libdoc.type,
                  'format': libdoc.doc_format,
                  'scope': libdoc.scope,
-                 'generated': generated,
+                 'generated': get_timestamp_for_doc(),
```
pekkaklarck, line 37: This changes the timestamp from current `2022-05-27T19:08:42Z` to `2022-05-27T19:07:15+00:00`. That ought to be the same thing, and mach `xs:dateTime` that we use in `libdoc.xsd` schema, but because external tools may depend on the exact format this makes the change backwards incompatible. It's not so big change it couldn't be done in RF 5.1, but need to still be mentioned in the release notes. I'll add a label to the issue so that we don't forget about it.

src/robot/libdocpkg/model.py
```diff
@@ -113,7 +113,7 @@ def to_dictionary(self):
             'name': self.name,
             'doc': self.doc,
             'version': self.version,
-            'generated': get_timestamp(daysep='-', millissep=None),
+            'generated': get_timestamp_for_doc(),
```
pekkaklarck, line 116: This changes the current timestamp from `2022-05-27 22:07:15` to `2022-05-27T19:07:15+00:00`. The old timestamp was naive and I guess it makes sense to have the timezone information there. The change is only in HTML output so it shouldn't cause real backwards compatibility problems. I don't particularly like the `T` separator in outputs targeted for humans, but don't care strongly about it.

src/robot/utils/robottime.py
```diff
@@ -316,6 +318,13 @@ def get_timestamp(daysep='', daytimesep=' ', timesep=':', millissep='.'):
     return TIMESTAMP_CACHE.get_timestamp(daysep, daytimesep, timesep, millissep)
 
 
+def get_timestamp_for_doc():
+    """Return a timestamp that honors `SOURCE_DATE_EPOCH`."""
+    ts = float(os.getenv('SOURCE_DATE_EPOCH', time.time()))
+    dt = datetime.datetime.fromtimestamp(round(ts), datetime.timezone.utc)
+    return dt.isoformat()
+
+
```
pekkaklarck, line 327: This utility is used only in two places and only by Libdoc. I don't think adding it to `robot.utils` is a good idea because it's part of our public API and we cannot change it in the future without backwards compatibility concerns. There's no good place to have utils in the `robot.libdocpkg` package, but I think it would be fine to just inline these few lines of code. It would then make it possible to have slightly different formatting in the timestamp as well.

pekkaklarck (changes requested):
Looks good but I think it's better to inline the timestamp generation code than have it in `robot.utils` that's part of our public API. Change to some tests seem unnecessary as well. See the more detailed comments for details.

src/robot/libdocpkg/model.py
pekkaklarck: Ooops, my comment was partly invalid. This changes also JSON spec files, not only HTML output. Thus this change is backwards incompatible the same way as the change to XML spec files. Also makes the `T` separator pretty much mandatory.

atest/robot/standard_libraries/operating_system/modified_time.robot
```diff
@@ -8,7 +8,7 @@ ${TESTFILE}       %{TEMPDIR}${/}robot-os-tests${/}f1.txt
 *** Test Cases ***
 Get Modified Time As Timestamp
     ${tc} =    Check Test Case    ${TESTNAME}
-    Should Match Regexp    ${tc.kws[0].msgs[0].message}    Last modified time of '<a href=.*</a>' is 20\\d\\d-\\d\\d-\\d\\d \\d\\d:\\d\\d:\\d\\d
+    Should Match Regexp    ${tc.kws[0].msgs[0].message}    Last modified time of '<a href=.*</a>' is \\d\\d\\d\\d-\\d\\d-\\d\\d \\d\\d:\\d\\d:\\d\\d
```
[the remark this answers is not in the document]
Apteryks: No, there're needed because with the ability to set the date all the way back to 1969 (epoch == 0), the date stamp prefix no longer is 20 and the tests would fail.

src/robot/utils/robottime.py
Apteryks: Since the timestamp gets output in various files, you can't change it at any time lightly anyway, so I don't see the concern to have the procedure generating it live in a public API place.  As both a maintainer and user, I'd prefer consistency in all dates (figure its format once, and expect it everywhere).

BUILD.rst
```diff
@@ -204,6 +204,9 @@ Creating distributions
 
 7. Documentation
 
+   - For a reproducible build, set the ``SOURCE_DATE_EPOCH``
+     environment variable to 1.
```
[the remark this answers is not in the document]
Apteryks: I've expound on SOURCE_DATE_EPOCH a bit, thanks!

src/robot/utils/robottime.py
pekkaklarck: Actually, there could probably be a `timestamp` property in the `LibraryDoc` model object itself containing this functionality. That would avoid duplication.

pekkaklarck:
One more thing to do is creating tests for this functionality. Need to add tests for JSON and XML specs where SOURCE_DATE_EPOCH is set to some know value before generating specs.

pekkaklarck (changes requested):
Things to do:
- Move the code to generate the timestamp from `robot.utils` to `robot.libdocpkg` to avoid it being part of the public API.
- Revert changes to tests under `atest/{robot,testdata}/standard_libraries` unless they are actually needed. They don't seem to be related to this change and mostly make tests less specific.
- Add test that `SOURCE_DATE_EPOCH` can be used to control the generation time in LIbdoc spec files. Should basically have same test for the generation time as normally but set the environment variable to some know value before running Libdoc. Then we should be able to validate that the exact generation time is what we expect. If you need help with tests, comment here or ping me on Slack.

Apteryks:
> Things to do:
> 
>     * Move the code to generate the timestamp from `robot.utils` to `robot.libdocpkg` to avoid it being part of the public API.

OK.  That seems simple enough.  I'll look into it.

>     * Revert changes to tests under `atest/{robot,testdata}/standard_libraries` unless they are actually needed. They don't seem to be related to this change and mostly make tests less specific.

These changes are actually needed, as I wrote here https://github.com/robotframework/robotframework/pull/4286/files#r883940643.  Other adjustments are needed due to the format having changed slightly.

>     * Add test that `SOURCE_DATE_EPOCH` can be used to control the generation time in LIbdoc spec files. Should basically have same test for the generation time as normally but set the environment variable to some know value before running Libdoc. Then we should be able to validate that the exact generation time is what we expect. If you need help with tests, comment here or ping me on Slack.

Thanks, I'll look into adding an extra test here as well.

src/robot/libdocpkg/xmlwriter.py
```diff
@@ -13,8 +13,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from datetime import datetime
-
+from robot.libdocpkg.output import get_timestamp_for_doc
 from robot.utils import XmlWriter
```
pekkaklarck, line 17: We've used relative imports with package internal imports. That would change these imports to:

```python
from robot.utils import XmlWritre

from .output import get_timestamp_for_doc
```

src/robot/libdocpkg/output.py
```diff
@@ -40,3 +42,14 @@ def __exit__(self, *exc_info):
                 os.remove(self._output_path)
             except OSError:
                 pass
+
+
+def get_timestamp_for_doc():
```
pekkaklarck, line 47: I'd prefer `get_generation_time` as the function name.

atest/robot/libdoc/spec_library.robot
```diff
@@ -1,4 +1,5 @@
 *** Settings ***
+Library           OperatingSystem
```
pekkaklarck, line 2: Ought to be already imported by the imported resource file.

atest/robot/libdoc/spec_library.robot
```diff
@@ -98,6 +99,28 @@ Keyword Source Info
...
+        Set Environment Variable    SOURCE_DATE_EPOCH    0
+        Run Libdoc And Parse Output    %{TEMPDIR}/Example.libspec
+        Generated Should Be    1970-01-01T00:00:00+00:00
+    FINALLY
+        IF    $current_source_date_epoch is None
+            Remove Environment Variable    SOURCE_DATE_EPOCH
+        ELSE
+            Set Environment Variable    SOURCE_DATE_EPOCH    ${current_source_date_epoch}
+        END
+    END
```
pekkaklarck, line 122: This is _really_ complicated. Why not just set `SOURCE_DATE_EPOCH` to a know value and validate that generation time is expected? I see no point in checking is the environment variable set earlier.

The environment variable should probably then be removed in teardown to avoid subsequent tests being affected. That obviously changes environment of the current process if `SOURCE_DATE_EPOCH` was set in the system earlier, but no acceptance test should rely on external environment variables like that. If considered important, that environment variable could be cleared automatically either by Libdoc tests or even in `atest/run.py` before starting execution.

pekkaklarck (changes requested):
The created test is _really_ complicated and needs to be simplified.

Would also be good to revert unrelated test changes. Or are they actually needed?

I also added some nitpickey comments about imports and function name, but I can change those after the merge myself as well.

atest/robot/libdoc/spec_library.robot
Apteryks: Clearing the value in atest/run.py should be doable, good idea. I'll try it.

atest/robot/libdoc/spec_library.robot
pekkaklarck: There's no single best practice but we have generally imported only resource files in our tests if those resource files anyway import libraries we need. One use for resource files with Robot is to avoid importing multiple libraries or resource files in a test case file to keep them simpler.

src/robot/libdocpkg/xmlwriter.py
pekkaklarck: Interesting, I don't remember seeing that recommendation in PEP-8 earlier. Anyway, our style has been doing relative inter-package imports and only use absolute imports when importing something from other top-level packages. I think it makes sense with packages with lot of sub-packages in general, but most importantly imports should use consistent style.

atest/robot/standard_libraries/operating_system/modified_time.robot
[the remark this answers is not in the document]
pekkaklarck: That would require changing system time wouldn't it? Why would you ever want to do that when running Robot's acceptance tests?

atest/robot/libdoc/spec_library.robot
Apteryks: I see!

atest/robot/libdoc/spec_library.robot
Apteryks: I've implemented a much simplified (albeit less clean) version in the latest push.

pekkaklarck (approved):

Issue #4303, opened 2022-04-07
Support disabling continue-on-failure mode


pekkaklarck:
When a keyword fails, Robot's default behavior is to stop executing the current test or a user keyword. There are, however, few cases where the remaining keywords in the test/keyword are run (i.e. the continue-on-failure mode is on):

1. When `robot:continue-on-failure` or `robot:recursive-continue-on-failure` tags have been used (#2285).
2. When executing keywords in test or keyword teardown.
3. When a test has a template, all top level iterations are run even if one of them would fail.

It would sometimes be convenient to be able to turn the continue-on-failure more off in the above cases. A concrete example is #4293, where TRY/EXCEPT behaves differently in teardowns and elsewhere. When `robot:recursive-continue-on-failure` was added, we [discussed](https://github.com/robotframework/robotframework/issues/2285#issuecomment-867528381) about adding `robot:no-continue-on-failure` to be able to disable it. It wasn't done then, but now would be a good time. For consistency reasons we should also support templates.

oboehmer:
I can take this, @pekkaklarck ..

pekkaklarck:
That would be great @oboehmer!

pekkaklarck:
I think the only design decision related to this is deciding should `robot:no-continue-on-failure´ be recursive or not. I got a feeling it should be recursive even though it would be inconsistent with `robot:continue-on-failure`. Alternatively we could add separate `robot:recursive-no-continue-on-failure` but that's getting pretty complicated.

oboehmer:
I would opt for it not to be recursive.. Not to continue is the default, so why should it be recursive? Let me create a PR shortly, we can discuss there..

oboehmer:
please check #4306

oboehmer:
one question is also if a test-level `robot:no-continue-on-failure` would propagate into the Teardown method or not?

```
No-continue-on-failure-2 in Teardown
    [Documentation]    FAIL 1
    [Tags]   robot:no-continue-on-failure
    [Teardown]   Run Keywords   Fail   1   AND   Fail  2
    Log    does not matter
```

pekkaklarck:
[...]

PR #4306, opened 2022-04-07
Support disabling continue-on-failure mode in teardown and with template

oboehmer:
Enable a user-keyword executed within teardown context to override the default continue-on-failure behaviour in teardown, i.e. 

```
No-continue-on-failure in Teardown
    [Documentation]    FAIL kw1a
    [Teardown]   Failure in user keyword with no-continue tag
    Log    does not matter

*** Keywords ***
Failure in user keyword with no-continue tag
    [Tags]   robot:no-continue-on-failure
    Fail   kw1a
    Log    This should not be executed
    Fail   kw1b
```

Also allow the user to disable default for continue-on-failure with templates. The below test case will abort the test case after the 2nd iteration (42!=43), the last values (Something and Different) are not compared

```
No-continue-on-failure with Template
    [Documentation]    FAIL    42 != 43
    [Tags]   robot:no-continue-on-failure
    [Template]    Should Be Equal
    Same         Same
    42           43
    Something    Different
```

This addresses #4303

### DIFF
--- a/atest/robot/running/continue_on_failure_tag.robot
+++ b/atest/robot/running/continue_on_failure_tag.robot
@@ -63,8 +63,14 @@ Recursive continue in test with Set Tags and two nested UK without tag
 Recursive continue in test with tag and two nested UK with and without tag
     Check Test Case    ${TESTNAME}
 
+Recursive continue in test with tag and UK with no-continue tag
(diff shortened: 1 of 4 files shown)


oboehmer:
closing the PR, will re-open once design decisions have been finalized

PR #4308, opened 2022-04-10
Implementation of "Sleep Between" keywords

sebastianciupinski:


pekkaklarck:
Could you clarify the use case for this new keyword? Assuming the use case is valid, I'd believe it's better to enhance the existing Sleep keyword than adding a new one. Tests would also be needed.

sebastianciupinski:
Adding new keyword instead of updating `Sleep` was caused that extending `Sleep` with varargs or kwargs could complicate `Sleep` pure simplicity.

Motivation is to just assure some randomization when writing RPA scripts.

Tests can be added when we agree that there is need for such keyword

pekkaklarck:
I don't consider this important enough to warrant adding a new BuiltIn keyword. Enhancing existing Sleep is possible if we can figure out a way to do it in backwards compatible manner and without making the usage too complicated.

Pratapreddy1188:
It could be useful sometimes and **Sleep Random** would be appropriate Naming, may be for now you could make it to your project specific , a function which could generate a random value between your max and min and wait/sleep for that duration.

pekkaklarck:
It's better to enhance the current Sleep keyword than adding a new one. It could either support ranges like `1s - 2s` or have some optional argument to control this. Please submit an issue about that and we can discuss the design there. Once we agree on the design, a new PR can be created.

PR #4313, opened 2022-04-13
Add robot:stop-on-failure / recursive-stop-on-failure to disable continue-on-failure in teardown/template

oboehmer:
This PR adds two new tags to control the continue-on-failure behaviour in teardown or test template, addressing #4303

### DIFF
--- a/src/robot/errors.py
+++ b/src/robot/errors.py
@@ -139,14 +139,14 @@ class ExecutionStatus(RobotError):
         if self.syntax or self.exit or self.skip or self.test_timeout:
             return False
         if templated:
-            return True
+            return context.continue_on_failure(default=True)
         if self.keyword_timeout:
             if context.in_teardown:
                 self.keyword_timeout = False
             return False
-        if context.in_teardown or context.continue_on_failure:
-            return True
-        return self.continue_on_failure
+        if context.in_teardown:
+            return context.continue_on_failure(default=True)
+        return context.continue_on_failure(default=False) or self.continue_on_failure
 
     def get_errors(self):
         return [self]
--- a/src/robot/running/context.py
+++ b/src/robot/running/context.py
@@ -126,14 +126,22 @@ class _ExecutionContext:
     def variables(self):
         return self.namespace.variables
 
-    @property
-    def continue_on_failure(self):
-        parents = ([self.test] if self.test else []) + self.user_keywords
+    def continue_on_failure(self, default=False):
+        if self.in_teardown:
+            parents = self.user_keywords
+        else:
+            parents = ([self.test] if self.test else []) + self.user_keywords
         if not parents:
+            return default
+        if 'robot:stop-on-failure' in parents[-1].tags:
             return False
         if 'robot:continue-on-failure' in parents[-1].tags:
             return True
-        return any('robot:recursive-continue-on-failure' in p.tags for p in parents)
+        if any('robot:recursive-stop-on-failure' in p.tags for p in parents):
+            return False
+        if any('robot:recursive-continue-on-failure' in p.tags for p in parents):
+            return True
+        return default
 
     @property
     def allow_loop_control(self):
--- a/atest/robot/running/continue_on_failure_tag.robot
+++ b/atest/robot/running/continue_on_failure_tag.robot
@@ -83,3 +83,6 @@ stop-on-failure in User Keyword
 
 stop-on-failure with Template
     Check Test Case    ${TESTNAME}
+
+stop-on-failure with Template and Teardown
+    Check Test Case    ${TESTNAME}
--- a/atest/robot/running/continue_on_failure_tag.robot
+++ b/atest/robot/running/continue_on_failure_tag.robot
@@ -63,8 +63,23 @@ Recursive continue in test with Set Tags and two nested UK without tag
 Recursive continue in test with tag and two nested UK with and without tag
     Check Test Case    ${TESTNAME}
 
+Recursive continue in test with tag and UK with stop-on-failure tag
+    Check Test Case    ${TESTNAME}
+
 Recursive continue in user keyword
     Check Test Case    ${TESTNAME}
 
 Recursive continue in nested keyword
+    Check Test Case    ${TESTNAME}
+
+stop-on-failure in keyword in Teardown
+    Check Test Case    ${TESTNAME}
+
+stop-on-failure in test case in Teardown
+    Check Test Case    ${TESTNAME}
+
+stop-on-failure in User Keyword
+    Check Test Case    ${TESTNAME}
+
+stop-on-failure with Template
     Check Test Case    ${TESTNAME}
--- a/atest/testdata/running/continue_on_failure_tag.robot
+++ b/atest/testdata/running/continue_on_failure_tag.robot
@@ -238,6 +238,17 @@ stop-on-failure with Template
     42           43
     Something    Different
 
+stop-on-failure with Template and Teardown
+    [Documentation]    FAIL    42 != 43\n\nAlso teardown failed:\n${HEADER}\n\n
+    ...    1) 1\n\n
+    ...    2) 2
+    [Tags]   robot:stop-on-failure
+    [Teardown]   Run Keywords   Fail   1   AND   Fail  2
+    [Template]    Should Be Equal
+    Same         Same
+    42           43
+    Something    Different
+
 *** Keywords ***
 Failure in user keyword with tag
     [Arguments]    ${run_kw}=No Operation
--- a/atest/testdata/running/continue_on_failure_tag.robot
+++ b/atest/testdata/running/continue_on_failure_tag.robot
@@ -188,6 +188,14 @@ Recursive continue in test with tag and two nested UK with and without tag
     Failure in user keyword with tag     run_kw=Failure in user keyword without tag
     Fail   This should be executed
 
+Recursive continue in test with tag and UK with stop-on-failure tag
+    [Documentation]    FAIL ${HEADER}\n\n
+    ...    1) kw1a\n\n
+    ...    2) This should be executed
+    [Tags]   robot:recursive-continue-on-failure
+    Failure in user keyword with stop-on-failure tag
+    Fail   This should be executed
+
 Recursive continue in user keyword
     [Documentation]    FAIL ${HEADER}\n\n
     ...    1) kw1a\n\n
@@ -203,6 +211,32 @@ Recursive continue in nested keyword
     ...    2) kw1b
     Failure in user keyword without tag     run_kw=Failure in user keyword with recursive tag
     Fail   This should not be executed
+
+stop-on-failure in keyword in Teardown
+    [Documentation]    FAIL    Teardown failed:\nkw1a
+    [Teardown]   Failure in user keyword with stop-on-failure tag
+    No Operation
+
+stop-on-failure in test case in Teardown
+    [Documentation]    FAIL    Teardown failed:\n${HEADER}\n\n
+    ...    1) 1\n\n
+    ...    2) 2
+    [Tags]   robot:stop-on-failure
+    [Teardown]   Run Keywords   Fail   1   AND   Fail  2
+    No Operation
+
+stop-on-failure in User Keyword
+    [Documentation]    FAIL    Teardown failed:\nKeyword teardown failed:\nkw1a
+    [Teardown]   User Keyword Teardown
+    No Operation
+
+stop-on-failure with Template
+    [Documentation]    FAIL    42 != 43
+    [Tags]   robot:stop-on-failure
+    [Template]    Should Be Equal
+    Same         Same
+    42           43
+    Something    Different
 
 *** Keywords ***
 Failure in user keyword with tag
@@ -226,6 +260,17 @@ Failure in user keyword with recursive tag
     Fail   kw1b
     Log    This should be executed
     Run Keyword   ${run_kw}
+
+Failure in user keyword with stop-on-failure tag
+    [Tags]   robot:stop-on-failure
+    Fail   kw1a
+    Log    This should not be executed
+    Fail   kw1b
+
+User Keyword Teardown
+    [Tags]   robot:stop-on-failure
+    [Teardown]   Run Keywords    Fail  kw1a   AND   Fail   kw1b
+    No Operation
 
 FOR loop in in user keyword with tag
     [Tags]   robot:continue-on-failure


src/robot/running/context.py
```diff
@@ -126,14 +126,22 @@ def in_teardown(self):
...
         return self.namespace.variables
 
-    @property
-    def continue_on_failure(self):
-        parents = ([self.test] if self.test else []) + self.user_keywords
+    def continue_on_failure(self, default=False):
+        if self.in_teardown:
+            parents = self.user_keywords
+        else:
+            parents = ([self.test] if self.test else []) + self.user_keywords
```
pekkaklarck, line 133: Is the reason to handle teardowns differently that then parent test possibly having `robot:(recursive-)stop-on-failure` doesn't affect teardowns? I guess that makes sense, but then again we could consider making it possible to alter the default teardown behavior with that tag a feature. Because stopping is the default behavior (outside teardowns), I see no reason to enable the stop-on-failure on test level otherwise.

src/robot/running/context.py
```diff
@@ -126,14 +126,22 @@ def in_teardown(self):
...
+            return default
+        if 'robot:stop-on-failure' in parents[-1].tags:
             return False
         if 'robot:continue-on-failure' in parents[-1].tags:
             return True
-        return any('robot:recursive-continue-on-failure' in p.tags for p in parents)
+        if any('robot:recursive-stop-on-failure' in p.tags for p in parents):
+            return False
+        if any('robot:recursive-continue-on-failure' in p.tags for p in parents):
+            return True
```
pekkaklarck, line 143: Doesn't this work wrong if a keyword enables recursive stop-on-failure mode and then a keyword it uses enables recursive continue-on-failure mode? Shouldn't the latter, i.e. the closest parent, then have precedence? Changing the algorithm as follows ought to fix that:
```python
for parent in reversed(parents):
    if 'robot:recursive-continue-on-failure' in parent.tags:
        return True
    if 'robot:recursive-stop-on-failure' in parent.tags:
        return False
```
The above algorithm would also have a benefit that parents weren't traversed twice so it ought to be a bit faster.

Looking at tags twice with each parent is still likely to down handling failures. There were some reports related to performance degradation when we enabled controlling continue-on-failure with tags. That mainly shows if you have a lot of failures (e.g. a loop or Wait Until Keyword Succeeds) and that ought to be pretty rare. Should do some performance testing after this PR is merged and see could the situation be enhanced. Possibly looking at tags could be made faster in cases like this where the value isn't a pattern.

pekkaklarck: One thing to decide is how to handle situation where a test or a keyword has both more than one of the tags controlling the continue-on-failure mode. It should be an error, but us detecting it would require adding additional checks and I'm not sure are they worth it. Just leaving the situation as it is could be fine. I wouldn't call it a backwards incompatible change if we made it stricter later. Linters could also easily check that.

src/robot/errors.py
```diff
@@ -139,14 +139,14 @@ def can_continue(self, context, templated=False):
...
         if self.keyword_timeout:
             if context.in_teardown:
                 self.keyword_timeout = False
             return False
-        if context.in_teardown or context.continue_on_failure:
-            return True
-        return self.continue_on_failure
+        if context.in_teardown:
+            return context.continue_on_failure(default=True)
+        return self.continue_on_failure or context.continue_on_failure(default=False)
```
pekkaklarck, line 149: Adding `default=False` argument to `continue_on_failure` looks like a good idea.

I got an idea that the three last lines could be simplified to this:
```python
return self.continue_on_failure or context.continue_on_failure(default=context.in_teardown)
```
Passing `context.in_teardown` to `context.continue_on_failure` is a bit odd, though, as `context` obviously has that information already. Perhaps we should just use
```python
return self.continue_on_failure or context.continue_on_failure()
```
here and then at the end of the `continue_on_failure` method change
```python
return default
```
to
```python
return default or self.in_teardown
```

pekkaklarck (changes requested):
Looks good but there seems to be a bug in handling nested usage of `robot:recursive-continue-on-failure` and `robot:recursive-stop-on-failure`. Handling situation where these tags are set in test level when failure occurs in teardown could possibly be changed as well.

I didn't look at the tests yet. Changes above require changes to them anyway.

This feature will also need documentation.

src/robot/running/context.py
oboehmer: Here I followed your [guidance](https://github.com/robotframework/robotframework/issues/4303#issuecomment-1096889806):
> - Neither stop-on-failure variants should propagate to teardowns. We can think them having robot:recursive-continue-on-failure implicitly. To enable the stop-on-failure mode in teardowns, these special tags need to be used explicitly.
but we can reconsider?

src/robot/running/context.py
oboehmer: Hmm..  I figured the feature was designed to STOP continue-on-failure mode, so I felt that `recursive-stop-on-failure` in **ANY** parent would propagate to the bottom of the stack, no matter what. The only way to overwrite it would be a specific `robot:continue-on-failure`, but not a recursive-continue-on-failure..  
This explicit preference (stop overriding continue) would also handle the case of both tags being present in a test case (stop is checked first and hence takes precedence).. 
Maybe we should discuss on slack?

src/robot/errors.py
oboehmer: Somehow I feel it is cleaner to leave the in_teardown check in `can_continue` so the `continue_on_failure` is purely about checking the tags and no other side-effects..  
I think `context.continue_on_failure(default=context.in_teardown)` looks odd only at first glance.. changing it to `context.continue_on_failure(default=context.in_teardown is True)` would be more explicit, but I reckon this would void the optimization (not sure how big such an optimization would be)?

src/robot/running/context.py
oboehmer: I now implemented your suggested algorithm (`for parent in reversed(parents)`, etc.), however first checking stop, then continue.. adding test cases for this one now..

src/robot/errors.py
oboehmer: there is actually a behavioral difference: In the current case, context.in_teardown overrides continuable keywords, so in the current implementation the user-keyword stops, ignoring continuable keyword. 

```
stop-on-failure with run-kw-and-continue failure in keyword in Teardown
    [Documentation]     FAIL    Teardown failed:\nkw10a
    [Teardown]   run-kw-and-continue failure in user keyword with stop tag
    No Operation

*** Keywords ***
run-kw-and-continue failure in user keyword with stop tag
    [Tags]   robot:stop-on-failure
    Run Keyword And Continue On Failure    Fail    kw10a
    Log    This is not executed
    Fail    kw10b
```

Combining the two statements into one line 
```
return self.continue_on_failure or context.continue_on_failure(default=context.in_teardown)
```
would continue the execution in the keyword.

src/robot/running/context.py
```diff
@@ -126,14 +126,20 @@ def in_teardown(self):
...
+        if 'robot:stop-on-failure' in parents[-1].tags:
             return False
         if 'robot:continue-on-failure' in parents[-1].tags:
             return True
-        return any('robot:recursive-continue-on-failure' in p.tags for p in parents)
+        for parent in reversed(parents):
+            if 'robot:recursive-stop-on-failure' in parent.tags:
+                return False
+            if 'robot:recursive-continue-on-failure' in parent.tags:
+                return True
```
pekkaklarck, line 141: It would be possible to combine checking modes into single place and remove the `if not parents` block at the same time:

```python
first = True
for parent in reversed(parents):
    if ('robot:recursive-continue-on-failure' in parent.tags
            or first and 'robot:continue-on-failure' in parent.tags):
        return True
    if ('robot:recursive-stop-on-failure' in parent.tags
            or first and 'robot:stop-on-failure' in parent.tags):
        return False
    first = False
```

Could also use `enumerate` and check the index instead of using the `first` flag. Adding separate helper methods for checking tags that could be considered as well:

```python
for index, parent in enumerate(reversed(parents)):
    if self._continue_by_tags(parent.tags, index == 0):
        return True
    if self._stop_by_tags(parent.tags, index == 0):
        return False
```

These are just ideas, the current code is good as it is and can be fine-tuned later if really needed.

src/robot/errors.py
oboehmer: Ok, understood.. not sure how common continuable keywords are, but if clearly documented people should be able to handle this.
changes implemented through 4f3fe30ab

src/robot/running/context.py
oboehmer: I like the optimization suggestion, took the enumerate one without helper..

src/robot/running/context.py
```diff
@@ -126,14 +126,20 @@ def in_teardown(self):
     def variables(self):
         return self.namespace.variables
 
-    @property
-    def continue_on_failure(self):
+    def continue_on_failure(self, default_continue=False):
```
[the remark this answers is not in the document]
oboehmer: I reconsidered and changed back to `default`..

doc/userguide/src/ExecutingTestCases/TestExecution.rst
```diff
@@ -455,8 +455,88 @@ can be used. The below examples executes all the keywords listed.
...
+In the example below, the 2nd teardown2 keyword will not be executed in case
+the first teardown1 keyword fails.
+
+.. sourcecode:: robotframework
+
+   *** Test Cases ***
+   Test
+       [Teardown]    Run Keywords    teardown1    AND    teardown2
+       [Tags]    robot:stop-on-failure
+       No Operation
```
pekkaklarck, line 495: Is this example correct? Doesn't `robot:stop-on-failure` only work on the level where it has been set? It would probably be better anyway to have an example with a separate keyword used as a teardown like:
```
*** Test Cases ***
Example
    No Operation
    [Teardown]    Teardown Keyword

*** Keywords ***
Teardown Keyword
    [Tags]    robot:stop-on-failure
    Keyword 1
    Keyword 2
```

Notice also that our style is to have `[Teardown]` at the end of a test/keyword.

doc/userguide/src/ExecutingTestCases/TestExecution.rst
```diff
@@ -455,8 +455,88 @@ can be used. The below examples executes all the keywords listed.
...
+       Log   this is also executed
+
+   User Keyword 2
+       [Tags]    robot:stop-on-failure
+       Should be Equal   5   6
+       Log   this is not executed
+
+The `robot:stop-on-failure` and its recursive variant
+`robot:recursive-stop-on-failure` can also be used to alter
+the default continue behavior of Teardowns and Templates:
```
pekkaklarck, line 484: Assuming suitable sections exist, it would be good to add links to sections about teardowns and teamplates.

pekkaklarck, line 484: Assuming suitable sections exist, it would be good to add links to sections about teardowns and templates.

doc/userguide/src/ExecutingTestCases/TestExecution.rst
pekkaklarck, line 495: Is this example correct? Doesn't `robot:stop-on-failure` only work on the level where it has been set? It would probably be better anyway to have an example with a separate keyword used as a teardown like:
```
*** Test Cases ***
Example
    No Operation
    [Teardown]    Teardown Keyword

*** Keywords ***
Teardown Keyword
    [Tags]    robot:stop-on-failure
    Keyword 1
    Keyword 2
```

Notice also that our style is to have `[Teardown]` at the end of a test/keyword.

doc/userguide/src/ExecutingTestCases/TestExecution.rst
```diff
@@ -455,8 +455,88 @@ can be used. The below examples executes all the keywords listed.
...
+   Test
+       [Tags]    robot:stop-on-failure
+       Run Keyword and Continue on Failure    Fail    failure
+       Log    this is executed
+
+
 The `robot:continue-on-failure` and `robot:recursive-continue-on-failure`
 tags are new in Robot Framework 4.1.
+The `robot:stop-on-failure` and `robot:recursive-stop-on-failure`
+tags are new in Robot Framework 5.1.
```
pekkaklarck, line 539: This new note, and the old note above, should be turned into "real" notes by using the `.. note::` syntax. There ought to be examples in how to use it in the same file.

Should also mention that continue-on-failure doesn't work with WHILE loops prior to RF 5.1. There could be a separate note here or earlier where using these with FOR and WHILE loops was discussed, or this could be mentioned in the note about adding continue-on-failure tags in RF 4.1.

doc/userguide/src/ExecutingTestCases/TestExecution.rst
```diff
@@ -466,6 +546,9 @@ continue on failure mode is automatically on in `test and suite
 teardowns`__. In practice this means that in teardowns all the
 keywords in all levels are always executed.
 
+The reserved tags  `robot:stop-on-failure` and `robot:recursive-stop-on-failure`
+can be used to alter this behavior.
```
pekkaklarck, line 550: Link to the section discussing this would be nice.

doc/userguide/src/ExecutingTestCases/TestExecution.rst
```diff
@@ -476,6 +559,9 @@ make it sure that all the different combinations are tested. In this
 usage continuing is limited to the top-level keywords, and inside them
 the execution ends normally if there are non-continuable failures.
 
+The reserved tags  `robot:stop-on-failure` and `robot:recursive-stop-on-failure`
+can be used to alter this behavior.
```
pekkaklarck, line 563: Link to the section discussing this would be nice.

pekkaklarck (approved):

pekkaklarck:
Latest changes look great! My (hopefully) final comments/questions:

1. Are there tests for the nested continue/stop usage? For example, setting recursive-continue-on-failure on test level and then later overriding it with (recursive-)stop-on-failure? There are many tests but I didn't notice one covering this.
2. Related the above, test names could possibly be enhanced a bit to make it easier to see what's actually tested by looking at tests on `atest/robot` side. Perhaps the somewhat large suite could also be split into smaller pieces. I know it's hard to keep tests easy to understand when there are so many of them!
3. I added some comments about the documentation. Hopefully you can take a look at them.

doc/userguide/src/ExecutingTestCases/TestExecution.rst
oboehmer: this example is correct. stop-on-failure is visible to the `Run Keywords`, so in below `Fail 2` would not be executed. 

```
    Test
       [Tags]    robot:stop-on-failure
       No Operation
       [Teardown]    Run Keywords    Fail   1    AND    Fail   2
```
but agree that a teardown keyword is a better example

doc/userguide/src/ExecutingTestCases/TestExecution.rst
oboehmer: I added also the notes about feature details into the `.. note::` box.. please review.

doc/userguide/src/ExecutingTestCases/TestExecution.rst
oboehmer: done

doc/userguide/src/ExecutingTestCases/TestExecution.rst
oboehmer: done.. it reads a bit strange, though.. not sure if rst allows something like [text](link)

pekkaklarck:
Huge thanks for a great PR @oboehmer!

PR #4327, opened 2022-04-29
Include source file path on xunit report output

luiz1361:
This includes the source file path to the xunit report output.  This is equivalent to the foramat of RspecJunitFormatter.

### DIFF
--- a/src/robot/reporting/xunitwriter.py
+++ b/src/robot/reporting/xunitwriter.py
@@ -70,6 +70,7 @@ class XUnitFileWriter(ResultVisitor):
         self._writer.start('testcase',
                            {'classname': test.parent.longname,
                             'name': test.name,
+                            'file': test.source,
                             'time': self._time_as_seconds(test.elapsedtime)})
         if test.failed:
             self._writer.element('failure', attrs={'message': test.message,


luiz1361:
Hi @pekkaklarck, thanks for the very quick response. I have created a new issue and detailed the use case and the reference from another working test framework RSPEC. 
Also detailed the reason for the feature and the reason to be on test cases instead of suites.
I'm already working based on this change sucessfully on my environment, but it would be great if it was on mainstream.

rikerfi:
Hi Pekka and Luiz.

we have talked about this topic a few times, but always we get back to the sad fact there is no official standard xunit output. The best reference I have found is JUnit team's `xsd`-file for testing purposes. Here: [jenkins-junit.xsd](https://github.com/junit-team/junit5/blob/main/platform-tests/src/test/resources/jenkins-junit.xsd) Better references are very much appreciated.

During RF5.0 development we planned to add `file`and `lineno` information, but that was dropped as we thought it will introduce RF specific xunit output dialect.

Also, it was discussed that possible new dialects like this PR, should use `prerebotmodifier` to make specific xunit output.

PR #4337, opened 2022-05-09
setup.py and src/robot/version.py: change version to 5.0

Pokisiekk:
Signed-off-by: Kamil Pokornicki <kamil.pokornicki@3mdeb.com>


pekkaklarck:
Versions are set to 5.0 in the v5.0 tag.

Issue #4335, opened 2022-05-06
Pass more information about control structures to `start/end_keyword` listener methods


osrjv:
### Why

In listener API v2, control structures such as IF or FOR use the same `start_keyword` and `end_keyword` methods as regular keywords. Additional fields specific for those statements are lost, such as the conditional statements for an IF branch, or the assigned variables for a FOR loop.

Parsing these is possible, to some degree, from the string representation marked as the name for these "keywords". However having them available directly in the attributes would help many use-cases.

### What

I propose that additional attributes are available based on the `type` attribute, which should be backwards compatible to previous versions.

The attributes would be as follows (same as in the models):

**FOR**: `variables`, `flavor`, `values`
**WHILE**: `condition`, `limit`
**IF / ELSE_IF**: `condition`
**EXCEPT**: `patterns`, `pattern_type`, `variable`
**RETURN**: `values`

### How

The attributes/arguments for listeners are defined [here](https://github.com/robotframework/robotframework/blob/master/src/robot/output/listenerarguments.py#L128). A look-up for extra attributes could be done based on the `repr_args` field of the control structure model, for instance. Alternatively a separate map of type to extra attributes could be maintained.

pekkaklarck:
Sounds good to me. Are you @osrjv interested to create PR?

osrjv:
I think I can make a draft PR with a suggestion and we can continue the discussion there.

pekkaklarck:
Are you @osrjv still working with the PR?

osrjv:
> Are you @osrjv still working with the PR?

Sorry, I was busy with vacation season and other priorities after your response to the PR. I'll try to pick this up again now and get it finished.

pekkaklarck:
Thanks for a great PR @osrjv!

PR #4341, opened 2022-05-10
Expose attributes from control structures to listeners

osrjv:
A draft PR for issue #4335 to start the conversation. As mentioned in the original issue, the idea is to add attributes from control structures to the `start_keyword`/`end_keyword` listener methods which change based on the `type` attribute value.

My main question before continuing is as follows: is it alright that we maintain a separate map of type to extra attributes? The information is a duplicate of the `repr_args` field of the corresponding models, but I'm not sure if we want to overload the semantics of that field. It would also be too easy to accidentally break the listener API contract if that link is not clear.

Another small detail which makes this particular implementation worse is that the `ForIteration` and `WhileIteration` models have the same type value but different attributes.


pekkaklarck:
Sorry for a late review. The code looks fine and having a separate mapping for extra attributes looks good as well. I thing differences between `WhileIteration` and `ForIteration` can be handled simply with `hasattr`:

```python
attrs = {...}
if kw.type in self._type_attributes:
    attrs.update({name: self._get_attribute_value(kw, name)
                  for name in self._type_attributes[kw.type] if hasattr(kw, name)})
return attrs
```

I added `if` to the example above to avoid the dict compression in the common case when dealing with normal keywords as a performance optimization. Even an empty compression takes some time and these listener methods are called _very_ often.

pekkaklarck (approved):
Looks awesome! The only issue I see is that there doesn't seem to a note that these extra attributes are new in RF 5.1. I'll add it myself either directly to this PR or after merging it.

atest/testresources/listeners/VerifyAttributes.py
```diff
@@ -0,0 +1,138 @@
+import os
+
+try:
+    basestring
+    long
+except NameError:
+    basestring = str
+    long = int
```
pekkaklarck, line 8: Python 2 compatibility could be removed altogether. I'll do it myself.

PR #4342, opened 2022-05-10
Use proper main_thread API from Python 3.4: https://docs.python.org/3…

pianofab:
….4/library/threading.html#threading.main_thread


pekkaklarck:
Thanks! Definitely better way to check this.

PR #4344, opened 2022-05-18
Possibility to change keyword status (to failed) with listeners v3 'end_keyword' 

AVaitkunas:
Added possibility to fail keyword execution with listener v3 method 'end_keyword'. Issue #3296

Use case: 
For one of our current projects, there is a need to monitor certain conditions after each keyword and stop test execution if these conditions aren't satisfied.


pekkaklarck:
Closing this because `start/end_keyword` listener v3 methods are now finally implemented.

PR #4345, opened 2022-05-18
Keyword visibility modifiers for resource files (#430)

### LINKED ISSUE #430: Keyword visibility modifiers for resource files
> _Originally submitted to [Google Code](http://code.google.com/p/robotframework/issues/detail?id=430) by tookko on 19 Nov 2009_

Complicated user keywords in resource files are typically built by 
breaking the desired functionality into smaller, lower level user keywords 
which are then used by the higher level user keyword. For maintainability 
purposes it's often desirable that the lower level keywords are not used 
directly from test cases, but only the higher level ones.

While this can be tackled softly by usage of documentation tags, it would 
be useful to have a [Private] modifier for keywords, which would make it 
not possible to use the keyword from outside the same resource file.

robinmatz:
Resolves #430.
- User keywords can now be marked as private by tagging them `robot:private`.
- Using them in resources within the same file is fine and intended
- Using them in a test case in the same file or in a user keyword in another file will cause a warning
- Keywords tagged as `robot:private` are removed from libdoc's html output
- They are, however, still included in libdoc's spec files, for other tools to pick up on them
- If there is both a public and a private User Keyword with the same name in the current scope, Robot Framework will execute the public one

### DIFF
--- a/src/robot/running/namespace.py
+++ b/src/robot/running/namespace.py
@@ -320,8 +320,9 @@ class KeywordStore:
         if not found:
             return None
         if len(found) > 1:
-            found = self._remove_private_keywords(found)
             found = self._get_runner_based_on_search_order(found)
+        if len(found) == 2:
+            found = self._filter_private_runner(*found)
         if len(found) == 1:
             return found[0]
         self._raise_multiple_keywords_found(name, found)
@@ -339,11 +340,28 @@ class KeywordStore:
             return found[0]
         self._raise_multiple_keywords_found(name, found)
 
-    def _remove_private_keywords(self, runners):
-        for runner in runners:
-            if 'robot:private' in runner.tags:
-                runners.remove(runner)
-        return runners
+    def _filter_private_runner(self, runner1, runner2):
+        if all(['robot:private' in runner.tags for runner in [runner1, runner2]]):
+            return [runner1, runner2]
+        if 'robot:private' in runner1.tags:
+            private, public = runner1, runner2
+        elif 'robot:private' in runner2.tags:
+            private, public = runner2, runner1
+        self._public_and_private_keyword_conflict_warning(public, private)
+        return [public]
+
+    def _public_and_private_keyword_conflict_warning(self, public, private):
+        warning = Message(
+            f"There were both public and private keyword found with the name '{public.name}', "
+            f"'{public.longname}' being public and '{private.longname}' being private. "
+            f"The public keyword is used. To select explicitly, and to get rid of this warning, "
+            f"use either '{public.longname}' or '{private.longname}'.",
+            level='WARN'
+        )
+        if public.pre_run_messages:
+            public.pre_run_messages.append(warning)
+        else:
+            public.pre_run_messages = [warning]
 
     def _get_runner_based_on_search_order(self, runners):
         for libname in self.search_order:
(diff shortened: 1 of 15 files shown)


robinmatz:
Hi @pekkaklarck , just a quick reminder that this PR is ready for code review

src/robot/running/namespace.py
```diff
@@ -338,6 +339,42 @@ def _get_runner_from_libraries(self, name):
             return found[0]
         self._raise_multiple_keywords_found(name, found)
 
+    def _handle_private_runners(self, runners):
+        sum_public = sum("robot:private" not in runner.tags for runner in runners)
+        sum_private = sum("robot:private" in runner.tags for runner in runners)
```
pekkaklarck, line 344: It's a bit questionable to use `sum()` with Boolean values but I guess it's ok here. The latter line could be simplified to:
```python
sum_private = len(runners) - sum_public
```

src/robot/running/namespace.py
```diff
@@ -338,6 +339,42 @@ def _get_runner_from_libraries(self, name):
...
+
+        if sum_public == len(runners) or sum_private == len(runners) or sum_public > 1:
```
pekkaklarck, line 346: The first condition doesn't seem to be necessary. We only get to this function if there were multiple matches, so just having `sum_public > 1` ought to be enough to cover also the `sum_public == len(runners)` case.

We generally haven't used empty lines between functions. If there's a need to use them for grouping code, we have typically created helper functions instead. That could be considered hare as well or possibly the empty lines could just be removed.

src/robot/running/namespace.py
```diff
@@ -338,6 +339,42 @@ def _get_runner_from_libraries(self, name):
...
+            return runners
+
+        for runner in runners:
+            if 'robot:private' not in runner.tags:
+                public_runner = runner
+                private_runners = runners.copy()
+                private_runners.remove(public_runner)
+                break
```
pekkaklarck, line 354: Probably we could initially filter runners so that we have list of public and private runners instead of `sum_xxx` containing just their counts. Then this loop wouldn't needed at all.

src/robot/running/namespace.py
```diff
@@ -338,6 +339,42 @@ def _get_runner_from_libraries(self, name):
...
+        self._public_and_private_keyword_conflict_warning(public_runner, private_runners)
+        return [public_runner]
+
+    def _public_and_private_keyword_conflict_warning(self, public_runner, private_runners):
+        private_runners_str = ""
+        for runner in private_runners:
+            if runner != private_runners[-1]:
+                private_runners_str += f"'{runner.longname}' / "
+            else:
+                private_runners_str += f"'{runner.longname}'"
```
pekkaklarck, line 365: This looks a bit complicated way to concatenate names with ` /  `. Just using something like
```python
private_runner_str = ' / '.join(r.longname for r in private_runners)
```
ought to handler it as a one-liner. I'm also not certain is using ` / ` as a separator between private keywords here that good in general. Just listing _all_ runners like `'First', 'Second' or 'Third` would be simpler to implement and would look, in my opinion, clearer. I can test that in practice myself.

pekkaklarck (approved):
Looks great! Implementation is good, there are comprehensive tests, and also documentation has been updated. I had some somewhat nitpickey comments about the code style but I can change the code myself after merging this.

PR #4350, opened 2022-05-30
Make parsing 17% faster.

fabioz:
I've spent a couple of hours optimizing some low-hanging fruits in the parsing (while for RF this may not be a huge gain, the language servers which use the parsers can gain quite a bit since parsing is a bottleneck on big documents).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ __pycache__
 .classpath
 .settings
 .jython_cache
+/.mypy_cache/
--- a/src/robot/parsing/lexer/blocklexers.py
+++ b/src/robot/parsing/lexer/blocklexers.py
@@ -52,14 +52,15 @@ class BlockLexer(Lexer):
         return lexer
 
     def lexer_for(self, statement):
-        for cls in self.lexer_classes():
-            lexer = cls(self.ctx)
-            if lexer.handles(statement):
+        for cls in self.RESOLVED_LEXER_CLASSES:
+            if cls.handles(self.ctx, statement):
+                lexer = cls(self.ctx)
                 return lexer
         raise TypeError("%s did not find lexer for statement %s."
                         % (type(self).__name__, statement))
 
-    def lexer_classes(self):
+    @classmethod
+    def lexer_classes(cls):
         return ()
 
     def lex(self):
@@ -80,7 +81,8 @@ class FileLexer(BlockLexer):
     def lex(self):
         self._lex_with_priority(priority=SettingSectionLexer)
 
-    def lexer_classes(self):
+    @classmethod
+    def lexer_classes(cls):
         return (SettingSectionLexer, VariableSectionLexer,
                 TestCaseSectionLexer, KeywordSectionLexer,
                 CommentSectionLexer, ErrorSectionLexer,
@@ -95,64 +97,78 @@ class SectionLexer(BlockLexer):
 
 class SettingSectionLexer(SectionLexer):
 
-    def handles(self, statement):
-        return self.ctx.setting_section(statement)
+    @classmethod
+    def handles(cls, ctx, statement):
+        return ctx.setting_section(statement)
 
-    def lexer_classes(self):
+    @classmethod
+    def lexer_classes(cls):
         return (SettingSectionHeaderLexer, SettingLexer)
 
 
 class VariableSectionLexer(SectionLexer):
 
-    def handles(self, statement):
-        return self.ctx.variable_section(statement)
+    @classmethod
+    def handles(cls, ctx, statement):
+        return ctx.variable_section(statement)
 
-    def lexer_classes(self):
+    @classmethod
+    def lexer_classes(cls):
         return (VariableSectionHeaderLexer, VariableLexer)
 
 
 class TestCaseSectionLexer(SectionLexer):
 
-    def handles(self, statement):
-        return self.ctx.test_case_section(statement)
+    @classmethod
+    def handles(cls, ctx, statement):
+        return ctx.test_case_section(statement)
 
-    def lexer_classes(self):
+    @classmethod
+    def lexer_classes(cls):
         return (TestCaseSectionHeaderLexer, TestCaseLexer)
 
 
 class KeywordSectionLexer(SettingSectionLexer):
 
-    def handles(self, statement):
-        return self.ctx.keyword_section(statement)
+    @classmethod
+    def handles(cls, ctx, statement):
+        return ctx.keyword_section(statement)
 
-    def lexer_classes(self):
+    @classmethod
+    def lexer_classes(cls):
         return (KeywordSectionHeaderLexer, KeywordLexer)
 
 
 class CommentSectionLexer(SectionLexer):
 
-    def handles(self, statement):
-        return self.ctx.comment_section(statement)
+    @classmethod
+    def handles(cls, ctx, statement):
+        return ctx.comment_section(statement)
 
-    def lexer_classes(self):
+    @classmethod
+    def lexer_classes(cls):
         return (CommentSectionHeaderLexer, CommentLexer)
 
 
 class ImplicitCommentSectionLexer(SectionLexer):
 
-    def handles(self, statement):
+    @classmethod
+    def handles(cls, ctx, statement):
         return True
 
-    def lexer_classes(self):
+    @classmethod
+    def lexer_classes(cls):
         return (CommentLexer,)
 
 
 class ErrorSectionLexer(SectionLexer):
 
-    def handles(self, statement):
+    @classmethod
+    def handles(cls, ctx, statement):
         return statement and statement[0].value.startswith('*')
 
-    def lexer_classes(self):
+    @classmethod
+    def lexer_classes(cls):
         return (ErrorSectionHeaderLexer, CommentLexer)
 
 
@@ -177,9 +193,10 @@ class TestOrKeywordLexer(BlockLexer):
             self._name_seen = True
         else:
             while statement and not statement[0].value:
-                statement.pop(0).type = None    # These tokens will be ignored
+                statement.pop(0).type = None  # These tokens will be ignored
 
-    def lexer_classes(self):
+    @classmethod
+    def lexer_classes(cls):
         return (TestOrKeywordSettingLexer, BreakLexer, ContinueLexer,
                 ForLexer, InlineIfLexer, IfLexer, ReturnLexer, TryLexer,
                 WhileLexer, KeywordCallLexer)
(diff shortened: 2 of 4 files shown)


src/robot/parsing/lexer/blocklexers.py
```diff
@@ -52,14 +52,15 @@ def input(self, statement):
         return lexer
 
     def lexer_for(self, statement):
-        for cls in self.lexer_classes():
-            lexer = cls(self.ctx)
-            if lexer.handles(statement):
+        for cls in self.RESOLVED_LEXER_CLASSES:
```
fabioz, line 55: This was the major improvement: not needing to create instances to use `handles` and not creating temporary tuples from the call to `lexer_classes()`.

This meant that the `lexer_classes` and `handles` had to be changed to `classmethods`.

src/robot/parsing/lexer/tokenizer.py
```diff
@@ -72,49 +72,52 @@ def _split_from_pipes(self, line):
...
-            tokens = self._remove_non_data(tokens)
-        return tokens, has_data and not continues
-
-    def _handle_comments(self, tokens):
+        # Handle comments
         has_data = False
-        commented = False
-        for token in tokens:
+
+        iter_in_tokens = iter(tokens)
```
fabioz, line 78: 2nd improvement was not iterating so many times though the tokens in `_cleanup_tokens`.

pekkaklarck:
Very interesting! I haven't yet tested changes locally, but I believe your measurements. Tests have also succeeded on CI so changes shouldn't have broken anything either.

Some general comments about the changes:

1. Since there are two separate enhancements, two separate PRs would've been a bit easier to review. Changes are relatively simple, considering that this is optimization, so we ought to manage this way as well.

2. Resolving lexer classes just once is a clever idea. It would be great if there was a cleaner way to accomplish that than a module level function call, though. Especially the fact that the function in `blocklexers` apparently modifies also classes in `statementlexers` is kind of ugly. I was thinking we could possibly use a metaclass instead, but that probably won't work because all lexer classes may not be available yet at the class creation time. Another alternative would be resolving classes when `lexer_classes` is called the first time:
   ```python
   def lexer_classes(self):
       if self._resolved_lexer_classes is None:
           self._resolved_lexer_classes = self._lexer_classes()
       return self._resolved_lexer_classes
   ```
   If the check for `self._resolved_lexer_classes` is considered expensive, perhaps we could even rewrite the whole method:
   ```python
   def lexer_classes(self):
       classes = self._lexer_classes()
       self.lexer_classes = lambda: classes
       return classes
   ```
   I think the above solutions would have same performance as the current proposal but that would need to be tested. I agree performance is important here so even somewhat ugly but fast solution is fine. If we'd stick with the current proposal, the `RESOLVED_LEXER_CLASSES` should be defined already in the base `Lexer` class with a comment explaining where it is populated.

3. Avoiding unnecessary iteration of tokens in `_cleaup_tokens` looks good too. I especially like how you handle comments. That said, the resulting method is now pretty long and there seems to be parts that aren't related to each others. Would splitting it to helper methods slow down parsing speed too much? I'd prefer that over comments explaining that the code is doing in general because it eases maintenance.


Let me know what you think about my proposals related to points 2. and 3. above. We could also do it so that I merge this PR and then do those changes myself.

fabioz:
> Since there are two separate enhancements, two separate PRs would've been a bit easier to review. Changes are relatively simple, considering that this is optimization, so we ought to manage this way as well.

I'll keep that in mind for future PRs.

> Especially the fact that the function in `blocklexers` apparently modifies also classes in `statementlexers` is kind of ugly.

I can't really see that happening (i.e.: `def lexer_classes` is only found in `blocklexers.py` -- `statementlexers.py` was changed only to convert the `handles` to a classmethod).

Another thing that happened is that the method lookup became an attribute access which is faster (so, if the approach is changed to still have a method it'll probably regress a bit from the performance in the patch)...

> Would splitting it to helper methods slow down parsing speed too much? I'd prefer that over comments explaining that the code is doing in general because it eases maintenance.

Regarding `_cleanup_tokens`, I usually tend to inline methods if the performance is really critical in some session (and I think it is in this case), but if you'd rather regress a bit in the performance to keep the code a bit cleaner in this case, that's ok for me (also, splitting usually prevents some optimizations as it makes it harder to see the full code, so, in this case I started expanding all the code to then started looking for things which could be enhanced -- I think there are additional opportunities there, but I didn't have more time to dig further).

The current profile I have is below (so, `_cleanup_tokens` is the 3rd there) -- keep in mind that the profile itself adds some overhead, which makes function calls even slower, so, to make final measurements the timing should be gotten without the profile overhead...

image

As a note, the code I used to measure the performance is below (if you want to get a profile picture you can get https://www.pyvmmonitor.com/download.html, start it and uncomment the code related to it -- you may have to update the `sys.path.append` if you don't install it in the default location):

```
from robot.api import get_model
import time

header = """
*** test case ***
"""

body = """
some keyword
    no operation    ${some var}
    no operation

""" * 10000

full = header + body
print(len(body.splitlines()))
count = 10

_count = 3
_full = '''
*** test case ***
*** keywords ***
some keyword   #comment  
    no operation    #  comment  

'''


def main():
    check()
    
# import sys
# sys.path.append('C:\\Program Files\\Brainwy\\PyVmMonitor 2.0.2\\public_api')
# import pyvmmonitor
# @pyvmmonitor.profile_method
def check():
    values = []
    for i in range(count):
        t = time.time()
        get_model(full)
        values.append(time.time() - t)
        print('%s of %s' % (i, count))
    
    m = min(values)
    values.remove(m)
    
    m = max(values)
    values.remove(m)
    
    avg = sum(values) / len(values)
    print('%.2fs' % avg)


if __name__ == '__main__':
    main()
```

p.s.: Feel free to change the PR in any way you'd like to get it integrated (I'm saying this because it's usually faster for a maintainer just change and merge than going back and forth, but if you'd like me to do the changes, that's ok too).

p.s.2: Another thing that crossed my mind was using some optional accelerator modules using cython (but this can add some burden in maintaining build scripts, so, I wasn't sure it was worth investigating).

p.s.3: There are definitely other low-hanging fruits there... Image showing profile by cumulative time:

image

pekkaklarck:
I tested this locally and was able to reproduce ~10% performance enhancement with the data set I used. That's very nice!

I tested changing the code so that `lexer_classes` itself resolves classes only once like this:

```diff
-    @classmethod
-    def lexer_classes(cls):
+    def lexer_classes(self):
+        classes = self._lexer_classes()
+        self.lexer_classes = lambda: classes
+        return classes
+
+    def _lexer_classes(self):
         return ()
```

That change didn't affect the performance but avoided the somewhat ugly global modifications of the classes. To my surprise the performance didn't change when I commented out the `self.lexer_classes = lambda: classes` line which basically reverts the logic to what it was initially. Could it be that resolving classes multiple times doesn't actually have meaningful performance penalty? The change to make `Lexer.handles` a classmethod could be responsible for most of the benefits. Could you @fabioz test this with your data?

fabioz:
From my tests here it's just a bit slower. The test timing with my data/machine goes from around 1.58s -> 1.61s, so, it accounts for a bit less than 2% of the savings to me -- most gains there really seem to be from not creating the class instances and not from the function call/tuple creation.

I agree that this is probably what makes the code uglier. I'll leave it up to you to decide whether you think it's worth it or not.

fabioz:
> ```diff
> -    @classmethod
> -    def lexer_classes(cls):
> +    def lexer_classes(self):
> +        classes = self._lexer_classes()
> +        self.lexer_classes = lambda: classes
> +        return classes
> +
> +    def _lexer_classes(self):
>          return ()
> ```

p.s.: have you changed all the classes so that `_lexer_classes` is overridden instead of `lexer_classes`? Because otherwise, the `lexer_classes` is the method still overridden and this did nothing.

pekkaklarck:
Good point that trying to be too clever with overriding the called function can cause problems with Python's own optimization especially when there are so many enhancements planned for that in Python 3.11+. I think it's best to just leave `lexer_classes` as it was because performance benefits are so small. Converting `handles` to a class method avoids creating instances unnecessarily and seems to have much bigger impact.

pekkaklarck:
I played with the tokenizer changes locally. Changes in this PR had smallish but measurable performance benefits that seemed to surprisingly be mostly due to method inlining. As I commented earlier, this change made the method somewhat long and complex, and I tried to find alternatives for similar performance benefits. Most of the changes caused barely measurable enhancements, but the following change had pretty much exactly the same effect as the change in this PR:
```diff
diff --git a/src/robot/parsing/lexer/tokenizer.py b/src/robot/parsing/lexer/tokenizer.py
index 1784cbec0..39311bd59 100644
--- a/src/robot/parsing/lexer/tokenizer.py
+++ b/src/robot/parsing/lexer/tokenizer.py
@@ -72,8 +72,7 @@ class Tokenizer:
         yield rest, True
 
     def _cleanup_tokens(self, tokens, data_only):
-        has_data = self._handle_comments(tokens)
-        continues = self._handle_continuation(tokens)
+        has_data, continues = self._handle_comments_and_continuation(tokens)
         self._remove_trailing_empty(tokens)
         if continues:
             self._remove_leading_empty(tokens)
@@ -82,31 +81,27 @@ class Tokenizer:
             tokens = self._remove_non_data(tokens)
         return tokens, has_data and not continues
 
-    def _handle_comments(self, tokens):
+    def _handle_comments_and_continuation(self, tokens):
         has_data = False
+        continues = False
         commented = False
         for token in tokens:
             if token.type is None:
                 # lstrip needed to strip possible leading space from first token.
                 # Other leading/trailing spaces have been consumed as separators.
                 value = token.value.lstrip()
-                if value and not commented:
+                if commented:
+                    token.type = Token.COMMENT
+                elif value:
                     if value[0] == '#':
+                        token.type = Token.COMMENT
                         commented = True
-                    else:
+                    elif not has_data:
+                        if value == '...':
+                            token.type = Token.CONTINUATION
+                            continues = True
                         has_data = True
-                if commented:
-                    token.type = Token.COMMENT
-        return has_data
-
-    def _handle_continuation(self, tokens):
-        for token in tokens:
-            if token.value == '...' and token.type is None:
-                token.type = Token.CONTINUATION
-                return True
-            elif token.value and token.type != Token.SEPARATOR:
-                return False
-        return False
+        return has_data, continues
 
     def _remove_trailing_empty(self, tokens):
         for token in reversed(tokens):
```

The above change basically combines functions for handling comments and continuations together which avoids one pass through the tokens. Also this is a bit more complicated than earlier, but looks pretty ok to me anyway. Could you @fabioz test how these changes affect performance with your tests?

.gitignore
```diff
@@ -29,3 +29,4 @@ __pycache__
 .classpath
 .settings
 .jython_cache
+/.mypy_cache/
```
pekkaklarck, line 32: What does the leading `/` mean?

fabioz: It means to ignore it only at the top-level (IDE did it for me... I guess that removing the leading `/` to ignore everywhere would be ok too).

pekkaklarck:
2% is quite little especially when shown only by a single benchmark. Could also be that Python 3.11 with various enhancements to performance changes these measurements. Anyway, I'll merge this PR as-is and then look at the code again. I'll certainly remove resolving classes and even more certainly preserve `handles` as a classmethod. I'll take another look at tokenizing and decide what to do with it. When I'm done, I hope you can try out the code and experiment with possible new changes.

One thing to test would be adding `__slots__` to lexer classes. It mainly affects memory usage and it can be that when parsing individual files that doesn't give any benefits. Could also try adding `__slots__` to the parsing model objects, but they are based on `ast.AST` that at least in Python 3.10 doesn't itself have `__slots__`. I think that was about to change but haven't tested with Python 3.11.

pekkaklarck:
I merged the PR and then did some of the changes discussed above. In my tests there was no measurable performance difference before and after my changes and the code is, arguably, a bit cleaner. Anyway, hopefully you @fabioz can test the latest changes and report if there are big differences or that if you find out other possible enhancements. Thanks also for the contribution!

PR #4363, opened 2022-06-09
Maintenance: XUnitFileWriter code streamlined

rikerfi:
Small things add up.

-Risto

### DIFF
--- a/src/robot/reporting/xunitwriter.py
+++ b/src/robot/reporting/xunitwriter.py
@@ -39,22 +39,15 @@ class XUnitFileWriter(ResultVisitor):
         self._writer = xml_writer
 
     def start_suite(self, suite):
-        tests, failures, skipped = self._get_stats(suite.statistics)
+        stats = suite.statistics  # Accessing property only once.
         attrs = {'name': suite.name,
-                 'tests': tests,
+                 'tests': f'{stats.total}',
                  'errors': '0',
-                 'failures': failures,
-                 'skipped': skipped,
+                 'failures': f'{stats.failed}',
+                 'skipped': f'{stats.skipped}',
                  'time': self._time_as_seconds(suite.elapsedtime),
                  'timestamp' : self._starttime_to_isoformat(suite.starttime)}
         self._writer.start('testsuite', attrs)
-
-    def _get_stats(self, statistics):
-        return (
-            str(statistics.total),
-            str(statistics.failed),
-            str(statistics.skipped)
-        )
 
     def end_suite(self, suite):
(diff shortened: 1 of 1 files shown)


pekkaklarck (changes requested):
Thanks, cleanup is always appreciated! This change has a performance penalty, though, so the code needs to be changed a bit.

src/robot/reporting/xunitwriter.py
```diff
@@ -70,7 +63,7 @@ def visit_test(self, test):
         self._writer.start('testcase',
                            {'classname': test.parent.longname,
                             'name': test.name,
-                            'time': self._time_as_seconds(test.elapsedtime)})
+                            'time': f'{test.elapsedtime / 1000 :.3f}'})
```
pekkaklarck, line 66: In my opinion `_time_as_seconds` was a good helper method because it made it clear what the code does. `f'{test.elapsedtime / 1000 :.3f}'` is likely not as obvious.

Now that I look at this again, I got a feeling that `f'{millis / 1000:.3f}'`is kind-of abusing f-strings. There's no template, it's only used for formatting and using `format(millis / 1000, '.3f')` would be better. Alternatively could also use `str(round(millis / 1000, 3))`.

src/robot/reporting/xunitwriter.py
```diff
@@ -39,23 +39,16 @@ def __init__(self, xml_writer):
         self._writer = xml_writer
 
     def start_suite(self, suite):
-        tests, failures, skipped = self._get_stats(suite.statistics)
+        stats = suite.statistics
```
pekkaklarck, line 42: I'm not a big fan of commenting code in general, I prefer code that can be understood without them, but here a comment like `# Accessing property only once.` would be good. Otherwise someone could refactor this code so that this assignment is inlined.

pekkaklarck (commented):
Looks good but in my opinion could still be enhanced a bit.

pekkaklarck:
The performance difference between different formatting solutions is too small to matter at all. Also inlining helper methods for the small benefit of avoiding the method call overhead generally makes no sense. There is code, also in Robot, where performance is so important that it wins over cleaner code, but this is not that kind of code.

src/robot/reporting/xunitwriter.py
rikerfi: I was also thinking that removing `_time_as_seconds` helper might be too much. I'll play around with `format(millis / 1000, '.3f')` and `str(round(millis / 1000, 3))` when getting back the helper.

PR #4367, opened 2022-06-16
User guide: Update inheritance example

mikahanninen:
The current version of the inheritance example does not work, because `SeleniumLibrary` needs `keyword` decoration for implemented methods to be recognized as Robot Framework keywords.

Either update according to my changes or switch to some other example that does not require decoration.

https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#using-inheritance


pekkaklarck:
Thanks for the fix. It would probably be better to use some other library, a library that uses the normal API and not the dynamic API, but at least I don't have time to for that now.

PR #4377, opened 2022-06-21
Update languages.py

leeuwe:
PT by @HelioGuilherme66 
BT-br by @HelioGuilherme66 
DE by @Snooz82 and @Noordsestern
NL by @pimjansen and @leeuwe
CS by @MoreFamed
PL by @bhirsz
TR by @yusufcanb
UK by @dies
ZN-cn by @nixuewei


src/robot/conf/languages.py
```diff
@@ -110,7 +110,7 @@ def settings(self):
         return settings
 
 
-class En(Language):
+class EN(Language):
```
pekkaklarck, line 113: Our class naming convention has followed `TitleCase` format also with abbreviations like `HtmlToXml`. I guess using `EN` instead of `En` would be fine in this context, but it causes problems with language variants like `PT-BR`. I'd prefer them to be in format like `PtBr` and not `PTBR`.

src/robot/conf/languages.py
```diff
@@ -170,3 +170,265 @@ class Fi(Language):
...
+    tags = 'Štítky'
+    setup = 'Příprava'
+    teardown = 'Ukončení'
+    template = 'Šablona'
+    timeout = 'Časový limit'
+    arguments = 'Argumenty'
+    return_ = 'Vrací'
+    bdd_prefixes = {'Pokud', 'Když', 'Pak', 'A', 'Ale'}
+
+class DE(Language):
```
pekkaklarck, line 203: Should have two empty rows between classes here and elsewhere.

src/robot/conf/languages.py
```diff
@@ -170,3 +170,265 @@ class Fi(Language):
...
+    setup = 'Inicialização'
+    teardown = 'Finalização'
+    template = 'Modelo'
+    timeout = 'Tempo Limite'
+    arguments = 'Argumentos'
+    return_ = 'Retornar'
+    bdd_prefixes = {'Dado', 'Quando', 'Então', 'E', 'Mas'}
+
+from robot.conf import Language
+class PT-br(Language):
```
pekkaklarck, line 320: The import isn't needed and the `PT-br` class name is a syntax error. It should be changed to `PtBr`. Probably should update the script that generates this Python code as well.

src/robot/conf/languages.py
```diff
@@ -170,3 +170,265 @@ class Fi(Language):
...
+    tags = 'Tags'
+    setup = 'Setup'
+    teardown = 'Teardown'
+    template = 'Template'
+    timeout = 'Timeout'
+    arguments = 'Arguments'
+    return_ = 'Return'
+    bdd_prefixes = {'Given', 'When', 'Then', 'And', 'But'}
+
+class ZH-CN(Language):
```
pekkaklarck, line 407: Another invalid class name.

src/robot/conf/languages.py
```diff
@@ -170,3 +170,265 @@ class Fi(Language):
...
+    tags = 'Etiketler'
+    setup = 'Kurulum'
+    teardown = 'Teardown'
+    template = 'Şablon'
+    timeout = 'Zaman Aşımı'
+    arguments = 'Argümanlar'
+    return_ = 'Return'
+    bdd_prefixes = {'Diyelim ki', 'Eğer ki', 'O zaman', 'Ve', 'Ancak'}
+
+class UK(Language):
```
pekkaklarck, line 378: Many people are likely to confuse this Ukrainian language code with United Kingdom. There are likely also other codes that aren't that well known so we probably should add the full name of the language somewhere. I don't think we need it in the code so it could be a comment or, probably better, a docstring. Do you have any opinions?

src/robot/conf/languages.py
```diff
@@ -170,3 +170,265 @@ class Fi(Language):
...
+    setting_headers = {'Setting', 'Settings', 'Settings', 'Settings'}
```
pekkaklarck, line 379: Markers that aren't translated could also be omitted altogether. Listing them again shouldn't cause problems, though, so leaving them here is fine as well. I guess it could make it easier for others to propose translations.

pekkaklarck (changes requested):
There are syntax errors and some style violations in the code. We should also discuss about adding the full language name somewhere because language codes aren't too explicit.

src/robot/conf/languages.py
leeuwe: now handled by the newer version of crowdin.py

leeuwe:
@pekkaklarck Finally resolved the conflicts!
PT by @HelioGuilherme66
BR by @HelioGuilherme66 
NL by @pimjansen and @leeuwe
CS by @MoreFamed
UK by @dies

src/robot/conf/languages.py
leeuwe: I opted for Pt-Br. Should that still be changed to PtBr or is the code updated?

bhirsz:
@leeuwe You tagged me, but I don't see polish there? It still needs some updates though (the translation)

leeuwe:
> @leeuwe You tagged me, but I don't see polish there? It still needs some updates though (the translation)

Yes, initially, Polish was done if I remember correctly, but after some changes it wasn't 100% translated and I left it out. I should've written explainatory commit messages. My bad

src/robot/conf/languages.py
```diff
@@ -202,12 +271,156 @@ class Fi(Language):
...
+    tags = 'Marker'
+    setup = 'Vorbereitung'
+    teardown = 'Nachbereitung'
+    template = 'Vorlage'
+    timeout = 'Zeitlimit'
+    arguments = 'Argumente'
+    bdd_prefixes = {'Angenommen', 'Wenn', 'Dann', 'Und', 'Aber'}
+
+
+class Pt-Br(Language):
```
pekkaklarck, line 359: This is invalid syntax. `-` needs to be removed.

src/robot/conf/languages.py
```diff
@@ -202,12 +271,156 @@ class Fi(Language):
     test_timeout = 'Testin Aikaraja'
     task_timeout = 'Tehtävän Aikaraja'
     test_tags = 'Testin Tagit'
+    task_setup = 'Tehtävän Alustus'
+    task_teardown = 'Tehtävän Purku'
+    task_template = 'Tehtävän Malli'
+    task_timeout = 'Tehtävän Aikaraja'
```
pekkaklarck, line 277: - Task settings are here now twice.
- Setup/teardown have conflicting terms in different places. Setup is using both `Alustus` and `Ylösajo` and teardown uses `Purku` and `Alasajo`. I don't have a super strong opinion on which to use, but the usage should be consistent. I think it's now best to use `Alustus` and `Purku` as that's what tests use. Could then be changed later if needed.

leeuwe: Fixed. I will need to reflect it in Crowdin to have it all in one place

pekkaklarck:
The description doesn't seem to match what's actually included. Most likely that's due to changes to settings that can be translated. Anyway, the PR actually adds these languages:
- PT
- PT-BR
- DE
- NL
- CS
- FR

pekkaklarck (approved):

PR #4396, opened 2022-07-12
Adds Portuguese pt-pt translations.

HelioGuilherme66:
Tests are failing because of three words values.

@pekkaklarck In the cases we have an article there will be three words, and seems the parser is not linking it ;)

I tried the pt-pt variant and it was not accepted. Will the pt-br be accepted, for the class PtBr?

Example: 
Test Cases --> Cases of Test --> Casos de Teste

### DIFF
--- a/atest/robot/parsing/translations.robot
+++ b/atest/robot/parsing/translations.robot
@@ -20,6 +20,14 @@ Invalid
     ...    .*${USAGE TIP}
     Should Match Regexp    ${result.stderr}    (?s)^\\[ ERROR \\] ${error}$
 
+Portuguese language test case
+    Run Tests    --lang pt    parsing/portuguese_test.robot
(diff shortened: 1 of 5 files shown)


HelioGuilherme66:
I used Fedora 36 and Python 3.10.5.

pekkaklarck:
Could you test does using "Casos De Teste" work?

pekkaklarck:
The problem with `Casos de Teste` was that it wasn't in title case. I changed to code so that also this format works.

PT is added by #4377 along with other languages so I'll close this PR. Need to think should we add tests for translations like this PR did, though.

pekkaklarck:
My fix above didn't actually fix the problem with headers, it fixed it only with setting names. Better fix coming soon. See #4411 for details.

PR #4405, opened 2022-07-14
add optional default argument to Get From Dictionary

### LINKED ISSUE #4398: Collections: `Get From Dictionary` should accept a default value
[Pop From Dictionary](https://robotframework.org/robotframework/latest/libraries/Collections.html#Pop%20From%20Dictionary) has support for returning a default value if the key isn't found in the dictionary. e.g.
```robot
Pop From Dictionary  ${myDict}  myKey  myDefaultValue
````

I think it'd be great if `Get From Dictionary` also support having a default value passed to it. 🙂

e.g.
```robot
Get From Dictionary  ${myDict}  myKey  myDefaultValue
```

oboehmer:
fixes #4398


pekkaklarck:
Looks good! Docs need a note that the argument is new in RF 5.1 but I can add it myself.

PR #4416, opened 2022-07-21
Change german Variables to plural form

d-biehl:
In german the plural form of the english word "Variables" is "Variablen". Maybe we can change this ;-)


pekkaklarck:
Have you talked about this with @Snooz82 and @Noordsestern? Also, it would be good to change this at Crowdin as well.

Snooz82 (approved):

Snooz82:
I approve this change. Daniel is right!
I will change it in crowding as well.

Issue #4418, opened 2022-07-25
Dictionaries insider lists in YAML variable files not converted to DotDict objects


jiangliangnihao2022:
Yaml file: 
```yaml
x:
  a: a0
  b: b0
y:
  - a: a1
    b: b1
  - a: a2
    b: b2
```

Robot file:
```robotframework
*** Setting ***
Variables      ./data.yaml

*** Test Cases ***
Import Variables By Yaml Test
    ${type_x}=    Evaluate    type($x)
    Log To Console    \nx type is ${type_x}

    ${y_0}=    Set Variable    ${y}[0]
    ${type_y_0}=    Evaluate    type($y_0)
    Log To Console    y[0] type is ${type_y_0}\n
```
Print log:                                        .
**x type is <class 'robot.utils.dotdict.DotDict'>
...y[0] type is <class 'dict'>****

 I think y[0] also should be DotDict.  would you fix this issue, because Variable y[0] can not use more feature like .(dot)

===advice===
update YamlImporter._set function in src/robot/variables/filesetter.py
before:
```python
   def _dot_dict(self, value):
       if is_dict_like(value):
           value = DotDict((n, self._dot_dict(v)) for n, v in value.items())
       return value
```
after:
```python
    def _dot_dict(self, value):
        if is_dict_like(value):
            value = DotDict((n, self._dot_dict(v)) for n, v in value.items())
        if isinstance(value, list):
            value = [ self._dot_dict(v) for v in value ]
        return value
```

pekkaklarck:
True, the code doesn't currently take into account dicts inside lists. The proposed solution looks good but could possibly be changed to
```python
        if is_dict_like(value):
            return DotDict((n, self._dot_dict(v)) for n, v in value.items())
        if is_list_like(value):
            return [self._dot_dict(v) for v in value]
        return value
```
Using `is_list_like` instead of `isinstance` is consistent with `is_dict_like`. Alternatively both could use ´isinstance`, but then there could be problems is PyYAML would return some custom list/dict objects.

Are you @jiangliangnihao2022 interested to provide a PR? It would need tests in addition to the fix. Since this is a bug, I can fix it before RF 5.1 myself as well.

jiangliangnihao2022:
Hi  Pekka,
I am glad to receive your quick reply. Thanks!
I also think your revision is more reasonable. So I'm sure you can do better.
About PR, I am not familiar with what is PR.


[...]

PR #4424, opened 2022-08-07
Fix issue 4418

sunday2:
fix bug #4418

### DIFF
--- a/atest/testdata/variables/valid.yaml
+++ b/atest/testdata/variables/valid.yaml
@@ -14,3 +14,6 @@ dict:
   key with spaces: value with spaces
 nested dict:
   dict: *alias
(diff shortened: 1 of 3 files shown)


pekkaklarck:
Oh no, I had already implemented this last week when I had some coding time in a train. Unfortunately network connection wasn't great so I didn't push changes and haven't been working since. I pushed the changes now and there's obviously a conflict. Fixing it doesn't make sense because your changes are pretty much exactly what I did as well. I thus need to close this PR but thanks for contribution anyway. Hopefully you can find something else to work on later.

I try to be better in mentioning in issues if I start working with them to avoid this in the future. You may want to do the same so that others, including I, don't work with the same issue.

PR #4430, opened 2022-08-12
add flags argument to all regexp kws in Builtin and String

### LINKED ISSUE #4429: Allow passing flags to regexp related keywords using explicit `flags` argument
The use of regex flags as part of the expression (i.e. `(?i)fooBAR` to do a case-insensitive lookup) is an advanced regex feature not obvious to all users. It would be convenient to also pass optional arguments to `Get Regexp Matches` and other regexp keywords, i.e.

```
${result}=   Get Regexp Matches     FOObar  foo    flags=I
${result}=   Get Regexp Matches     FOO\n123\nbar  foo.*bar    flags=DOTALL|I
```

oboehmer:
closes #4429


src/robot/libraries/BuiltIn.py
```diff
@@ -1303,7 +1304,7 @@ def should_match(self, string, pattern, msg=None, values=True,
             raise AssertionError(self._get_string_msg(string, pattern, msg,
                                                       values, 'does not match'))
 
-    def should_match_regexp(self, string, pattern, msg=None, values=True):
+    def should_match_regexp(self, string, pattern, msg=None, values=True, flags=''):
```
pekkaklarck, line 1307: I think the default should be `None`. This obviously applies to all keywords.

src/robot/libraries/String.py
```diff
@@ -363,15 +364,16 @@ def get_lines_matching_regexp(self, string, pattern, partial_match=False):
         | ${lines} = | Get Lines Matching Regexp | ${result} | Reg\\\\w{3} example |
         | ${lines} = | Get Lines Matching Regexp | ${result} | Reg\\\\w{3} example | partial_match=true |
         | ${ret} =   | Get Lines Matching Regexp | ${ret}    | (?i)FAIL: .* |
+        | ${ret} =   | Get Lines Matching Regexp | ${ret}    | FAIL: .* | flags=IGNORECASE
```
pekkaklarck, line 367: `|` missing at the end. It probably actually doesn't matter, but it should be used for consistence as above lines have it.

src/robot/utils/misc.py
```diff
@@ -115,3 +115,20 @@ def isatty(stream):
         return stream.isatty()
     except ValueError:    # Occurs if file is closed.
         return False
+
+def parse_re_flags(flags_as_str):
+    result = 0
+    if flags_as_str:
+        unknown = []
+        for f in flags_as_str.split('|'):
```
pekkaklarck, line 123: We haven't generally used one character variable names except with comprehensions. Using `for flag in flags.split('|')` or possibly `for name in flags.split('|')` would be better.

src/robot/utils/misc.py
```diff
@@ -115,3 +115,20 @@ def isatty(stream):
         return stream.isatty()
     except ValueError:    # Occurs if file is closed.
         return False
+
+def parse_re_flags(flags_as_str):
+    result = 0
+    if flags_as_str:
```
pekkaklarck, line 121: Using
```python
if not flags:
    return result
```
would avoid one indentation level with code below.

src/robot/utils/misc.py
```diff
@@ -115,3 +115,20 @@ def isatty(stream):
...
+            try:
+                flag = getattr(re, f.upper().strip())
+                if isinstance(flag, re.RegexFlag):
+                    result |= flag
+                else:
+                    unknown.append(f)
+            except AttributeError:
+                unknown.append(f)
+        if unknown:
+            raise ValueError('unknown flag{}: {}'.format(plural_or_not(unknown), seq2str(unknown)))
```
pekkaklarck, line 133: - Should use f-string here.
- Error messages should be proper English and thus should start with `U` not `u`.
- Could consider changing the message to 'Unknown regexp flag' to make it more explicit.
- It's nice to report all unknown tags in one go, but I'm not sure how much benefits it actually has over just raising this exception when the first unknown value is encountered. Since you already have implemented it, we can as well keep it.

pekkaklarck (changes requested):
I did a somewhat quick look at the code and added some comments that were partly pretty nitpickey. :+1: for adding unit tests for the new utility. I'll take a more detailed look at the acceptance tests and documentation later.

pekkaklarck (approved):
Looks great! Few more "New in RF 5.1" notes could be added but I can do that myself after merging this.

oboehmer:
> 

I didn’t add this because all keywords reference the Should Match Regexp one for general usage, and there it’s mentioned..

pekkaklarck:
Yeah, I noticed that. Then again, not everyone is looking for those other keywords and having a "new in x.y" note at the end doesn't really cost anything.

Issue #4433, opened 2022-08-16
Convert and validate collection contents when using generics in type hints


JFoederer:
The [supported conversions](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#toc-entry-525) section of the documentation states:

> Also types in in the typing module that map to the supported concrete types or ABCs (e.g. List) are supported. With generics also the subscription syntax (e.g. List[int]) works, **but no validation is done for container contents**.

It seems like a good enhancement to include these conversions as well, but the current code actually also contains a defect in this area. The `converter_for` method of `TypeConverter` strips the `__origin__` information for all types but `Union`. As a result it is not possible to write any custom converters for generic types. Asuming you would typically need the nested types to do the proper conversions.

![image](https://user-images.githubusercontent.com/32476108/184913083-e18d2303-de46-4f99-b5d2-42c119ec27b7.png)

A pull request will follow to clarify the issue and demonstrate the suggested enhancement.

pekkaklarck:
Interesting! I don't have time to look at this now, but I add this to RF 5.1 scope to not forget about it. Would be a nice enhancement to get into that release.

JFoederer:
That would be great! I will try a Linux system with python 3.6 as well, to see what the unit test failures are about over there. I also gained some better insights in the LibDoc failures. A proposal will follow.

pekkaklarck:
I've been busy with non-work stuff lately (mostly studying interpreters and compilers so it's been highly related) and haven't had time to look at this yet. I try to get beta out today but will look at this afterwords. This is in v5.1 milestone and won't be forgotten.

JFoederer:
No problem, thanks for the update. I'll just consider it a compliment that you don't see it as a risk to integrate it that late in the process ;-) Although I also hope the tests I added are clear enough to show the exact impact.

The current status is that the Libdoc issues are solved without regression or changes. The Python 3.6 unit tests are fixed, but the 3.6 acceptance tests aren't. I haven't tried 3.7, but was able to run successful from Python 3.8 upward on Windows and Linux. The offer to disable the tests for Python 3.6 starts to look tempting. The issue originates from somewhere in the surrounding code, where the None-converter gets in the way. I lacked the time to deep dive into this part.

pekkaklarck:
This has been now implemented based on the excellent PR #4434 by @JFoederer. Thanks a lot for your hard work!

pekkaklarck:
[...]

PR #4434, opened 2022-08-16
Automatic type conversion for generic types

JFoederer:
This PR is related to issue #4433. It adds support for converting nested types like List[int] or Dict[str, bool]. It also resolves the defect where type information is lost by retaining the original type information.


JFoederer:
I still have 2 open questions about this pull request.

- There are a few failing tests in the LibDoc suite. It was unclear to me if we need to update the tests, because a limitation was resolved, or that these reflect a requirement and the code needs to change.
- Some of the *invalid* tests now have a very long error message in them. I would like to shorten them, but didn't succeed (yet)

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
+            raise ValueError
+        if self.nested_types:
+            if TypeConverter.converter_for(self.nested_types[0],
+                                           languages=self.languages):
+                for i, elem in enumerate(converted_list) :
+                    converter = TypeConverter.converter_for(self.nested_types[0],
+                                                            languages=self.languages)
+                    if converter:
+                        elem_info = f"List[{type_name(self.nested_types[0])}]"
+                        converted_list[i] = converter.convert(elem_info, elem,
```
pekkaklarck, line 426: This code seems to be responsible on that 
> Argument 'List[integer]' got value '3.14' (float) that cannot be converted to integer

error message that is a bit confusing because the actual argument is the whole list. I think the error could be improved by catching the exception raised by `converter.convert` and then re-raising it with more info. Good error messages are very important, but I believe it's better to get other code cleaned up before looking at this too much.

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
+    _convert = _non_string_convert
 
 @TypeConverter.register
 class DictionaryConverter(TypeConverter):
     type = dict
     abc = Mapping
+    abc = (Mapping, MutableMapping)
     type_name = 'dictionary'
-    aliases = ('dict', 'map')
+    aliases = ('dict', 'dictionary', 'map')
```
pekkaklarck, line 471: Are the above changes to `abc` and `aliases` really needed?

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
     value_types = (str, Mapping)
 
-    def _non_string_convert(self, value, explicit_type=True):
-        if issubclass(self.used_type, dict) and not isinstance(value, dict):
-            return dict(value)
-        return value
+    def __init__(self, type_, custom_converters=None, languages=None):
+        super().__init__(type_, languages=languages)
+        self.nested_types = getattr(type_, '__args__', ())
```
pekkaklarck, line 476: Custom converters need to be passed forward.

Here we could validate that `len(self.nested_types)` is either 0 or 2 (like with `list` we could check it's 0 or 1). There wouldn't be a need to look that again later anymore.

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
+        if isinstance(value, str):
+            converted_dict = self._literal_eval(value, dict)
+        elif is_dict_like(value):
+            converted_dict = dict(value)
+        else:
+            raise ValueError
+        if self.nested_types and len(self.nested_types) == 2:
+            for key, elem in converted_dict.copy().items():
+                key_type = self.nested_types[0]
+                value_type = self.nested_types[1]
```
pekkaklarck, line 491: Having
```python
key_type, value_type = self.nested_types
```
before the loop would be better. Possibly that assignment could be done already in `__init__` and then `self.nested_types` wouldn't be needed at all. Other converters should obviously use the same logic in that case.

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
+                d_info = f"Dict[{type_name(key_type)}, {type_name(value_type)}]"
+                key_converter  = TypeConverter.converter_for(key_type,
+                                                             languages=self.languages)
+                elem_converter = TypeConverter.converter_for(value_type,
```
pekkaklarck, line 495: `value_converter` would be more consistent name.

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
+                if elem_converter:
+                    converted_elem = elem_converter.convert(d_info, elem, explicit_type)
+                else:
+                    converted_elem = elem
+                replace_elem = type(converted_elem) != type(elem) 
+                if converted_key != key:
+                    converted_dict.pop(key)
+                    replace_elem = True
+                if replace_elem:
+                    converted_dict[converted_key] = converted_elem
```
pekkaklarck, line 511: Modifying existing dictionary looks pretty complicated. Why not start with an empty dictionary that is populated based on (possibly) converted keys and values? Other converters would probably benefit from that approach as well.

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
+
+    _convert = _non_string_convert
 
 
 @TypeConverter.register
 class SetConverter(TypeConverter):
     type = set
     abc = Set
     type_name = 'set'
+    aliases = ('set',)
```
pekkaklarck, line 522: Shouldn't be needed.

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
     value_types = (str, Container)
 
+    def __init__(self, type_, custom_converters=None, languages=None):
+        super().__init__(type_, languages=languages)
+        self.nested_types = getattr(type_, '__args__', ())
```
pekkaklarck, line 527: Custom converters should be passed forward.

Could check that `__args__` is either empty or contains exactly one argument already here.

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
+        if isinstance(value, str):
+            converted_set = self._literal_eval(value, set)
+        elif is_list_like(value):
+            converted_set = set(value)
+        else:
+            raise ValueError
+        if self.nested_types:
+            for elem in converted_set:
+                converter = TypeConverter.converter_for(self.nested_types[0],
+                                                        languages=self.languages)
```
pekkaklarck, line 542: Converter could be created only once before the loop.

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
+                if converter:
+                    converted_set.remove(elem)
+                    t_name = type_name(self.nested_types[0])
+                    converted_set.add(converter.convert(f"Set[{t_name}]", elem,
+                                                        explicit_type))
+        return converted_set
 
-    def _convert(self, value, explicit_type=True):
-        return self._literal_eval(value, set)
+    _convert = _non_string_convert
```
pekkaklarck, line 550: Shouldn't be done.

src/robot/libdocpkg/datatypes.py
```diff
@@ -58,9 +58,11 @@ def for_type(cls, type, converters):
         elif not converter.type:
             return cls(cls.CUSTOM, converter.type_name, converter.doc,
                        converter.value_types)
-        else:
+        elif converter.type in STANDARD_TYPE_DOCS:
             return cls(cls.STANDARD, converter.type_name,
                        STANDARD_TYPE_DOCS[converter.type], converter.value_types)
+        else:
+            return None
```
pekkaklarck, line 65: Why are these changes needed?

pekkaklarck (changes requested):
Great work but changes are still needed. See comments for details and don't hesitate to ask details here or on Slack.

pekkaklarck:
New tests for generics with custom converters like `List[UsDate]` should probably be added to the `custom_converters.robot` suite. The `CustomConverters.py` library used there contains converts that can be used.

atest/testdata/keywords/type_conversion/annotations_with_typing.robot
```diff
@@ -9,15 +10,18 @@ List
...
+    List with ints            [1, ${2}, 1313, -42]        [1, 2, 1313, -42]
+    List with enums           ['foo', 'bar']              [MyEnum.foo, MyEnum.bar]
 
 Invalid list
     [Template]                Conversion Should Fail
     List                      [1, oops]                                        error=Invalid expression.
     List                      ()                                               error=Value is tuple, not list.
-    List with params          ooops                       type=list            error=Invalid expression.
+    List with ints            ooops                       type=list            error=Invalid expression.
+    List with ints            [1, ${2}, 3.14, -42]        type=list            error=Argument 'List[integer]' got value '3.14' (float) that cannot be converted to integer: Conversion would lose precision.
```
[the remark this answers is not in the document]
JFoederer: Yes, this is quite a tricky one. Difficult to make a concise message that still has the good info. In an earlier version I had also included the index of the element for which the conversion failed. I removed that because it felt too lengthy. I ended up with this balance where the user is shown the actual failing conversion, but has to go through the list their selves to see where it is.

doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
```diff
@@ -1166,8 +1166,8 @@ argument is converted to the concrete type.
 
 Also types in in the typing_ module that map to the supported concrete
 types or ABCs (e.g. `List`) are supported. With generics also the subscription
-syntax (e.g. `List[int]`) works, but no validation is done for container
-contents.
+syntax (e.g. `List[int]`) works. When using subscripting, validation and
+conversion is extended to the container contents (New in Robot Framework 5.1).
```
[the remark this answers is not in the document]
JFoederer: Feel free to update as you feel fit. Keeping the text as-is does not seem appropriate. It explicitly states that no conversion is done.

src/robot/libdocpkg/datatypes.py
JFoederer: Due to a defect in the existing code. While testing I ran into a KeyError in what is now line 63. The code assumed that if there is a converter and it has a type, that it must be in STANDARD_TYPE_DOCS. At one point it wasn't.

To be honest, I am not sure anymore if this was only to improve error reporting during work in progress, or that there are still active paths in the running code that reach this point. Anyway, the KeyError was not very helpful, so I kept the improvement regardless.

src/robot/running/arguments/typeconverters.py
```diff
@@ -15,7 +15,8 @@
 
 from ast import literal_eval
 from collections import OrderedDict
-from collections.abc import ByteString, Container, Mapping, Sequence, Set
+from collections.abc import ByteString, Container, Sequence, Set, \
+                            Mapping, MutableMapping
```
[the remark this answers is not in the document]
JFoederer: Ok

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
     type_name = 'list'
     abc = Sequence
     value_types = (str, Sequence)
+    aliases = ('list',)
```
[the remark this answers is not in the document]
JFoederer: I think you are right.

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
 
-    def no_conversion_needed(self, value):
-        if isinstance(value, str):
+    def __init__(self, type_, custom_converters=None, languages=None):
+        super().__init__(type_, languages=languages)
```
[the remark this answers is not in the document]
JFoederer: Well spotted, I completely missed that. I have not done anything with custom converters yet in this PR. Also not sure yet what the impact might be. I will start by fixing the constructors.

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
+
+    def no_conversion_needed(self, value):
+        return not self.nested_types \
+               and isinstance(value, self.used_type) \
+               and not isinstance(value, str)
 
     def _non_string_convert(self, value, explicit_type=True):
-        return list(value)
+        if isinstance(value, str):
+            converted_list = self._literal_eval(value, list)
```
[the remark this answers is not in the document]
JFoederer: agreed

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
+        elif is_list_like(value):
+            converted_list = list(value)
+        else:
+            raise ValueError
+        if self.nested_types:
+            if TypeConverter.converter_for(self.nested_types[0],
+                                           languages=self.languages):
+                for i, elem in enumerate(converted_list) :
+                    converter = TypeConverter.converter_for(self.nested_types[0],
+                                                            languages=self.languages)
```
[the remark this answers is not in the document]
JFoederer: With the max line length limitation I have started using way shorter names than I generally like. Not an issue here though. I changed to index.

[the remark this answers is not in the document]
JFoederer: I took the converter out of the loop. The question about List[int, str] is a good one. I don't have a straight answer. My first thought it that you should use a tuple instead of a list when mixing types. Not sure what Python supports.

src/robot/running/arguments/typeconverters.py
JFoederer: I agree that good error messages are very important. Something was mentioned on this elsewhere as well. Let's continue the discussion there.

JFoederer:
Thanks @pekkaklarck, good review! I did what I could within the time I had available now. Not all review comments are processed yet. I hope I can spend some more time on it later this week.

pekkaklarck:
Don't have time to look at the changes now and probably not before Friday, but a quick comment about the syntax we support. I somewhat strongly believe we should allow only one type with lists and sets and require two types with dicts. With tuples we can allow any number of types, but the tuple used as an argument should then have that amount of values. Python doesn't seem to care so you can use `list[int, str, bool]` and even `dict[int, str, bool]`, but I believe we should be stricter. Using `list[int|str|bool]` should work, though.

src/robot/running/arguments/typeconverters.py
JFoederer: agreed

src/robot/libdocpkg/datatypes.py
pekkaklarck: If it's a recognized converted, Libdoc should have docs for it and `return None` is thus risky. I agree `KeyError` isn't good either convenient (that occurred recently after I added `pathlib.Path` converter and forgot to add docs), but it would be better to use something like `raise ValueError(f'No documentation for {type(converter).__nama__}.')` instead of `return None´.

src/robot/running/arguments/typeconverters.py
```diff
@@ -390,16 +391,42 @@ class ListConverter(TypeConverter):
...
+    def no_conversion_needed(self, value):
+        return not self.nested_types \
+               and isinstance(value, self.used_type) \
+               and not isinstance(value, str)
 
     def _non_string_convert(self, value, explicit_type=True):
-        return list(value)
+        if not is_list_like(value):
+            raise ValueError
+        return self.__nested_conversion(value, explicit_type)
```
pekkaklarck, line 414: If I remember this logic right, this method isn't called if this converter doesn't handle the value and in that case `is_list_like` check would be unnecessary. That would allow moving all code from  `__nested_conversion` here.

pekkaklarck:
Converters could probably create converters for their values already in `__init__` similarly as `CombinedConverter` initializes `self.converters`.

With lists and sets it could be something like this:
```python
types = getattr(type_, '__args__', ())
if not types:
    self.item_converter = None
elif len(types) == 1:
    self.item_converter = TypeConverter.converter_for(types[0], custom_converters, languages)
else:
    raise TypeError    # Or some other exception. This would need own tests as well.
```

Dicts would need to initialize both `self.key_converter` and `self.value_converter`. Tuples would need to store all:

```python
types = getattr(type_, '__args__', ())
if not types:
    self.item_converters = None
else:
    self.item_converters = [TypeConverter.converter_for(t, custom_converters, languages) for t in types]
```

src/robot/libdocpkg/datatypes.py
JFoederer: I would have to reproduce the issue to really remember what was going on. If I recall correctly it had something to do with new composed types being created at runtime from existing types, for which Libdoc didn't find the docs. Returning None had in that case identical behavior as when there isn't any converter (line 57). I think it still showed the name, but no link to the type's details.

src/robot/running/arguments/typeconverters.py
```diff
@@ -389,61 +389,165 @@ class ListConverter(TypeConverter):
...
+            converted_tuple = self._literal_eval(value, tuple)
+        elif is_list_like(value):
+            converted_tuple = tuple(value)
+        else:
+            raise ValueError
+        if self.nested_types:
+            tmp_list = list(converted_tuple)
+            for i, elem in enumerate(converted_tuple):
+                converter = TypeConverter.converter_for(self.nested_types[i],
+                                                        languages=self.languages)
```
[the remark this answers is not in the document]
JFoederer: Yes, that is indeed a good idea. Note that for tuples it is not mandatory to list all types. Partial annotations by listing the first type(s) only is permitted.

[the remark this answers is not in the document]
pekkaklarck: Is it documented somewhere that if you have e.g. `tuple[int, int, int]` type checkers should allow also `(0, 0, 0, 0)`? If not, I think we should be strict and require exactly the same amount of items as there are types listed. Otherwise there can be annoying runtime errors if you have e.g. `a, b, c = arg`. If you don't want such strict validation, you can use `list[int]` or `list[int|other|types]`.

src/robot/libdocpkg/datatypes.py
pekkaklarck: In that case it would probably be best to replace `return None` with some meaningful exception. Most likely it occurs only if there's a bug in our code (all recognized converters should have docs!) and a good exception makes understanding the situation better. If there are in fact legitimate cases where a converter shouldn't have docs, then this code is fine but there should be a unit test validating it.

src/robot/running/arguments/typeconverters.py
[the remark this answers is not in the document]
JFoederer: Never mind. I recalled reading this somewhere, but when looking it up again I came up empty. This is the official Python docu and it makes no such mention: https://docs.python.org/3/library/typing.html#typing.Tuple
Also, it is always possible to use the _Any_ type and still be complete.

So in short, I fully agree with your initial comment on this topic.

src/robot/libdocpkg/datatypes.py
JFoederer: I am clearly missing something obvious. This very basic example already fails to show the type link. It also doesn't raise the exception though.

`attempt.py`
```
class MyStuff:
    """bla bla"""

class attempt:
    def dummy(self, x: MyStuff):
        pass
```

It shows Keyword dummy with argument `x <MyStuff>`, but it does not link to show '_bla bla_'

src/robot/running/arguments/typeconverters.py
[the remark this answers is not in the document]
pekkaklarck: The docs you referred mention `tuple[int, ...]` that we need to support. We could do it so that if the last type is `Ellipsis`, we allow tuples with more values that actual types. Not sure should we validate/convert those extra values or just leave them as-is. Related to that, we need to decide how to handle `tuple[int, str, ...]`.

src/robot/libdocpkg/datatypes.py
pekkaklarck: Libdoc doesn't work like this. It shows type info only for types it converts. That means std types like `int` and custom types that are separately registered.

pekkaklarck:
I prototyped converting `TypedDict` items (#4477) and had a similar problem as here with properly reporting errors when converting an individual item fails. My solution was adding new `kind='Argument'` argument to `convert()` and passing it to `_handle_error` where `f"Argument '{name}'` was then changed to `f"{kind} '{name}'`. After that I just needed to pass `kind='Item'` to `convert()` when converting items. The resulting error messages look like this:
> ValueError: Argument 'argument' got value '{'x': 'bad'}' that cannot be converted to Example: Item 'x' got value 'bad' that cannot be converted to integer.

In my opinion that error is pretty ok and most likely same approach could be used with generics as well.

JFoederer:
That does indeed look ok. Good addition! Maybe be a bit careful in duplicating the dictionary or list. They can get large and draw attention away from the item actually causing the failure. Especially now that those details are also properly included. Not sure if we have all the information available at the right time to make such decision from code. If not, this looks totally fine to me.

JFoederer:
> Converters could probably create converters for their values already in `__init__` similarly as `CombinedConverter` initializes `self.converters`.

I was unsure how _dynamic_ the types and converters could be and when in time the constructors are actually called. I put them in the convert method to be on the safe side. But to be honest, without checking the potential performance impact.

pekkaklarck:
Ok, let's go with the simple error reporting enhancement I proposed above. That's something that can be enhanced in future versions if needed. Right now we need to concentrate on getting the main functionality working well and good enough error messages are, well, good enough. Changes to the actual conversion logic in the future would have backwards compatibility concerns, but enhancing error reporting is totally safe. All this said, it will be interesting to test how this works with nested collections.

atest/testdata/keywords/type_conversion/annotations_with_typing.robot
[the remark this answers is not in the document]
xylix: Perhaps 
> Argument 'List[integer]' included value '3.14' (float) that cannot be converted to integer

Using "included" instead of "got" would make it more explicit that it was only one part of the argument that caused the problem.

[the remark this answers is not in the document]
pekkaklarck: Yeah, it would be better. We already discussed this elsewhere in these comments and agreed on a small enhancement that basically changes the erro to
> Item '1' got value ...
where `1` is the index. This is wrapped into another similar error which ought to make the final message pretty good. Anyway, error reporting can certainly be enhanced once the main functionality is in.

src/robot/libdocpkg/datatypes.py
JFoederer: The real issue here seems to be that I was using the same construct as inside Robot in an external library. So, using `@TypeConverter.register` instead of `ROBOT_LIBRARY_CONVERTERS=` or `@library(converters=`. Functionally it works, but it somehow has a side effect when using Libdoc. On the other hand, I am unable to reach the nested_type info when using the custom converter approach.

pekkaklarck:
Due to there being hurry to get the next release out, @JFoederer and I agreed that I continue with this PR. I checked it out locally, but unfortunately there were some issues with older Python versions, especially Python 3.6, and I couldn't easily figure out what caused them. In the end I started a new branch where I implemented this functionality based on this PR and the discussion we had above. That turned out to work well and in the end it didn't anymore make sense to try updating this PR.

I have now committed my changes to master which makes this PR outdated. The PR was _very_ helpful and I'd like to thank @JFoederer for his excellent work. I hope you can test how the new implementation works to see are there any holes that we'd still need to fill.

PR #4445, opened 2022-08-24
Fix missing variable error msg when importing library / variable in --dryrun mode

jnhyperion:
Here's a minimal example to reproduce this issue.
[example.zip](https://github.com/robotframework/robotframework/files/9415174/example.zip)

run command `robot --dryrun example/`

Before fix:

there will be the error msg raise:

[ ERROR ] Error in file '/example/test.robot' on line 2: Variable '${module}' not found.
[ ERROR ] Error in file '/example/test.robot' on line 3: Replacing variables from setting 'Variables' failed: Variable '${var}' not found.

After fix:
there will not error msg

Because in many cases we will set the variables dynamically and in `--dryrun` mode, robot does not parsing the variables, and we do not expect to see these kind of errors.


pekkaklarck:
I'm actually not sure is this a good idea at least with libraries and resource files. If importing them fails, keywords they contain won't be found and there will be failures also in dry run. The error about failed import makes it easier to understand them. This could be safely done with variable files, though, because we don't anyway look at variables in dry-run. I'm not sure is that worth the effort though.

If we agree to do this change, this PR would need to have tests and an issue tracking it should be submitted as well.

pekkaklarck:
No comments since September. Closing.

PR #4446, opened 2022-08-28
Enhancements for per-file language config without bleeding languages

d-biehl:
Tried to implement the file language configuration so that no language bleeding into other files occurs. 

The 2 test cases that fail must be updated or removed.

### DIFF
--- a/src/robot/running/builder/parsers.py
+++ b/src/robot/running/builder/parsers.py
@@ -16,6 +16,7 @@
 import os
 from ast import NodeVisitor
 
+from robot.conf.languages import Languages
 from robot.errors import DataError
 from robot.output import LOGGER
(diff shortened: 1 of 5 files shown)


pekkaklarck:
Although this doesn't look as bad as I was afraid it would, there are still quite many changes and dependencies between various parts. I really think we should try to avoid such direct dependencies and instead pass context objects around during execution. That's obviously a bigger task but something we should do at some point anyway.

The above said, I'm not totally against adding this kind of change if the current behavior turns out to be problematic. I don't think it's important enough to warrant delaying RF 6.0 (previously known as RF 5.1) release, though.

pekkaklarck:
I'm closing this PR because:
- It's pretty complicated.
- There are conflicts.
- Language configuration leaking hasn't turned out to be too big a problem.

We can return to this later if needed.

PR #4452, opened 2022-09-05
Added countdown timer capability to the public logging API for test libraries

shashank404error:
- `logger.countdown(3)` will sleep for 180 seconds and print the countdown timer to the standard output
- looks like this in console - robotframework_ss
- Could be of help when two test cases depends on a time based event and you could know how much time will it take for the next test to execute.


pekkaklarck:
This isn't a generic enough functionality to be added into the core. You can very easily have your own utility function that does this using `logger.console`.

Issue #4294, opened 2022-04-01
Drop Python 3.6 and 3.7 support


pekkaklarck:
Python 3.6 [reached end-of-life](https://endoflife.date/python) in December 2021. We don't yet have concrete estimation when RF 6.0 will be released, or even when the development starts, but my guesstimate is that it will happen earliest in early 2023. Supporting Python version that itself hasn't been supported for over a year doesn't make much sense for us at that point.

At the time of this writing, April 2022, about 15% of [Robot Framework downloads](https://pypistats.org/packages/robotframework) still use Python 3.6, but I expect that number to decline pretty fast in the future. It could be that some Linux distributions still support Python 3.6 and have it as the default Python version, but even with them installing a newer Python ought to be pretty easy.

A concrete benefit of dropping Python 3.6 is that it allows us to use [datetime.fromisoformat](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat) that would help with #4258. Other Python 3.7 features that would likely be useful are [dataclasses](https://docs.python.org/3/library/dataclasses.html) and [importlib.resources](https://docs.python.org/3/library/importlib.html#module-importlib.resources).

pekkaklarck:
Python 3.7 will reach its end-of-life in [June 2023](https://peps.python.org/pep-0537/) and it's clear RF 7.0 won't be released before that. We can thus remove support of both Python 3.6 and Python 3.7 in RF 7.0.

The main reason we have supported Python 3.6 quite long after its end-of-life in [December 2021](https://peps.python.org/pep-0494/) is that it is the default Python version in RHEL 8 that is supported until [May 2029](https://endoflife.date/rhel). RHEL uses Python 3.9 by default and with newer RHEL versions upgrading Python ought to be easier as well. It's unlikely that other long-term support Linux distributions would use Python 3.7 by default and removing its support as soon as it's not officially supported ought to be fine.

pekkaklarck:
Commit b092961a4833b5e6b4c75c449f017c5e3a75f186 made Python 3.8 the minimum version and removed code supporting older versions. Some compatibility code may have been left behind, but it can be removed when encountered.

The main thing that hasn't been done is updating documentation. I have already updated the User Guide locally, but we also need to update README.rst and INSTALL.rst. A problem with them is that changes would be visible on the GitHub project pages immediately and RF 7 isn't actually available until at the end of the year. Probably best to leave them as they are now and update them when we the first release candidate is out. That means we cannot close this issue until that either.

PR #4474, opened 2022-09-22
Update the minimum release notes in the README

MarkHoo:
The minimum version should be modified to 3.7+. Because the 3.6 version has already ended support. Version 3.7 will also end support on June 27 next year.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,7 +49,7 @@ you can simply run::
 For more detailed installation instructions, including installing Python, see
 `<INSTALL.rst>`__.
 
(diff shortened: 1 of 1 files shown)


pekkaklarck:
We support Python 3.6 still until RF 6.0 (#4294). The main reason is that although the Python devs themselves don't support it anymore, it is still supported on various long term support Linux distributions. For example, Python 3.6 is the default Python on RHEL 8 that is in full support until May 2024, has end of life in 2029, and likely has some extended support period still after that.

Issue #4390, opened 2022-07-05
Add and document translations


pekkaklarck:
RF ~5.1~ 6.0 starts localization efforts with these ~two~ three enhancements:

- #4096
- #519
-  #4400 

The issues above cover adding the support for translations. This issue exists for these purposes:
- Actually adding translations for various languages.
- Documenting what languages we support in the User Guide and elsewhere.
- Collecting information about who has provided translations so that we can acknowledge them in the release notes.

The translation efforts have been lead by @leeuwe who also set a [Crowdin page](https://robotframework.crowdin.com/robot-framework) for collaboration. We also have a dedicated `#localization` channel on our [Slack](http://slack.robotframework.org).

pekkaklarck:
As mentioned above, we use [Crowdin](https://robotframework.crowdin.com/robot-framework) for collaboration and the #localization Slack channel for general discussion. You may also create your own language specific Slack channels for discussion if needed.

Translations at Crowdin are stored to YAML files in format `Language.yml`. This format isn't directly supported by Robot Framework, but you can convert them to suitable format using our [crowdin.py](https://gist.github.com/pekkaklarck/fd348b7eb59f35a26f2fdc1e7b0f008d) script. If you have a language file from Crowdin like `Czech.yml`, you can convert it with the script like this:

    python crowdin.py Czech.yml Czech.py

After that you can use the generated language file with Robot Framework like this:

    robot --language Czech.py example.robot

The generated class can also be added to Robot's [languages](https://github.com/robotframework/robotframework/blob/master/src/robot/conf/languages.py) module via a pull request to get it included as a built-in language.

pekkaklarck:
PR #4377 added translations for these languages:
- PT by @HelioGuilherme66 
- BT-BR by @HelioGuilherme66 
- DE by @Snooz82 and @Noordsestern
- NL by @Heatzone87 and @leeuwe
- CS by @MoreFamed
- FR by @lesnake

Thanks all for contributions!

HelioGuilherme66:
> **BT**-BR by @HelioGuilherme66
PT-BR

:)

pekkaklarck:
Two new translations:

- Thai (TH) by @up1 
- Polish (PL) by @bhirsz

pekkaklarck:
Two more translations:

- Spanish by Miguel Angel Apolayo Mendoza
- Simplified Chinese by @mawentao119 and @nixuewei

[...]

PR #4488, opened 2022-09-29
Updated Chinese Simplified and added Traditional

leeuwe:
provided by Ming Ni https://github.com/nixuewei


pekkaklarck:
There seem to be quite a lot of changes to ZhCn. Based on #4390 the original translations were provided by @mawentao119 and @nixuewei. Since @nixuewei as also provided these updates, I believe they are fine.

Issue #4490, opened 2022-09-30
Time string parsing for micro and nanoseconds


mikkuja:
RF can parse time from multiple different input formats. One of them is [time string](https://robotframework.org/robotframework/latest/libraries/DateTime.html#Time%20string). Currently smallest time specifiers are for milliseconds. However, we are working with electrical measurement devices and would need to input times down to nanosecond accuracy.

We use time strings in standard keywords

```robotframework
${time}=    Convert Time    250 ns
```

and in our own keywords that uses [argument conversions](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#supported-conversions).

Considerations:
- RF typically converts times to float as seconds.
- It is already possible to define sub millisecond times eg. using floats `1E-6` or decimals in string format `0.001 ms`.
- In some places accuracy is limited either to one second or one millisecond. See [Millisecond handling](https://robotframework.org/robotframework/latest/libraries/DateTime.html#Millisecond%20handling).
- Python [datetime.timedelta](https://docs.python.org/3/library/datetime.html#timedelta-objects) only supports up to microsecond accuracy.
  - Custom argument converter is needed for nanosecond accuracy. That can be easily implemented using [robot.Datetime API](https://robotframework.org/robotframework/latest/libraries/DateTime.html#Programmatic%20usage).
- DateTime converters can also output time in str formats like `1min 2s 345ms`. At this time we don't need this conversion and would probably prefer seconds as decimals anyway `1min 2.3456s`.

pekkaklarck:
PR #4491 implements support for micro and nanoseconds when parsing time strings. I don't understand how I missed that when I looked at this issue earlier. The PR looks so good that I guess we can still include it in RF 6.0. Other possible enhancements discussed above then need separate issues.

pekkaklarck:
PR #4491 added support for parsing nano and microseconds and I enhanced docs a bit in the above commit. This issue can be considered done and new issues submitted for other related enhancements.

Thanks for contribution @mikkuja!

PR #4491, opened 2022-09-30
Time string parsing for micro and nanoseconds

mikkuja:
Implementation for #4490

Adds following specifiers for time string parsing:
- `microseconds`, `microsecond`, `us`, `μs`
- `nanoseconds`, `nanosecond`, `ns`


pekkaklarck (approved):
This looks very good. Only thing missing is documenting this also in the User Guide and adding a note to documentation that this is new in RF 6.0. I can do that myself.

PR #4496, opened 2022-10-05
rework public Languages API fixes #4494

### LINKED ISSUE #4494: Enhance public `robot.api.Languages` and `robot.api.Language` APIs
In order to support languages reasonably in RobotCode (and other tools), it would be nice if we could talk about the Language API again.

1. To give the user a good error message which language or custom language was not loaded or found, it should be possible to load individual languages and import custom language files.
    - In the current implementation I can either create the `Languages` class in the hope that it works, in case of error I can not easily filter out the language that does not work.
    - `add_language` can't load user defined language files

2. There should be a public API that returns all available languages.
    - already exists, is just not public

3. The process of loading custom language files should be separated from loading the real language.
    - Before loading the real languages, the custom language files should be loaded, only then you can load the language
    - I think this should also be done with the command line parameters, there should be 2 arguments, one for loading the custom language files and the current one for the real language

4. it should be possible to create a `Languages` instance that does not insert English and the deprecated single forms of the headers automatically
   - I currently have to filter it out again if I want to show only the specified languages e.g. in the completion.

Let's discuss ;-)

d-biehl:
Example of a better public `Languages` API


d-biehl:
How can I rerun the checks or/and download the artifacts of the failed runs?
I cannot reproduce the errors on my local machine. And I'm not able to download the artifact off failed checks, to get the real error messages

d-biehl:
ok, I got it.

src/robot/conf/languages.py
```diff
@@ -20,27 +20,31 @@
 
 
 class Languages:
+    """Keeps a list of languages and unifies the translations in the properties.
 
-    def __init__(self, languages=None):
+    Parameters:
+        languages: a language or a list of languages. Can be the name, the code or an instance.
+        add_default: if True the default language (En) and some aliases is (Default: True)
```
pekkaklarck, line 27: :+1: for doc strings. Should preferably use the [Sphinx style](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html) for specifying parameters.

src/robot/conf/languages.py
```diff
@@ -67,36 +71,57 @@ def _get_languages(self, languages):
...
+        if os.path.exists(file_or_module):
+            file_or_module = os.path.abspath(file_or_module)
+        file_or_module = Importer('language file').import_module(file_or_module)
+
+        available = {}
+        for _, lang in inspect.getmembers(file_or_module, is_language):
+            available[normalize(lang.__name__)] = lang
+            if lang.__doc__:
+                available[normalize(lang.__doc__.splitlines()[0])] = lang
+        return available
```
pekkaklarck, line 124: What's the reason to return languages as a dict instead of a list? List would be easier for our code in `_get_languages` but is a dict better for external usage? Dict containing languages both by code (class name) and name (docstring) also feels a bit strange because in `_get_languages` we then add languages with both code and name twice.

If there's no strong reason to return a dict, I'd change this back to a list of instances. External users could then easily convert it to a dict if needed. Alternatively external users could first call this method and then `get_available_languages` that returns also languages imported here.

As a style issue, we haven't in general used empty lines inside methods or functions unless there are nested functions. If the code gets so long that empty lines are needed to split it to parts, we have added helper methods/functions instead.

utest/api/test_languages.py
```diff
@@ -102,6 +119,40 @@ def test_duplicates_are_not_added(self):
...
+        data = join(abspath(dirname(__file__)), 'orcish_languages.py')
+
+        languages = [(v.name, v.code) for v in Languages(data)]
+
+        self.assertIn(("OrcishLoud", "or-CLOU"), languages)
+        self.assertIn(("Orcish Quiet", "or-CQUI"), languages)
+        self.assertIn(("English", "en"), languages)
+
+
+
```
pekkaklarck, line 155: :+1: for comprehensive unit tests. Unnecessary empty lines should be removed, though.

src/robot/conf/languages.py
```diff
@@ -67,36 +71,57 @@ def _get_languages(self, languages):
...
                     and issubclass(member, Language)
                     and member is not Language)
-        if os.path.exists(lang):
-            lang = os.path.abspath(lang)
-        module = Importer('language file').import_module(lang)
-        return [value() for _, value in inspect.getmembers(module, is_language)]
+
+        if os.path.exists(file_or_module):
+            file_or_module = os.path.abspath(file_or_module)
+        file_or_module = Importer('language file').import_module(file_or_module)
```
pekkaklarck, line 117: Would `path_or_name` be a better name than `file_or_module`? `file` often refers to a open file object that we don't support. `module` could also refer to an actual module object when we actually except module name.

The value assigned above should be `module` because it is an actual imported module.

pekkaklarck (changes requested):
Looks very good. Apart from smallish style issues, the only thing I think that should be changed is the return value of `import_language_module`. Or do you have a strong reason for returning a dict of classes instead of a list of instances like earlier?

src/robot/conf/languages.py
```diff
@@ -55,48 +62,77 @@ def _add_language(self, lang):
...
             if isinstance(lang, Language):
                 returned.append(lang)
             else:
+                module, lang_part = lang.rsplit(":", 1) if ":" in lang else (None, lang)
+                if module is not None:
+                    try:
+                        available.update(self.import_languages_module(module))
+                        lang = lang_part
+                    except DataError:
+                        pass
```
pekkaklarck, line 79: What's this? Is the idea to support something like `ManyLangs.py:SomeLang` where `ManyLangs` would contain multiple languages and only `SomeLang` should be taken into use? I don't consider that too useful. You can easily have separate language modules for each language if needed. I'm not totally against adding such functionality, but it would then need separate acceptance tests and documentation and that's not worth the effort in my opinion. Let's stick with API enhancements and try to get this done so that we get rc 2 out soon ASAP.

src/robot/conf/languages.py
```diff
@@ -55,48 +62,77 @@ def _add_language(self, lang):
...
         available = {}
         for lang in Language.__subclasses__():
             available[normalize(lang.__name__)] = lang
             if lang.__doc__:
                 available[normalize(lang.__doc__.splitlines()[0])] = lang
         return available
 
-    def _import_languages(self, lang):
+    @staticmethod
+    def import_languages_module(path_or_name):
```
pekkaklarck, line 118: This should be `import_language_module` without the `s`.

src/robot/conf/languages.py
pekkaklarck: This issue is still open.

pekkaklarck (changes requested):

d-biehl:
`import_language_module` returns the same things like `get_available_languages` but only for a module. One entry for the code and one entry for the name of language from __doc__.

d-biehl:
Yes, it helps,  with that it is possible to use the same mechanism to select the possible language class that needs to be instanciated as if we use `get_available_languages`. Or To update the results from `import_language_module` with `import_language_module`
like this:

```python
langs = Languages.get_available_languages()
for m in lang_modules:
   langs.update(Languages.import_language_module(import_language_module)
```

pekkaklarck:
Is there a reason you need to use
```python
langs = Languages.get_available_languages()
for m in lang_modules:
   langs.update(Languages.import_language_module(m)
```
instead of
```python
for m in lang_modules:
   Languages.import_language_module(m)
langs = Languages.get_available_languages()
```
?

d-biehl:
Yes it is, because I want to be as close as possible to the robotframework implementation and the language argument is either a language name or a language code or module name. So I have to look in available languages if there is name or code otherwise I have to import the language and look there, if I found something, I can add the result of my available languages instead of calling get_available_languages again.

pekkaklarck:
So your use case is something like this:
```python
available = Languages.get_available_languages()
for lang in languages:
    if lang not in available:
        available.update(Languages.import_language_module(lang))
```
How's that better than this:
```python
available = Languages(languages).get_available_languages()
```

pekkaklarck:
It should also be decided should `get_available_languages` return language codes with or without possible hyphen. We have, for example `PtBr` that has code `pt-BR`. In the current implementation `_get_available_languages` doesn't use the hyphen which is fine in internal usage, but in a public method I believe the hyphen should be there. Similarly we need to decide are codes and names case and space normalized or not.

This makes me think should this method really be public or not. Perhaps it would be better if external users just used `list(Languages(...))` to get language instances and then used them as needed. If there's a need to check is a language available, we could then add something like `Languages.get_language()` that would accept language code, w/ or w/o hyphen, or name as an argument the same way as `Language.from_name()`. Not sure would `Languages.get_language()` be that useful when we already have `Language.from_name()`, though.

pekkaklarck:
The more I think about this, the less I see real needs to expose `get_available_languages` publicly when `list(Languages(...)` already works. Could you explain the use case for it? If it only simplifies some use case a little, and doesn't allow some new use case, then I think we should keep it private. I'm fairly certain we need to enhance `Languages` in future releases and the smaller public API it has, the easier it is to change.

I'm not sure is exposing `import_language_module` publicly needed either when just passing all languages to `Languages()` when you initialize it works as well. Not needing to recreate the instance if new language modules are to be loaded would be convenient, though, but the current implementation of `import_language_module` doesn't actually do that. I think it would be better to enhance the current `add_language` so that it works also with language modules. Do you have some concrete use cases where `import_language_module` would be better?

Controlling is English loaded or not is definitely a useful enhancement.

d-biehl:
Ok, example:

The user defines these `languages` (see code) in the settings of VSCode. I want to load all languages that i can load. Independent of whatever happens, wrong syntax, the pythonpath ist not correct or whatever, I want to load all Languages I can and for the languages that I cannot load RobotCode should give an error in the problems tab.

Here the code (it's copied from robotcode, but simplified to be runnable)

```
from typing import List
from robot.conf.languages import Languages
from robot.utils.normalizing import normalize

languages = ["fi", "pt-BR", "wr-ONG", "de", "CustomLang", "Chinese Traditional"]

available_languages = Languages.get_available_languages()
diagnostics: List[str] = []

found_languages: List[str] = []
for lang in languages:
    normalized_lang = normalize(lang, "-")

    language = available_languages.get(normalized_lang, None)
    if language is None:
        try:
            available_languages.update(Languages.import_language_module(lang))
            language = available_languages.get(normalized_lang, None)
        except DataError:
            pass
    if language is None:
        diagnostics.append(f"{lang} is not a valid language definition or language module.")
    else:
        found_languages.append(lang)

print(diagnostics)
print(found_languages)
```

If I only have:

```
available = Languages(languages).get_available_languages()
```

I get an exceptions after the first wrong language, And I know where in the list of languages is the error.

d-biehl:
And why I must create an instance of the `Language` class to get the available languages? 
The actual implemention of [_get_available_languages ](https://github.com/robotframework/robotframework/blob/ec2e07a8a903d6fcf536e0a5c1507102dff24a84/src/robot/conf/languages.py#L83)is defined as instance method, it defines `self` but did not use it.

pekkaklarck:
Thanks for the example, it's very useful. Based on it seems that just enhancing `Languages.add_language` so that it works also with language modules would work and make the code a lot simpler:
```python
from typing import List
from robot.api import Languages

languages = ["fi", "pt-BR", "wr-ONG", "de", "CustomLang", "Chinese Traditional"]
available = Languages()
diagnostics: List[str] = []

found_languages: List[str] = []
for lang in languages:
    try:
        available.add_language(lang)
    except ValueError as err:
        diagnostics.append(f"{lang} is not a valid language definition or language module.")

print(diagnostics)
print(list(available))
```
The above code already works otherwise, but custom language modules aren't supported. Enhancing `add_language` to support that is easy.

pekkaklarck:
You don't need to create an instance of `Language` or `Languages` to get available languages. You can simply call `Language.__subclassess__()` that's already public.

pekkaklarck:
Are there some other use case where making English optional and enhancing `add_language` as discussed above wouldn't be enough? If not, I'd say that's the approach to use. I already tested that enhancing `add_language` is easy:

```python
    def add_language(self, name):
        try:
            languages = [Language.from_name(name)]
        except ValueError:
            try:
                languages = self._import_languages(name)
            except DataError:
                # Should we include the error from importing to the error we raise?
                raise ValueError(f'Language {name} not found nor importable as a module.')
        for lang in languages:
            self._add_language(lang)
```

d-biehl:
Ok, see my comment in the issue ;-)

Even if I think (please do not be angry) we could do better ;-) Can I finalize the PR, but only tomorrow evening. If too late for you, ... let's talk tommorow.

pekkaklarck:
If enhancing `add_language` and not exposing new methods is fine for you, I can take your PR and do those changes. I can either continue with the PR or just merge it and continue from there. As I commented the issue, I'd like to keep the API small now because it needs to be enhanced in the future anyway. More convenience can be added later when the API is more stable.

If you have time, it would be awesome if you could look at #4497. I got an implementation for that ready and can commit if it is considered good enough.

d-biehl:
ok, implemented your add_language proposal, but move get_available_languages and import_language_module to the global namespace, so that these functions are not public api, but because in robotcode I use already the not public api it's my turn if something changed.
What about this. 

I missed 1 unittest will fix this later when I'm back from the doctor....

pekkaklarck:
No, I don't want these methods to be in the module scope as public functions either. That makes no sense from Robot's point of view and I don't want to make them public as a convenience for external tools because it adds maintenance burden for us. I've told that multiple times. My position is not going to change unless you have a concrete example where these methods are needed and don't only avoid you writing a line or two more code on RobotCode side.

This issue has now delayed RF 6.0 so long that I _really_ need to get it closed. Are you willing to revert your changes related to making these two methods public? If yes, and you can do that pretty soon today, I can still wait for that. Otherwise I can do the following API changes myself:

- Make English optional.
- Enhance `add_language` to work with language modules.

pekkaklarck (approved):

d-biehl:
You were too fast ;-) I just optimized a little bit and merged duplicate codes. And the add_language also allows to add a language instance.

Maybe we can merge that too. Do I have to open a new PR for this?

d-biehl:
Parser "needs" a ValueError thats why I leave it so. CommandLine Option needs a DataError to get the correct error text.

pekkaklarck:
I don't see the final changes you mentioned anywhere. Were there other new features than `add_language` accepting `Language` instances? I can add that myself. Since this PR is already merged, let's use the issue for discussion about possible changes.

d-biehl:
See here https://github.com/d-biehl/robotframework/commit/fa63db3c4db7383845f27600260b472b65c0cc76

Because the PR was closed, it is not shown there. I did not have the time this evening to make a new PR, but maybe you see what I did.
Mainly add_language does the same as get_languages. I extract the same parts to a new method.

pekkaklarck:
I added support for using `Language` instances with `add_language`. Other changes didn't seem to add new functionality.

PR #4507, opened 2022-10-12
Added Arabic

leeuwe:
provided by Haliem @Alpha-Centauri-00 
https://github.com/Alpha-Centauri-00


Alpha-Centauri-00:
Adding Arabic Language would need a huge changes, regarding to the right-to-left issue. which vscode and pycharm dont support. 
so writing in Arabic using Robot-framework, will be a challenge and hard to read. not to mention that log.html and report.html should also rebuild.
In addition, English is the second language in most Arab countries. and everyone coding in English
that's why Arabic in this contest is not necessary.

leeuwe:
As commented by @Alpha-Centauri-00 and in allignment with discussion on Slack, for now we'll close this PR, as more work is needed. This is currently out of scope.

pekkaklarck:
I agree adding support for right-to-left languages doesn't make much sense until these other issues are resolved. I hope we could resolve those issues, though!

PR #4508, opened 2022-10-13
Correct the compound words for German.

d-biehl:
Some German words are not correctly written together, I have corrected that. @Snooz82, is that ok for you?


Snooz82 (approved):

Snooz82:
Approved

PR #4526, opened 2022-11-02
Add Polish Booleans

leeuwe:
Boolean translations for polish by Jerzy Głowacki


pekkaklarck:
Unfortunately I missed these so that they didn't make it to RF 6.0.1. Will be added in 6.0.2 (or 6.1 or 7.0, whichever the next release will be).

PR #4556, opened 2022-12-06
add a newline to start_test output message

TrN003:
`robot.output.console.verbose.start_test` should only output the name and docs to the console. However, it does not put a newline at the end of the line, so the first use of `Log To Console` lands on that line as well.

robot version: 5 and up
python version: 3.10.4
OS: Windows 10 Enterprise

Given the test cases:
Screenshot 2022-12-06 083907

the output looks like this:
Screenshot 2022-12-06 083840

The proposed change makes it look like this:
Screenshot 2022-12-06 083618

 believe this looks cleaner and is what a user would expect of logging.
The `--dotted` and `--quiet` options are unaffected.

The downside of this change is that tests that don't log to console print the name of the test twice:
Screenshot 2022-12-06 091538

Fixing that would take more invasive changes.

### DIFF
--- a/src/robot/output/console/verbose.py
+++ b/src/robot/output/console/verbose.py
@@ -46,6 +46,7 @@ class VerboseOutput:
 
     def start_test(self, test):
         self._writer.info(test.name, test.doc)
+        self._writer.stdout('\n')
         self._running_test = True
 
     def end_test(self, test):


pekkaklarck:
It's true that messages logged to console don't work well with the default console output. Just adding a newline has the problem, also mentioned above, that test names are there twice. In my opinion that is a lot bigger issue than the problem with logged messages so this isn't a good fix. Most likely it would be best to design the whole console output from scratch so that logged messages are taken into account.

PR #4561, opened 2022-12-11
Possibility to set a custom message when WHILE limit is reached

asaout:
Hello everyone,

I have been using the WHILE loop since the release of RF 5.0 and I was thinking that it would be interesting to let users set a custom message when the limit is reached.

As a result, I worked on that enhancement in that pull request. The idea is to add a third argument named `limit_exceed_message` to the WHILE loop. This new argument must be placed after the `limit` argument. If not, the error `Second WHILE loop argument must be 'limit', got ...` is raised.

Here is an example :

```
Default behaviour
    WHILE    True    limit=3
        Log     Test
    END

Behaviour when limit_exceed_message is set
    WHILE    True    limit=3    limit_exceed_message=Custom error message
        Log     Test
    END
```
image

I also added acceptance and unit tests corresponding to this enhancement. I also added a paragraph in the "Limiting WHILE loop iterations" section of the user guide.

Do you think this functionality is interesting ? If yes, what do you think I should improve/change ? Maybe changing the name of the new argument (as I'm not fluent in English) ?

Thank you very much for taking the time to read this pull request.


pekkaklarck:
Interesting idea and on a _very_ quick look the PR also looks really good.

Could you submit an issue about this functionality? That would be a good place to discuss the feature itself including what the option name should be. If we agree this is a good idea, we can then continue more technical discussion as part of this PR.

If we add this, we need think are there other similar features that we could add. We don't need to add everything in one go, but the design should be extendible. One feature I have been thinking is making it possible to control what to do if the limit is reached. In same cases it could be convenient if execution wouldn't fail but instead would continue after the loop in that case. Anyway, that discussion don't belong here and such functionality would need its own issue.

pekkaklarck:
Sorry for review taking time @asaout. We've been busy with other RF 6.1 issues, but now its time to look at the WHILE related issues. We are planning to create RF 6.1 alpha soon and after that @yanne will concentrate on WHILE.

Although @yanne will take the main responsibility related to WHILE issues, I'll review this PR now. I already noticed there are several conflicts due to changes to related code, but I would expect them to be pretty easy to resolve. After a quick initial look this PR looks really good, but there are some things to be enhanced still. If you @asaout are still interested to help after this long delay, let us know and we'll give you time to make changes and can review this again. Otherwise @yanne or I can finish this issue.

src/robot/parsing/model/statements.py
```diff
@@ -968,15 +968,20 @@ class WhileHeader(Statement):
...
                   Token(Token.SEPARATOR, separator),
                   Token(Token.ARGUMENT, condition)]
         if limit:
             tokens.extend([Token(Token.SEPARATOR, indent),
                            Token(Token.OPTION, f'limit={limit}')])
+        if on_limit_message:
+            tokens.extend([Token(Token.SEPARATOR, indent),
+                           Token(Token.OPTION,
+                                 f'on_limit_message={on_limit_message}'
+                                 )])
```
pekkaklarck, line 984: Closing parens should be at the end of the previous line for consistency with other code.

src/robot/parsing/model/statements.py
```diff
@@ -986,17 +991,30 @@ def condition(self):
 
     @property
     def limit(self):
-        value = self.get_value(Token.OPTION)
-        return value[len('limit='):] if value else None
+        values = self.get_values(Token.OPTION)
+        for value in values:
+            if value.startswith('limit='):
+                return value[len('limit='):]
+        return None
```
pekkaklarck, line 998: I added `Statement.get_option()` to ease getting options with `FOR IN ZIP` loop and updated this code as well so that it is nowadays just:
```python
@property
def limit(self):
    return self.get_option('limit')
```
The change has obviously caused a conflict, sorry for that, but the good news is that the `on_limit_message` property can be implemented using the same approach.

src/robot/parsing/model/statements.py
```diff
@@ -986,17 +991,30 @@ def condition(self):
...
         values = self.get_values(Token.ARGUMENT)
         if len(values) == 0:
             self.errors += ('WHILE must have a condition.',)
-        if len(values) == 2:
-            self.errors += (f"Second WHILE loop argument must be 'limit', "
+        if len(values) == 2 or len(values) == 3:
+            self.errors += (
+                            f"WHILE loop arguments must be 'limit' "
+                            f"or 'on_limit_message', "
                             f"got '{values[1]}'.",)
```
pekkaklarck, line 1016: Reporting all possible options here if WHILE has more than one condition, or an invalid option, can get a bit complicated. For example, the above error message shows only the second argument, not possible third. I'm not sure how much effort should be spent on this. I'd be fine with something simple like below, but a better error listing all possible options certainly would have benefits.

```python
if len(values) > 1:
    self.errors +) ('WHILE cannot have more than one condition, got {seq2str(values)}.',)
```

src/robot/running/bodyrunner.py
```diff
@@ -341,14 +341,20 @@ def run(self, data):
         error = None
         run = False
         limit = None
-        loop_result = WhileResult(data.condition, data.limit, starttime=get_timestamp())
+        loop_result = WhileResult(data.condition, data.limit,
+                                  data.on_limit_message,
+                                  starttime=get_timestamp()
+                                  )
```
pekkaklarck, line 347: Closing paren should be on the same line as the last item for consistency.

src/robot/running/bodyrunner.py
```diff
@@ -596,9 +602,14 @@ def _run_finally(self, data, run):
...
-    def create(cls, limit, variables):
+    def create(cls, limit, on_limit_message, variables):
         if not limit:
-            return IterationCountLimit(DEFAULT_WHILE_LIMIT)
+            return IterationCountLimit(DEFAULT_WHILE_LIMIT,
+                                       on_limit_message
+                                       )
+        if on_limit_message:
+            on_limit_message = variables.replace_string(
+                on_limit_message)
```
pekkaklarck, line 612: Variables should be replaced in the beginning of `create`. Now the `IterationCountLimit` returned above gets wrong message.

src/robot/running/bodyrunner.py
```diff
@@ -610,18 +621,23 @@ def create(cls, limit, variables):
...
             raise DataError(f'Invalid WHILE loop limit: {err.args[0]}')
         else:
-            return DurationLimit(secs)
+            return DurationLimit(secs, on_limit_message)
 
-    def limit_exceeded(self):
-        raise ExecutionFailed(f"WHILE loop was aborted because it did not finish "
-                              f"within the limit of {self}. Use the 'limit' argument "
-                              f"to increase or remove the limit if needed.")
+    def limit_exceeded(self, on_limit_message):
```
pekkaklarck, line 632: Why is the message passed here? Couldn't it be accessed like `self.on_limit_message`?

src/robot/running/bodyrunner.py
```diff
@@ -596,9 +602,14 @@ def _run_finally(self, data, run):
 class WhileLimit:
```
pekkaklarck, line 602: Should add 
```python
def __init__(self, on_limit_message=None):
    self.on_limit_message = on_limit_message
```
and then in child class `__init__`s use `super().__init__(on_limit_message)`.

pekkaklarck (changes requested):
Great work! Sorry again for late review and for conflicts caused by the wait. If you are still interested to finalize this, please take a look at the comments. Otherwise we can continue from here.

asaout:
Hello @yanne @pekkaklarck, 

Thank you for the review ! 
I resolved the conflicts and tried to add the requested changes in my last commit.

yanne:
Manually merged, this is now in master

Issue #4595, opened 2023-01-13
Add ${KEYWORD NAME} to BuildIn variables


Obluchatel:
There are ${TEST NAME} and ${SUITE NAME} variables in BuiltIn library, but there is no ${KEYWORD NAME} variable in BuiltIn library. 

It would be very handy to have it, because currently you need to use CustomLibrary to assign it:
Example:
https://stackoverflow.com/questions/42508905/find-keyword-name-or-keyword-name-stack-in-robot-framework

Implementation is quite simple, since context.py already has ${TEST NAME} declaration:
    def start_test(self, test):
        self.test = test
        self._add_timeout(test.timeout)
        self.namespace.start_test()
        self.variables.set_test('${TEST_NAME}', test.name)
        self.variables.set_test('${TEST_DOCUMENTATION}', test.doc)
        self.variables.set_test('@{TEST_TAGS}', list(test.tags))
        
The same can be done for ${KEYWORD NAME}, something like:
    def start_keyword(self, keyword):
        self._started_keywords += 1
        **self.variables.set_keyword('${KEYWORD_NAME}', keyword.name)**
        ...
        
That variable will be very useful for [Teardown] e.g. Log   ${KEYWORD_NAME} ${result}   console=True

pekkaklarck:
I agree this would be a nice feature and this has also been requested already long time ago. We have issue #1042 about it but it was closed for two reasons:
- There was no activity for very long time.
- If `${KEYWORD NAME}` is implemented the same way as `${TEST NAME}` and others, it would override possible variables with that name which would be very badly backwards incompatible.

This new issue demonstrates there's interest again, but the second point is still valid. `${KEYWORD NAME]` is such a common and generic name that there certainly are tests/tasks using it. Even if there weren't, anyone using it in the future would get the variable overwritten when a keyword starts.

This should be implemented so that `${KEYWORD NAME}` exists in the global variable namespace and is looked from there only after looking variables first from local, test and suite namespaces. Unfortunately our variable namespaces are implemented stupidly so that each lower level namespace is copied from the higher level namespace and thus we only look from one namespace. If we add `${KEYWORD NAME}` to that namespace, it overrides possible earlier value.

In practice this cannot be implemented until variable namespaces are implemented better. That's something I think would be a good idea to do anyway, so I leave this issue open.

sebastianciupinski:
can be tracked and implemented with https://github.com/robotframework/robotframework/issues/4140

PR #4597, opened 2023-01-13
Added ${KEYWORD_NAME} variable 

Obluchatel:
Implemented enhancement described here:
https://github.com/robotframework/robotframework/issues/4595


pekkaklarck:
Having `${KEYWORD NAME}` would be nice, but this implementation would be so badly backwards incompatible it's not possible. For more details see https://github.com/robotframework/robotframework/issues/4595#issuecomment-1383123186. Notice also that a PR would need tests and documentation.

Issue #2535, opened 2017-01-31
Support `--argumentfile` case-insensitively and document that shortening it does not work


pekkaklarck:
While implementing #2420, I noticed that `--argumentfile` option doesn't work case-insensitively like `--ArgumentFile` similarly as other long options do. Additionally using shorter but unique variants like `--argumentf` don't work either. Both of these limitations are due to `--argumentfile` being handled using its own logic before using Python's `getopt` that actually handles argument parsing. Case-insensitivity support is pretty easy to add to our own code, but handling shortened variants is not. My plan is to enhance this situation as follows:

- Support `--ArgumentFile` style.
- Document in the User Guide that `--argumentf` style doesn't work.
- If possible/easy, explicitly check that `--argumentf`style wasn't used.

pekkaklarck:
Tasks listed in the original description done by the commits above. This issue is done.

PR #4608, opened 2023-01-17
Support long command line options with hyphens like --pre-run-modifier

### LINKED ISSUE #4547: Support long command line options with hyphens like `--pre-run-modifier`
To make long option names like `--prerunmodifier` easier to read, Robot handles them case-insensitively so that also `--PreRunModifier` works. This is an old feature and works nicely, but it's inconsistent with the `--my-long-option` format other tools typically use nowadays. We could easily support also that format simply by removing hyphens from long option values at the same time as we convert them to lower case.

Removing hyphens would make `--prerunmodifier`, `--PreRunModifier` and `--pre-run-modifier` all equivalent. It would also allow stupid usages like `--p-r-eru-n-mod-i-fier`, but we support also `--PReruNmodIfier` so it shouldn't be a big deal. The leading two hyphens obviously shouldn't be removed.

Supporting multiple different formats is perhaps not a good idea in general, though, and it could be better to require using a single format instead. I personally prefer the `--my-long-option` format over the other alternatives because it is consistent with shorter options like `--option` and also what other tools typically use. I'm not sure is requiring that worth the deprecation effort, but making that format valid now certainly helps if we ever want to do that.

Vincema:
Hello everyone,

This PR fixes #4547.

It adds support for the usual CLI options as well as the special "--argumentfile" option (#2535).

I added a few unit tests and edited some existing acceptance tests.
I also updated the user guide to briefly describe the new behavior.

Thank you in advance for taking the time to review this PR, and please let me know if I need to fix anything.

### DIFF
--- a/atest/robot/cli/console/colors_and_width.robot
+++ b/atest/robot/cli/console/colors_and_width.robot
@@ -17,7 +17,7 @@ Console Colors On
     Outputs should have ANSI colors when not on Windows
 
 Console Colors ANSI
-    Run Tests With Colors    --ConsoleColors AnSi
+    Run Tests With Colors    --Console-Colors AnSi
     Outputs should have ANSI colors
 
 Invalid Console Colors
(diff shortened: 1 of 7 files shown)


pekkaklarck:
Thanks for an awesome PR! Code, tests and docs all look great. Great that you noticed that argument files need special attention and also added tests for them. The only missing thing is a note in the UG that this is new feature in RF 6.1 but I can add it myself. I'll also check should this be mentioned in the `--help` text.

PR #4612, opened 2023-01-18
Zipapp

franzhaas:
Hi,

this is the pull request for the technique to create zipapps with robotframework.

I left the pyproject.toml, and the pdm.lock in the repository, as they are helpful to me, but I would not be surprised if they are not useful to anyone else and need to be removed...

There were very few changes necessary to get this feature working. Thus, the way it is implemented at the moment is not very intrusive, making the code unnecessarily complex.

Thanks to Carmen.

Best regards,
Franz

PS
The importlib.resources.open_text function used in this improvement was introduced in python 3.7. But as there is a backport available for python 3.6, i believe this is still a reasonable solution.

unfortunately this function got depricated in python 3.11, but the suggested alternative did not work for me on python 3.10... I still believe this to be a reasonable improvement.

PPS
I just pushed a small change to accommodate pre importlib.resources pythons (3.6, 3.7 maybe...) I failed to find the point to add the requirement for "importlib-resources" for python 3.6, and maybe also 3.7, I currently don't have access to python 3.7 and 3.11 to check. Also I haven't checked on anything other than windows.


franzhaas:
@pekkaklarck I believe i addressed all the things we discussed, could you give it another look?

src/robot/htmldata/template.py
```diff
@@ -12,18 +12,47 @@
...
+            idx -= 1
+        else:
+            idx += 1
 
-    def __iter__(self):
-        with open(self._path, encoding='UTF-8') as file:
-            for line in file:
-                yield line.rstrip()
+    parts = ("." + item.replace(".", "") for item in parts)
+    modulepart = "robot.htmldata" + "".join(parts)
```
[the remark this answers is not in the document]
pekkaklarck: Is there a reason why this has been changed from a class to a function?

pekkaklarck (changes requested):
Looks good but some things still to be done:
1. Cleanup as instructed in individual comments.
2. Add copyrights to added `__init__.py` files.
3. Remove `src/robot/htmldata/testdata/__init__.py` unless it's really needed. This directory ought to be only used by tests and shouldn't be even included in the installation package.
4. Add something about zipapps to the installation instructions. As we discussed earlier, more detailed documentation can be in a separate project.

src/robot/htmldata/template.py
```diff
@@ -12,18 +12,47 @@
...
+    with warnings.catch_warnings():
+        # This is necessay as open_text is deprecated started from python 
+        # 3.11, but the alternatives given did create issues in zipapps...
```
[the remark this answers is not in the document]
franzhaas: This is true, the problem is that this method fails in zipapps for python version 3.9. And does not exist before that.

I found a solution for the medium term here, which does not resort to deprecated code. And uses importlibresources backports for zipapps in python 3.9 and earlier.

src/robot/htmldata/template.py
```diff
@@ -12,18 +12,47 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-
+import pathlib
 import os
-from os.path import abspath, dirname, join, normpath
+import warnings
```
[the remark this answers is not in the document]
franzhaas: ok, donne

src/robot/htmldata/template.py
```diff
@@ -12,18 +12,47 @@
...
+        # This code is here to optionally suport zipapps with python 3.6,
+        # under the condition that the backport importlib-resources is
+        # installed.
+        from importlib_resources import open_text
+    except:
+        # use our own...
 
+        def open_text(modulepath, resource_part, encoding='utf-8'):
+            base_dir = pathlib.Path(__file__).parent.parent.parent
+            resource_path = base_dir / pathlib.Path(modulepath.replace(".", os.sep)) / pathlib.Path(resource_part)
```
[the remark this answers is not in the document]
franzhaas: I changed this to use nondepricated functions instead...

src/robot/pythonpathsetter.py
```diff
@@ -13,7 +13,11 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-"""Module that adds directories needed by Robot to sys.path when imported."""
+"""Module that adds directories needed by Robot to sys.path when imported.
+
+This is suposed to run if the robotframework is running but the robotframework is not in the module search path...
+This should not execute in a zipapp.
+"""
```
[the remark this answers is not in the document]
franzhaas: Well i left this inforamtion, as it was helpful for me... i tried to generalize it...

src/robot/htmldata/template.py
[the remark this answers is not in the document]
franzhaas: Thanks for the tip, yes these function calls help. 

@moving from class to function well, I had the impression that this piece of code looks and feels and is used like a function, and therefore should be a function.

But no hard fact arguments, and no strong position.

INSTALL.rst
```diff
@@ -320,3 +320,24 @@ __ https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtua
...
+--------------------
+Zipapps (https://docs.python.org/3/library/zipapp.html) are technique to distribute all the python code of  asolution in a single file, which can be executed using a python interpreter. The same zipapp file can be run on multiple plattforms. An example of using pdm 
+(https://pdm.fming.dev/latest/) with the packer extension (https://github.com/frostming/pdm-packer) to create a zipapp would be.
+
+.. sourcecode:: bash
+
+    $ pdm init 
+    $ pdm add robotframework
+    $ #If the target is python 3.6: pdm add importlib_resources         
+    $ pdm pack -m robot:run_cli
```
pekkaklarck, line 337: Is `pdm` needed to be able to use zipapps for anything useful? If not, I'd prefer an example using just the standard `zipapp` here. There could be a separate example using `pdm`, but perhaps an external project with more documentation that could be referenced here would be even better.

INSTALL.rst
```diff
@@ -320,3 +320,24 @@ __ https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtua
...
+
+At this point you have created a zipapp. This zipapp can be uesed like this.:
+
+.. sourcecode:: bash
+
+    $ python zipappname.pyz example.robot
```
pekkaklarck, line 343: Does the above `pdm` example really create `zipappname.pyz`? If yes, it would be good to be explicitly mentioned.

src/robot/htmldata/lib/__init__.py
pekkaklarck, line 1: Copyrights needed here and in other newly added `__init__.py` files.

src/robot/htmldata/template.py
```diff
@@ -14,16 +14,29 @@
...
+import pathlib
+import sys
 
+if sys.version_info < (3, 10) and not os.path.exists(__file__):
+    # Try importlib resources backport as prior to python 3.10
+    # importlib.resources.files was not zipapp compatible...
+    try:
+        from importlib_resources import files
+    except ImportError:
+        raise ImportError("robotframework outside of filesystem (zipapp?) requires importlib resources backport on python < 3.10")
```
pekkaklarck, line 26: Style issues:
- Too long line. Our convention is max 88 characters.
- The message is somewhat confusing. You could try rephrasing it, but I can update it later as well.

src/robot/htmldata/template.py
```diff
@@ -14,16 +14,29 @@
 #  limitations under the License.
 
 import os
-from os.path import abspath, dirname, join, normpath
+import pathlib
+import sys
 
+if sys.version_info < (3, 10) and not os.path.exists(__file__):
```
pekkaklarck, line 20: Since we are using `pathlib` in this module, `Path(__file__).exists()` would be more consistent here. Alternatively could just use `os.path` everywhere.

src/robot/htmldata/template.py
```diff
@@ -14,16 +14,29 @@
...
-        with open(self._path, encoding='UTF-8') as file:
-            for line in file:
-                yield line.rstrip()
+def HtmlTemplate(filename):
+    module, filename = os.path.split(os.path.normpath(filename))
+    module = 'robot.htmldata.' + module
+    
+    with files(module).joinpath(filename).open('r', encoding="utf-8") as f:
+        for item in f:
+            yield item.rstrip()
```
pekkaklarck, line 42: This looks good otherwise, but I still don't understand what was the reason to change this from a class to a function. In the current usage there's no difference, but having a function named in `CamelCase` is always strange. The function could obviously be renamed, but a class works here as well and allows further configuration in the future if needed. Unless you have concrete reasons for the change, please change this back to a class.

src/robot/pythonpathsetter.py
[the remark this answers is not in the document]
pekkaklarck: The explanation is still a bit confusing but arguably better than the original. The real reason for this module is that it allows executing Robot like `python path/to/robot/run.py`. That's not too relevant nowadays when you can also run a directory like `python path/to/robot`, but this is still needed with e.g. with `python path/to/robot/libdoc.py`. When Robot is used normally, this module isn't imported. I can enhance the docs later to make this more explicit.

utest/htmldata/test_htmltemplate.py
```diff
@@ -17,7 +17,17 @@ def test_lines_do_not_have_line_breaks(self):
...
+        def check_for_correct_exception(args):
+            try:
+                list(args)
+            except IOError:
+                pass
+            except ModuleNotFoundError:
+                pass
+            except Exception as e:
+                fail("unexpcected Exception: " + str(e))
+
```
pekkaklarck, line 29: Use `assert_raises((ImportError, IOError), ...)` instead.

pekkaklarck (changes requested):
Looks good but little tuning still needed. The biggest task is probably documentation.

utest/htmldata/test_htmltemplate.py
franzhaas: donne

src/robot/htmldata/template.py
franzhaas: donne

src/robot/htmldata/template.py
franzhaas: I tried to express that this applies to all environments where robotframework is not installed on a file system. zipapps, pyoxidizer etc... I tried to improve, please take over if it is still not good enough.

INSTALL.rst
franzhaas: No pdm is not needed, but it is the tool I use on a daily basis for a reason....

venv should also work, also shiv or pex...  poetry I would need to ask around... but completely without tool support zipapps are not practical. I am happy if we just state the option and link to marketsquare....

pekkaklarck (approved):
Looks good. Some doc tuning needed but I can do that myself after merging and testing how zipapps work in general.

Issue #1283, opened 2014-06-29
Possibility to use custom parser for test data


spooning:
> _Originally submitted to [Google Code](http://code.google.com/p/robotframework/issues/detail?id=1283) by jussi.vastimo on 21 Nov 2012_

We have many test cases defined for different framework. test cases are defined via xml file. 

I have made my own reader and added it to populators READERS dictionary. This works for me. Is it possible add this kind of parameter for Framework class?
from .xmlreader import XmlReader

READERS = {'html': HtmlReader, 'htm': HtmlReader, 'xhtml': HtmlReader,
           'tsv': TsvReader , 'rst': RestReader, 'rest': RestReader,
           'txt': TxtReader, 'xml': XmlReader}

rjsmith:
Hi,
Is there any prospect of this ever getting added to the robot framework core? The hard work has already been done for you by Chris Prinos ...... I would really find it useful so that I can run RF tests directly from Microsoft Office files (using my bolt-on docx and xlsx parser)

pekkaklarck:
The situation with this is very much open. On one hand we can see that supporting new test data formats would be useful, but on the other hand they do add maintenance burden and can also limit syntax choices in the future.

I would prefer alternative formats to be supported via an interface so that they could be developed independently similarly as test libraries, listeners and other tools. Obviously you can already now convert data in any format to format supported by Robot before running tests, but I would prefer it to be possible to use a programmatic API. Ideally you should be able to specify something like `--dataparser xsl:ExcelParser` to register a custom parser for `.xsl` files.

A twist in this planning is that we have been thinking could we in RF 3.0 drop HTML test data support and support only the plain text format (+ TSV which is nearly 100% compatible with it). Having a way to register custom parsers would make that change less backwards-incompatible and thus more doable.

Both supporting alternative parsers and possibly dropping built-in support for HTML format are something we need to ask community opinion about. Sending an email about the latter to robotframework-users is already on my to-do-after-RF-2.9-alpha-1 list and I can ask opinions about the former at the same time. We obviously can continue discussion here already before that.

rjsmith:
Hi,
Thanks for replying, much appreciated!
[...]

PR #4614, opened 2023-01-18
Custom parsers

d-biehl:
This is the result of our RoboCon sprint to define and implement the custom parser API see #1283

What's missing at moment are the unit test cases for that API.

We need to talk about the Documentation statement value property. Can call the the from_params with a mutliline text with 2 paragraphs, but the value property deletes `\n' characters, so I get only one paragraph. Is this correct?


pekkaklarck:
Thanks, I hopefully have time to look at this during the weekend or latest next week. Some quick comments already now:

- I didn't fully understand the issue with `Documentation.from_params` but there certainly can be bugs. That probably deserves its own issue.
- As we discussed at the sprints, the parser API itself will be relatively simple with just few methods like `def parse_suite_file(path: Path, defaults: Defaults)`. That then makes our `Default` object part of the public API and it definitely needs some cleanup. I can take care of that myself.
- Same tests related to reStructuredText syntax are failing on CI. Most likely the parser changes have affected the reST parser. Can you take a look?

d-biehl:
created the issue #4670 to fix the Documentation thing independently.

next step is finding out how custom resource files can be parsed. Seems a little bit complication because they where imported by a static IMPORTER class in the namespace. @pekkaklarck do you have a quick idea?

src/robot/running/builder/parsers.py
```diff
@@ -34,12 +38,25 @@ def parse_suite_file(self, source: Path, defaults: Defaults = None):
...
+    @property
+    def extensions(self):
+        raise NotImplementedError
+
+    @property
+    def resource_extensions(self):
+        raise NotImplementedError
+
+    def _get_curdir(self, source):
+        return str(source.parent).replace('\\', '\\\\') if self.process_curdir else None
```
pekkaklarck, line 50: If we want this to be part of the public API, it shouldn't have `_`.

pekkaklarck:
I commented this PR already some time ago and submitted some design decision question to the [original issue](https://github.com/robotframework/robotframework/issues/1283#issuecomment-1508212701) a week ago. There were no comments from anyone, so I decided to go with the design I had proposed. The initial implementation is now ready (d973fb86b34f7290ccc2377889415698d527f3c6) and it makes this PR obsolete. There's still some [work to do before this feature is ready](https://github.com/robotframework/robotframework/issues/1283#issuecomment-1517705142), but I plan to get all that done early next week. If you @d-biehl have a change to test the new API it would be great!

PR #4623, opened 2023-01-27
Only return the source as the directory if it's actually a directory. Fixes #4607

### LINKED ISSUE #4607: `robot.running.model.Import` using wrong `directory` if `source` is set to a non-existing file.
When the `source` for a `robot.running.model.Import` is set internally to a non existing file, the `directory` considered is returning the `source` unchanged, which isn't correct.

It should always use the dirname.

The background is that the interactive console creates in-memory structure (so that the proper cwd would be expected but with a non-existing file internally) and this is now failing because it considers the directory the same as the source.

fabioz:


fabioz:
This should fix the breakage I have...

The old code was:

```
        if not self.source:
            return None
        if os.path.isdir(self.source):
            return self.source
        return os.path.dirname(self.source)
```

and the behavior was changed to check if the file exists to return the dirname, but that doesn't seem right (and fails when the source is pointing to a non-existing file as expected just to set the context).

@pekkaklarck if you could merge this soon I'd really appreciate it (because I had to disable my own tests with the robot framework master due to this regression and I'd really like to turn those back on).

src/robot/running/model.py
```diff
@@ -743,7 +743,12 @@ def source(self) -> Path:
     @property
     def directory(self) -> Path:
         source = self.source
-        return source.parent if source and source.is_file() else source
+        if not source:
+            return source
+        
+        if source.is_dir():
+            return source
+        return source.parent
```
pekkaklarck, line 751: We don't typically use empty lines between functions and for consistency reasons it would be good to remove the empty line here as well. Separate `if`s are fine for me, but it would also be possible to write this as a ternary `if/else` like the original code:
```python
return source if not source or source.is_dir() else source.parent
```

pekkaklarck (changes requested):
Changing the behavior to match the old functionality makes sense. No need to make backwards incompatible changes without any benefits. A unit tests should be added to make sure this isn't changed back again, though.

PR #4629, opened 2023-02-01
(🎁) Support only vararg in custom converters

KotlinIsland:
This allows registering a custom converter that only has vararg arguments

### DIFF
--- a/atest/testdata/keywords/type_conversion/custom_converters.robot
+++ b/atest/testdata/keywords/type_conversion/custom_converters.robot
@@ -69,6 +69,9 @@ Failing conversion
     Conversion should fail    Strict    wrong type
     ...    type=Strict    error=TypeError: Only Strict instances are accepted, got string.
 
+Only vararg
+    Only var arg    10    10
+
 With library as argument to converter
     String    ${123}
 
--- a/atest/robot/keywords/type_conversion/custom_converters.robot
+++ b/atest/robot/keywords/type_conversion/custom_converters.robot
@@ -33,6 +33,9 @@ Failing conversion
 `None` as strict converter
     Check Test Case    ${TESTNAME}
 
+Only vararg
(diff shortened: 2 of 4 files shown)


atest/testdata/keywords/type_conversion/custom_converters.robot
```diff
@@ -69,6 +69,9 @@ Failing conversion
     Conversion should fail    Strict    wrong type
     ...    type=Strict    error=TypeError: Only Strict instances are accepted, got string.
 
+Only vararg
+    Only var arg    10    ${10}
```
pekkaklarck, line 73: This test needs a counterpart under `atest/robot`. See `atest/README.rst` for details.

src/robot/running/arguments/argumentspec.py
```diff
@@ -55,6 +55,10 @@ def minargs(self):
     def maxargs(self):
         return len(self.positional) if not self.var_positional else sys.maxsize
 
+    @property
+    def first_positional(self):
+        return self.positional and self.positional[0] or self.var_positional or None
```
pekkaklarck, line 60: This property is somewhat confusingly named and doesn't feel too useful in general. Better to remove it and inline this code to the only place where it's used.

pekkaklarck (changes requested):
Separate test is needed to validate that if a converter accepts varargs, the second argument is the library itself. For other changes see the inline comments.

atest/testdata/keywords/type_conversion/custom_converters.robot
KotlinIsland: Thanks, done!

src/robot/running/arguments/argumentspec.py
KotlinIsland: Thanks, done!

PR #4641, opened 2023-02-05
(🎁) deprecate python3.7

### LINKED ISSUE #4637: Deprecate Python 3.7
We have already deprecated Python 3.6 (#4295) and plan to remove support for it in RF 7.0 (#4294). Python 3.7 will reach its end-of-life in [June 2023](https://peps.python.org/pep-0537/) and it's clear RF 7.0 won't be released before that. In this situation it makes sense to drop also Python 3.7 support in RF 7.0 and thus we need to deprecate it now.

Similarly as with Python 3.6 deprecation, it doesn't make sense to emit deprecation warnings or make any other code changes. We only need to make sure info about supported versions in the README and in the UG is correct and mention this in RF 6.1 release notes.

KotlinIsland:
Does this need to wait for master to reach 7.0?

- resolves #4637

### DIFF
--- a/.github/workflows/acceptance_tests_cpython.yml
+++ b/.github/workflows/acceptance_tests_cpython.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ 'ubuntu-latest', 'windows-latest' ]
-        python-version: [ '3.7', '3.8', '3.9', '3.10', '3.11', 'pypy-3.8' ]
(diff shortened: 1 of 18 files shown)


pekkaklarck:
The idea of  #4637 is just to explicitly mention that RF 7 won't support Python 3.7 anymore. RF 6.x will still support it so there shouldn't be any code changes, changes to CI, etc. This PR seems to be making such changes and cannot thus be merged right now. We could keep it open until we actual drop Python 3.7 support, but I'm afraid there are so much changes before that resolving conflicts won't make sense. Should we just close this instead?

PR #4643, opened 2023-02-05
(🎁) Add an option for strict type conversions

### LINKED ISSUE #4060: no error when passing wrong type to keyword that takes a bool
```robot
*** Settings ***
Library           Sandpit.foo

*** Test Cases ***
test adsf
    foo   asdfasdf
```
```py
from robot.api import deco

@deco.keyword
def foo(value: bool):
    ...
```

the dry run doesn't pick this up, and at runtime the value doesn't get converted to a bool either:
image

### LINKED ISSUE #4628: Support registering default argument converter
I would expect the following code to report an error or warning of some kind, given that the data being passed to this function is completely invalid.

```py
def foo(a: Callable[[int], str]):
    print(a(1).upper())
```
```robot
foo    hi
```

KotlinIsland:
The code changes in this suck hard, so I really need some better direction in order to make am acceptable PR.

- Resolves #4628
- Resolves #4060

Also makes `bool` conversion strict.

### DIFF
--- a/src/robot/conf/settings.py
+++ b/src/robot/conf/settings.py
@@ -478,7 +478,8 @@ class RobotSettings(_BaseSettings):
                        'ConsoleWidth'       : ('consolewidth', 78),
                        'ConsoleMarkers'     : ('consolemarkers', 'AUTO'),
                        'DebugFile'          : ('debugfile', None),
-                       'Language'           : ('language', [])}
+                       'Language'           : ('language', []),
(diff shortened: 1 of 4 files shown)


pekkaklarck:
As I commented #4628, I don't like the idea to make this configurable from the command line at all. I found it much better to be able to specify the behavior per library. Also, as I commented #4060, Boolean conversion absolutely won't be made strict by default.

PR #4661, opened 2023-02-17
4583 custom suite name

otemek:
My proposal of implementation for suite names.
Any questions, ideas, or remarks are welcome as it's my first contribution to the RF project.


pekkaklarck (changes requested):
On a quick look this looks really promising. I need to take a bit closer look at the code changes and added tests later, but I already noticed these things that need to be addressed:

- The setting name should be just `Name`, not `Suite Name`, to be consistent with `Documentation` and `Metadata` that are suite specific as well. It won't then be consistent with `Suite Setup/Teardown`, but they need the `Suite` prefix to separate them from `Test Setup/Teardown`.
- The setting needs to be translateable. This requires adding `name_setting = None` to the `Language` base class in the `robot.conf.languages` module and English translation to the `En` class. You can add Finnish translation `Nimi` there as well.
- This functionality needs to be documented in the User Guide. This can be done separately after merging this PR as well.
-

otemek:
Thanks for the help and I hope I understood it correctly because I've  left the token name SUITE_NAME and changed only the value for it as there is already a token "NAME" for tokenizing e.g. keyword names (am I right?)

pekkaklarck:
I mainly meant the name of the setting you use in the data. I noticed you already changed it to `Name` which is great.

What token name/type to use is a good question. It didn't occur to me there's already `NAME` token that's apparently used with settings. I'd in general prefer having less token types and would be fine reusing `NAME` also with suite names, but we already have `TESTCASE_NAME` and `KEYWORD_NAME` so probably it's best to add `SUITE_NAME` as well. The value of the token type shouldn't really matter, but we've in general used values that match the name like
```python
    TESTCASE_NAME = 'TESTCASE NAME'
    KEYWORD_NAME = 'KEYWORD NAME'
```
so I guess this should be
```python
    SUITE_NAME = 'SUITE NAME'
```

pekkaklarck (changes requested):
The added tests look pretty comprehensive, but unfortunately they aren't structured correctly. As the [atest/README.rst](https://github.com/robotframework/robotframework/tree/master/atest#readme) file explains, tests on `atest/robot` side should just validate that features tested on `atest/testdata` side work correctly. Basically all files using the new `Name` setting should be on the `atest/testdata` side and on the `atest/robot` side you'd have tests that validate that the name is set correctly.

There's existing [atest/robot/parsing/suite_settings.robot](https://github.com/robotframework/robotframework/blob/master/atest/robot/parsing/suite_settings.robot) file for testing other settings. It runs [atest/testdata/parsing/suite_settings.robot](https://github.com/robotframework/robotframework/blob/master/atest/testdata/parsing/suite_settings.robot) and validates everything works correctly. It currently checks that `${SUITE.name}` is set correctly based on the file name, but you could add `Name` setting to the file and see that the name is set accordingly. A problem with this approach is that it would be good to have an explicit test for getting suite name from a file name as well. In addition to that, it's a good idea to test that `Name` works also with `__init__.robot` files and that with directories the name is got from the directory name by default.

An alternative would be having a separate `atest/robot/parsing/suite_name.robot` file that tests all aforementioned combinations. Also it would need to run some tests suites where `Name` is used (and not used). Instead of creating new dummy suites, it would probably be better to add `Name` to few suites under [atest/testdata/misc/suites](https://github.com/robotframework/robotframework/tree/master/atest/testdata/misc/suites) and run it. Also other tests run `atest/testdata/misc/suites` so you may need to update them after making changes, but that shouldn't be too big a task.

I didn't yet look at the code in detail but on a very quick look it looks really good. I'll review it in more detail once tests have been restructured. If you have any questions about the somewhat complicated test setup or anything else, don't hesitate to ask here or on the #devel channel on our Slack.

atest/testdata/parsing/suite_names/__init__.robot
```diff
@@ -0,0 +1,4 @@
+*** Settings ***
+Force Tags      suite_naming
+
+Suite Name      Parent Init Suite
```
pekkaklarck, line 4: This setting should be just `Name`, not `Suite Name`.

No need to use tags here. If there was a need, should use `Test Tags` instead of deprecated `Force Tags`. Finally, we don't generally use empty lines between settings and following the same style would keep tests consistent.

atest/testdata/parsing/suite_names/child_tests01/child0102.robot
```diff
@@ -0,0 +1,13 @@
+*** Settings ***
+Documentation       Suite without custom name but with custom name in init
+
+Suite Name          Child 0102 Suite
+
+
+*** Test Cases ***
+Verify That Suite Name is not a file like name
+    Should Not Be Equal    ${SUITE_NAME}    Robot.Parsing.Parent Init Suite.Child Tests01.Child0101
+    Should Not Be Equal    ${SUITE_NAME}    Robot.Parsing.Parent Init Suite.Child Suite01 Name.Child0101
```
pekkaklarck, line 10: Also this test is unnecessary. Removing it and just leaving `Verify Suite Name` makes the test simpler. Same comment applies to latter suites with similar structure.

The empty line between settings is unnecessary as well.

src/robot/conf/languages.py
```diff
@@ -307,6 +309,7 @@ class En(Language):
     variables_setting = 'Variables'
     documentation_setting = 'Documentation'
     metadata_setting = 'Metadata'
+    name_setting = "Name"
```
pekkaklarck, line 312: It's interesting that the setting has name `Name` here but it still needs to be `Suite Name` in data. Most likely this is due to lexer having `Suite Name`. To make sure the used name matches, existing translation tests should also use `Name` setting. It ought to be enough to use `Nimi` (the Finnish translation) in `atest/testdata/parsing/translations/finnish/tests.robot` and enhance `Validate Translations` keyword in `atest/robot/parsing/translations.robot` so that it validates `${SUITE.name}` similarly as e.g. `${SUITE.doc}`.

src/robot/parsing/lexer/tokens.py
```diff
@@ -49,6 +49,7 @@ class Token:
     KEYWORD_NAME = 'KEYWORD NAME'
 
     DOCUMENTATION = 'DOCUMENTATION'
+    SUITE_NAME = 'SUITE NAME'
```
pekkaklarck, line 52: It's a bit unfortunate we apparently need to use `SUITE_NAME` here instead of `NAME` that's already used elsewhere.

utest/parsing/test_lexer.py
```diff
@@ -43,6 +43,7 @@ def test_common_suite_settings(self):
 Test Timeout      1 day
 Force Tags        foo    bar
 Keyword Tags      tag
+Suite Name        Custom Suite Name
```
pekkaklarck, line 46: Change to `Name` also in unit tests.

pekkaklarck (changes requested):
Code looks good but some things still to be done:
1. There are now tests only on the `atest/testdata` side which means they aren't actually run as part of the normal acceptance test run. Separate `atest/robot/parsing/suite_names.robot` file is needed to run them and to validate that suite names are set correctly. If you need help with this, ping me on Slack or ask help here.
2. As commented separately, it would be good to test also translation.
3. The code uses `Suite Name` in some places and `Name` in others. Should consistently use `Name` everywhere.
4. Documentation in the User Guide is missing. That can be done separately once this PR is otherwise ready.

otemek:
Hi, thanks @pekkaklarck for extensive review, but i haven't finished re-work after your previous review. I know I shouldn't commit unfinished job, but I was in the middle of doing it (analyze what's exactly is going on there...) and than had to switch to another task... nevertheless, I'll try to complete this thing on the nearest weekend

atest/testdata/parsing/suite_names/__init__.robot
otemek, line 1: to be deleted as for that already existing suite will be used

atest/testdata/parsing/suite_names/child_tests01/__init__.robot
otemek, line 1: to be deleted as for testing this feature - already existing suite will be used

atest/testdata/parsing/suite_names/child_tests01/child0101.robot
otemek, line 1: to be deleted as for testing this feature - already existing suite will be used

atest/testdata/parsing/suite_names/child_tests01/child0102.robot
otemek, line 1: to be deleted as for testing this feature - already existing suite will be used

atest/testdata/parsing/suite_names/child_tests02/child0201.robot
otemek, line 1: to be deleted as for testing this feature - already existing suite will be used

atest/testdata/parsing/suite_names/child_tests02/child0202.robot
otemek, line 1: to be deleted as for testing this feature - already existing suite will be used

atest/testdata/parsing/suite_names/suite_w_name.robot
otemek, line 1: to be deleted as for testing this feature - already existing suite will be used

atest/testdata/parsing/suite_names/suite_wo_name.robot
otemek, line 1: to be deleted as for testing this feature - already existing suite will be used

atest/robot/core/filter_by_names.robot
```diff
@@ -1,6 +1,7 @@
 *** Settings ***
 Documentation      Tests for --test and --suite options.
 Resource           atest_resource.robot
+Test Tags    filter_by_names
```
pekkaklarck, line 4: Please remove. We don't generally use tags like this. We do use tags for other purposes, though, and unnecessary tags like this are just a distraction. If tags like this were used, the value should be aligned to match other setting values.

atest/robot/parsing/suite_names.robot
```diff
@@ -0,0 +1,23 @@
+*** Settings ***
+Documentation     Run testdata and validate that suite names are set correctly
+Suite Setup     Run Tests    ${EMPTY}    misc/suites
+Resource        atest_resource.robot
+# Test Tags       suite_naming
```
pekkaklarck, line 5: 1. Please remove the commented Test Tags line. No need to have such tags and we avoid commented out code as well.
2. Align setting values.

atest/robot/parsing/suite_names.robot
```diff
@@ -0,0 +1,23 @@
...
+    Should Be Equal As Strings    ${Suite.suites[2].suites[1].name}    Sub2
+
+Cutom Suite Name With Parent init
+    Should Be Equal As Strings    ${Suite.suites[3]}    Parent Suite Name
+    Should Be Equal As Strings    ${Suite.suites[3].suites[0].name}    Sub.Suite.4
+    Should Be Equal As Strings    ${Suite.suites[3].suites[2].name}    SubSuite333
+
+Multi Root Custom Suite Name
+    Should Be Equal As Strings    ${Suite.suites[3].suites[1].name}    SubParentSuite3Name
+    Should Be Equal As Strings    ${Suite.suites[3].suites[1].suites[0].name}    Another Custom Sub Suite
```
pekkaklarck, line 23: 1. What does "Multi Root" mean here?
2. Add newline at the end of the file.

atest/testdata/misc/suites/subsuites2/__init__.robot
```diff
@@ -0,0 +1,2 @@
+*** Settings ***
+Name    Parent Suite Name
```
pekkaklarck, line 2: 1. It would probably be a good idea to include the word "Custom" in all suite names to make it explicit they are custom names. You used that style with most other suite names.
2. Add newline at the end of the file.

These comments apply also to other suites.

src/robot/conf/languages.py
```diff
@@ -305,6 +307,7 @@ class En(Language):
     library_setting = 'Library'
     resource_setting = 'Resource'
     variables_setting = 'Variables'
+    name_setting = "Name"
```
pekkaklarck, line 310: For consistency single quotes would be better than double quotes. Same also with other `name_settings`.

src/robot/parsing/lexer/settings.py
```diff
@@ -109,7 +110,8 @@ def _lex_error(self, setting, values, error):
     def _lex_setting(self, setting, values, name):
         self.settings[name] = values
         # TODO: Change token type from 'FORCE TAGS' to 'TEST TAGS' in RF 7.0.
-        setting.type = name.upper() if name != 'Test Tags' else 'FORCE TAGS'
+        setting_type_map = {'Test Tags': "FORCE TAGS", "Name": "SUITE NAME"}
+        setting.type = setting_type_map.get(name, name.upper())
```
pekkaklarck, line 114: - Nice way to handle the fact that the setting name and type don't match.
- For consistency it would be good to use either single or double quotes in the dictionary. We've generally used single quotes (and ought to even mention that in contribution guidelines).

pekkaklarck (changes requested):
Awesome work! I think all my comments were about style. Two other things to do in addition to them:

1. Add localization test. You could just add e.g. `# language: Polish` into some of the data files and use use `Nazwa` there instead of `Name`. Possibly existing localization tests could also be enhanced, but I can do that separately afterwards as well.
2. Add documentation to the User Guide. I can take care of this as well.

I hope you can take a look my comments and possibly also the two other points above. If you are busy, I can also merge this PR and make these changes myself.

src/robot/parsing/lexer/settings.py
otemek: switched to single quote

src/robot/conf/languages.py
otemek: switched to single quotes

atest/testdata/misc/suites/subsuites2/__init__.robot
otemek: changed name to `Custom parent Suite Name`
+ new EOL

atest/robot/core/filter_by_names.robot
otemek: test tags removed. Just forgot to remove them as I'm using those to debug only those particular tests using "--include" switch

atest/robot/parsing/suite_names.robot
otemek: tag removed
settings aligned

atest/robot/parsing/suite_names.robot
```diff
@@ -0,0 +1,23 @@
+*** Settings ***
+Documentation     Run testdata and validate that suite names are set correctly
+Suite Setup     Run Tests    ${EMPTY}    misc/suites
+Resource        atest_resource.robot
+# Test Tags       suite_naming
+
+*** Test Cases ***
+Simple Custom Suite Name
+    Should Be Equal As Strings    ${Suite.suites[2].name}    Subsuites
```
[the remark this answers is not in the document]
otemek: - correct variable name to use uppercase as the global variable
- changed `Should be Equal As Strings` to `Should be Equal`
- removed word `Simple` from test case names

atest/robot/parsing/suite_names.robot
otemek: Changed name from `Multi Root' to 'Multi Init' to indicate that the suite name is placed under multiple folders with `__init__.robot` initialization file with setting `Name`. Can it be like that?

utest/parsing/test_lexer.py
otemek: Changed to `Name`

src/robot/parsing/model/statements.py
```diff
@@ -426,6 +426,24 @@ def from_params(cls, values, separator=FOUR_SPACES, eol=EOL):
...
 
 
+@Statement.register
+class SuiteName(Fixture):
+    type = Token.SUITE_NAME
+
+    @classmethod
+    def from_params(cls, value, separator=FOUR_SPACES, eol=EOL):
+        return cls([
+            Token(Token.SUITE_NAME, 'Suite Name'),
```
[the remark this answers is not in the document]
otemek: Changed to `Name`

src/robot/parsing/lexer/settings.py
```diff
@@ -32,7 +32,8 @@ class Settings:
         'Test Timeout',
         'Test Template',
         'Timeout',
-        'Template'
+        'Template',
+        'Suite Name'
```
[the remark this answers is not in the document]
otemek: changed to `Name`

src/robot/conf/languages.py
```diff
@@ -168,6 +168,7 @@ class Language:
     variables_setting = None
     documentation_setting = None
     metadata_setting = None
+    name_setting = None
```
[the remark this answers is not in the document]
otemek: moved before `documentation_setting`

src/robot/conf/languages.py
otemek: Changed to `Name` in  data

src/robot/parsing/lexer/tokens.py
otemek: Probably not the best time to ask, but after going through this PR one more time I feel like I'm confused about which Token should i use... either reuse `Token.NAME` or the new one `Token.SUITE_NAME` :smile:

atest/robot/parsing/suite_names.robot
pekkaklarck: I don't understand what you mean with "placed under multiple folders". Is the file inside a sub folder that's executed? If yes, then something like `Custom child suite name` would be better. Anyway, the exact name doesn't matter that much.

atest/robot/core/filter_by_names.robot
pekkaklarck, line 4: Why wasn't this removed?

atest/robot/parsing/suite_names.robot
pekkaklarck, line 5: Why wasn't this removed?

pekkaklarck:
I enabled running tests on CI and several tests failed due to changes to suite names under `atest/testdata/misc`. That's as expected but these tests obviously need to be fixed. Can you @otemek take care of that?

It also seems that although my comments have been marked as resolved, changes aren't seen here. Did you forget to commit and push changes?

Other than the above this PR is ready to be merged. Still need to write documentation, but I already promised to do that earlier.

atest/robot/parsing/suite_names.robot
otemek: renamed to `Custom Child Suite Name`

otemek:
Sorry about those delays/code not pushed/lags/etc... but as You probably have noticed - I have quite a big problem with finding some time for me to completely take care of this so I don't know if I'll be able to finish this in a reasonable time. If there is anyone else who can make it quicker then I'm totally fine with passing it to someone else :)
P.S. Sorry for the problems

pekkaklarck:
No worries! We all have our time pressures and with some new features there's surprisingly lot to do to get everything done. If you don't have time to continue, that's totally fine. We can either continue from here or post-pone this to RF 6.2.

One reason there are so many tests to be fixed is that you added a new test under `atest/testdata/misc/suites`. That breaks all tests running that directory and checking test counts. I'm not sure is the new test needed in this case, though. I believe it would have been possible to just change suite names which wouldn't have had such a big effect on other tests. Unless that new test is really needed, I'd just remove the whole `subsubsuites3`.

pekkaklarck:
I looked at the changes again and this is _so_ close to be ready that I believe we should get this in still in RF 6.1.

As I wrote above, the new test under `atest/testdata/misc/suites` is the reason why so many existing tests need to be changed. You seem to have already fixed many of them and I now enabled CI again to see are there still failures. If there are, removing the new test may be easier way forward than fixing the remaining failures. Some changes would need to be reverted after removing the test but that is likely a smaller task.

pekkaklarck:
I try to get those tests fixed and fixes pushed to this same PR!

pekkaklarck:
I fixed test and pushed changes here. Looking at the history, there are now some commits that are related to recent changes in master, not to actual changes in this PR. I don't know are they due to me pushing these changes without first merging/rebasing locally or were they here already. Well, if tests pass I'll merge this PR and squash all changes. Hopefully that makes the history clean as well.

pekkaklarck:
Rebasing the PR branch locally fixed the history but then needed to force-push changes here. Changes look good and I'll merge this if the CI run succeeds.

otemek:
Thanks for the help and for making it done

pekkaklarck:
Thanks for your contribution!

PR #4664, opened 2023-02-23
Run Keyword And Warn Message

Franz333:
I have this functionality in my user-defined keyword and I thought about submitting it here.

When we are dealing with non blocking bugs (in the context of an application), we want them not to block the execution of the test and be visible in the logs so we know they are still pending. We then use the Builtin `Run Keyword And Warn On Failure`. It is great because it gives visibility and a quick link to the Keyword from within the "Test Execution Errors" section.

Though, this Keyword does not provide the ability to add a custom message to give further information and context. For example, when dealing with many bugs across an application - some of them impacting several areas and therefore present in several test suites - it's nice to be able to display the ticket id and/or link for a quick access to the bug description. 

This Keyword would be an enhancement of `Run Keyword And Warn On Failure` providing the ability to add a message when the keyword fails or succeeds.

image

image

This raises the question of log levels and the ability to customise the label and the colors.

### DIFF
--- a/src/robot/libraries/BuiltIn.py
+++ b/src/robot/libraries/BuiltIn.py
@@ -2085,6 +2085,26 @@ class _RunKeyword(_BuiltInBase):
         status, message = self.run_keyword_and_ignore_error(name, *args)
         if status == 'FAIL':
             logger.warn("Executing keyword '%s' failed:\n%s" % (name, message))
+        return status, 
+        
+    @run_keyword_variant(resolve=0)
+    def run_keyword_and_warn_message(self, success_msg, failure_msg, name, *args):
+        """Runs the given keyword with given arguments and logs a warning.
+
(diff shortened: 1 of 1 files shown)


pekkaklarck:
I'm not sure is this a useful enough enhancement to warrant adding a new BuiltIn keyword especially when there are other alternatives. One alternative is using the newish `TRY/EXCEPT` structure, possibly wrapped to a generic user keyword, and other is using the skip-on-failure functionality. The latter actually sounds like a really good match for your particular need. Is there a reason it wouldn't work for you?

pekkaklarck:
I still believe this isn't important enough feature to warrant a new BuiltIn keyword. Enhancing the current `Run Keyword And Warn On Failure` would be fine, but that is likely pretty complicated without breaking backwards compatibility.

PR #4668, opened 2023-03-01
Handle dangling END, ELSE, and EXCEPT in lexer/parser

yanne:


atest/testdata/parsing/same_setting_multiple_times.robot
```diff
@@ -28,11 +28,18 @@ Use Defaults
...
-    [Documentation]    T1
+    [Documentation]    FAIL Several failures occurred:\n\n
+    ...  1) Setting 'Documentation' is allowed only once. Only the first value is used.\n\n
+    ...  2) Setting 'Tags' is allowed only once. Only the first value is used.\n\n
+    ...  3) Setting 'Setup' is allowed only once. Only the first value is used.\n\n
+    ...  4) Setting 'Teardown' is allowed only once. Only the first value is used.\n\n
+    ...  5) Setting 'Teardown' is allowed only once. Only the first value is used.\n\n
+    ...  6) Setting 'Template' is allowed only once. Only the first value is used.\n\n
+    ...  7) Setting 'Timeout' is allowed only once. Only the first value is used.\n\n
+    ...  8) Setting 'Tags' is allowed only once. Only the first value is used.
```
pekkaklarck, line 39: Now that these errors prevent the whole execution, the "Only the first value is used." part in the error message is unnecessary and actually pretty confusing.

atest/robot/output/listener_interface/lineno_and_source.robot
```diff
@@ -12,225 +12,225 @@ ${RESOURCE FILE}     ${LISTENER DIR}/lineno_and_source.resource
 *** Test Cases ***
 Keyword
     START    KEYWORD          No Operation                    6    NOT SET
-    END      KEYWORD          No Operation                    6    PASS
+    \END      KEYWORD          No Operation                    6    PASS
```
pekkaklarck, line 15: Quite a few ENDs needed to be escaped...

atest/testdata/parsing/test_case_settings.robot
```diff
@@ -180,10 +180,6 @@ Timeout
     [Timeout]    1d
     No Operation
 
-Timeout with message
-    [Timeout]    666    Message not supported since RF 3.2
-    No Operation
```
pekkaklarck, line 185: Is there a test for using timeout with more than one value somewhere else? If not, this probably should be added back, renamed to something like "Invalid timeout" and validation on the `robot` side adjusted accordingly.

atest/testdata/parsing/test_case_settings.robot
```diff
@@ -214,14 +210,19 @@ Multiple settings
...
     [Metadata]    Not valid.
     [Arguments]    Not valid.
     No Operation
 
 Small typo should provide recommendation
+    [Documentation]    FAIL
+    ...    Non-existing setting 'Doc U ment a tion'. Did you mean:
+    ...    ${SPACE*4}Documentation
     [Doc U ment a tion]    This actually worked before RF 3.2.
     No Operation
```
pekkaklarck, line 228: `No Operation` could be changed to something like `Fail    Should not be run` in cases like this.

src/robot/model/__init__.py
```diff
@@ -27,7 +27,7 @@
 
 from .body import BaseBody, Body, BodyItem, Branches
 from .configurer import SuiteConfigurer
-from .control import For, While, If, IfBranch, Try, TryBranch, Return, Continue, Break
+from .control import For, While, If, IfBranch, Try, TryBranch, Return, Continue, Break, Error
```
pekkaklarck, line 30: Imports should be organized alphabetically.

src/robot/model/visitor.py
```diff
@@ -105,7 +105,8 @@ def visit_test(self, test: TestCase):
 from typing import TYPE_CHECKING
 
 from .body import BodyItem
-from .control import Break, Continue, For, If, IfBranch, Return, Try, TryBranch, While
+from .control import (Break, Continue, For, If, IfBranch, Return, Try,
+                      TryBranch, While, Error)
```
pekkaklarck, line 109: Imports should be organized alphabetically.

src/robot/model/visitor.py
```diff
@@ -490,6 +491,29 @@ def end_break(self, break_: Break):
         """
         self.end_body_item(break_)
 
+    def visit_error(self, error: Error):
+        """Visits ERROR elements."""
```
pekkaklarck, line 495: Doc like this would be better:
```python
"""Visits body items resulting from invalid syntax.

Examples include syntax like ``END`` or ``ELSE`` in wrong place and
invalid setting like ``[Invalid]``.
"""
```
Earlier docs use style like `BREAK` because such syntax actually exists, but we don't have any explicit `ERROR` syntax.

src/robot/parsing/lexer/blocklexers.py
```diff
@@ -200,7 +201,7 @@ def _handle_name_or_indentation(self, statement):
     def lexer_classes(self):
         return (TestOrKeywordSettingLexer, BreakLexer, ContinueLexer,
                 ForLexer, InlineIfLexer, IfLexer, ReturnLexer, TryLexer,
-                WhileLexer, KeywordCallLexer)
+                WhileLexer, SyntaxErrorLexer, KeywordCallLexer)
```
pekkaklarck, line 204: `SyntaxErrorLexer` should handle also `BREAK` and `CONTINUE` to allow removing `BreakLexer` and `ContinueLexer` from here. It probably could also handle `RETURN`, but if `ReturnLexer` is removed here then `KeywordLexer` needed to add it to its `lexer_classes`.

src/robot/parsing/lexer/statementlexers.py
```diff
@@ -334,3 +334,17 @@ class BreakLexer(TypeAndArguments):
...
     def handles(cls, statement: list, ctx: TestOrKeywordContext):
         return statement[0].value == 'BREAK'
+
+
+class SyntaxErrorLexer(TypeAndArguments):
+    token_type = Token.ERROR
+
+    @classmethod
+    def handles(cls, statement: list, ctx: TestOrKeywordContext):
+        return statement[0].value in {'END', 'ELSE', 'ELSE IF','EXCEPT'}
```
pekkaklarck, line 344: As mentioned above, `BREAK`, `CONTINUE` and possibly also `RETURN` should be added here.

src/robot/parsing/model/blocks.py
```diff
@@ -19,7 +19,7 @@
 from robot.utils import file_writer, is_pathlike, is_string
 
 from .statements import (Break, Continue, KeywordCall, ReturnSetting, ReturnStatement,
-                         Statement, TemplateArguments)
+                         Statement, TemplateArguments, Error)
```
pekkaklarck, line 22: Order.

src/robot/result/__init__.py
```diff
@@ -39,6 +39,6 @@
 
 from .executionresult import Result
 from .model import (Break, Continue, For, ForIteration, If, IfBranch, Keyword, Message,
-                    Return, TestCase, TestSuite, Try, TryBranch, While, WhileIteration)
+                    Return, TestCase, TestSuite, Try, TryBranch, While, WhileIteration, Error)
```
pekkaklarck, line 42: Order.

src/robot/result/model.py
```diff
@@ -419,6 +419,44 @@ def doc(self):
...
+        self.endtime = endtime
+        self.body = None
+
+    @setter
+    def body(self, body):
+        """Child keywords and messages as a :class:`~.Body` object.
+
+        Typically empty. Only contains something if running BREAK has failed
+        due to a syntax error or listeners have logged messages or executed
+        keywords.
```
pekkaklarck, line 440: `BREAK` reference is invalid here. Also, isn't there typically a message so the body isn't typically empty? There aren't "child keywords", though.

src/robot/result/model.py
```diff
@@ -419,6 +419,44 @@ def doc(self):
         return ''
 
 
+@Body.register
+class Error(model.Error, StatusMixin, DeprecatedAttributesMixin):
```
pekkaklarck, line 423: Shouldn't `Error` be registered also to `Branches` and `Iterations`? Otherwise something like
```robotframework
*** Test Case ***
Invalid syntax in iteration
    WHILE    True
        ELSE
    END

Invalid syntax in branch
    IF    True
        EXCEPT
    END
```
probably won't be handled properly. These cases most likely require new tests.

src/robot/result/xmlelementhandlers.py
```diff
@@ -267,6 +267,15 @@ def start(self, elem, result):
         return result.body.create_break()
 
 
+@ElementHandler.register
+class ErrorHandler(ElementHandler):
+    tag = 'error'
+    children = frozenset(('status', 'msg', 'value'))
+
+    def start(self, elem, result):
+        return result.body.create_error()
```
pekkaklarck, line 276: Is this enough to handle values? Are there tests for invalid syntax going through Rebot?

src/robot/running/builder/transformers.py
```diff
@@ -330,6 +334,10 @@ def visit_If(self, node):
     def visit_Try(self, node):
         TryBuilder(self.kw).build(node)
 
+    def visit_Error(self, node):
+        self.kw.body.create_error(lineno=node.lineno,
+                                  values=node.values, error=format_error(node.errors))
+
 
 class ForBuilder(NodeVisitor):
```
pekkaklarck, line 342: Won't `ForBuillder` and other similar builders need `visit_Error` as well?

src/robot/running/model.py
```diff
@@ -43,7 +43,7 @@
 from robot.model import BodyItem, create_fixture, Keywords, ModelObject
 from robot.output import LOGGER, Output, pyloggingconf
 from robot.result import (Break as BreakResult, Continue as ContinueResult,
-                          Return as ReturnResult)
+                          Return as ReturnResult, Error as ErrorResult)
```
pekkaklarck, line 46: Order.

pekkaklarck (changes requested):
Great work! Some things to fix and cleanup still. See inline comments for details.

atest/testdata/parsing/test_case_settings.robot
yanne: Yes, there is a separate suite about timeouts, this test is completely redundant

src/robot/result/xmlelementhandlers.py
yanne: Yes it is, they would not end in the log otherwise (or in the result model that is used in the tests).

atest/testdata/parsing/same_setting_multiple_times.robot
yanne: This will by the way be different with suite settings, they will still be only logged to console but do not cause runtime error

Issue #4089, opened 2021-09-16
Support asynchronous functions and methods as keywords


KotlinIsland:
Say I want to have an `async def` as a keyword implementation, currently it will fail silently:
```py
from asyncio import gather

async def foo():
    log("start")
    await gather(something(1), something(2), something(3))
    log("done")
```
```robot
*** Test Cases ***
test
    foo
```
## actual result:
Nothing from the coroutine is executed and nothing appears in the log:
![image](https://user-images.githubusercontent.com/65446343/133693521-aa5f0cbe-7061-4233-82b9-3dfbebf30e3d.png)

### command line output:
```
RuntimeWarning: coroutine 'foo' was never awaited
```

## Workaround
```py
from asyncio import run, gather

def foo():
    run(_foo())


async def _foo():
    log("start")
    await gather(something(1), something(2), something(3))
    log("done")
```
Now it works, BUT if the event loop is already started by something else then this won't work :(

you would have to add some nonsense to start a second loop like:
```py
def run(coro):
    t = Thread(target=lambda: asyncio.run(coro))
    t.start()
    t.join()
```

# Request
Make async keywords await by default.

Another option would be to add an `Await Keyword` keyword, but I don't like the sound of that

pekkaklarck:
It has never even occurred to me to use async functions as keywords, but I guess that's something we could support now that we are dropping Python 2 support in RF 5.0. How should that work in practice? Should Robot check that if a keyword is async and then automatically wait for it?

Are you interested to implement this support? Notice that it might be better to wait until Python 2 support has been removed because that's likely to change to related code a bit.

KotlinIsland:
Maybe it could be as simple as if the result of calling the keyword is a coroutine then await it. Doesn't need to drop 2 support to do that right?

KotlinIsland:
I'd definitely be interested in implementing it.

pekkaklarck:
We are anyway dropping Python 2 support in RF 5.0.

KotlinIsland:
# 😳

Could you please add python 3 support back in as It's what I use myself.

pavlemutic:
Hey guys, any updates on this? :)

pekkaklarck:
We have  PR #4677 implementing this and on a very quick look it looks good. I don't have time for more detailed review now, but I add this issue to RF 6.1 scope and try to have time to look at it this week.

pekkaklarck:
[...]

PR #4677, opened 2023-03-08
Asyncio runner

ygorpontelo:
Issue #4089

### DIFF
--- a/atest/robot/keywords/async_tests.robot
+++ b/atest/robot/keywords/async_tests.robot
@@ -12,5 +12,17 @@ Basic Async Works
 Works Using Gather
     Check Test Case    ${TESTNAME}
 
+Builtin Async Sleep Keyword Works
+    Check Test Case    ${TESTNAME}
+
 Long Async Tasks Run In Background
     Check Test Case    ${TESTNAME}
+
+Builtin Call From Library Works
+    Check Test Case    ${TESTNAME}
+
+Builtin Gather Keyword Works
+    Check Test Case    ${TESTNAME}
+
+Builtin Gather Keyword Fails If Keyword Is Not Async
+    Check Test Case    ${TEST NAME}    status=FAIL    message=Keyword 'No Operation' is not async
--- a/atest/robot/keywords/async_tests.robot
+++ b/atest/robot/keywords/async_tests.robot
@@ -1,0 +1,16 @@
+*** Settings ***
+Suite Setup      Run Tests    ${EMPTY}    keywords/async_keywords.robot
+Resource         atest_resource.robot
+
+*** Test Cases ***
+Works With Asyncio Run
+    Check Test Case    ${TESTNAME}
+
+Basic Async Works
+    Check Test Case    ${TESTNAME}
+
+Works Using Gather
+    Check Test Case    ${TESTNAME}
+
+Long Async Tasks Run In Background
+    Check Test Case    ${TESTNAME}
--- a/atest/testdata/keywords/AsyncLib.py
+++ b/atest/testdata/keywords/AsyncLib.py
@@ -11,6 +11,9 @@ class Hanger:
             self.ticks.append('tick')
             await asyncio.sleep(2)
 
+async def just_wait():
+    await asyncio.sleep(2)
+
 class AsyncLib:
 
     def __init__(self) -> None:
@@ -28,7 +31,8 @@ class AsyncLib:
         return asyncio.run(inner())
     
     async def can_use_gather(self):
-        asyncio.gather()
+        tasks = [just_wait() for _ in range(5)]
+        await asyncio.gather(*tasks)
 
     async def create_hanger(self):
         hanger = Hanger()
--- a/atest/testdata/keywords/AsyncLib.py
+++ b/atest/testdata/keywords/AsyncLib.py
@@ -1,5 +1,8 @@
 import asyncio
 
+from robot.libraries.BuiltIn import BuiltIn
+
+
 class Hanger:
 
     def __init__(self) -> None:
@@ -9,7 +12,7 @@ class Hanger:
     async def start_async_process(self):
         while True:
             self.ticks.append('tick')
-            await asyncio.sleep(2)
+            await asyncio.sleep(0.5)
 
 async def just_wait():
     await asyncio.sleep(2)
@@ -41,3 +44,6 @@ class AsyncLib:
     
     async def stop_task_from_hanger(self, hanger):
         hanger.task.cancel()
+
+    async def run_keyword_using_builtin(self):
+        return await BuiltIn().run_keyword("Basic Async Test")
--- a/atest/testdata/keywords/AsyncLib.py
+++ b/atest/testdata/keywords/AsyncLib.py
@@ -1,0 +1,39 @@
+import asyncio
+
+class Hanger:
+
+    def __init__(self) -> None:
+        self.task = None
+        self.ticks = []
+    
+    async def start_async_process(self):
+        while True:
+            self.ticks.append('tick')
+            await asyncio.sleep(2)
+
+class AsyncLib:
+
+    def __init__(self) -> None:
+        pass
+
+    async def basic_async_test(self):
+        await asyncio.sleep(2)
+        return 'Got it'
+
+    def async_with_run_inside(self):
+        async def inner():
+            await asyncio.sleep(2)
+            return 'Works'
+
+        return asyncio.run(inner())
+    
+    async def can_use_gather(self):
+        asyncio.gather()
+
+    async def create_hanger(self):
+        hanger = Hanger()
+        hanger.task = asyncio.create_task(hanger.start_async_process())
+        return hanger
+    
+    async def stop_task_from_hanger(self, hanger):
+        hanger.task.cancel()
--- a/atest/testdata/keywords/AsyncLib.py
+++ b/atest/testdata/keywords/AsyncLib.py
@@ -41,3 +41,8 @@ class AsyncLib:
 
     async def run_keyword_using_builtin(self):
         return await BuiltIn().run_keyword("Basic Async Test")
+
+    async def create_task_with_loop(self):
+        loop = asyncio.get_event_loop()
+        task = loop.create_task(self.basic_async_test())
+        return await task
--- a/atest/testdata/keywords/AsyncLib.py
+++ b/atest/testdata/keywords/AsyncLib.py
@@ -8,40 +8,34 @@ class Hanger:
     def __init__(self) -> None:
         self.task = None
         self.ticks = []
-    
+
     async def start_async_process(self):
         while True:
             self.ticks.append('tick')
-            await asyncio.sleep(0.5)
+            await asyncio.sleep(0.01)
 
-async def just_wait():
-    await asyncio.sleep(2)
 
 class AsyncLib:
 
-    def __init__(self) -> None:
-        pass
-
     async def basic_async_test(self):
-        await asyncio.sleep(2)
+        await asyncio.sleep(0.1)
         return 'Got it'
 
     def async_with_run_inside(self):
         async def inner():
-            await asyncio.sleep(2)
+            await asyncio.sleep(0.1)
             return 'Works'
-
         return asyncio.run(inner())
-    
+
     async def can_use_gather(self):
-        tasks = [just_wait() for _ in range(5)]
+        tasks = [asyncio.sleep(0.1) for _ in range(5)]
         await asyncio.gather(*tasks)
 
     async def create_hanger(self):
         hanger = Hanger()
         hanger.task = asyncio.create_task(hanger.start_async_process())
         return hanger
-    
+
     async def stop_task_from_hanger(self, hanger):
         hanger.task.cancel()
 
--- a/doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
+++ b/doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
@@ -1946,8 +1946,8 @@ __ `Specifying argument types using @keyword decorator`_
 Asynchronous keywords
 ~~~~~~~~~~~~~~~~~~~~~
 
-Starting from Robot Framework 6.1, it is possible to run async functions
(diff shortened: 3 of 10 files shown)


atest/robot/keywords/async_tests.robot
```diff
@@ -0,0 +1,28 @@
+*** Settings ***
+Suite Setup      Run Tests    ${EMPTY}    keywords/async_keywords.robot
```
pekkaklarck, line 2: `async_tests.robot` should be renamed to `async_keywords.robot`. We are running keywords, not tests, asynchronously and that name would also match the file under `atest/testdata`.

atest/robot/keywords/async_tests.robot
```diff
@@ -0,0 +1,28 @@
...
+    Check Test Case    ${TESTNAME}
+
+Builtin Call From Library Works
+    Check Test Case    ${TESTNAME}
+
+Builtin Gather Keyword Works
+    Check Test Case    ${TESTNAME}
+
+Builtin Gather Keyword Fails If Keyword Is Not Async
+    Check Test Case    ${TEST NAME}    status=FAIL    message=Keyword 'No Operation' is not async
```
pekkaklarck, line 28: Typically better to include status information into the documentation of the executed suite like
```
[Documentation]    FAIL    Expected error message
```

atest/testdata/keywords/AsyncLib.py
```diff
@@ -0,0 +1,49 @@
...
+            self.ticks.append('tick')
+            await asyncio.sleep(0.5)
+
+async def just_wait():
+    await asyncio.sleep(2)
+
+class AsyncLib:
+
+    def __init__(self) -> None:
+        pass
```
pekkaklarck, line 23: This `__ini__` is unnecessary and just adds noise. Also, there should be two empty rows between classes/functions on the module level (PEP-8).

atest/testdata/keywords/AsyncLib.py
```diff
@@ -0,0 +1,49 @@
...
+
+    async def basic_async_test(self):
+        await asyncio.sleep(2)
```
pekkaklarck, line 26: Are this long sleeps needed? To keep execution as fast as possible, sleeps like this  should be as short as possible without making tests flakey.

doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
```diff
@@ -1943,6 +1943,60 @@ __ `Specifying argument types using @keyword decorator`_
 
 .. note:: Automatic type conversion is new in Robot Framework 3.1.
 
+Asynchronous Keywords
+~~~~~~~~~~~~~~~~~~~~~
+
+Starting from Robot Framework 6.0.3, it is possible to run async functions
```
pekkaklarck, line 1949: Next release is going to be 6.1.

src/robot/libraries/BuiltIn.py
```diff
@@ -2526,6 +2527,64 @@ def _get_suite_in_teardown(self, kwname):
                                % kwname)
         return self._context.suite
 
+    @run_keyword_variant(resolve=0, dry_run=True)
+    async def gather_async_keywords(self, *keywords):
```
pekkaklarck, line 2531: I'm not sure is this keyword a good addition to the BuiltIn library. It looks like something that may need to evolve in the future, and making such changes in BuiltIn is hard. It also looks like there would be needs for more similar keywords and having all async keywords in a single library would be better.

Instead of having this keyword in BuiltIn, I believe it would be better to have a separate Async library. In the beginning it would probably be best if it would be an external library, because that allows faster development independent from Robot's release schedule. External libraries don't have such strict backwards compatibility concerns as Robot's standard libraries either. Once the library gets more mature and proves generally useful, it could then be included as one of the standard libraries.

src/robot/libraries/BuiltIn.py
```diff
@@ -2945,6 +3004,43 @@ def _sleep_in_parts(self, seconds):
             if remaining <= 0:
                 break
             time.sleep(min(remaining, 0.01))
+    
+    async def async_sleep(self, time_, reason=None):
```
pekkaklarck, line 3008: As I commented above, I believe we shouldn't add async keywords to BuiltIn. A separate library that's first developed externally and possibly later made a standard library would be better.

In this particular case, I'd also like to know what's the use case for this keyword? In what context would you need to have an async sleep and not a normal sleep?

src/robot/running/context.py
```diff
@@ -22,6 +23,7 @@ class ExecutionContexts:
 
     def __init__(self):
         self._contexts = []
+        self.event_loop_ref = None
```
pekkaklarck, line 26: I don't think this is needed. It ought to be enough if actual context objects have the reference.

src/robot/running/context.py
```diff
@@ -39,12 +41,19 @@ def namespaces(self):
         return (context.namespace for context in self)
 
     def start_suite(self, suite, namespace, output, dry_run=False):
-        ctx = _ExecutionContext(suite, namespace, output, dry_run)
+        if self._contexts:
+            self.event_loop_ref = self._contexts[-1]._event_loop
+        ctx = _ExecutionContext(suite, namespace, output, dry_run, self.event_loop_ref)
```
pekkaklarck, line 46: I believe the above could be written like this to avoid the need to have `self.event_loop_ref`:
```python
event_loop = self._contexts[-1]._event_loop if self._contexts else None
ctx = _ExecutionContext(..., event_loop)
```

src/robot/running/context.py
```diff
@@ -39,12 +41,19 @@ def namespaces(self):
...
         return ctx
 
     def end_suite(self):
-        self._contexts.pop()
+        ctx = self._contexts.pop()
+        if not self.event_loop_ref and ctx._event_loop:
+            # handle event loop being created after ctx creation
+            self.event_loop_ref = ctx._event_loop
+        if not self._contexts and self.event_loop_ref:
+            self.event_loop_ref.close()
```
pekkaklarck, line 56: Wouldn't it be enough to have this instead:
```python
ctx = self._contexts.pop()
if not self._contexts and ctx._event_loop:
    ctx.event_loop.close()
```

src/robot/running/context.py
```diff
@@ -67,6 +76,7 @@ def __init__(self, suite, namespace, output, dry_run=False):
         self.timeout_occurred = False
         self.steps = []
         self.user_keywords = []
+        self._event_loop = event_loop
```
pekkaklarck, line 79: This attribute needs to be accessed by `ExecutionContexts` so it shouldn't have `_` prefix. I believe it could be just `self.event_loop` and then there could be a method like `get_running_event_loop()` (or a property like `running_event_loop`) that creates a new loop and returns that.

pekkaklarck (changes requested):
Tests and code looks very good. I added some comments but nothing that would be hard to fix. Also documentation looks good on a very quick look but I need to read it more thoroughly later.

The main open design decision is what to do with the added async keywords in BuiltIn. As I commented separately, I believe a separate Async library is better place for such keywords and I'd rather see such a library developed externally first. What are your thoughts about this?

src/robot/running/context.py
ygorpontelo: This unfortunately doesn't close the loop correctly, i tried a different aproach now. Instead of passing the reference directly, i made a class to handle things. Tell me what you think when you check the commit.

src/robot/libraries/BuiltIn.py
ygorpontelo: Makes sense to be a library. I can create even more keywords there to cover asyncio functionality, will see to it soon.

src/robot/running/context.py
```diff
@@ -13,15 +13,36 @@
...
+        if self._loop_ref is None:
+            self._loop_ref = asyncio.new_event_loop()
+        return self._loop_ref
+
+    def run_until_complete(self, coro):
+        return self.running_event_loop.run_until_complete(coro)
+
+    def close_loop(self):
+        if self._loop_ref:
+            self._loop_ref.close()
```
pekkaklarck, line 38: An explicit class for handling the event loop is a good idea. No need for it to start with `_`, though. We used that style earlier to indicate a class was private not not anymore.

doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
```diff
@@ -1943,6 +1943,63 @@ __ `Specifying argument types using @keyword decorator`_
...
+        await asyncio.sleep(5)
+
+Running an async function inside asyncio.run() works, but it's not recommended
+as this will close the event loop as soon as the function finishes. You can also
+get the reference of the loop by running asyncio.get_running_loop(), note however
+that you should be careful when modifying how the event loop runs, for example
+calling close() on it will make it impossible to run other coroutines, let Robot
+Framework manage the loop. If you have any function or resource that requires the
+event loop, even though await is not used, you have to define your function as
+async to have the event loop available.
```
pekkaklarck, line 1969: 1. References to code should be surrounded with backticks like
    ```
    function inside `asyncio.run()` works
    ```

2. We have generally used style `it is` instead of `it's` in this somewhat formal text.

src/robot/running/librarykeywordrunner.py
```diff
@@ -104,7 +106,13 @@ def _run_with_output_captured_and_signal_monitor(self, runner, context):
...
         try:
             STOP_SIGNAL_MONITOR.start_running_keyword(context.in_teardown)
-            return runner()
+            runner_result = runner()
+            if asyncio.iscoroutine(runner_result):
+                try:
+                    asyncio.get_running_loop()
+                except RuntimeError:
+                    runner_result = context.event_loop.run_until_complete(runner_result)
+            return runner_result
```
pekkaklarck, line 115: What's the point of [asyncio.get_running_loop()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop) here? It looks like it would be used for checking is there a running loop, but I would have though that the possibly returned loop would then be used for something.

There's also a problem that `asyncio.get_running_loop()` is new in Python 3.7 and we still support Python 3.6. I don't mind our new async support not working with Python 3.6, but there should be an explicit error in that case.

pekkaklarck (changes requested):
A separate `EventLoop` class is a really good idea. Most of my comments are related to documentation formatting and I could fix them myself as well, but the `asyncio.get_running_loop` usage looks a bit strange. We can discuss about that on Slack as well.

src/robot/running/librarykeywordrunner.py
ygorpontelo: The check is important because we can run keywords inside an async function, like the builtIn example. So if we try to make the event loop run again, it will throw an error.  So if a running loop exists, we just need to return the coroutine. I would like to have it together with the If condition, but there's no function to check without raising an exception. I can make a wrapper and name it is_loop_running or something like that if the intention is not clear.

pekkaklarck: I see. The thing that looks strange is just returning the `runner_result` if there's no exception. It's clear there's an event loop in that case, but I would have though you need to then use that loop for something here explicitly. Perhaps a comment could be added to explain this.

As I wrote above, I'm fine making this a Python 3.7+ feature but then there should be an explicit exception. On Slack you mentioned that you found some private `asyncio` method that exists also in Python 3.6 and could be used here instead. I don't think using a private method is good long term solution, but it could certainly used if `get_running_loop()` doesn't exist. In that case checking is there a loop should certainly be done in a helper method.

src/robot/running/librarykeywordrunner.py
```diff
@@ -104,7 +104,14 @@ def _run_with_output_captured_and_signal_monitor(self, runner, context):
     def _run_with_signal_monitoring(self, runner, context):
         try:
             STOP_SIGNAL_MONITOR.start_running_keyword(context.in_teardown)
-            return runner()
+            runner_result = runner()
+            if (context.asynchronous.is_coroutine(runner_result) and not
+                context.asynchronous.is_loop_running()):
```
pekkaklarck, line 109: I believe this violates PEP-8. The second line of the expression should be indented more and, less importantly, `and not` moved there. One easy way to avoid these issues would be adding new helper that checks both is the result a coroutine and is a loop running. I'm just not sure is that worth the effort (or what that helper should be named).

src/robot/running/librarykeywordrunner.py
```diff
@@ -104,7 +104,14 @@ def _run_with_output_captured_and_signal_monitor(self, runner, context):
...
+                runner_result = (context
+                                 .asynchronous
+                                 .event_loop
+                                 .run_until_complete(runner_result))
```
pekkaklarck, line 113: This looks a bit odd. Perhaps
```python
loop = context.asynchronous.event_loop
return loop.run_until_complete(runner_result)
```
would be better.

`context.asynchronous.event_loop` is also pretty long compared to earlier ´context.event_loop`, but having all the async stuff under `context.asynchronous` definitely has benefits as well. Probably best leave it like it is.

doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
```diff
@@ -1943,6 +1943,63 @@ __ `Specifying argument types using @keyword decorator`_
...
+
+    @keyword
+    async def examples():
+        tasks = [task_async() for _ in range(10)]
+        results = await asyncio.gather(*tasks)
+
+        background_task = asyncio.create_task(task_async()) # create a task in background
+        await background_task
+
+        await BuiltIn().run_keyword("Some async keyword")
```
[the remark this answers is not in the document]
ygorpontelo: Not really common, will remove it.

pekkaklarck:
I got an error here when merging this PR but it seems changes were nevertheless committed in a9160a8d53d68d121f7731a6e98e9eefc6a179da. Not sure what happened, but feels like GitHub bug. I'll leave this open for a little time to see is the issue resolved and this PR properly closed as merged. If not, I'll close this later.

pekkaklarck (approved):

Issue #4570, opened 2022-12-19
Add type information to `TestSuite` structure


pekkaklarck:
The `TestSuite` structure would benefit from more type hints that would allow IDEs to provide better completion support. That would help with writing code using the visitor API and listener API v3, especially if #4569 and #4568 are implemented.

For most parts adding type info ought to be pretty easy and at least those parts should preferably be done already in RF 6.1. It is possible that `TestSuite` code using our `@setter` is too dynamic to editors to understand and just adding type info to isn't enough. If that's the case, we need to consider changing such code so that it uses the standard `@property.setter` instead. That may be too big a task for RF 6.1, though.

pekkaklarck:
I added Mypy compatible type hints to our `setter` in 810afc8. Based on discussion our Slack with @Serhiy1, that helps also with VSCode that uses pyrigh. Typing can be still enhanced (should at least use `setter(Generic[T, V])` instead of `setter(Generic[V])`), but this is a good start nevertheless.

Unfortunately the added typing didn't help at all with PyCharm and I submitted issue [PY-59658](https://youtrack.jetbrains.com/issue/PY-59658) about that to their tracker. That issue was closed as a duplicate of [PY-29523](https://youtrack.jetbrains.com/issue/PY-29523) and there seemed to be also other related issues. Notice that you may need 

PyCharm also has problems with descriptors being used together with `__slots__`. Basically when we have a descriptor like
```python
@setter
def source(self, source) -> Path:
    return Path(source) if isinstance(source, str) else source
```
then assignment like `self.source = source` in `__init__` cause errors. I was about to submit an issue about that as well, but it turned out that there's already [PY-29770](https://youtrack.jetbrains.com/issue/PY-29770) about that.

I'm not surprised with PyCharm having bugs, but it's pretty worrying that both PY-29523 and PY-29770 have been reported already five years ago and there hasn't been much interest towards them. We cannot thus expect them to be fixed in the near future. If you are PyCharm user and interested in better experience when developing Robot Framework or tooling using its APIs, I recommend you to vote the referenced issues. You need to log in to Youtrack to vote, but you can authenticate using your GitHub account. I don't have high hopes for voting to help, but there's not much else we can do.

pekkaklarck:
[...]

PR #4678, opened 2023-03-10
Add some type hinting to the classes inside models. 

Serhiy1:
Completes type hinting in the test suite structure #4570

The files that got type hints added, where mainly files that I have interacted with as an end user. 

The PR is fully Compatible with Compatible with python 3.6 onwards.

### DIFF
--- a/src/robot/model/testsuite.py
+++ b/src/robot/model/testsuite.py
@@ -14,6 +14,7 @@
 #  limitations under the License.
 
 from pathlib import Path
+from typing import TYPE_CHECKING, Iterable
 
 from robot.utils import setter
 
@@ -27,6 +28,9 @@ from .modelobject import ModelObject
 from .tagsetter import TagSetter
 from .testcase import TestCase, TestCases
 
+if TYPE_CHECKING:
+    from robot.model.visitor import SuiteVisitor
+    from robot.model.tags import Tags
 
 class TestSuite(ModelObject):
     """Base model for single suite.
@@ -37,11 +41,13 @@ class TestSuite(ModelObject):
     test_class = TestCase    #: Internal usage only.
     fixture_class = Keyword  #: Internal usage only.
     repr_args = ('name',)
-    __slots__ = ['parent', '_name', 'doc', '_setup', '_teardown',  'rpa',
+    __slots__ = ['parent', '_name', 'doc', '_setup', '_teardown', 'rpa',
                  '_my_visitors']
 
-    def __init__(self, name: str = '', doc: str = '', metadata: dict = None,
-                 source: Path = None, rpa: bool = False, parent: 'TestSuite' = None):
+    def __init__(self, name: str = '', doc: str = '',
+                 metadata: 'Metadata | dict | None' = None,
+                 source: 'Path | str | None' = None, rpa: bool = False,
+                 parent: 'TestSuite | None' = None):
         self._name = name
         self.doc = doc
         self.metadata = metadata
@@ -55,7 +61,7 @@ class TestSuite(ModelObject):
         self._my_visitors = []
 
     @staticmethod
-    def name_from_source(source: Path):
+    def name_from_source(source: 'Path | str | None') -> str:
         if not source:
             return ''
         if not isinstance(source, Path):
@@ -67,7 +73,7 @@ class TestSuite(ModelObject):
         return name.title() if name.islower() else name
 
     @property
-    def _visitors(self):
+    def _visitors(self) -> 'list[SuiteVisitor]':
         parent_visitors = self.parent._visitors if self.parent else []
         return self._my_visitors + parent_visitors
 
@@ -83,37 +89,37 @@ class TestSuite(ModelObject):
                 or ' & '.join(s.name for s in self.suites))
 
     @name.setter
-    def name(self, name):
+    def name(self, name: str):
         self._name = name
 
     @setter
-    def source(self, source):
+    def source(self, source: 'Path | str | None') -> 'Path | None':
         return source if isinstance(source, (Path, type(None))) else Path(source)
 
     @property
-    def longname(self):
+    def longname(self) -> str:
         """Suite name prefixed with the long name of the parent suite."""
         if not self.parent:
             return self.name
         return f'{self.parent.longname}.{self.name}'
 
     @setter
-    def metadata(self, metadata):
+    def metadata(self, metadata) -> Metadata:
         """Free test suite metadata as a dictionary."""
         return Metadata(metadata)
 
     @setter
-    def suites(self, suites):
+    def suites(self, suites) -> 'TestSuites':
         """Child suites as a :class:`~.TestSuites` object."""
         return TestSuites(self.__class__, self, suites)
 
     @setter
-    def tests(self, tests):
+    def tests(self, tests) -> TestCases:
         """Tests as a :class:`~.TestCases` object."""
         return TestCases(self.test_class, self, tests)
 
     @property
-    def setup(self):
+    def setup(self) -> Keyword:
         """Suite setup as a :class:`~.model.keyword.Keyword` object.
 
         This attribute is a ``Keyword`` object also when a suite has no setup
@@ -138,11 +144,11 @@ class TestSuite(ModelObject):
         ``suite.keywords.setup``.
         """
         if self._setup is None and self:
-            self._setup = create_fixture(None, self, Keyword.SETUP)
+            self._setup : Keyword = create_fixture(None, self, Keyword.SETUP)
         return self._setup
 
     @setup.setter
-    def setup(self, setup):
+    def setup(self, setup: 'Keyword | None'):
         self._setup = create_fixture(setup, self, Keyword.SETUP)
 
     @property
@@ -161,17 +167,17 @@ class TestSuite(ModelObject):
         return bool(self._setup)
 
     @property
-    def teardown(self):
+    def teardown(self) -> Keyword:
         """Suite teardown as a :class:`~.model.keyword.Keyword` object.
 
         See :attr:`setup` for more information.
         """
         if self._teardown is None and self:
-            self._teardown = create_fixture(None, self, Keyword.TEARDOWN)
+            self._teardown: Keyword = create_fixture(None, self, Keyword.TEARDOWN)
         return self._teardown
 
     @teardown.setter
-    def teardown(self, teardown):
+    def teardown(self, teardown: 'Keyword | None'):
         self._teardown = create_fixture(teardown, self, Keyword.TEARDOWN)
 
     @property
@@ -236,7 +242,10 @@ class TestSuite(ModelObject):
             return True
         return any(s.has_tests for s in self.suites)
 
-    def set_tags(self, add=None, remove=None, persist=False):
+    def set_tags(self,
+                  add: 'Tags | str | Iterable[str] | None' = None,
+                  remove: 'Tags | str | Iterable[str] | None' = None,
+                  persist: bool = False):
         """Add and/or remove specified tags to the tests in this suite.
 
         :param add: Tags to add as a list or, if adding only one,
@@ -251,8 +260,10 @@ class TestSuite(ModelObject):
         if persist:
             self._my_visitors.append(setter)
 
-    def filter(self, included_suites=None, included_tests=None,
-               included_tags=None, excluded_tags=None):
+    def filter(self, included_suites: 'str | Iterable[str] | None' = None,
+               included_tests: 'str | Iterable[str] | None' = None,
+               included_tags: 'str | Iterable[str] | None' = None,
+               excluded_tags: 'str | Iterable[str] | None' = None):
         """Select test cases and remove others from this suite.
 
         Parameters have the same semantics as ``--suite``, ``--test``,
@@ -301,7 +312,7 @@ class TestSuite(ModelObject):
     def __str__(self):
         return self.name
 
-    def to_dict(self):
+    def to_dict(self) -> dict:
         data = {'name': self.name}
         if self.doc:
             data['doc'] = self.doc
@@ -322,7 +333,7 @@ class TestSuite(ModelObject):
         return data
 
 
-class TestSuites(ItemList):
+class TestSuites(ItemList[TestSuite]):
     __slots__ = []
 
     def __init__(self, suite_class=TestSuite, parent=None, suites=None):


src/robot/model/body.py
```diff
@@ -122,7 +130,7 @@ def create(self):
     def create_keyword(self, *args, **kwargs):
         return self._create(self.keyword_class, 'create_keyword', args, kwargs)
 
-    def _create(self, cls, name, args, kwargs):
+    def _create(self, cls:Union[Type[C], None], name: str, args, kwargs) -> C:
```
pekkaklarck, line 133: - `cls` can only be `None` is there's a bug in our code so it shouldn't be included in type hints.
- This method creates body items so I believe this should be `_create(self, cls: Type[BodyItem], name: str, args: list, kwargs: dict) -> BodyItem`
- This is mostly internal code so I'm not sure are type hints needed in general. If the above works, it may be simple enough to be useful, but I believe more complex type hints with `TypeVar` aren't worth the effort.

src/robot/model/body.py
```diff
@@ -222,15 +230,15 @@ class Body(BaseBody):
     pass
 
 
-class Branches(BaseBody):
+class Branches(BaseBody, Generic[C]):
     """A list-like object representing IF and TRY branches."""
     __slots__ = ['branch_class']
 
-    def __init__(self, branch_class, parent=None, items=None):
+    def __init__(self, branch_class: Type[C], parent=None, items=None):
```
pekkaklarck, line 237: `branch_class` cannot be anything (i.e. generic). It actually only be `IfBranch` or `TryBranch`, but just using `BodyItem` could be fine as well.

src/robot/model/control.py
```diff
@@ -14,10 +14,10 @@
 #  limitations under the License.
 
 from robot.utils import setter
-
 from .body import Body, BodyItem, Branches
 from .keyword import Keywords
 
+from typing import Any, Dict
 
 @Body.register
```
pekkaklarck, line 22: - Standard module imports should be before internal imports.
- Our style is to have an empty row between absolute imports from top level module (`from robot.xxx import ...`) and relative imports from the same package (`from .xxx import ...`).
- Two empty lines needed before class declaration (and a class decorator is part of the declaration).

src/robot/model/control.py
```diff
@@ -117,7 +117,7 @@ def _include_in_repr(self, name, value):
         return name == 'condition' or value is not None
 
     def to_dict(self):
-        data = {'type': self.type}
+        data: Dict[str, Any] = {'type': self.type}
```
pekkaklarck, line 120: This doesn't look too useful, especially inside a method. It would be better to add a type hint to the method, but I believe just `dict` is good enough unless we'd use a `TypedDict` to properly document the structure. Using a `TypedDict` with all `to_dict` methods would be a huge task and not in the scope of this PR. It would probably be better to use something like pydantic for documenting the `to_dict/json` and `from_dict/json` methods anyway.

src/robot/model/control.py
```diff
@@ -313,6 +313,7 @@ class Return(BodyItem):
     type = BodyItem.RETURN
     repr_args = ('values',)
     __slots__ = ['values']
+    body: Body
```
pekkaklarck, line 316: `Return`, `Continue` and `Break` don't have a `body` except on the result side.

src/robot/model/fixture.py
```diff
@@ -15,10 +15,16 @@
...
+    from robot.running import model
+    from robot import result
+    from robot.model import keyword
+    F = TypeVar("F", keyword.Keyword, result.Keyword, model.Keyword)
 
-def create_fixture(fixture, parent, fixture_type):
+def create_fixture(fixture: Union["F", None], parent: object, fixture_type: str) -> "F":
     # TestCase and TestSuite have 'fixture_class', Keyword doesn't.
-    fixture_class = getattr(parent, 'fixture_class', parent.__class__)
+    fixture_class: Type["F"] = getattr(parent, 'fixture_class', parent.__class__)
```
pekkaklarck, line 27: This looks _really_ complicated for a method that's only used internally. If this was to be typed, `@overload` would probably help, but I believe it's not worth the effort.

src/robot/model/keyword.py
```diff
@@ -15,9 +15,14 @@
 
 import warnings
 
+
```
pekkaklarck, line 18: Remove.

src/robot/model/keyword.py
```diff
@@ -15,9 +15,14 @@
...
 from .body import Body, BodyItem
 from .itemlist import ItemList
 
+from typing import TYPE_CHECKING
+if TYPE_CHECKING:
+    from robot.model.visitor import SuiteVisitor
```
pekkaklarck, line 24: Standard library import in wrong place.

src/robot/model/keyword.py
```diff
@@ -29,12 +34,15 @@ class Keyword(BodyItem):
     repr_args = ('name', 'args', 'assign')
     __slots__ = ['_name', 'args', 'assign', 'type']
 
-    def __init__(self, name='', args=(), assign=(), type=BodyItem.KEYWORD, parent=None):
+
+    def __init__(self, name:str='', args: tuple=(), assign: tuple=(), type=BodyItem.KEYWORD, parent=None):
```
pekkaklarck, line 38: - `name:str=''` -> `name: str = ''`
- Should add type hints to all parameters.
- `args` and `assign` probably could also be lists, but I believe `tuple` is fine as a type hint. Can be changed to something more generic like `Sequence` later if needed.

src/robot/model/keyword.py
```diff
@@ -29,12 +34,15 @@ class Keyword(BodyItem):
...
         self._name = name
         self.args = args
         self.assign = assign
         self.type = type
         self.parent = parent
+        self.body: "Body"
+        self.teardown: "Keyword"
 
     @property
     def name(self):
```
pekkaklarck, line 48: Return type hint missing.

src/robot/model/keyword.py
```diff
@@ -65,7 +73,7 @@ def to_dict(self):
         return data
 
 
-class Keywords(ItemList):
+class Keywords(ItemList[Keyword]):
```
pekkaklarck, line 76: Although this class is called `Keywords`, it actually contains `BodyItems`. Not a huge deal either way, it will be removed in RF 7.

src/robot/model/testcase.py
```diff
@@ -34,8 +39,11 @@ class TestCase(ModelObject):
     repr_args = ('name',)
     __slots__ = ['parent', 'name', 'doc', 'timeout', 'lineno', '_setup', '_teardown']
 
-    def __init__(self, name='', doc='', tags=None, timeout=None, lineno=None,
-                 parent=None):
+    def __init__(self, name: str = '', doc: str = '',
+                 tags: Union[List[str], str, None] = None,
```
pekkaklarck, line 43: `tags` can also be `Tags` instance. It could also be a tuple or any other iterable but I'm not sure is that relevant.

src/robot/model/testcase.py
```diff
@@ -86,7 +94,7 @@ def setup(self):
         return self._setup
 
     @setup.setter
-    def setup(self, setup):
+    def setup(self, setup: Keyword):
```
pekkaklarck, line 97: `setup` can also be `None`.

src/robot/model/testcase.py
```diff
@@ -114,7 +122,7 @@ def teardown(self):
         return self._teardown
 
     @teardown.setter
-    def teardown(self, teardown):
+    def teardown(self, teardown: Keyword):
```
pekkaklarck, line 125: Can also be `None`.

src/robot/model/testcase.py
```diff
@@ -162,15 +170,15 @@ def longname(self):
     def source(self):
```
pekkaklarck, line 170: Return type hint missing here and possibly elsewhere.

src/robot/model/testcase.py
```diff
@@ -180,21 +188,21 @@ def to_dict(self):
...
     def __init__(self, test_class=TestCase, parent=None, tests=None):
         super().__init__(test_class, {'parent': parent}, tests)
 
     def _check_type_and_set_attrs(self, test):
         test = super()._check_type_and_set_attrs(test)
-        for visitor in test.parent._visitors:
+        for visitor in test.parent._visitors:   #type: ignore
             test.visit(visitor)
-        return test
+        return test
```
pekkaklarck, line 208: Files should end with a newline.

src/robot/model/testsuite.py
```diff
@@ -40,8 +44,11 @@ class TestSuite(ModelObject):
     __slots__ = ['parent', '_name', 'doc', '_setup', '_teardown',  'rpa',
                  '_my_visitors']
 
-    def __init__(self, name: str = '', doc: str = '', metadata: dict = None,
-                 source: Path = None, rpa: bool = False, parent: 'TestSuite' = None):
+    def __init__(self, name: str = '', doc: str = '',
+                 metadata: Union[dict, None] = None,
```
pekkaklarck, line 48: `metadata` can also be `Metadata`.

src/robot/model/testsuite.py
```diff
@@ -50,12 +57,12 @@ def __init__(self, name: str = '', doc: str = '', metadata: dict = None,
         self.rpa = rpa        #: ``True`` when RPA mode is enabled.
         self.suites = None
         self.tests = None
-        self._setup = None
-        self._teardown = None
+        self._setup: Union[Keyword, None] = None
+        self._teardown: Union[Keyword,  None] = None
```
pekkaklarck, line 61: Not part of the public API so type hints probably aren't needed.

src/robot/model/testsuite.py
```diff
@@ -50,12 +57,12 @@ def __init__(self, name: str = '', doc: str = '', metadata: dict = None,
...
         self._my_visitors = []
 
     @staticmethod
-    def name_from_source(source: Path):
+    def name_from_source(source: Union[Path, str, None]):
```
pekkaklarck, line 65: Return type missing.

src/robot/model/testsuite.py
```diff
@@ -83,37 +90,37 @@ def name(self):
...
 
     @setter
-    def tests(self, tests):
+    def tests(self, tests) -> TestCases:
         """Tests as a :class:`~.TestCases` object."""
         return TestCases(self.test_class, self, tests)
 
     @property
-    def setup(self):
+    def setup(self) -> Union[Keyword, None]:
```
pekkaklarck, line 123: `setup` and `teardown` are never `None` when accessed.

src/robot/model/testsuite.py
```diff
@@ -142,7 +149,7 @@ def setup(self):
         return self._setup
 
     @setup.setter
-    def setup(self, setup):
+    def setup(self, setup: Keyword) -> Union[Keyword, None]:
```
pekkaklarck, line 152: - Passed in `setup` can be `None`.
- This method doesn't return anything.

src/robot/model/visitor.py
```diff
@@ -159,10 +159,10 @@ def visit_test(self, test: TestCase):
         """
         if self.start_test(test) is not False:
             if test.has_setup:
-                test.setup.visit(self)
+                test.setup.visit(self)  # type: ignore
             test.body.visit(self)
             if test.has_teardown:
-                test.teardown.visit(self)
+                test.teardown.visit(self) # type: ignore
```
pekkaklarck, line 165: I don't like `# type: ignore` comments in general, but in these cases they shouldn't even be needed.

pekkaklarck (changes requested):
Great work but lot of things to enhance still. General problems:
- Some style violations especially related to imports.
- In some case types added to attributes that don't even exist.
- Return types and also parameter types missing in many places in public APIs.
- Some of the internal type hints look so complicated that they may not be useful.

Fixing everything in this PR and then re-reviewing is a pretty big task. I believe it would be best to start from a single module like `robot.model.testsuite` and type it thoroughly. Once we get that done, we set standard that can then be followed when typing other modules.

src/robot/model/testsuite.py
Serhiy1: Fixed

src/robot/model/testsuite.py
Serhiy1: Removed

src/robot/model/testsuite.py
Serhiy1: Fixed

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,35 @@
...
+				"python.analysis.diagnosticSeverityOverrides": {
+					"reportOptionalCall": "none"
+				}
+			}
+		}
+	}
+
+	// Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
+	// "remoteUser": "root"
+}
```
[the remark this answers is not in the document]
Serhiy1: This was accidentally added removing

src/robot/model/visitor.py
Serhiy1: I will remove these. Unfortunately, most IDEs don't understand that there is a valid check beforehand.  has_teardown and has_setup will need to implement [typeguards](https://peps.python.org/pep-0647) but they are not available until the very latest versions of python

Serhiy1:
> Fixing everything in this PR and then re-reviewing is a pretty big task. I believe it would be best to start from a single module like robot.model.testsuite and type it thoroughly. Once we get that done, we set standard that can then be followed when typing other modules.

@pekkaklarck I think I have gone through all the requested changes inside `robot.model.testsuite`

src/robot/model/visitor.py
pekkaklarck: `test.teardown` and others ought to have type hint `Keyword` so there shouldn't be errors calling `test.teardown.visit(self)`. At least earlier the type hint was `Keyword | None` which would have issues.

The `has_teardown` check is there to avoid accessing `test.teardown` property that creates the underlying `Keyword` object. It wouldn't be a problem otherwise, but creating lot of such objects unnecessarily consumes memory.

src/robot/model/testsuite.py
```diff
@@ -27,6 +27,10 @@
 from .tagsetter import TagSetter
 from .testcase import TestCase, TestCases
 
+from typing import Any, TYPE_CHECKING, Dict, List, Union
+if TYPE_CHECKING:
+    from robot.model.visitor import SuiteVisitor
+
```
pekkaklarck, line 33: - Standard module imports should be before internal imports.
- We discussed on #devel about using `'x | y'` `'list[x]'` style instead of  `Union[x, y]` and `list[x]`. That would avoid lot of `from typing` imports, although in this case we anyway needed to import `TYPE_CHECKING`.

src/robot/model/testsuite.py
```diff
@@ -40,8 +44,11 @@ class TestSuite(ModelObject):
     __slots__ = ['parent', '_name', 'doc', '_setup', '_teardown',  'rpa',
                  '_my_visitors']
 
-    def __init__(self, name: str = '', doc: str = '', metadata: dict = None,
-                 source: Path = None, rpa: bool = False, parent: 'TestSuite' = None):
+    def __init__(self, name: str = '', doc: str = '',
+                 metadata: Union[Metadata, None] = None,
+                 source: Union[Path, None] = None,
+                 rpa: bool = False,
+                 parent: Union["TestSuite", None] = None):
```
pekkaklarck, line 51: - `metadata` can also be `dict` and  `source` can also be `str`.
- Using stringly types like `'Metadata | dict | None'` would probably be better than using `Union[Metadata, dict, None]`.
- Please use single quotes instead of double quotes.
- We need to decide how to handle long signatures like this one. Splitting all parameters to own lines would be a simple solution, but it should be done consistently (i.e. also `name` and `doc` on own lines) and such formatting takes lot of vertical space. I'd try having multiple params per line, but we can make a final decision a bit later. Notice that we use 88 as the maximum line length.

src/robot/model/testsuite.py
```diff
@@ -302,7 +315,7 @@ def __str__(self):
         return self.name
 
     def to_dict(self):
-        data = {'name': self.name}
+        data: Dict[str, Any] = {'name': self.name}
```
pekkaklarck, line 318: I don't think this helps. The method should get `-> dict`, though.

Serhiy1: I added the type hint to suppress some reported type errors from VScode, with the first assignment, it assumes that keys and values for the dict are always strings raising errors all over the place. 

Will remove though.

src/robot/model/testsuite.py
Serhiy1: Should I switch to stringifying all the type hints in the file?

pekkaklarck:
> @pekkaklarck I think I have gone through all the requested changes inside `robot.model.testsuite`

I added few more comments. I mentioned two things to decide in them:
1. Should we use stringified types like `'x | y'` and `'list[int]'` or use `Union`, `List`, etc? We discussed this on #devel and most people seemed to prefer the former. I also like that style more, especially with unions, and it avoids lot of `from typing` imports.
2. How to format long signatures. Having each parameter on its own line would be an easy solution, but such style takes lot of vertical space. I believe we should try fitting multiple params per line.

How do you feel about the above? If you agree with me, please update `testsuite` to use such style. We can then see how it feels and decide should we use that style also with other files.

The next question is how to merge only `testsuite` now. I think I cannot, at least here on web, merge only one file, so a new PR would probably be easiest.

Serhiy1:
Changes to `TestSuite.py`  
- Re-ordered imports
- Switched to stringified type hints 
- Changed init signature to maximise horizontal character count before newlines

As for, just emerging TestSuite.py, I will roll back and locally stash my changes and use them in a follow-up PR

Issue #4343, opened 2022-05-11
Collections. Support case-insensitive comparisons


AllanMedeiros:
When comparing Lists or Dictionaries would be great if we could ignore case on item, value or both.
Something similar was done in the past with other kws: https://github.com/robotframework/robotframework/issues/2439
This would apply to these kws:
[Dictionaries Should Be Equal](https://robotframework.org/robotframework/latest/libraries/Collections.html#Dictionaries%20Should%20Be%20Equal)
[Dictionary Should Contain Item](https://robotframework.org/robotframework/latest/libraries/Collections.html#Dictionary%20Should%20Contain%20Item)
[Dictionary Should Contain Key](https://robotframework.org/robotframework/latest/libraries/Collections.html#Dictionary%20Should%20Contain%20Key)
[Dictionary Should Contain Sub Dictionary](https://robotframework.org/robotframework/latest/libraries/Collections.html#Dictionary%20Should%20Contain%20Sub%20Dictionary)
[Dictionary Should Contain Value](https://robotframework.org/robotframework/latest/libraries/Collections.html#Dictionary%20Should%20Contain%20Value)
[Dictionary Should Not Contain Key](https://robotframework.org/robotframework/latest/libraries/Collections.html#Dictionary%20Should%20Not%20Contain%20Key)
[Dictionary Should Not Contain Value](https://robotframework.org/robotframework/latest/libraries/Collections.html#Dictionary%20Should%20Not%20Contain%20Value)

And List kws:
[List Should Contain Sub List](https://robotframework.org/robotframework/latest/libraries/Collections.html#List%20Should%20Contain%20Sub%20List)
[List Should Contain Value](https://robotframework.org/robotframework/latest/libraries/Collections.html#List%20Should%20Contain%20Value)
[List Should Not Contain Duplicates](https://robotframework.org/robotframework/latest/libraries/Collections.html#List%20Should%20Not%20Contain%20Duplicates)
[List Should Not Contain Value](https://robotframework.org/robotframework/latest/libraries/Collections.html#List%20Should%20Not%20Contain%20Value)
[Lists Should Be Equal](https://robotframework.org/robotframework/latest/libraries/Collections.html#Lists%20Should%20Be%20Equal)

pekkaklarck:
Sounds good to me. Are you @AllanMedeiros, or someone else, interested to create a PR? Should be pretty straightforward. The only design decision I see is deciding should with dicts this be applied both to keys and values.

[...]

PR #4690, opened 2023-03-14
Feature/4343 add ignore case to list and dictionary comparisons

MobyNL:
As mentioned in issue https://github.com/robotframework/robotframework/issues/4343, this is my first time contributing. I would love to have feedback and I'm still working on complying to coding principles


MobyNL:
Thanks for the feedback @pekkaklarck. Some nice learnings and certainly good suggestions/improvements. I'm gonna work on it!

MobyNL:
I updated and implemented some of the feedback mentioned above. Some additional comments/questions from my part:

1. _normalize is now a function in both classes _List and _Dictionary. Considering the function is identical in both, we could consider making a general helper for this. I was thinking how to best do this? There already is a normalizing.py in utils, which has a different approach than how it is now used in Collections.
2. I added a general function in _Dictionary, to check whether to ignore_case for key and/or value. I used a dictionary to return the values, because I did not always need both (and always returning both would sometimes result in unused variables). It is also possible to use multiple functions and split the variables, but I thought it was a nice approach. Maybe someone has additional comments.
3. I added some tests to check the sad flow when ignore_case=False (or empty), but we could also consider changing the test data for some of the existing tests of the base functions. However, that would mean you are starting to test 2 things at once for those testcases. 
For example: the test Dictionary Should Contain Item currently looks at a key x, while the dictionary has keys a, b and 3. If we changed x to A, we should also get a Fail on the same testcase. However, if something would be wrong in ignore_case, the test might wrongfully pass, compare to using an entirely different key.

I also tried to uphold the PEP-8 standards, and also used pylint with an adjusted max 88 characters per line. So hopefully it's a bit closed to the standards (if not fully compliant) than before

MobyNL:
I updated some changes for #4343. Please let me know if there are any more remarks, and whether I should change this from a draft pull request to a regular one

src/robot/libraries/Collections.py
```diff
@@ -698,60 +748,108 @@ def get_from_dictionary(self, dictionary, key, default=NOT_SET):
...
+            dictionary = self._normalize(dictionary)
+            key=self._normalize(key)
+            value=self._normalize(value)
+        elif should_ignore['key']:
+            dictionary = {self._normalize(k): dictionary[k] for k in dictionary}
+            key=self._normalize(key)
+        elif should_ignore['value']:
+            dictionary = {k: self._normalize(dictionary[k]) for k in dictionary}
+            value=self._normalize(value)
         assert_equal(dictionary[key], value,
```
pekkaklarck, line 811: This can be massively simplified by using the `Normalizer` class. There are also other places where it's useful but I won't mention it unless there's something special.

src/robot/libraries/Collections.py
```diff
@@ -698,60 +748,108 @@ def get_from_dictionary(self, dictionary, key, default=NOT_SET):
...
+        if should_ignore['value']:
+            dictionary = self._normalize(dictionary)
+            value=self._normalize(value)
         _verify_condition(value not in dictionary.values(),
                           f"Dictionary contains value '{value}'.",
                           msg)
 
-    def dictionaries_should_be_equal(self, dict1, dict2, msg=None, values=True):
+    def dictionaries_should_be_equal(self, dict1, dict2, msg=None, values=True,
+                                     ignore_case=False):
```
pekkaklarck, line 852: This keyword has got new `ignore_keys` argument since this PR was created. It creates a conflict, but it ought to be pretty easy to resolve.

Responsibility to actually ignore keys could be moved to the `Normalize` class. Probably not a good idea to do that as part of this PR, though.

src/robot/libraries/Collections.py
```diff
@@ -804,9 +916,14 @@ def _log_dictionary(self, dictionary):
...
 
-    def _keys_should_be_equal(self, dict1, dict2, msg, values):
+    def _keys_should_be_equal(self, dict1, dict2, msg, values, ignore_case):
         keys1 = self.get_dictionary_keys(dict1)
         keys2 = self.get_dictionary_keys(dict2)
+        if ignore_case:
+            keys1 = self._normalize(dict1)
+            keys2 = self._normalize(dict2)
+            dict1 = self._normalize(dict1)
+            dict2 = self._normalize(dict2)
```
pekkaklarck, line 926: There shouldn't be need to update this helper method at all. I believe normalization can be done already earlier. Same with the similar helper methods below.

src/robot/libraries/Collections.py
```diff
@@ -835,6 +961,29 @@ def _validate_dictionary(self, dictionary, position=1):
...
+        elif self._normalize(ignore_case)=="value":
+            ignore_case_key=False
+            ignore_case_value=True
+        elif self._normalize(ignore_case)=="both" or ignore_case is True:
+            ignore_case_key=True
+            ignore_case_value=True
+        else:
+            ignore_case_key=False
+            ignore_case_value=False
+        return {'key': ignore_case_key, 'value': ignore_case_value}
```
pekkaklarck, line 986: Both of these helper methods can be removed and normalization handled fully by the `Normalizer` class.

src/robot/libraries/Collections.py
```diff
@@ -1034,4 +1206,4 @@ def _get_matches_in_iterable(iterable, pattern, case_insensitive=False,
                       spaceless=is_truthy(whitespace_insensitive),
                       regexp=regexp)
     return [string for string in iterable
-            if is_string(string) and matcher.match(string)]
+            if is_string(string) and matcher.match(string)]
```
pekkaklarck, line 1209: Newline missing at the end of the file. This has actually caused a conflict because the above row was changed. That underlines one reason to always end files with newlines. Luckily fixing the conflict is easy.

pekkaklarck (changes requested):
This looks very good otherwise, but all keywords handling normalization separately adds lot of code even though there's a common `_normalize` helper. The `Normalizer` class I propose ought to remove most of that complexity.

I'm planning to create RF 6.1 rc 1 on Monday, June 5. There's not much time to get this PR merged before that, but because this change only affects `Collections` and ought to be pretty safe in general, I'm willing to make an exception and merge this after the release candidate is out. The final release is targeted for Monday, June 12, so there's not too much time for that either.

Notice also that there have been some changes to `Collections` after this PR was created. As the result there are some conflicts, but they ought to be easy to fix.

Let me know here or on Slack if you need any help!

MobyNL:
Thanks @pekkaklarck. I was a bit too busy to work on this but will take a look at the feedback in the coming days!

src/robot/libraries/Collections.py
```diff
@@ -304,27 +304,43 @@ def sort_list(self, list_):
         self._validate_list(list_)
         list_.sort()
 
-    def list_should_contain_value(self, list_, value, msg=None):
+    def list_should_contain_value(self, list_, value, msg=None, ignore_case=False):
         """Fails if the ``value`` is not found from ``list``.
 
         Use the ``msg`` argument to override the default error message.
+
+        See section `Ignore Case` for more information about ignoring case in lists.
```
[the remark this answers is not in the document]
MobyNL: I copied the text, but changed it to say It is new in Robot Framework 6.2 (may need work to match the correct version, if 6.1.1 or something is going to be used)

MobyNL:
@pekkaklarck I updated this, as I saw it had merge conflicts with the current version. And just a reminder that it's still open

src/robot/libraries/Collections.py
```diff
@@ -303,27 +303,34 @@ def sort_list(self, list_):
...
         list_.sort()
 
-    def list_should_contain_value(self, list_, value, msg=None):
+    def list_should_contain_value(self, list_, value, msg=None, ignore_case=False):
         """Fails if the ``value`` is not found from ``list``.
 
         Use the ``msg`` argument to override the default error message.
+
+        The ignore_case argument can be used to make comparison case-insensitive. 
+        See the Ignore case section for more details. It is new in Robot Framework 6.2.
```
MobyNL, line 312: If it is added to version 7.0 we need to change this documentation

src/robot/libraries/Collections.py
```diff
@@ -303,27 +303,34 @@ def sort_list(self, list_):
...
+                          f"{seq2str2(list_)} does not contain value '{value}'.", msg)
 
-    def list_should_not_contain_value(self, list_, value, msg=None):
+    def list_should_not_contain_value(self, list_, value, msg=None, ignore_case=False):
         """Fails if the ``value`` is found from ``list``.
 
         Use the ``msg`` argument to override the default error message.
+
+        The ignore_case argument can be used to make comparison case-insensitive. 
+        See the Ignore case section for more details. It is new in Robot Framework 6.2.
```
MobyNL, line 325: Same

src/robot/libraries/Collections.py
```diff
@@ -333,22 +340,26 @@ def list_should_not_contain_duplicates(self, list_, msg=None):
 
         This keyword works with all iterables that can be converted to a list.
         The original iterable is never altered.
+
+        The ignore_case argument can be used to make comparison case-insensitive. 
+        See the Ignore case section for more details. It is new in Robot Framework 6.2.
```
MobyNL, line 345: Same (last comment, more in code)

MobyNL:
> @pekkaklarck I updated this, as I saw it had merge conflicts with the current version. And just a reminder that it's still open

Maybe it could be added to version 7.0?

pekkaklarck:
I enabled running this PR on GitHub Actions and one test failed:
```
FAIL: Robot.Variables.Using Dict Variables.Multiple dict variables with same names multiple times
Status of 'Multiple dict variables with same names multiple times' should have been PASS but it was FAIL.

Error message:
Following keys have different values:
Key d: 4 != 1
Key a: 1 != 2
Key b: 2 != 3
Key c: 3 != 4
```
That test uses `Dictionaries Should Be Equal` with dicts that have items in different orders. I found where the bug is in the code and will add a comment there.

In addition to fixing the issue, it's a good idea to add a new test for this library that validates that `Dictionary Should Be Equal` doesn't require dicts to have same order. In practice this should succeed:
```
${dict1} =    Create Dictionary    a=1    b=2    c=3    d=4    e=5
${dict2} =    Create Dictionary    d=4    b=2    e=5    a=1    c=3
Dictionaries Should Be Equal    ${dict1}    ${dict2}
```

We could also enhance the keyword to actually check the order if both dicts are `OrderedDict` instances. That's relevant related to this PR and a separate issue and PR should be submitted instead.

pekkaklarck:
As I commented issue #4343, I won't have time for a proper review right now (and the aforementioned bug anyway needs to be fixed before merging this), but I added the issue to RF 7.0 scope that it won't be forgotten.

src/robot/libraries/Collections.py
```diff
@@ -842,16 +900,19 @@ def _keys_should_be_equal(self, dict1, dict2, msg, values):
...
                           msg, values)
 
-    def _yield_dict_diffs(self, keys, dict1, dict2):
-        for key in keys:
+    def _yield_dict_diffs(self, dict1, dict2, ignore_case):
+        normalize = Normalizer(ignore_case).normalize
+        for key1, key2 in zip(normalize(dict1), normalize(dict2)):
             try:
-                assert_equal(dict1[key], dict2[key], msg=f'Key {key}')
+                assert_equal(dict1[key1], dict2[key2], msg=f'Key {key1}')
```
pekkaklarck, line 915: This implementation isn't correct and causes a failure in an unrelated suite that uses `Dictionaries Should Be Equal` with dicts having items in different order. The old `for key in keys` approach avoided the issue and should be restored. `normalize` can then be used with each value like `assert_equal(normalize(dict1[key]), ...)`.

MobyNL: I fixed this, by resetting it to for key in keys, so this behaviour is fixed. Now differently ordered dictionaries also work. I also added some tests for this. I did this by change it to `for key in keys:`

I do however now encounter a problem at various tests because of key errors, that I haven't yet solved. Those can be solved by changing it to `for key in normalize(keys):`, but this poses a new problem as keys is list-like, meaning it will always normalize, unless `ignore_case=False`. I will work on this some more when I have time tomorrow

MobyNL: I have now also fixed the error mentioned above in a very ugly way. I will commit it for now but put in some more time tomorrow to see if a cleaner solution is possible

pekkaklarck: Right, if you have `{'A': 1, 'B': 2}` and `{'a': 1, 'b': 2}`, keys are equal case-insensitively, but comparing values by iterating over keys of either of the dicts and doing `d1[k] == d2[k]` fails. Keys are returned by the method comparing keys, and it could possibly return a list of two-tuples containing matching keys of both of the dicts.

pekkaklarck: Things get even more complicated if you have `{'a': 1, 'A': 2}` and `{'a': 2, 'A': 1}`. Keys are equalnand I guess we can consider values to match case-insensitively as well. Value comparison will fail, however, if comparison is done using key pairs `[('a', 'a'), ('A', 'A')]`. This gets even more "fun" with dicts having keys like `'abc', 'Abc', 'aBc', 'abC', 'ABc', 'AbC', 'aBC', 'ABC'` so that they have same values but with different keys. I have and idea how to handle this in ganeric manner, but I'm currently on mobile and writing code examples is too hard. I probably should test my idea as well.

MobyNL: Hmm, that would be a solution, but we don't know the actual pairs. The current `_keys_should_be_equal` just checks if the same keys exist in the dictionaries.

MobyNL: I made an example using `keys = [(key1, key2) for key1, key2 in
                zip(dict1.keys(), dict2.keys()) if normalize(key1) == normalize(key2)]`
It works, however, now a new challenge rises when amount of keys is not equal. So, still ongoing

MobyNL: Ah, I found it. In `dictionary_should_contain_sub_dictionary` we also create a set of keys. This should return the couples as well

MobyNL: This doesn't solve the problem yet for keys with double names. But I think that might be something that can't be solved. Because you don't know the actual pairs. If i create 2 dicts:

```
dict1 = {'abc': 3, 'AbC':4}
dict2 = {'ABC':4, 'Abc': 3}
```
It is now impossible to check if this is correct. I have 2 keys that are 3 and 2 that are 4, but i don't actually know if they are similar. Especially if they are also in a different order.

I think that if you have the identically named keys (except for case), that you shouldn't be able to use ignore_case. So we could add an error message when multiple keys are present.

pekkaklarck: I forgot to test it, but I believe this algorithm would solve the above problem:
1. Collect all values with matching keys to lists. In the above case we'd get lists `[3, 4]` and [4, 3]`.
2. Validate that lists have same items.This wouldn't be that easy because we cannot simply sort lists because they could be unsortable nor could we use sets because values aren't necessarily hashable.

Even though we could make the above work, I'm not sure is it worth the effort or a good idea in general. Making it an error to have multiple keys that normalize to the same value could be better.

MobyNL: I have now added an error in `_keys_should_be_equal`. I used:
`len(set(normalize(dict1).keys()))!=len(dict1.keys())`
I think this is valid, since normalize changes dict_like objects to a dictionary. So I can use sets too, after the normalizing

MobyNL:
I also ran the atests locally for collections and using_dict_variables that had a fail earlier. They all still pass, as well as the newly created ones.

MobyNL:
> I also ran the atests locally for collections and using_dict_variables that had a fail earlier. They all still pass, as well as the newly created ones.

Now the rest has also run. Everything except for the Telnet testcases passed. But those have never worked locally. I guess I need some installation to have those working too

pekkaklarck:
Making it an error if dicts have duplicate keys after normalization is a good solution for handling cases like `{'a': 1, 'A': 2}` vs `{'A': 1, 'a': 2}`. There are some style issues in the code, but it's probably easiest if I merge this PR and fix those myself. Thanks a lot for contribution @MobyNL!

pekkaklarck:
While making the small enhancements I mentioned above, I noticed `Dictionary Should Contain Item` doesn't work correctly in all cases. The problem occurs, for example, with a dictionary `{'X': 1}` when using key `x` or `X` because the keyword tries to get a normalized key from the original dictionary before normalizing the value like `normalize(dictionary[normalize(key)])`. Trying to fix by normalizing the whole dictionary like `normalize(dictionary)[normalize(key)]` fixes this situation, but breaks if someone has used `normalize=key` or `normalize=value`.

I believe a proper fix requires adding separate `normalize_key` and `normalize_value` methods to `Normalizer` to allow getting the compared value like `normalizer.normalize_dict(dictionary)[normalizer.normalize_key(key)]` and then normalizing the expected value like `normalizer.normalize_value(value)`. I've already done this locally and it seems to work. I'll commit it along with other cleanup I mentioned above (and other cleanup I do this somewhat old library at the same time).

pekkaklarck:
As I commented #4343, this library now has some keywords that accept `case_insensitive` and others that accept `ignore_case`. The latter is used also by BuiltIn and I like it more also otherwise. I'll change the others to accept it as well, but for backwards compatibility reasons `case_insensitive` should still work for the time being.

PR #4703, opened 2023-03-26
created keyword to convert list to string

Indraja22:
Created a new keyword in the string library to convert list to string.

### DIFF
--- a/src/robot/libraries/String.py
+++ b/src/robot/libraries/String.py
@@ -802,3 +802,9 @@ class String:
         if not message:
             message = default_template % tuple(safe_str(item) for item in items)
         raise AssertionError(message)
+
+    def convert_list_to_string(self, list, separator):
+        """Converts list to string.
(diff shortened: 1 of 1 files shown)


Indraja22:
Created a new keyword in the string library to convert list to string.

pekkaklarck:
The proposed keyword is very close to the existing `Catenate`, but the separator being a separate argument could make the usage more convenient in some usages. There are, however, various problems with the proposal:
- I don't think `Convert List To String` is that good name because the main functionality is joining list items with the given separator. Something like `Join List` would be better.
- Added keyword lacks tests.
- Documentation needs to be improved. I believe the example doesn't even work.
- Should the keyword convert non-string values to strings? What about possible nested lists?
- Each enhancement needs an issue for tracking purposes. Issues are also good for discussing about the design like keyword name and what to do with non-string items.

If you @Indraja22 are still interested in this, please submit an issue about the enhancement. Once we agree how the keyword should work, this PR can be adjusted. If you need help e.g. with tests, I can help.

PR #4714, opened 2023-04-01
adding a simple test for the zipapp usage of robotframework

franzhaas:
There is an ugly hack to detect which python versions are supposed to work with zipapps in github actions...

Once the minimum python version for robotframework is 3.7.:

We can eliminate this hack provided we support zipapps natively in robotframework (using importlib.resources.open_text for python < 3.9 and importlib.resources.files for python >= 3.9).

The choice to use importlib.resources.open_text for backwards compatibility once 3.6 is no longer supported would reduce the usage of __file__ in the part of robotframework which gets distributed as a wheel to.:

pythonpathsetter.py
utils/importer.py (using getattr)

Both places are not problematic for zipapps.

### DIFF
--- a/.github/workflows/acceptance_tests_cpython.yml
+++ b/.github/workflows/acceptance_tests_cpython.yml
@@ -92,7 +92,7 @@ jobs:
         run: |
           python -m zipapp -m "robot:run_cli" src -o robot.pyz -c
           python robot.pyz --output NONE atest/testdata/core/test_suite_dir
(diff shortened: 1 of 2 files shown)


pekkaklarck:
I'm closing this PR for two reasons:
- Tests should be integrated to our existing unit test or acceptance system and they should be executable locally.
- We already have a unit test that all `__file__` usages are safe (e5082cf5da).

PR #4719, opened 2023-04-05
Add type hints for model.testcase

Serhiy1:
1. Added type hints to `model.testcase`
2. minor change to `model.testsuite` to keep consistent with the convention of no spaces inside stringified type hints

### DIFF
--- a/src/robot/model/testcase.py
+++ b/src/robot/model/testcase.py
@@ -122,7 +122,7 @@ class TestCase(ModelObject):
         return self._teardown
 
     @teardown.setter
-    def teardown(self, teardown: 'Keyword|None'):
+    def teardown(self, teardown: 'Keyword|Mapping|None'):
         self._teardown = create_fixture(teardown, self, Keyword.TEARDOWN)
 
     @property
@@ -201,9 +201,9 @@ class TestCase(ModelObject):
 class TestCases(ItemList[TestCase]):
     __slots__ = []
 
-    def __init__(self, test_class: Type[TestCase]= TestCase,
+    def __init__(self, test_class: Type[TestCase] = TestCase,
                  parent: 'TestSuite|None' = None,
-                  tests: 'Sequence[TestCase|Mapping]' = ()):
+                 tests: 'Sequence[TestCase|Mapping]' = ()):
         super().__init__(test_class, {'parent': parent}, tests)
 
     def _check_type_and_set_attrs(self, test):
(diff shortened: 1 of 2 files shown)


src/robot/model/testsuite.py
```diff
@@ -148,7 +148,7 @@ def setup(self) -> Keyword:
         return self._setup
 
     @setup.setter
-    def setup(self, setup: 'Keyword | None'):
+    def setup(self, setup: 'Keyword|None'):
```
pekkaklarck, line 151: Good catch! As I mentioned above, we should add `Mapping` here as well. I let you to do that to avoid creating a conflict.

pekkaklarck (changes requested):
Looks great! Few simple things to be fixed, though.

src/robot/model/testcase.py
```diff
@@ -189,14 +198,16 @@ def to_dict(self):
         return data
 
 
-class TestCases(ItemList):
+class TestCases(ItemList[TestCase]):
     __slots__ = []
 
-    def __init__(self, test_class=TestCase, parent=None, tests=None):
+    def __init__(self, test_class: Type[TestCase]= TestCase,
+                 parent: 'TestSuite|None' = None, tests: Sequence[TestCase] = ()):
```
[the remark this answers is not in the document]
Serhiy1: Adding Mapping here raises an error from vscode. Looks like `ItemList` needs to be updated to use the newly agreed-upon conventions (stringified type hints, using mapping instead of dict, etc).

src/robot/model/testcase.py
```diff
@@ -104,21 +112,21 @@ def has_setup(self):
...
         See :attr:`setup` for more information.
         """
-        if self._teardown is None and self:
+        if self._teardown is None:
             self._teardown = create_fixture(None, self, Keyword.TEARDOWN)
         return self._teardown
 
     @teardown.setter
-    def teardown(self, teardown):
+    def teardown(self, teardown: 'Keyword|None'):
```
pekkaklarck, line 125: `Mapping` missing.

src/robot/model/testcase.py
```diff
@@ -189,14 +198,17 @@ def to_dict(self):
...
 
 
-class TestCases(ItemList):
+class TestCases(ItemList[TestCase]):
     __slots__ = []
 
-    def __init__(self, test_class=TestCase, parent=None, tests=None):
+    def __init__(self, test_class: Type[TestCase]= TestCase,
+                 parent: 'TestSuite|None' = None,
+                  tests: 'Sequence[TestCase|Mapping]' = ()):
```
pekkaklarck, line 206: Inconsistent indentation.

PR #4722, opened 2023-04-05
Add on_limit option to WHILE

### LINKED ISSUE #4562: Possibility to continue execution after WHILE limit is reached
Hello everyone,

I'm creating this issue to discuss about a potential evolution of the WHILE loop behaviour when the limit is reached. As a reminder, in the current implementation of the WHILE loop, when the limit is reached, execution fails with the following message `WHILE loop was aborted because it did not finish within the limit of 5 iterations. Use the 'limit' argument to increase or remove the limit if needed.` (if the limit is 5 iterations).

In this PR #4561, I suggested to add a third argument in order to be able to set a custom error message when the limit is reached. @pekkaklarck suggested to use this third argument to control what to do if the limit is reached and, for example, allow execution to continue even if the limit is reached.

I have been thinking about it and as a result, I imagined the following scenarios :

```
*** Test Cases ***
Continue after limit is reached
    [Documentation]  The execution will continue after the WHILE loop even if the limit is reached
    WHILE    True    limit=5    limit_reached_behavior=CONTINUE
        Log     Test
    END

Fail with custom error message if limit is reached
    [Documentation]  The execution will fail with custom error message if the limit is reached
    WHILE    True    limit=5    limit_reached_behavior=Fail    Custom error message
        Log     Test
    END

Run keyword and continue if limit is reached
    [Documentation]  The execution will execute the keyword and continue execution if the limit is reached
    WHILE    True    limit=5    limit_reached_behavior=Example keyword
        Log     Test
    END

*** Keywords ***
Example keyword
    Log     Example
```

If `limit_reached_behavior` equals `CONTINUE`, the execution simply continues if the limit is reached. 
If `limit_reached_behavior` doesn't equal `CONTINUE`, the keyword given in `limit_reached_behavior` is executed.

What do you think of such scenarios ? Feel free to suggest other scenarios !

I am not very comfortable with the idea of using `CONTINUE` to continue execution if the limit is reached because the string `CONTINUE` is already used to stop executing the current loop iteration and continues to the next one.

It would also be very interesting to discuss what the option name should be.

Such scenarios also beg the following question : Where and How should we display the keyword executed after reaching the limit in `log.html` ?

Thank you very much for taking the time to read this issue !

yanne:
This implementation allows two values, `pass`, where the execution is continued even when the limit is exceeded and `fail`, (the default) where the loop fails when the limit is exceeded.

Fixes #4562


src/robot/running/bodyrunner.py
```diff
@@ -654,23 +661,34 @@ def create(cls, limit, on_limit_message, variables):
...
-                                  f"it did not finish "
-                                  f"within the limit of {self}. "
-                                  f"Use the 'limit' argument to "
-                                  f"increase or remove the limit if needed.")
+            raise LimitExceeded(self.on_limit,
+                                f"WHILE loop was aborted because "
+                                f"it did not finish "
+                                f"within the limit of {self}. "
+                                f"Use the 'limit' argument to "
+                                f"increase or remove the limit if needed.")
```
pekkaklarck, line 691: It looks like this could be formatted to take a little less vertical space. Just making individual text lines longer (up to 88 chars) ought to be enough, but formatting like
```python
raise LimitExceeded(
    ...
)
```
is an option as well.

src/robot/running/bodyrunner.py
```diff
@@ -631,18 +638,18 @@ def _run_finally(self, data, run):
...
+        self.on_limit = on_limit
         self.on_limit_message = on_limit_message
 
     @classmethod
-    def create(cls, limit, on_limit_message, variables):
+    def create(cls, limit, on_limit, on_limit_message, variables):
         if on_limit_message:
-            on_limit_message = variables.replace_string(
-                on_limit_message)
+            on_limit_message = variables.replace_string(on_limit_message)
```
pekkaklarck, line 648: Possible error in replacing variables should probably be handled better and would need a test. Not really related to `on_limit` but could be fixed as part of this change anyway.

src/robot/running/bodyrunner.py
```diff
@@ -716,3 +734,14 @@ class NoLimit(WhileLimit):
...
+
+class LimitExceeded(ExecutionFailed):
+
+    def __init__(self, on_limit, message):
+        super().__init__(message)
+        self.on_limit = on_limit
+
+    @property
+    def on_limit_pass(self):
+        return self.on_limit == 'pass'
```
pekkaklarck, line 747: Comparing strings here feels a bit unsafe. Possibly `on_limit_pass` should be given directly to `__init__` i.e. the usage would be `LimitExceeded(self.on_limit == 'pass', message)`.

pekkaklarck (changes requested):
Looks good. My comments were mostly about style, but it looks like continuable failures wouldn't be properly handled and few other tests could be added as well.

doc/userguide/src/ExtendingRobotFramework/ListenerInterface.rst
```diff
@@ -270,6 +270,8 @@ it. If that is needed, `listener version 3`_ can be used instead.
    |                  |                  |                                                                |
    |                  |                  | * `condition`: The looping condition.                          |
    |                  |                  | * `limit`: The maximum iteration limit.                        |
+   |                  |                  | * `on_limit`: What to do if the limit is exceeded.             |
+   |                  |                  | Valid values are `pass` and `fail`. New in RF 6.1.             |
```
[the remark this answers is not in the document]
yanne: For some reason the diff here has two lines, but in reality it is in a single line.

atest/robot/running/while/on_limit.robot
```diff
@@ -1,11 +1,29 @@
 *** Settings ***
-Suite Setup       Run Tests    ${EMPTY}    running/while/on_limit_message.robot
+Suite Setup       Run Tests    ${EMPTY}    running/while/on_limit.robot
 Resource          while.resource
 
 *** Test Cases ***
+On limit pass without limit defined
+    Check Test Case    ${TESTNAME}
```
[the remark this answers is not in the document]
yanne: Not sure about the second part, can of course be done. It's probably quite rare that one would execute the loop exactly 10000 times.

src/robot/running/bodyrunner.py
```diff
@@ -410,13 +412,18 @@ def run(self, data):
...
+                    if isinstance(failed, LimitExceeded):
+                        limit_exceeded = failed
                     errors.extend(failed.get_errors())
                     if not failed.can_continue(ctx, self._templated):
                         break
                 iter_result = loop_result.body.create_iteration(starttime=get_timestamp())
                 if not self._should_run(data.condition, ctx.variables):
                     break
-            if errors:
+            should_raise = not (limit_exceeded and limit_exceeded.on_limit_pass)
```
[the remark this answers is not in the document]
yanne: The first case actually works in the current implementation, but the second one fails.

src/robot/running/bodyrunner.py
yanne: Not sure if it matters where the unsafe string comparison is?

PR #4723, opened 2023-04-06
Add type hints for robot.model.keywords

Serhiy1:


src/robot/model/keyword.py
```diff
@@ -13,11 +13,16 @@
...
 
+from typing import Sequence, TYPE_CHECKING, Type
 import warnings
 
 from .body import Body, BodyItem
 from .itemlist import ItemList
 
+if TYPE_CHECKING:
+    from robot.model.visitor import SuiteVisitor
+    from robot.model.testsuite import TestSuite
```
pekkaklarck, line 24: Imports should be in alphabetical order. #nitpick

src/robot/model/keyword.py
```diff
@@ -29,43 +34,46 @@ class Keyword(BodyItem):
     repr_args = ('name', 'args', 'assign')
     __slots__ = ['_name', 'args', 'assign', 'type']
 
-    def __init__(self, name='', args=(), assign=(), type=BodyItem.KEYWORD, parent=None):
+    def __init__(self, name: str = '', args: Sequence[str] = (),
+                 assign: Sequence[str] = (), type: str = BodyItem.KEYWORD,
+                 parent: 'TestSuite|None' = None):
```
pekkaklarck, line 39: `Keyword` parent can be at least `TestSuite`, `TestCase`, and `Keyword`, but I believe it could also be `For`, `If`, etc. Probably the typing should be just `BodyItem|None` here.

src/robot/model/keyword.py
```diff
@@ -29,43 +34,46 @@ class Keyword(BodyItem):
...
-    def __str__(self):
+    def __str__(self) -> str:
         parts = list(self.assign) + [self.name] + list(self.args)
         return '    '.join(str(p) for p in parts)
 
-    def to_dict(self):
-        data = {'name': self.name}
+    def to_dict(self) -> dict:
+        data = {}
+        data['name'] = self.name
```
pekkaklarck, line 68: I noticed that PyCharm complains about this and recommends using `data =  {'name': self.name}` instead. As we know, that causes issues with VScode. PyCharm warning is less severe than VScode error, so this format is better. Adding type hint like `data: 'dict[str, Any]' = {'name': self.name}` like you at one point did could also be considered. You can choose.

pekkaklarck (changes requested):
Looks good otherwise but `parent` type is wrong.

src/robot/model/keyword.py
Serhiy1: `data: 'dict[str,list|str]' =  {'name': self.name}` also makes vscode happy without needing any imports

src/robot/model/keyword.py
pekkaklarck: Good point. Could also have `ModelObject`, but that's too broad.

src/robot/model/keyword.py
```diff
@@ -13,11 +13,17 @@
...
+from typing import Sequence, TYPE_CHECKING, Type
 import warnings
 
 from .body import Body, BodyItem
 from .itemlist import ItemList
 
+if TYPE_CHECKING:
+    from robot.model.testcase import TestCase
+    from robot.model.testsuite import TestSuite
+    from robot.model.visitor import SuiteVisitor
```
pekkaklarck, line 25: Using `from .testcase import TestCase` would be more consistent with our import style. If there's nothing else the change, I merge this PR and this can be changed later.

pekkaklarck (approved):

src/robot/model/keyword.py
Serhiy1: `TestCase` and `testSuite` will also need to be updated as well to match this import style

PR #4724, opened 2023-04-06
Add type hints for model.body

Serhiy1:


src/robot/model/body.py
```diff
@@ -120,39 +136,39 @@ def create(self):
...
         )
 
-    def create_keyword(self, *args, **kwargs):
+    def create_keyword(self, *args, **kwargs) -> 'Keyword':
         return self._create(self.keyword_class, 'create_keyword', args, kwargs)
 
-    def _create(self, cls, name, args, kwargs):
+    def _create(self, cls: 'Type[C]|None', name: str, args, kwargs) -> C:
         if cls is None:
             raise TypeError(f"'{full_name(self)}' object does not support '{name}'.")
```
pekkaklarck, line 144: I don't think `cls` here should have `None` as a valid type. Passing `None` here would be a bug and we even raise a TypeError in that case.

`args` and `kwargs` should have typing like `Iterable` and `Mapping`, respectively.

src/robot/model/body.py
```diff
@@ -199,7 +215,7 @@ def _filter(self, types, predicate):
             items = [item for item in items if predicate(item)]
         return items
 
-    def flatten(self):
+    def flatten(self) -> 'list[BodyItem|Try|If]':
```
pekkaklarck, line 218: This method effectively removes all `Try` and `If` objects from the returned list so they shouldn't be listed in the return type either.

src/robot/model/body.py
```diff
@@ -223,15 +240,18 @@ class Body(BaseBody):
     pass
 
 
-class Branches(BaseBody):
+class Branches(BaseBody, Generic[T]):
```
pekkaklarck, line 243: This is wrong. `Try` and `If` objects have `body` that has type `Branches`. `Branches` doesn't contain `Try` and `If` objects, it contains `TryBranch` and `IfBranch` objects.

Also `__init__` below has wrong typing due to this.

pekkaklarck (changes requested):
This is a bit trickier module to type properly. See separate comments for details.

src/robot/model/body.py
```diff
@@ -43,7 +54,7 @@ class BodyItem(ModelObject):
     __slots__ = ['parent']
 
     @property
-    def id(self):
+    def id(self) -> 'str|None':
```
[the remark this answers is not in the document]
Serhiy1: Returning an empty string feels like the most sensible approach.  But there is a comment that makes me apprehensive about changing it. 

```py
# This algorithm must match the id creation algorithm in the JavaScript side
# or linking to warnings and errors won't work.
```

src/robot/model/body.py
```diff
@@ -57,14 +68,17 @@ def id(self):
             return 'k1'
         return self._get_id(self.parent)
 
-    def _get_id(self, parent):
+    def _get_id(self, parent: 'TestSuite|TestCase|T') -> str:
```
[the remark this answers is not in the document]
Serhiy1: `T` is bound to the `Try` and `If` classes. As far as I can tell those are the only two body items with a `.body` attribute.

that's needed to type 
```python
if hasattr(parent, 'body'):
    steps.extend(step for step in parent.body.flatten()
                         if step.type != self.MESSAGE)
```

pekkaklarck:
Have you @Serhiy1 had time to look at my comments? Now that RF 6.1 beta 1 is finally out, I plan to concentrate on TestSuite type hints and it probably is a good idea to start by merging this PR. If you don't have time to look at my comments, I can fix issues myself.

src/robot/model/body.py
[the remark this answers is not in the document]
pekkaklarck: On Javascript side the structure is flattened so that there are no `If` or `Try` roots.

Because `id` should never be accessed from `If/Try` roots, I believe it's best to change them to raise a TypeError in that case.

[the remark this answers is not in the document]
Serhiy1: The TypeChecker mainly complains about the following clause 
```
    @property
    def id(self) -> 'str':
        if not self:
            return None
```

I'll update this to return an empty string

src/robot/model/body.py
Serhiy1: I added the None, as the type checker complains about the `xxx_class` Class attributes Since they are initialised with `None` . 

image
 
Not accepting `None` here causes all of the `create_xxx` keywords to report typing errors.

Serhiy1: Alternatively, I can do the following, and remove `None`

image

src/robot/model/body.py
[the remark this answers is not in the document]
Serhiy1: Changing this to return a string caused test failure, So I've reverted.

src/robot/model/body.py
pekkaklarck: Somewhat annoying to have such big `if/else`, but it's probably better than the other alternative. I can merge it but we can try to come up with something better later.

src/robot/model/body.py
```diff
@@ -120,39 +147,40 @@ def create(self):
...
 
-    def create_for(self, *args, **kwargs):
+    def create_keyword(self, *args, **kwargs) -> 'Keyword':
+        return self._create(self.keyword_class, 'create_keyword', args, kwargs)
+
+    def create_for(self, *args, **kwargs) -> 'For':
         return self._create(self.for_class, 'create_for', args, kwargs)
 
-    def create_if(self, *args, **kwargs):
+    def create_if(self, *args, **kwargs) -> 'If':
```
pekkaklarck, line 162: Having `robot.model.If` as a return type is a bit problematic, because then also `robot.running` and `robot.result` side model objects use that and not their own `If`. It's nevertheless an improvement over not having a return type. I know your are working with enhancements for `create` signature and we can return to this later.

src/robot/model/body.py
```diff
@@ -57,15 +68,16 @@ def id(self):
...
-            steps.append(parent.setup)
+            steps.append(parent.setup)  # type: ignore # Use Protocol with RF 7
         if hasattr(parent, 'body'):
-            steps.extend(step for step in parent.body.flatten()
+            steps.extend(step for step in
+                         parent.body.flatten()  # type: ignore # Use Protocol with RF 7
                          if step.type != self.MESSAGE)
         if getattr(parent, 'has_teardown', False):
-            steps.append(parent.teardown)
+            steps.append(parent.teardown) # type: ignore # Use Protocol with RF 7
```
pekkaklarck, line 80: PEP-8 says there should be at least two spaces before an inline comment, but we generally prefer four. I also believe the "second comment" mentioning Protocol should have four spaces before it as well to make it stand out. I can fix these myself.

pekkaklarck (approved):

PR #4726, opened 2023-04-07
add xmlschema to dev dependencies

JonathanArns:
I was unable to run the acceptance test suite locally (fedora, python 3.11) with the currently defined dependencies in `requirements-dev.txt`. In my case, `xmlschema` was needed to in addition to the other dependencies.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,3 +7,4 @@ wheel
 docutils
 pygments >= 2.8
 pydantic
+xmlschema


pekkaklarck:
Dependencies related to acceptance tests are in requirements files under the `atest` directory. For more details see:
https://github.com/robotframework/robotframework/blob/master/atest/README.rst#dependencies

We possibly could mention this in the `requirements-dev.txt` and I'm sure related docs could also be enhanced otherwise. `xmlschema` shouldn't be added to `requirements-dev.txt`, though, so I close this PR.

PR #4727, opened 2023-04-07
Item assignment support: ${x}[key] = Keyword

### LINKED ISSUE #4546: Support item assigment with lists and dicts like `${x}[key] =    Keyword`
These aren't currently supported:
```
${dict} =    Create Dictionary    key=initial
${dict}[key] =    Set Variable    new

${list} =    Create List    one    two
${list}[0] =    Set Variable    new
```
Especially setting dictionary items would be useful, but there certainly are cases where setting list items is convenient as well. In fact, this syntax should work with any object that has `__setitem__`.

Things to take into account:
- It should be possible to specify the key/index as a variable like `${x}[${y}]`.
- It should be possible to give list indices as strings and as integers.
- Lists should support slices.
- Error message should be clear if the item is immutable (tuple, string, immutable mapping, ...).

This would be a convenient feature, but because it's possible to manipulate lists and dicts with Collections keywords I don't consider this high priority. See also #4545 that proposes supporting `${x${y}} =  Keyword` and also contained this enhancement in it initially.

Vincema:
Hello everyone,

This PR fixes #4546. 
Adds item assignment support from keyword returns. Here a few examples of capabilities:
```
${dict}[key] =    Set Variable  value    # dict assign
${dict}[key1][key2] =    Set Variable    value    # nested assign

${list}[0] =     Set Variable  value    # list assign using str index
${list}[${0}] =     Set Variable  value    # list assign using int index
${list}[${other_variable}] =     Set Variable  value    # list assign using variable as index
${list}[1:-1] =     Evaluate  [1,2,3,4]    # list assign using slice as index (passing iterable value)

${dict}[key]   ${list}[0]    @{list}[1] =    Evaluate    (1, 2, 3, 4, 5)    # multiple variables assign
${obj}[1] =      Set Variable     value    # object that implements __setitem__

${str}[1] =     Set Variable     value    # will raise an error stating: "Variable '${str}' of type str does not support item assignment: variable should be mutable and should support a '__setitem__' method."
${list}[] =    Set Variable     value     # will raise an error stating: "Setting value to list variable '${list} at index [] failed: ValueError: invalid literal for int() with base 10: ''
```

I updated the UG and the Collections docs but it might need a bit more polish and details.

Please let me know how I could improve this PR and hopefully make it through to the v6.2 🙏


pekkaklarck:
RF 6.1 is so much late that we won't have time for RF 6.2 anymore this spring. Once RF 6.1 is out, we'll continue more active development only in the autumn. On a quick look this PR seems to be in so good shape that waiting until autumn would be annoying. If you @Vincema are interested, we can still try to push this to RF 6.1. We try to get it release ASAP, but there's at least the next week still time. I'll review the code changes shortly, and you can then decide do you want to try getting this into 6.1 or not.

src/robot/variables/assigner.py
```diff
@@ -158,21 +203,29 @@ def resolve(self, return_value):
...
-    def __init__(self, variables):
-        self._variables = variables
-        self._min_count = len(variables)
+    def __init__(self, assignments):
+        # _assignment contains a list of
+        # ... tuples: (variable, tuple of corresponding selectors)
+        self._assignments = []
+        for assigment in assignments:
+            match: VariableMatch = search_variable(assigment, ignore_errors=True)
+            self._assignments.append((match.name, match.items))
```
pekkaklarck, line 227: I got a feeling saving names and items to separate lists would be better. At least the default `_resolve` below could use just `return list(zip(self._names, self._items, return_value))`, but I'm not sure what that would do for `ScalarsAndListReturnValueResolver`.

src/robot/variables/search.py
```diff
@@ -114,13 +114,15 @@ def is_list_variable(self):
...
+                  allow_assign_mark=False, allow_nested=False, allow_items=False):
         if allow_assign_mark and self.string.endswith('='):
             match = search_variable(self.string[:-1].rstrip(), ignore_errors=True)
-            return match.is_assign()
+            return match.is_assign(allow_items=allow_items)
         return (self.is_variable()
                 and self.identifier in '$@&'
-                and not self.items
+                and not (self.identifier == '&' and self.items)
+                and (allow_items or not self.items)
```
pekkaklarck, line 125: We probably should allow item assignment only with `$` variables. This code already prohibits it with `&` variables, but I don't think it makes sense with `@` variables either.

pekkaklarck (changes requested):
Awesome work! I only looked at the code, but it looks really good. I added several comments, but most of them were related to style. If there are style related comments you don't understand or agree with, you can ignore them and I can take care of them afterwards if I consider them important.

I hope we could get this included still into RF 6.1 because otherwise there's a risk that future changes cause conflicts.

src/robot/utils/text.py
```diff
@@ -79,12 +79,13 @@ def _count_virtual_line_length(line):
     return lines if not remainder else lines + 1
 
 
-def format_assign_message(variable, value, cut_long=True):
+def format_assign_message(variable, items, value, cut_long=True):
```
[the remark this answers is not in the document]
Vincema: Fixed it ✅
Also removed the explicit None value to the function calls.

src/robot/utils/text.py
```diff
@@ -79,12 +79,13 @@ def _count_virtual_line_length(line):
...
     formatter = {'$': safe_str, '@': seq2str2, '&': _dict_to_str}[variable[0]]
     value = formatter(value)
     if cut_long:
         value = cut_assign_value(value)
-    return '%s = %s' % (variable, value)
+    decorated_items = "".join(map(lambda item: "[%s]" % item, items or tuple()))
+    return '%s%s = %s' % (variable, decorated_items, value)
```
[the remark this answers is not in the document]
Vincema: Fixed it ✅
I used your code sample.

src/robot/variables/assigner.py
```diff
@@ -134,6 +137,48 @@ def _variable_supports_extended_assign(self, var):
...
+            return int(index)
+        if index.count(':') > 2:
+            raise ValueError
+        return slice(*[int(i) if i else None for i in index.split(':')])
+
+    def _variable_type_supports_item_assign(self, var):
+        return (hasattr(var, "__setitem__") and callable(var.__setitem__))
+
+    def _item_assign(self, name, items, value, variables):
+        decorated_items = list(map(lambda item: "[%s]" % item, items))
```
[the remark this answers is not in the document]
Vincema: Fixed it ✅

src/robot/variables/assigner.py
```diff
@@ -134,6 +137,48 @@ def _variable_supports_extended_assign(self, var):
...
+        var = variables.replace_scalar("%s%s" % ('$' + name[1:],
+                                                "".join(decorated_items[:-1])))
```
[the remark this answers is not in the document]
Vincema: I used your second idea and I didn't need to add a condition on the nested variable.
```python3
*nested, item = items
decorated_nested_items = ''.join(f'[{item}]' for item in nested)
var = variables.replace_scalar(f'${name[1:]}{decorated_nested_items}')
```

I couldn't find a better name that `nested`...

Anyway, the whole function looks way neater with your suggestion. ✅

src/robot/variables/assigner.py
```diff
@@ -134,6 +137,48 @@ def _variable_supports_extended_assign(self, var):
...
+                f"Variable '{name}{''.join(decorated_items[:-1])}' of type {var_type} "
+                f"does not support item assignment: variable "
+                f"should be mutable and should support a '__setitem__' method."
+                )
+
+        selector = variables.replace_scalar(items[-1])
+
+        try:
+            if isinstance(var, list):
+                selector = self._parse_list_index(selector)
```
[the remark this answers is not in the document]
Vincema: 1. Actually yes `MutableSequence` makes more sense than `list`.
2. I edited the code to try handle `ValueError` from `_parse_sequence_index` (renamed). Now if the parsing fails, the unparsed index is used. I'm not sure in which case the parsing to an `int` or `slice` would fail but then succeed on accessing a `MutableSequence`. But I guess such cases such exist... Anyways I followed your suggestion. ✅

Regarding the test coverage, I believe the atests added in this PR cover the different paths.

src/robot/variables/assigner.py
```diff
@@ -134,6 +137,48 @@ def _variable_supports_extended_assign(self, var):
...
+            var[selector] = value
+        except (ValueError, IndexError, TypeError):
+            var_type = type_name(var)
+            raise VariableError(
+                f"Setting value to {var_type} variable "
+                f"'{name}{''.join(decorated_items[:-1])}' "
+                f"at index [{items[-1]}] failed: {get_error_message()}"
+                )
```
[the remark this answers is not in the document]
Vincema: Fixed ✅

src/robot/variables/assigner.py
```diff
@@ -158,21 +203,29 @@ def resolve(self, return_value):
 
 class OneReturnValueResolver:
 
-    def __init__(self, variable):
-        self._variable = variable
+    def __init__(self, assignment):
+        match: VariableMatch = search_variable(assignment, ignore_errors=True)
+        # _assignment contains the tuple: (variable, tuple of corresponding selectors)
+        self._assignment = (match.name, match.items)
```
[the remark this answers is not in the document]
Vincema: 1. TBH, I cannot remember why I set this `True`, and probably not for a good reason. I removed it and let the default value.
2. I've split the names and items in all the resolvers. I agree 100% with you, now everything is neat and much easier to understand.

src/robot/variables/assigner.py
Vincema: I've done as you said for both `_MultiReturnValueResolver` and `ScalarsAndListReturnValueResolver`. See next comment.

src/robot/variables/assigner.py
```diff
@@ -225,23 +280,26 @@ def _validate(self, return_count):
...
+
+    def _split_assignments(self, assignments):
+        list_index = [a[0][0] for a in assignments].index('@')
+        return (assignments[:list_index],
+                assignments[list_index],
+                assignments[list_index+1:])
+
+    def _split_return(self, return_value, before_assignment, after_assignment):
+        list_start = len(before_assignment)
+        list_end = len(return_value) - len(after_assignment)
```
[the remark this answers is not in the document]
Vincema: I rewrote the `_resolve` method for `ScalarsAndListReturnValueResolver`.
I believe it's less complicated now, with less function calls and intermediate variables.

Please take a look at the new implementation and let me know what you think. 🙏

src/robot/variables/search.py
Vincema: By allowing the `@`, we can actually use syntax like this one:
```robot
${mylist}=        Create List      ${1}    ${2}    ${3}
${mylist}[0]      @{mylist}[1]=    Set Variable     ${{ [9, 8, 7, 6] }}
Should Be Equal    ${mylist}    ${{ [9, [8, 7, 6], 3] }}
```
which is the equivalent of the Python:
```python3
mylist = [1, 2, 3]
mylist[0], *mylist[1] = [9, 8, 7, 6]
mylist   # [9, [8, 7, 6], 3]
```

It's tested in `atest/testdata/variables/return_values.py` by the tests:
* `Multiple item assigns to scalars and list`
* `Multiple item assigns to scalars and list slice`
* `Multiple item assigns to list only`

Do you think we can keep this feature?

Vincema:
Thank you so much for taking the time to review this PR. 
I've tried my best to improve the code as per your comments.

Please let me know when/if you need me🙏🙏

src/robot/variables/search.py
pekkaklarck: I didn't actually know `mylist[0], *mylist[1] = [9, 8, 7, 6]` is valid in Python. That's pretty obscure and I'm not sure does it make sense for us to support it. Do you have examples where this syntax would be useful in Python or with Robot?

It's great that you added tests for this functionality. It makes me want to just merge this as-is, but I need to think about this a bit more. I think I'll ask from other on the #devel channel on our Slack as well.

Vincema: TBH, I don't remember seeing this syntax anywhere before too...
I guess it could be useful to unpack a returned iterable into dict with one-liner like so:
```python
mydict["name"], *mydict["args"] = get_sysargv()
```
But it's a bit far fetched...


For your other question, until today we got the following result:
```robot
${x} =       Create List    a    b    c
@{x}[0] =    Create List    x    y    x   
# ${x} => [['x', 'y', 'z'], 'b', 'c']
# @{x}[0] is equivalent to ${x}[0]
```
I agree with you it doesn't make much sense.
**So I added today the logic in the `assigner.py` file to forbid this. User will get the error "Cannot assign a list variable with items."**


I can think of two options:
1. We keep today's logic and allow syntax like below:
```robot
@{x}[a]  ${y}[b]  ${z}[c]=   ...
```

2. We prevent this syntax. Then I can remove the logic to handle this and add a simple condition in `search.py` to prevent any `@` variable from using items => _**_least confusing solution_**_.

pekkaklarck: I asked about the
```
    ${x}    @{y}[1] =    Keyword
```
syntax on Slack and the only comment I got was very positive. I can also see it as a logic extension to the syntax we otherwise support and, although it gets pretty complicated, see no real harm with it. Let's add it.

pekkaklarck: If we allow `@{x}[0]` syntax when assigning multiple variables, I believe we should also allow
```
@{x}[0] =    Keyword 1
&{y}[key] =    Keyword 2
```
They should work exactly like
```
${x}[0] =    Keyword 1
${y}[key] =    Keyword 2
```
except that
- the first one with `@` should require the return value to be a list-like and convert it to a list, and
- the second one with `&` should require the return value to be a dict-like and convert it to a `DotDict`.
 
The above is exactly the difference between assigning
```
${x} =    Keyword 1
${y} =    Keyword 2
```
and
```
@{x} =    Keyword 1
&{y} =    Keyword 2
```
With this design the behavior ought to be consistent.

Vincema: The commit 53aa8ca adds support for the syntax:
```robot
@{x}[0] =      Keyword 1
&{y}[key] =    Keyword 2
```
They should now be consistent with the behavior of:
```robot
@{x} =    Keyword 1
&{y} =    Keyword 2
```
Please see the 4 last test cases in `return_values.robot`. Let me know if something must be improved 🙏.

pekkaklarck (approved):
Awesome work!

PR #4735, opened 2023-04-12
Add explicit command line option to limit which files are parsed

### LINKED ISSUE #4687: Add explicit command line option to limit which files are parsed
When execution just some tests from a big directory structure, parsing only relevant files instead of the whole structure makes parsing faster. We don't currently have any explicit option for limiting parsing like that, but when the `--suite` option is used files not matching the specified suite are ignored. This functionality won't work too well in the future, though, if we make it possible to change the suite name using the `Name` setting (#4583). It's also likely that not too many users know about this functionality and the underlying logic to convert file and directory names to suite names for matching purposes is also rather complicated.

A good way to solve all the above problems is adding a separate command line option or options for controlling what files to parse. Once we have such functionality, we can remove the current performance optimization from `--suite`, a separate option makes the feature easier to discover, and also the code gets simpler. There are some design decisions to be made, though:

- What should the option be named? I was thinking `--includefiles`, but it has the same prefix as `--include` which would mean that shortened usages like `--incl` that used to be unique wouldn't work anymore. Perhaps `--parsefiles` or just `--files` would be better.
- Should the value be a literal value or should we accept glob patterns? I believe supporting patterns is a good idea.
- Should the value be just a file name or a relative or absolute path to it? I see benefits with both so possibly we could support both usages.
- Should the value include file extension? I believe it should be allowed but preferably not required. If that gets too complicated, always requiring it is probably better.
- Should there be separate option for excluding files? It would make some use cases easier, but I'm not sure is that worth the added complexity.
- Should this option affect only files or also directories? If the option would match a directory, we could automatically include all files under it, recursively. On the other hand, just matching files would be easier to implement and explain.
- Should we have separate options for matching files and directories? That's probably too much, especially if we'd have separate options for including and excluding. If directories are considered important, probably better to implement this so that the same option work with both files and directories.

This feature should be implemented before RF 7 where we'd like to change `--suite` so that it doesn't affect parsing. That would give users who need the performance optimization time to update their scripts.

JonathanArns:
This PR implements the `--files` option as discussed in [this issue](https://github.com/robotframework/robotframework/issues/4687) with one difference:
The `--files` option does not fully override `--extension`, but instead only acts as an additional filter, after files have been filtered by extension already. We implemented it this way, since fully overriding `--extension` when `--files` is set did not seem like a nice change, because `--extension` not only influences the selection of files, but also the selection of parsers, and possibly other things. It seems more reasonable to us to make these additional changes, once and if `--extension` eventually gets deprecated.

closes #4687


pekkaklarck:
Unfortunately I don't have time to review this until the custom parser functionality is done (#1283) and RF 6.1 beta 1 is released. I try to get both of them done this week, but due to some other errands I may not have enough time.

Another unfortunate thing is that the parser changes affect the same parts of the code as these changes and are very likely to cause conflicts. There won't be big structural changes, though, so applying your changes on top of them shouldn't be too big a task.

Do you have some timeline with your school project?

JonathanArns:
Thank you for the info! And don't worry about it too much, we do indeed have a timeline for the assignment (deadline on the 24th or April), but while getting the changes merged before the deadline helps a lot with passing, it is not completely mandatory, because we all know that everyone has stuff to do :)

doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
```diff
@@ -31,11 +31,21 @@ extensions, the :option:`--extension (-F)` option must be used to explicitly
...
 extensions. Matching extensions is case insensitive and the leading `.`
-can be omitted::
+can be omitted. You can additionally filter files before parsing using
+the :option:`--files (-f)` option with a `simple pattern`_. The option
+can be used multiple times to match multiple patterns. Arguments to the
+:option:`--files (-f)` option are case- and space-insensitive.
+If the :option:`--files (-f)` option matches files with a non-default extension,
+the :option:`--extension (-F)` option must be added in order for those files
+to also be parsed.
+::
```
pekkaklarck, line 41: Explaining `--extension` and `--files` in the same section is a good idea. Documentation of `--files` should get its own paragraph, though. Possibly there could also be separate subsections.

Another thing that's needed is a note about `--files` being new in RF 6.1. There can either be an explicit, `.. note` or the paragraph explaining `--files` can start with something like "Starting from Robot Framework 6.1 ...". There ought to be examples of both styles here.

pekkaklarck (approved):
Looks really good! Few things to do before this can be merged:

- Small update to documentation. Especially a note about this being new in RF 6.1.
- Fix conflicts due to changes when the new external parsing API was introduced. I believe conflicts are mostly due to added typing.
- Related to above, typing may need to be added to the new code as well.

Sorry for somewhat late review that also caused the aforementioned conflicts. If you @JonathanArns don't have time or interest to fix the above issues, I can do that myself as well.

JonathanArns:
@pekkaklarck Thanks for the review! @cetceeve and I should have time to fix everything up this week :)

pekkaklarck:
Sounds great! There have been reports that `--suite` not working properly with the new `Name` setting causes problem. Probably need to remove the `include_suites` functionality, but I want to get this merged before that to avoid more conflicts.

pekkaklarck:
I resolved conflicts using the web interface. Assuming tests pass, I'll merge this later today.

PR #4757, opened 2023-05-08
Add the @set_type_hints decorator

Serhiy1:
The decorator allows function definitions to shadow other function definitions. 

It has a similar behaviour as `@wraps`, but works purely with type hints.

ItemList working
image

Body working
image

Part of #4570

### DIFF
--- a/src/robot/model/itemlist.py
+++ b/src/robot/model/itemlist.py
@@ -17,7 +17,7 @@ from functools import total_ordering
 from typing import (Any, Iterable, Iterator, MutableSequence, overload, TYPE_CHECKING,
                     Type, TypeVar)
 
-from robot.utils import copy_signature, known_at_runtime, type_name
+from robot.utils import copy_signature, KnownAtRuntime, type_name
 
 from .modelobject import DataDict
 
@@ -46,7 +46,7 @@ class ItemList(MutableSequence[T]):
 
     __slots__ = ['_item_class', '_common_attrs', '_items']
     # TypeVar T needs to be applied to a variable to be compatible with @copy_signature
-    item_type: Type[T] = known_at_runtime
+    item_type: Type[T] = KnownAtRuntime
 
     def __init__(self, item_class: Type[T],
                  common_attrs: 'dict[str, Any]|None' = None,
--- a/src/robot/model/itemlist.py
+++ b/src/robot/model/itemlist.py
@@ -17,8 +17,7 @@ from functools import total_ordering
 from typing import (Any, Iterable, Iterator, MutableSequence, overload, TYPE_CHECKING,
                     Type, TypeVar)
 
-from robot.utils import type_name
-from ..utils import copy_signature
+from robot.utils import copy_signature, known_at_runtime, type_name
 
 from .modelobject import DataDict
 
@@ -46,8 +45,8 @@ class ItemList(MutableSequence[T]):
     """
 
     __slots__ = ['_item_class', '_common_attrs', '_items']
-    # TypeVar T needs to be applied to a variable to be compatible with @set_type_hints
-    item_type: Type[T] = Any
+    # TypeVar T needs to be applied to a variable to be compatible with @copy_signature
+    item_type: Type[T] = known_at_runtime
 
     def __init__(self, item_class: Type[T],
                  common_attrs: 'dict[str, Any]|None' = None,
@@ -60,7 +59,7 @@ class ItemList(MutableSequence[T]):
 
     @copy_signature(item_type)
     def create(self, *args, **kwargs) -> T:
-        """Create a new item using the provided aguments."""
+        """Create a new item using the provided arguments."""
         return self.append(self._item_class(*args, **kwargs))
 
     def append(self, item: 'T|DataDict') -> T:
--- a/src/robot/model/itemlist.py
+++ b/src/robot/model/itemlist.py
@@ -18,6 +18,7 @@ from typing import (Any, Iterable, Iterator, MutableSequence, overload, TYPE_CHE
                     Type, TypeVar)
 
 from robot.utils import type_name
+from ..utils import copy_signature
 
 from .modelobject import DataDict
 
@@ -45,6 +46,8 @@ class ItemList(MutableSequence[T]):
     """
 
     __slots__ = ['_item_class', '_common_attrs', '_items']
+    # TypeVar T needs to be applied to a variable to be compatible with @set_type_hints
+    item_type: Type[T] = Any
 
     def __init__(self, item_class: Type[T],
                  common_attrs: 'dict[str, Any]|None' = None,
@@ -55,8 +58,9 @@ class ItemList(MutableSequence[T]):
         if items:
             self.extend(items)
 
+    @copy_signature(item_type)
     def create(self, *args, **kwargs) -> T:
-        """Create a new item using the provided arguments."""
+        """Create a new item using the provided aguments."""
         return self.append(self._item_class(*args, **kwargs))
 
     def append(self, item: 'T|DataDict') -> T:
--- a/src/robot/model/testsuite.py
+++ b/src/robot/model/testsuite.py
@@ -156,7 +156,7 @@ class TestSuite(ModelObject):
 
     @setter
     def suites(self, suites: 'Sequence[TestSuite|DataDict]') -> 'TestSuites[TestSuite]':
-        return TestSuites["TestSuite"](self.__class__, self, suites)
+        return TestSuites['TestSuite'](self.__class__, self, suites)
 
     @setter
     def tests(self, tests: 'Sequence[TestCase|DataDict]') -> TestCases[TestCase]:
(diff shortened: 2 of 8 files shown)


src/robot/model/itemlist.py
```diff
@@ -45,6 +46,8 @@ class ItemList(MutableSequence[T]):
     """
 
     __slots__ = ['_item_class', '_common_attrs', '_items']
+    # TypeVar T needs to be applied to a variable to be compatible with @set_type_hints
+    item_type: Type[T] = Any
```
pekkaklarck, line 50: - Comment uses old decorator name.
- It might be a good idea to more explicitly mention that `copy_signature` doesn't work without this. There are no automated tests to verify it, so it's easy to break it.
- `Any` is a type and using it as a value here looks a bit odd. The value cannot be `None` or `object()` because they aren't compatible with `Type[T]`, but `type(object)` ought to work. I'm not sure is it any better than `Any`, though. Your call.

src/robot/model/itemlist.py
```diff
@@ -55,8 +58,9 @@ def __init__(self, item_class: Type[T],
         if items:
             self.extend(items)
 
+    @copy_signature(item_type)
     def create(self, *args, **kwargs) -> T:
-        """Create a new item using the provided arguments."""
+        """Create a new item using the provided aguments."""
```
pekkaklarck, line 63: Typo.

src/robot/model/testsuite.py
```diff
@@ -155,12 +155,12 @@ def metadata(self, metadata: 'Mapping[str, str]|None') -> Metadata:
         return Metadata(metadata)
 
     @setter
-    def suites(self, suites: 'Sequence[TestSuite|DataDict]') -> 'TestSuites':
-        return TestSuites(self.__class__, self, suites)
+    def suites(self, suites: 'Sequence[TestSuite|DataDict]') -> 'TestSuites[TestSuite]':
+        return TestSuites["TestSuite"](self.__class__, self, suites)
```
pekkaklarck, line 159: We've generally used single quotes with strings (unless the string itself contains them). For consistency it would be good to use it here as well.

src/robot/model/testsuite.py
```diff
@@ -375,11 +375,12 @@ def to_dict(self) -> 'dict[str, Any]':
             data['suites'] = self.suites.to_dicts()
         return data
 
+TS = TypeVar("TS", bound=TestSuite)
```
pekkaklarck, line 378: See comments related to `TC`.

src/robot/result/model.py
```diff
@@ -37,10 +37,14 @@
...
 from collections import OrderedDict
 from itertools import chain
+from typing import Sequence
 import warnings
 
 from robot import model
 from robot.model import BodyItem, create_fixture, Keywords, Tags, TotalStatisticsBuilder
+from robot.model.modelobject import DataDict
+from robot.model.testcase import TestCases
+from robot.model.testsuite import TestSuites
```
pekkaklarck, line 47: All these should be importable directly from `robot.model`. If not, they need to be exposed.

src/robot/result/model.py
```diff
@@ -740,6 +744,14 @@ def elapsedtime(self):
         return sum(child.elapsedtime for child in
                    chain(self.suites, self.tests, (self.setup, self.teardown)))
 
+    @setter
+    def suites(self, suites: 'Sequence[TestSuite|DataDict]') -> 'TestSuites[TestSuite]':
+        return TestSuites['TestSuite'](self.__class__, self, suites)
+
+    @setter
+    def tests(self, tests: 'Sequence[TestCase|DataDict]') -> TestCases['TestCase']:
```
pekkaklarck, line 752: - Quotes around `TestCase` in `-> TestCases['TestCase']` probably aren't needed.
- I think `-> 'TestSuites[TestSuite]'` could also be `-> TestSuites['TestSuite']`. I'm not certain which is better, though.

src/robot/running/model.py
```diff
@@ -45,6 +45,8 @@
 from robot.conf import RobotSettings
 from robot.errors import BreakLoop, ContinueLoop, DataError, ReturnFromKeyword
 from robot.model import BodyItem, create_fixture, DataDict, Keywords, ModelObject
+from robot.model.testcase import TestCases
+from robot.model.testsuite import TestSuites
```
pekkaklarck, line 49: Import from `robot.model` directly.

src/robot/running/model.py
```diff
@@ -526,6 +528,14 @@ def randomize(self, suites: bool = True, tests: bool = True,
         """
         self.visit(Randomizer(suites, tests, seed))
 
+    @setter
+    def suites(self, suites: 'Sequence[TestSuite|DataDict]') -> 'TestSuites[TestSuite]':
+        return TestSuites['TestSuite'](self.__class__, self, suites)
+
+    @setter
+    def tests(self, tests: 'Sequence[TestCase|DataDict]') -> TestCases['TestCase']:
+        return TestCases['TestCase'](self.test_class, self, tests)
```
pekkaklarck, line 537: Same comment about quotes as above.

src/robot/utils/typehints.py
```diff
@@ -0,0 +1,31 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+
+from typing import Any, Callable, TypeVar, Generic
+
+T = TypeVar('T', bound=Callable[..., Any])
```
pekkaklarck, line 19: Our style is to use only one empty row after copyrights but two after imports. #nitpick

src/robot/utils/typehints.py
```diff
@@ -0,0 +1,31 @@
...
+class copy_signature(Generic[T]):
+    """A decorator that applies the signature of `F` to any function that it decorates
+    see https://github.com/python/typing/issues/270#issuecomment-555966301 for source
+    and discussion.
+    """
+
+    def __init__(self, target: T) -> None: ...
+
+    def __call__(self, wrapped: Callable[..., Any]) -> T:
+        return wrapped  # type: ignore
```
pekkaklarck, line 31: Isn't this now unnecessarily complicated for our needs? Based on my test this seems to be enough:

```python
def copy_signature(target: T) -> Callable[..., T]:
    def decorator(func):
        return func
    return decorator
```

I left out typing from `decorator` on purpose. It doesn't seem to be needed, and it wouldn't anyway match the actual parameters or the return type.

pekkaklarck (changes requested):
Awesome stuff! Few things mainly to code style to be fixed, though. See the individual comments for details.

src/robot/utils/typehints.py
Serhiy1: I don't mind either way. But I think the class-based approach would be slightly more performant since It's not creating any functions.

src/robot/model/itemlist.py
Serhiy1: Since this pattern will be used in `model.Body` in a future PR, I decided to deal with this by making a dedicated type alias `known_at_runtime` which just equals `type(object)`

pekkaklarck: Sounds like a good idea. Since it's a type, using a name like `LaterKnownType`, `TypeKnowAtRunTime`, `TypeType` or something like that could be better.

src/robot/utils/typehints.py
pekkaklarck: In the class based approach an instance is created and I believe it's slower. I don't think performance matter either way so a simpler approach is probably better.

src/robot/utils/typehints.py
```diff
@@ -0,0 +1,35 @@
...
+#  limitations under the License.
+
+from typing import Any, Callable, TypeVar, Generic
+
+
+T = TypeVar('T', bound=Callable[..., Any])
+
+# Type Alias for objects that are only known at runtime. This should be Used as a
+# default value for generic classes that also use `@copy_signature` decorator
+known_at_runtime = type(object)
```
pekkaklarck, line 23: - As I commented elsewhere, types should be named like `KnownAtRuntime` or `TypeType`.
- I'm not certain is there a need to have this in utils, it could just be created where needed and in some cases `type(object)` could be enough without a named type alias. This is especially true if in some cases we can use more detailed type than `type(object)` with `copy_signature`.

src/robot/model/testsuite.py
```diff
@@ -375,11 +375,12 @@ def to_dict(self) -> 'dict[str, Any]':
             data['suites'] = self.suites.to_dicts()
         return data
 
+TS = TypeVar('TS', bound=TestSuite)
```
pekkaklarck, line 378: Should be moved to the top of the module.

src/robot/utils/typehints.py
Serhiy1: Updated the Name. My follow-up PR for `Body` will use this value, I can define it a second time there, but it seems redundant.

src/robot/running/model.py
pekkaklarck: This still needs to be done at some point. Not a blocker for merging this.

pekkaklarck (approved):

Issue #4609, opened 2023-01-18
If multiple keywords match, resolve conflict first using search order


jylind:
The changes for fixing issue #4366 changed the `Set Library Search Order` functionality in a way that it cannot be used to force keywords to be used from imported resources if a keyword with same name exists in the importing resource file (or a testsuite in the future version?)

We have an old library (develop originally for RF 3.0) that enable us to created mocked user keywords to be used in user keyword “unit tests”. With this we can replace some of the user keywords (in the same resource file or testsuite) in a user keyword under test with a mocks and only concentrate on the functionality on that keyword. Just the default unit test approaches.

Now that local keywords cannot be overridden from the imported resource files, even using Set Library Search Order, this approach has now become impossible.

Related issue.
BuiltIn library does not have keyword for getting currently set search order. This is a bit inconvenient for our library.  Fake keywords (and implementing resources/libraries) can be set in different stages, so it is a bit hard to keep track of currently active search order. It would be lot easier if it could be asked from robot itself.

I guess `Set Library Search Order` could be called with some dummy value (or None) to get the current value and then set it back (maybe with some modification) but it would look a bit cumbersome.
(In the original implementation we are using internals from BuiltIn to get this information but this is definitely not something that should be done)

ursah:
I mentioned our library search order issue in RoboCon 2023: Extending Robot Framework workshop - @pekkaklarck had a quick glance at code and it looked like a change to the order of if statements (determining the library to call) would solve the issue to our benefit.

Any chance to get the solution verified and in before RF 7?

pekkaklarck:
Do I get it right that you'd like search order to be used first and only then keywords searched from the same file? That ought to be fine and as @ursah commented shouldn't be too hard to implement. The [related code](https://github.com/robotframework/robotframework/blob/master/src/robot/running/namespace.py#L374) looks like
```python
        if len(handlers) > 1:
            handlers = self._prioritize_same_file_or_public(handlers)
            if len(handlers) > 1:
                handlers = self._select_best_matches(handlers)
                if len(handlers) > 1:
                    handlers = self._filter_based_on_search_order(handlers)
```
and it would be easy to change the order so that filtering based on search order is done first. Can you test that locally and see how the results look like? If that fixes your issue, I can see are there some acceptance tests that fail if the logic is changed and are those tests so important that the change isn't possible. Other things to take into account:

- The above code is related to finding keywords from a resource file. There's similar code for finding keywords from libraries that would need to be updated as well.
[...]

PR #4760, opened 2023-05-10
Issue/4609

ursa-h:
[Allow Set Library Search Order to override the local keywords precedence.](https://github.com/robotframework/robotframework/issues/4609#top)

Restore ability to better control the prioritized keywords through Library Search Order
- Prioritize search order even when there is another keyword that is
  - a better match 
  - local 
  - private
 
Enables simpler unit testing of keywords


pekkaklarck:
Is there some reason for this PR to be Draft? On a quick look it looks pretty good. Very nice work updating tests. Few comments about it anyway:

- New tests should be added for a situation where there are multiple matches after applying the search order. That can happen at least if there are multiple keywords having embedded arguments that match the used keyword in the preferred resource/library. Tests should be organized so that one of the keywords matches better than others. If there already are tests like this and I missed them, let me know.
- New tests should be added for a situation where there's an unresolvable conflict after applying the search order.
- Do you remember is there documentation somewhat that should be updated?
- Some of the edited files don't have newline at the end. Please add them to make the diff (and future diffs if there are new edits) cleaner.

RF 6.1 is already pretty massively late and we try to get it released ASAP. It would be nice to get this into that release, because otherwise it needs to wait until RF 6.2 or RF 7.0, whichever comes first. I hope we'd be close to the release candidate after the next week. Ping me on Slack if you have problems hitting that deadline.

ursa-h:
Thank you for the valuable comments @pekkaklarck 

4 tests added:
Best Search Order Controlled Match Wins In Resource
Best Search Order Controlled Match Wins In Library
Search order causes conflict within resource
Search order causes conflict within library

In documentation I changed only one sentence. I wonder if I missed something. It all looked like it now works.

doc/userguide/src/CreatingTestData/CreatingUserKeywords.rst
```diff
@@ -702,9 +702,9 @@ succeeds:
 
 Before looking which match is best, Robot Framework checks are some of the matching
 keywords implemented in the same file as the caller keyword. If there are such keywords,
-they are given precedence over other keywords. If there are still conflicts
-after looking for best matches, Robot Framework checks can they be
-resolved based on the `library search order`_.
+they are given precedence over other keywords. Alternatively, `library search order`_
```
ursa-h, line 705: Link does not seem to work. Could be replaced with `:name:`Set Library Search Order` from the BuiltIn_ library`

pekkaklarck:
Looks great! I enabled running tests and I'll merge this if they pass.

There are some whitespace issues with tests (missing newlines, tabs instead of spaces), but I can fix them myself. I'll also check the issue with link.

pekkaklarck:
The link in documentation worked fine. Notice that you need to generate the User Guide using the `doc/userguide/ug2html.py` script to see how changes actually look like. I already fixed whitespace issues locally.

PR #4764, opened 2023-05-12
Allow statistics table to grow with content

### LINKED ISSUE #4025: Auto adapt column width in log.html
Current view:

image

More readable view:

image

I would gladly provide a PR if desired.

With some fiddling around locally, I found that adding the following in https://github.com/robotframework/robotframework/blob/681d359d39adc8418873e3eec05adade185a166d/src/robot/htmldata/rebot/common.css should be sufficient:

```
table {
    table-layout: auto;
    white-space: nowrap;
    empty-cells: show;
    font-size: 1em;
}
```

jin-harmoney:
Fixes #4025 Setting a min-width keeps the table pretty when suite-names are short, while also allowing the table to for long suite-names. On smaller screens, there is still a word-wrap to avoid a horizontal scroller.

### DIFF
--- a/src/robot/htmldata/rebot/common.css
+++ b/src/robot/htmldata/rebot/common.css
@@ -140,7 +140,7 @@ h2 {
 }
 /* Statistics table */
 .statistics {
-    width: 65em;
+    min-width: 65em;
     border-collapse: collapse;
     empty-cells: show;
     margin-bottom: 1em;


pekkaklarck:
I like the idea in general, but there are some problems:
1. There are three statistics tables. The proposed change only adjust width of those tables that have more content than fits with the minimum width. As the result tables end up having different widths that doesn't look too good in my opinion. The solution should work that one table being adjusted would affect also others.
2. Also the header has the same width as tables. It should be adjusted as well.
3. There should also be maximum width. With a wide monitor these tables don't look good if they take the whole horizontal space. Text wrapping looks much better.
4. We are planning a total log and report rewrite and making bigger changes before that doesn't make sense. Adjusting styles and possibly adding some container divs ought to be fine, though.

jin-harmoney:
I added my feedback here: https://github.com/robotframework/robotframework/issues/4025#issuecomment-1576784215, but let's just agree to disagree. I'll close the PR and the ticket.

Thank you for taking my proposal into consideration.

PR #4766, opened 2023-05-13
Make model.BaseBody accept Generic parameters

Serhiy1:
- `Body` sub classes pass in correct types
-  Those types are used with @copy_signature to type the `create_xxx` methods
-  Added correctly typed `body` entries for `running.TestCase` and `result.TestCase`

**Examples of the PR working**
image

image

Unfortunately, One side effect of this PR is, that Pyright believes that `Body.register` methods are cyclical dependencies.   

image

### DIFF
--- a/src/robot/model/body.py
+++ b/src/robot/model/body.py
@@ -110,29 +110,16 @@ class BaseBody(ItemList[BodyItem], Generic[KW, F, W, I, T, R, C, B, M, E]):
     """Base class for Body and Branches objects."""
     __slots__ = []
     # Set using 'BaseBody.register' when these classes are created.
-
-    if TYPE_CHECKING:
-        keyword_class: Type[KW] = KnownAtRuntime
-        for_class: Type[F] = KnownAtRuntime
-        while_class: Type[W] = KnownAtRuntime
-        if_class: Type[I] = KnownAtRuntime
-        try_class: Type[T] = KnownAtRuntime
-        return_class: Type[R] = KnownAtRuntime
-        continue_class: Type[C] = KnownAtRuntime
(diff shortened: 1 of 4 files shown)


src/robot/model/body.py
```diff
@@ -95,22 +106,22 @@ def to_dict(self) -> DataDict:
...
+        for_class: Type[F] = KnownAtRuntime
+        while_class: Type[W] = KnownAtRuntime
+        if_class: Type[I] = KnownAtRuntime
+        try_class: Type[T] = KnownAtRuntime
+        return_class: Type[R] = KnownAtRuntime
+        continue_class: Type[C] = KnownAtRuntime
+        break_class: Type[B] = KnownAtRuntime
+        message_class: Type[M] = KnownAtRuntime
+        error_class: Type[E] = KnownAtRuntime
     else:
```
pekkaklarck, line 125: Is `if TYPE_CHECKING` still needed here? Isn't `Type[X] = KnownAtRuntime` always available?

src/robot/model/body.py
```diff
@@ -254,7 +275,8 @@ def flatten(self) -> 'list[BodyItem]':
         return steps
 
 
-class Body(BaseBody):
+class Body(BaseBody["Keyword", "For", "While", "If", "Try", "Return", "Continue",
+                    "Break", "Message", "Error"]):
```
pekkaklarck, line 279: I'm totally nitpicking, but we've generally used single quotes around strings when possible.

pekkaklarck (approved):
Looks awesome! Can this still be simplified a bit by removing that one `if TYPE_CHECKING`?

PR #4767, opened 2023-05-14
Enhance model.branches and result.Iterations typing

Serhiy1:
re-named `model.Branches` to `model.BaseBranches`

`model.BaseBranches` now takes an additional `BranchType` type parameter  

`result.Iterations` now takes an additional `IterationType` parameter

**If branch**
image

**try branch**
image

**for Iteration**
image

**while Iteration**
image

### DIFF
--- a/src/robot/model/control.py
+++ b/src/robot/model/control.py
@@ -14,16 +14,26 @@
 #  limitations under the License.
 
 import sys
-from typing import Any, cast, Sequence
+from typing import Any, cast, Sequence, TypeVar, TYPE_CHECKING
 if sys.version_info >= (3, 8):
     from typing import Literal
 
 from robot.utils import setter
 
-from .body import Body, BodyItem, BodyItemParent, Branches
+from .body import Body, BodyItem, BodyItemParent, BaseBranches
 from .keyword import Keywords
 from .modelobject import DataDict
 from .visitor import SuiteVisitor
+
+if TYPE_CHECKING:
+    from robot.model import Keyword, Message
+
+BT = TypeVar('BT', 'IfBranch', 'TryBranch')
+
+
+class Branches(BaseBranches['Keyword', 'For', 'While', 'If', 'Try', 'Return', 'Continue',
+                          'Break', 'Message', 'Error', BT]):
+    pass
 
 
 @Body.register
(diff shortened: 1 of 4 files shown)


src/robot/model/control.py
```diff
@@ -14,17 +14,27 @@
...
-from .body import Body, BodyItem, BodyItemParent, Branches
+from .body import Body, BodyItem, BodyItemParent, BaseBranches
 from .keyword import Keywords
 from .modelobject import DataDict
 from .visitor import SuiteVisitor
 
+if TYPE_CHECKING:
+    from robot.model import Keyword, Message
+
+BT = TypeVar('BT', 'IfBranch', 'TryBranch')
```
pekkaklarck, line 31: See earlier comments about `bound` usage and naming.

src/robot/result/model.py
```diff
@@ -41,22 +41,25 @@
...
+from robot.utils import copy_signature, get_elapsed_time, KnownAtRuntime, setter
 
 from .configurer import SuiteConfigurer
 from .messagefilter import MessageFilter
 from .modeldeprecation import deprecated, DeprecatedAttributesMixin
 from .keywordremover import KeywordRemover
 from .suiteteardownfailed import SuiteTeardownFailed, SuiteTeardownFailureHandler
 
+BT = TypeVar('BT', 'IfBranch', 'TryBranch')
+IT = TypeVar('IT', 'ForIteration', 'WhileIteration')
```
pekkaklarck, line 62: See earlier comments about `bound` usage and naming. If we'd use `TI` instead of `BT`, then `IT` here could be `FW` or `WF`. Perhaps the latter would be better because the former looks like an abbreviation of `Framework`.

pekkaklarck (changes requested):
See inline comments.

src/robot/result/model.py
```diff
@@ -67,20 +70,29 @@ class Body(model.BaseBody['Keyword', 'For', 'While', 'If', 'Try', 'Return', 'Con
     __slots__ = []
 
 
-class Branches(model.Branches):
+class Branches(model.BaseBranches['Keyword', 'For', 'While', 'If', 'Try', 'Return',
+                                  'Continue', 'Break', 'Message', 'Error', BT]):
```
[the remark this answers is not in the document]
Serhiy1: The final parameter is passed in on the `If` and `Try` class 

```python
branch_class = IfBranch
branches_class = Branches[branch_class]
```

```Python
branch_class = TryBranch
branches_class = Branches[branch_class]
```

PR #4772, opened 2023-05-24
Enhancement: Introducing Status filter in search

adiralashiva8:
This pull request introduces a new feature that enhances the Search functionality in the `report.html` file. With this enhancement, a `Status` filter has been added to enable users to easily find specific status test cases.

image

The Status filter provides a convenient way to filter and narrow down the search results based on the desired status of the test cases. Users can now specify the desired status (e.g., "PASS" "FAIL" "SKIP") and instantly retrieve the relevant test cases.

**All Status**:
image

**Only PASS**:
image

Please review.


pekkaklarck:
I've typically just sorted tests based on status, but having a separate filter wouldn't hurt. Because there are exactly three statuses, I don't think a free text search is a good idea, though. I think having checkboxes that are initially all set would be better. You could then easily uncheck statuses you aren't interested in. In practice it would look something like this:

    Status:      [x] PASS    [x] SKIP    [x] FAIL

What do you think about it?

adiralashiva8:
@pekkaklarck I'm able to add checkbox as suggested (renamed label to `Test Status` to avoid conflict with stats field labeled as `Status`). 

**_Without Filter_**
image

**_With Filter_**
image

But I'm unable to render `Test Status` checkbox selection, something need to be done on `renderSelector`
 https://github.com/robotframework/robotframework/blob/3f1c3ddaeb823cbb5627866e214c300ff6d4865f/src/robot/htmldata/rebot/report.html#L238

Please suggest

pekkaklarck:
Sorry, I don't understand the problem. Could you clarify it?

I think the label should be just `Status`. I don't see big issue with the other label also being `Status`, but we could consider renaming it to `Total Status` to match.

PR #4773, opened 2023-05-24
Enhance setup and teardown typing for TestSuite and TestCase

Serhiy1:
`running.TestSuite` and `result.TestSuite` now return correct `setup` and `teardown` keyword types
`running.TestCase` and `result.TestCase` now also return correct `setup` and `teardown` keyword types

Part of #4570

### DIFF
--- a/src/robot/model/__init__.py
+++ b/src/robot/model/__init__.py
@@ -37,7 +37,7 @@ from .modifier import ModelModifier
 from .namepatterns import SuiteNamePatterns, TestNamePatterns, FileNamePatterns
 from .statistics import Statistics
 from .tags import Tags, TagPattern, TagPatterns
-from .testcase import TestCase, TestCases, ModelTestCase
-from .testsuite import TestSuite, TestSuites, ModelTestSuite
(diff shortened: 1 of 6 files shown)


pekkaklarck:
Looks good otherwise, but I don't understand why `SuiteVisitor` is changed. It is not meant to be used with  `robot.model` objects, but it has them in its type hints because it works both with `robot.running` and `robot.result` structures. This is explained in the [API docs](https://robot-framework.readthedocs.io/en/latest/autodoc/robot.model.html#type-hints).

The indented usage is extending `SuiteVisitor`, letting IDE to import type hints, and then updating imports as needed. That only works if type names don't change.

Serhiy1:
@pekkaklarck The type args need to be filled out. Otherwise, all of the type hints fail to produce any meaningful IDE autocomplete.  

image

VS 

image

pekkaklarck:
`robot.model.TestSuite` not having proper completion is ok as long as extending classes have it. The type hint for `SuiteVisitor` must be `TestSuite`.

PR #4778, opened 2023-05-31
Clarify code example

mneiferbag:
Update documentation setting to reflect that this is for keywords, not tests.


pekkaklarck:
Good catch! Thanks for the PR.

PR #4782, opened 2023-06-02
Add global caching to Matcher.match

vrpolakatcisco:
Tries to improves upon #3579 but makes it worse instead.


pekkaklarck:
#3579 was addressed otherwise.

Issue #4751, opened 2023-04-26
Using keyword Import Library result in unexpected behaviour


dries007:
I have a situation where 2 test cases would like to import the same library with different arguments.

My library controls the mocking of something and depending on the test case, some component of the system-under-test is present in the mock or not. This is not an issue in other test suites because there the mocked components are static for the while suite. This particular test suite verifies the response to a lack of one or more of the components, so only there the mock is variable. This is why I'd prefer to stick with being able to specify the settings at import time rather than having to add extra keywords to set up the mock.

The problem is that, when using the keyword `Import Library` as is, the library is imported for the whole suite. The second time the same library is imported, even with different arguments, nothing happens. I think because it ends up with the same name the importer assumes nothing needs to happen.

To make things more confusing, the `__init__` of the library actually is called with the given arguments, so my logging indicated that the library was being initialized correctly. It would also help if this was eliminated if the library is not going to be used anyway.

Ideally I'd like library imports to depend on the library scope, that way the library controls itself, but that would probably break comparability with other usecases. At the very least when an import does not actually do anything, I'd like to know about it. If possible a warning if the import arguments differ would already be a huge help. Now as far as I can find, nothing is logged anywhere.

My workaround for now is to import the libraries using `WITH NAME` and making sure the libraries are inert upon init. This is not ideal as it requires special attention that it counter-intuitive.

To aid in reproduction, I've attached a robot, library, and debug files as a zip:

[ImportLibraryIssue.zip](https://github.com/robotframework/robotframework/files/11330961/ImportLibraryIssue.zip)

For a quick overview without having to download & unzip:

![image](https://user-images.githubusercontent.com/1346725/234524933-d8ade002-9668-4458-ab9a-0851246b40d4.png)

[debug.txt](https://github.com/robotframework/robotframework/files/11330936/debug.txt)

```robotframework
*** Test Cases ***
TC1
    Import Library  TestLibrary     ABC
    Test Argument

TC2
    Import Library  TestLibrary     XYZ
    Test Argument
```

```python
from robot.api.logger import warn
class TestLibrary:
    def __init__(self, argument):
        warn(f'TestLibrary init with arg={argument!r}')
        self.argument = argument
    def test_argument(self):
        warn(f'TestLibrary test_argument with arg={self.argument!r}')
```

otemek:
I have the same case, so put my proposal here [#4804](https://github.com/robotframework/robotframework/pull/4804)

otemek:
[...]

PR #4804, opened 2023-06-22
#4751 reimport same lib with different arguments

otemek:
If that's still the case or reasonable to reimport the same library with different arguments then I've made a small change that does that (too simple to believe that it's enough - maybe I'm missing something/some bigger picture of it). IMHO it should allow me to use the "latest" version of an imported library. Anyway - it turns out that the Importer object [`robotframework/src/robot/running/importer.py:49`](https://github.com/robotframework/robotframework/blob/6a266bf852ad0525c350ea7412f7b33f81455e96/src/robot/running/importer.py#LL49C22-L49C22) returns a new instance of the same library with different arguments, but later the `Namespace` object checks if that library is imported (only via `name` attribute)  [`robotframework/src/robot/running/namespace.py:130`](https://github.com/robotframework/robotframework/blob/6a266bf852ad0525c350ea7412f7b33f81455e96/src/robot/running/namespace.py#LL130C11-L130C11) and if the names are the same - it does nothing.

If this case sounds like a bug/feature then I'd be glad if maintainers could verify this.
Thanks in advance

### DIFF
--- a/atest/robot/test_libraries/with_name.robot
+++ b/atest/robot/test_libraries/with_name.robot
@@ -6,7 +6,7 @@ Resource          atest_resource.robot
 Import Library Normally Before Importing With Name In Another Suite
     ${tc} =    Check Test Case    ${TEST NAME}
     Check Keyword Data    ${tc.kws[0]}    OperatingSystem.Should Exist    args=.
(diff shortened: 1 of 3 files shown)


otemek:
Closing cause I might have oversimplified this and not considered other scenarios where it's intentional to use the same instance...

PR #4806, opened 2023-06-23
4685 deprecate shortest as default for zip loop

MobyNL:


MobyNL:
Already solved by pekka

pekkaklarck:
Sorry, hadn't noticed this one.

PR #4823, opened 2023-07-17
(🐞) add `py.typed` file

### LINKED ISSUE #4822: (🐞) include a `py.typed` in the distribution (pep 561)
I am attempting to use some of the new code that includes type annotations, but unfortunately there is no `py.typed` file, so any type checker will ignore the annotations.

https://peps.python.org/pep-0561/

KotlinIsland:
- resolves #4822


DetachHead:
can this be merged? i'm trying to use the robot api to make a robot pytest plugin (#4834) and it would be nice if i could use the types

Issue #4808, opened 2023-06-28
Async keywords are not stopped when execution is stopped gracefully


jaltendorfer:
Stop a test execution during an async keyword with SIGINT or CTRL+C. Now the teardown with async keywords is executed, but the test keyword still runs and is not stopped.

**Example:**

Test python library:
```
import asyncio
from robot.api.deco import library, keyword
from robot.api import logger


@library
class TestLibrary:

    @keyword(name="Async Sleep")
    async def async_sleep(self, time: int, log_message: str):
        logger.info(f"{log_message} start", also_console=True)
        await asyncio.sleep(time)
        logger.info(f"{log_message} end", also_console=True)
```
Test robot suite:
```
*** Settings ***
Library    TestLibrary.py


*** Test Cases ***
Test
    TestLibrary.Async Sleep    ${5}    Sleep Test
    [Teardown]    TestLibrary.Async Sleep    ${10}    Sleep Teardown
```

Result:

![image](https://github.com/robotframework/robotframework/assets/138000046/848f7cb1-e359-44ba-b8f0-d77788e7544a)

pekkaklarck:
Teardowns are, by default, executed when execution is gracefully stopped. If you don't want that, you can use the `--skip-teardown-on-exit` command line option. Does that work for you? Do you see differences with normal and async keywords in this regard?

jaltendorfer:
The problem is not that the teardown is executed. The problem is, that the async keyword in the test case is not stopped immediately before executing the teardown. In the example result of the post before, you can see the log messages of the test case keyword in the teardown after a graceful stop of the execution.

Normal behaviour with graceful stop:
- Running keyword in the test case is stopped immediately
- Teardowns are executed

Current behaviour when using async keywords:
- Running keyword in the test case is not stopped
- Teardowns are executed and the test case keyword is still running (in background) until completion.

In my example in the post above, you can see the log message "Sleep Test" being logged in the teardown, although the execution was stopped and the testcase keyword should have been stopped immediately (only the teardown should be executed).

pekkaklarck:
I see. Apparently async methods would need separate handling to get them killed. Would be nice to get that fixed in RF 6.1.2.

I'm on holiday and won't have time to look at this in the near future. If someone with more async knowledge would be interested to look at this already now that would be great. Ping @ygorpontelo.

pekkaklarck:
[...]

PR #4835, opened 2023-08-02
if execution cannot continue, cancel task and its children

ygorpontelo:
This will only cancel the task that got the error. Other coroutines will still run normally.

We can try to clean up coroutines when closing the loop, but i think showing the logs of open tasks help developers manage the coroutines better, so i didn't include any clean up for that. Open to suggestions on this.

Adressing issue #4808


pekkaklarck:
Sounds good but there should be a test for this. Should there also be a note about this in the User Guide?

pekkaklarck:
Great work!

Issue #4803, opened 2023-06-22
Async dynamic & hybrid library API


WisniewskiP:
For our Lib we are using framework that is async and we have to await for results.
our lib works similar to Remote Lib to get the keywords and arguments but communication is done using asyncio.

with RF 6.1 run_keyword can be made async but functions for getting keywords/arguments/etc. can't.

as a workaround I'm getting context with EXECUTION_CONTEXTS.current and run async operations with context.asynchronous.run_until_complete().
this works but if the implementation of EXECUTION_CONTEXTS changes the library will break.

would it be possible to add async to get_* functions in Dynamic & Hybrid Lib interfaces?

ygorpontelo:
Hello, i helped implement async in robot. You shouldn't need to grab the context to work with async, can you elaborate more on what Robot is not being able to do? Like example of such keywords that is not being able to get the results.

WisniewskiP:
Hello,
running async keywords is not a problem. What I'm trying to do is run async functions inside `get_keyword_names`, `get_keyword_arguments`, etc.

as an example:
```robot
*** Settings ***
Library  test_async.py

*** Test Cases ***
Test 1
    async_keyword
```

library:
```python
from asyncio import sleep

class test_async:
    async def get_keyword_names(self):
        await sleep(1)
        return ["async_keyword"]

    async def async_keyword(self):
        await sleep(1)
        return "test"
```

When you try to start this test you will get:
```
Error in file '....\tests\async\test_async.robot' on line 2: Getting keyword names from library 'test_async' failed: Calling dynamic method 'get_keyword_names' failed: Return value must be a list of strings.
c:\Users\PawelWisniewski\ExecutionPlatform\ep\venv-3.10\lib\site-packages\robot\running\namespace.py:71: RuntimeWarning: coroutine 'test_async.get_keyword_names' was never awaited
  item.report_invalid_syntax(err.message)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

ygorpontelo:
May have found a way to make it async, follow the PR.

WisniewskiP:
Thank you for looking at it. This works as expected.

pekkaklarck:
This looks like a good enhancement and the implementation in the PR looks good as well. The PR lacks tests, though, and I believe this should also be mentioned in the User Guide. We are currently developing RF 7 with a plan to get it out in early December. I hope we could get this and #4808 done before that!

pekkaklarck:
[...]

PR #4836, opened 2023-08-07
run dynamic method in event loop if required

ygorpontelo:
Adressing request from #4803

At the moment, dynamic methods cannot be async and using asyncio.run() could result in problems because of separate event loops. We can try to grab the context and run the result in the event loop if the implementation is async.


pekkaklarck:
Looks good but needs a test and a note to the User Guide.

PR #4842, opened 2023-08-16
Add py.typed to declare api module typed

robinmackaij:


pekkaklarck:
Quick comments after a quick look:
1. This PR is too big, considering that the changes are non-trivial, to be reviewed in one go. Please split at least changes in `deco` to a separate PR. Either create a new PR with those changes or remove other changes from this one.
2. There are quite many issues in the aforementioned `deco` module. Some or related to code style, which we need to document better, but there are several cases where types are slightly wrong. I won't have time for a proper review right now and, as already mentioned, this PR is too big for that in general.
3. Related to style, I consider `-> None` rather useless noise with functions that return nothing in general and totally useless with `__init__` that even cannot return anything. I don't want them to be added to the our code base and hope no type checker actually requires them.

robinmackaij:
1. I'll see when I can find time to split the PR. My experience with GIT and cloned branches is close to non-existent, so it's not something I know how to do quickly.
2. With regards to code style, odds are that `black` or `isort` kicked in. Having the relevant settings (especially the non-default ones, e.g. string normalization) in a `pyproject.toml` would be welcome.
As for the way annotations should be done, as discussed on Slack, we should probably add a section on that to the `CONTRIBUTING.rst`. I've tried to trace back the types as much as possible, but there are probably some things I overlooked there.
3. The main type checkers do use them and flag for missing ones. `-> None` is not a special case for this purpose.
`mypy` will consider a function or method that has no typed arguments and no return type annotation to be untyped and it will not check inside such functions / methods. See also [the mypy documentation](https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods).
'pyright' tries to [infer the return type if it is not specified](https://github.com/microsoft/pyright/blob/main/docs/type-inference.md#return-type-inference). Of course, this will not detect discrepancies between intended return type and actual returns from the implementation. In addition, the overhead of inference is higher than that of validation of specified return type.
Personally, I have no issue with `-> None` annotations. To me, it's no different from e.g. an `-> int` annotation. The fact that Python returns `None` if nothing else is returned is an (implicit) implementation detail. Annotations are about making things explicit, so tools can reason about code without assumptions.

pekkaklarck:
Yeah, type style definitely needs to be documented. It has evolved over time so the code base isn't really consistent. I think the main issue was using e.g. `List[int]` instead of `'list[int]'`. Updating the latter to `list[int]` in somewhat near future will be easier than updating the former. With more complex types, especially type aliases, there may be needs to split nested types to multiple lines and then `List[...]` works better, though.

In my opinion we should add return type annotation to public APIs whenever we use `return` with a value. If there's no `return`, adding `-> None` is extra work and adds noise. It's also 8 characters which is close to 10% of our 88 character maximum line length. It would often require line wrapping which adds even more noise. I hope type checkers can be configured to not require `-> None`, but I don't want them included even if they wouldn't.

robinmackaij:
> Yeah, type style definitely needs to be documented. It has evolved over time so the code base isn't really consistent. I think the main issue was using e.g. `List[int]` instead of `'list[int]'`. Updating the latter to `list[int]` in somewhat near future will be easier than updating the former. With more complex types, especially type aliases, there may be needs to split nested types to multiple lines and then `List[...]` works better, though.

There is indeed some existing annotations in the `List[int]` or `Dict[..., ...]` format. Updating those to the preferred format can easily be done and it's not in that many cases. The exception is in `interfaces.py` that uses multi-line type aliases to be able to add comments to the individual lines.

> 
> In my opinion we should add return type annotation to public APIs whenever we use `return` with a value. If there's no `return`, adding `-> None` is extra work and adds noise. It's also 8 characters which is close to 10% of our 88 character maximum line length. It would often require line wrapping which adds even more noise. I hope type checkers can be configured to not require `-> None`, but I don't want them included even if they wouldn't.

The problem with leaving out `-> None` is equivalent to not typing the function / method at all for mypy and it'll emit errors from pyright. This is not something that can be configured in the type-checkers; those simply cannot know whether someone forgot to add a return annotation or someone did not annotate it intentionally.

If the annotations are not added on the RF side, the users have to add things like `# type: ignore` to the imported functions / methods to suppress the errors / warnings that mypy / pyright will emit.

Digging into the mypy / pyright issues in a number of the repos I work on makes me realize more and more that adding correct annotations requires a bit of a perspective shift. Take for example the current annotation on `trace` from `api.logger`:

`def trace(msg: str, html: bool = False) -> None:`

I didn't realize that that annotation actually leads to a typing issue on this usage:

```
from robot.api import logger

def print_some_dict() -> None:
    my_dict = {"my_key": 42}    
    logger.trace(my_dict)
```
`pyright`, `mypy` (and `sonarlint`) all flag the last line:
> Argument 1 to "trace" has incompatible type "dict[str, int]"; expected "str"

What `trace` (and the other logging methods) actually need as `msg` argument is not a string, but something that can be printed (e.g. has a `__str__`). So from an typing perspective, something like this:
```
class Printable(Protocol):
    def __str__(self) -> str: ...

def trace(msg: Printable, html: bool = False) -> None:
    ...
```
This is quite similar to typing an argument as a `list` vs typing it as a `Sequence` or an `Iterable`.

src/robot/api/deco.py
```diff
@@ -41,8 +53,28 @@ def exposed_as_keyword():
...
+def keyword(name: F) -> F:
+    ...
+
+
+# Decorator used with arguments, e.g. @keyword("Custom Keyword Name")
+@overload
+def keyword(
+    name: 'str | None' = None,
+    tags: Sequence[str] = (),
+    types: 'Sequence[str] | Mapping[str, Any]' = (),
```
pekkaklarck, line 67: This typing for types is wrong. Types can be specified using actual types, strings and tuples. Probably using union types created with `x | y` would work as well. And when types are given as a sequence, using `None` means that that argument doesn't have a type. Finally, using `types=None` means that argument conversion should be disabled altogether.

I guess there are three options:
1. Don't add typing to types at all. Just use `types: 'Seqeunce|Mapping|None' = ()`. This is correct but not very detailed.
2. Created a type alias like `TypeHint` that represents all possible ways a type can be specified properly. Then we ought to be able to use `types: Sequence[TypeHint|None]|Mapping[str, TypeHint]|None = ()`. This ought to be correct and detailed, but it is also rather complicated.
3. Same as 2. but instead of creating type aliases here, reuse type aliases from the `interfaces` module. It has both `TypeHint` for an individual type and `TypeHints` that covers Sequence and Mapping cases. With the latter we could reduce this to `types 'TypeHints|None' = ()` which looks pretty good.

I believe the option 3. is a clear winner here. Notice that at the moment `interfaces` has `Type` and `Types` instead of `TypeHint` and `TypeHints`, but I'll rename them to avoid confusion with `typing.Type`. There are also some other things that need to be fixed and I'll do that at the same time. That will likely cause conflict with this PR, but I try to fix the conflict afterwards.

src/robot/api/deco.py
```diff
@@ -41,8 +53,28 @@ def exposed_as_keyword():
...
+    ...
+
+
 @not_keyword
-def keyword(name=None, tags=(), types=()):
+def keyword(
+    name: 'str | Callable[..., Any] | None' = None,
+    tags: Sequence[str] = (),
+    types: 'Sequence[str] | Mapping[str, Any]' = (),
+) -> 'F | Callable[[F], F]':
```
pekkaklarck, line 77: With the above overloads, typing could be removed from this function altogether to keep it simpler.If typing is preserved, it should use type aliases and not complex constructs like `Callable[..., Any]`.

src/robot/api/deco.py
```diff
@@ -96,9 +128,38 @@ def decorator(func):
     return decorator
 
 
+C = TypeVar('C')
```
pekkaklarck, line 131: For consistency this could be bound to `type`.

src/robot/api/deco.py
```diff
@@ -96,9 +128,38 @@ def decorator(func):
...
+
+
+# Decorator used without arguments, i.e. @library()
+@overload
+def library(scope: Type[C]) -> Type[C]:
```
pekkaklarck, line 136: - `Type[C]` -> `'type[C]'`. This should be done also elsewhere.
- `scope: ...` -> `cls: ..., /`
- Remove unnecessary comment.

src/robot/api/deco.py
```diff
@@ -96,9 +128,38 @@ def decorator(func):
...
+    ...
+
+
+# Decorator used with arguments, e.g. @library(version="1.0.0")
+@overload
+def library(
+    scope: 'str | None' = None,
+    version: 'str| None' = None,
+    converters: 'Dict[Any, Callable[..., Any]] | None' = None,
```
pekkaklarck, line 145: Converter typing isn't correct. I believe it should be `dict[type, Callable[[Any], Any]]` in the common case. Converter methods have an optional second argument, though, that will be the library instance. I don't know what's the best way to type optional arguments, but most likely this will require a type alias. Alternatively we could use just `dict[type, Callable]` which at least doesn't cause false negatives.

src/robot/api/exceptions.py
```diff
@@ -31,7 +31,7 @@ class Failure(AssertionError):
     """
     ROBOT_SUPPRESS_NAME = True
 
-    def __init__(self, message, html=False):
+    def __init__(self, message: str, html: bool = False) -> None:
```
pekkaklarck, line 34: Please remove `-> None` from here and everywhere else. We can consider adding it to public APIs if we at some point aim for 100% type coverage, but at the moment they just add distraction in general and are totally useless with `__init__`s.

src/robot/api/interfaces.py
```diff
@@ -287,7 +287,7 @@ class StartSuiteAttributes(TypedDict):
     id: str
     longname: str
     doc: str
-    metadata: dict
+    metadata: Dict[str, Any]
```
pekkaklarck, line 290: `Dict` -> `'dict'`

src/robot/api/logger.py
```diff
@@ -66,12 +66,16 @@ def my_keyword(arg):
 """
 
 import logging
+from typing import Literal
 
 from robot.output import librarylogger
 from robot.running.context import EXECUTION_CONTEXTS
 
+LOGLEVEL = Literal['TRACE', 'DEBUG', 'INFO', 'CONSOLE', 'HTML', 'WARN', 'ERROR']
+CONSOLE_STREAM = Literal['stdout', 'stderr']
```
pekkaklarck, line 75: - Should have two empty lines after imports.
- I believe our code handles log levels case-insensitively. This code requires them to be in upper case, but I guess that's fine.
- I also believe our code handles stream names case-insensitively. Having streams in lower case while levels are in upper cases is a bit inconsistent, but I guess that's ok.
- Both of these seem to be used only once. At least the latter one is so simple that it could just be inlined.

src/robot/api/logger.py
```diff
@@ -89,27 +93,27 @@ def write(msg, level='INFO', html=False):
...
     else:
         logger = logging.getLogger("RobotFramework")
-        level = {'TRACE': logging.DEBUG // 2,
+        level_int = {'TRACE': logging.DEBUG // 2,
                  'DEBUG': logging.DEBUG,
                  'INFO': logging.INFO,
                  'CONSOLE': logging.INFO,
                  'HTML': logging.INFO,
                  'WARN': logging.WARN,
                  'ERROR': logging.ERROR}[level]
```
pekkaklarck, line 102: Inconsistent indentation.

pekkaklarck (changes requested):
There are some stylistics things to be changed but in some cases typing is also wrong and needs to be fixed. I was contemplating should I merge this anyway and just update the code afterwards, but in the end decided to write comments.

As I commented otherwise, I plan to enhance typing of the `interfaces` module and that will likely cause a conflict. I already resolved some conflict there and hopefully can resolve the new ones as well.

src/robot/api/deco.py
robinmackaij: I agree that 3. is preferred here. I would propose adding a `robot.annotations` module and move the `TypeHints` and other long / complex annotations there.

That annotations module would then be a place to look for "complex types / interfaces / Protocols" that are common within the robot codebase. I'm thinking of the annotations we'd see in e.g. the listener interface also. Extra documentation on the why of certain signatures could also be centralized then. In addition, it may prevent circular import issues.

Would you support adding such an annotations module?

src/robot/api/logger.py
robinmackaij: I traced them back to their sources and followed the casing used there. I believe the casing for `CONSOLE_STREAM` traces back to the Python standard library and is annotated / documented as shown.

I'm pretty sure this all is indeed case-insensitive, but I don't know of a way to make the `Literal` annotation case-insensitive. In the end it boils down to whether or not the type-checkers complain about `Trace` and `trace` and not about `TRACE` while `tRacE` is actually acceptable at runtime.

robinmackaij:
> I wasn't sure what's the best way to type `@keyword` and asked about that on the [Python forum](https://discuss.python.org/t/questions-related-to-typing-overload-style/38130). I learned that we can use `def keyword(func: F, /)` instead of `def keyword(name: F)` which is convenient. I also learned how we could use `Protocol` and `ParamSpec` for typing `@keyword` so that also `robot_name` and other special attributes would be recognized. The needed code is pretty complicated, but this is a public API and having a concrete example of using `ParamSpec` in our code base would be nice. Unfortunately `ParamSpec` requires Python 3.10 so we cannot use it yet. I don't consider this important enough to warrant conditional typing making the code even more complex.

Indeed interesting and insightful responses. Eric's solution is very nice, I'll see how far we can go with Python 3.8, but we should definitely preserve these insights for when we can use his solution.

If introducing an `robot.annotations` module is agreed upon, having the required conditional typing in there might be a good solution, especially if there's types in other places of the codebase that has similar annotation issues.

src/robot/api/deco.py
pekkaklarck: Having a separate module for annotations is fine in general, but I'm not sure do we have enough needs for it yet. If we only need to reuse `TypeHints`, this module importing it from `interfaces` is simpler. I believe we should get this PR merged and look st how many reusable types we have after that.

pekkaklarck:
I believe conditional typing logic would make the code complex even of it would be in a separate module. Tools actually inspecting `robot.name` are so rare, that doing that very likely isn't worth the effort and resulting complexity. Adding a comment with a reference to the aforementioned discussion and note about getting back to this when Python 3.10 is our minimum version is likely a better solution. If you want to test the conditional logic that's fine, but let's first get this PR merged in simpler form.

src/robot/api/deco.py
```diff
@@ -14,9 +14,21 @@
...
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Mapping,
+    Sequence,
+    Type,
+    TypeVar,
+    overload,
+)
```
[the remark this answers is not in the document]
robinmackaij: Corrected to single line. All are currently used.

src/robot/api/deco.py
```diff
@@ -14,9 +14,21 @@
...
 
+F = TypeVar("F", bound=Callable[..., Any])
```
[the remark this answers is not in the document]
robinmackaij: Since we can't use `ParamSpec` yet, I would prefer `Callable[..., Any]` since `mypy` does flag a plain `Callable`: `error: Missing type parameters for generic type "Callable"  [type-arg]`

src/robot/api/deco.py
```diff
@@ -41,8 +53,28 @@ def exposed_as_keyword():
 not_keyword.robot_not_keyword = True
 
 
+# Decorator used without arguments, i.e. @keyword()
+@overload
+def keyword(name: F) -> F:
```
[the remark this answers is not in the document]
robinmackaij: Done

src/robot/api/deco.py
robinmackaij: Changed to option 3.

src/robot/api/deco.py
robinmackaij: Removing the annotations from the actual function results in a lot of typing errors; the type-checkers check if the signatures of all the defined `overload`s match up with the signature / annotations of the overloaded function.

Do you have a suggestion for the names for type aliases we could use here? It currently looks like this:
```python
def keyword(
    name: 'Callable[[F], F] | str | None' = None,
    tags: Sequence[str] = (),
    types: 'TypeHints | None' = (),
) -> 'F | Callable[[F], F]':
```
I think `Callable[[F], F]` is what we would need an alias with a clear name for.

src/robot/api/exceptions.py
robinmackaij: Removing them from the `__init__` does not result in type-checking errors (a special case in mypy, actually), but removing the `-> None` annotations from the functions in `logger` and `interfaces` introduces 35 mypy errors (`[no-untyped-def]`). This is with `disallow_untyped_defs = true` OR `strict=true` in the mypy settings.

I personally feel we should aim for 100% annotations on external APIs. If we want to support users using type-checkers on their codebase, we need to be complete even if it's of little value for ourselves / within the robot codebase itself.

I've removed the return types on the `__init__`s and I'll remove the `-> None` returns if that's the direction / decision for now.

src/robot/api/logger.py
robinmackaij: > * Should have two empty lines after imports.

Done in the different modules

> * I believe our code handles log levels case-insensitively. This code requires them to be in upper case, but I guess that's fine.
> * I also believe our code handles stream names case-insensitively. Having streams in lower case while levels are in upper cases is a bit inconsistent, but I guess that's ok.

Handling is indeed case-insensitive. I haven't found a way to have case-insensitive annotations for Literals, so following the casing as used in the Python docs is the preferred option in my option.

> * Both of these seem to be used only once. At least the latter one is so simple that it could just be inlined.

Inlined means a line break (more than 88 characters) but I guess that's preferred to the `CONSOLE_STREAM` alias that, in hindsight is not a really good name since it's actually referring to `sys.stdout` and `sys.stderr` (not really related to console, essentially).

src/robot/api/logger.py
robinmackaij: Corrected

robinmackaij:
> I believe conditional typing logic would make the code complex even of it would be in a separate module. Tools actually inspecting `robot_name` are so rare, that doing that very likely isn't worth the effort and resulting complexity. Adding a comment with a reference to the aforementioned discussion and note about getting back to this when Python 3.10 is our minimum version is likely a better solution. If you want to test the conditional logic that's fine, but let's first get this PR merged in simpler form.

Comment added. I tried a version-dependent solution but that runs into other problems, meaning pretty much the entire module content would be in both the `if` and the `else` of the Python version check.

src/robot/api/deco.py
```diff
@@ -14,9 +14,19 @@
...
 
+from robot.api.interfaces import TypeHints
 
-def not_keyword(func):
+
+# Current annotations report `attr-defined` errors. This can be solved once Python 3.10
+# becomes the minimum version (error-free conditional typing proved too complex).
+# See: https://discuss.python.org/t/questions-related-to-typing-overload-style/38130
+F = TypeVar("F", bound=Callable[..., Any])
+Converters = dict[type, Callable[[Any], Any]]
```
pekkaklarck, line 26: Converter can actually also accept two arguments. I don't know is there any nicer syntax than `Callable[[Any], Any] | Callable[[Any, Any], Any]` for that.

src/robot/api/deco.py
pekkaklarck: I guess the current code is fine. We can experiment with an alias for `Callable[[F], F]`, though.

pekkaklarck:
This looked so good that I merged it now. Typing can still be enhanced in the future. Thanks @robinmackaij!

src/robot/api/deco.py
```diff
@@ -41,8 +51,26 @@ def exposed_as_keyword():
...
+    tags: Sequence[str] = (),
+    types: 'TypeHints | None' = (),
+) -> Callable[[F], F]:
+    ...
+
+
 @not_keyword
-def keyword(name=None, tags=(), types=()):
+def keyword(
+    name: 'Callable[[F], F] | str | None' = None,
```
pekkaklarck, line 70: Should this be `F | str | None`? I've already changed it like this locally so no need to update the already merged PR or anything like that.

pekkaklarck:
I noticed that Pylance on VSCode is complaining about setting class attributes in `@library`. I guess that could be avoided by using a protocol (they are new in Python 3.8), but these class attributes are optional and I don't know is it possible to create a protocol with optional attributes. This problem shouldn't affect anyone actually using `@library` so it's not too severe.

src/robot/api/deco.py
```diff
@@ -96,9 +124,38 @@ def decorator(func):
...
+# Decorator used with arguments, e.g. @library(version="1.0.0")
+@overload
+def library(
+    scope: 'str | None' = None,
+    version: 'str | None' = None,
+    converters: 'Converters | None' = None,
+    doc_format: 'str | None' = None,
+    listener: 'Type[Any] | None' = None,
+    auto_keywords: bool = False,
+) -> Callable[[Type[C]], Type[C]]:
```
pekkaklarck, line 145: Needing to use `Type[C]` everywhere is a bit annoying. I believe that if the `TypeVar` is bound to `type` that isn't needed. At least it seems to work fine when I tested it locally.

src/robot/api/deco.py
robinmackaij: I think `Callable[[F], F]` is correct since it's the first overload. I hope to find time on Friday to evaluate the alpha against some repos.

robinmackaij:
> I noticed that Pylance on VSCode is complaining about setting class attributes in `@library`. I guess that could be avoided by using a protocol (they are new in Python 3.8), but these class attributes are optional and I don't know is it possible to create a protocol with optional attributes. This problem shouldn't affect anyone actually using `@library` so it's not too severe.

The thought occurred to me that the added property on `keyword` and `library` being "obscured" is actually beneficial in our intended use case; these properties are intended to be used internally by robot and users creating Python libraries / keywords using the decorators should probably not be accessing these properties at all. The type-checking errors will be reported on the robot repo side, type-checking in user projects using the decorators should not raise errors, unless a user would actually try to access the added properties.

PR #4879, opened 2023-09-25
Issue/4846 remove deprecated attributes in result model

yanne:


src/robot/output/xmllogger.py
```diff
@@ -266,10 +266,12 @@ def _write_list(self, tag, items):
             self._writer.element(tag, item)
 
     def _write_status(self, item):
-        attrs = {'status': item.status,
-                 'start': item.start_time.isoformat() if item.start_time else None,
-                 'elapsed': str(item.elapsed_time.total_seconds())}
-        self._writer.element('status', item.message, attrs)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
pekkaklarck, line 270: This shouldn't be needed in the end. FIXME to remind about that would be good.

pekkaklarck (changes requested):

PR #4895, opened 2023-10-11
Improvement for Robot "Get Length" with len() Function own

NoobiDev94:
During my work on API endpoint validation, I needed know the length of an integer variable. It was then that I realized that the "Get Length" function couldn't count the characters in variables of types int, bool, and float. In discussions with my team, we noticed this need in several projects. Sometimes, it's necessary to validate the number of digits in an integer variable to ensure it doesn't exceed the limit of a system's business rule.

I'm submitting a Pull Request with this potential improvement. Now, when using "Get Length," it will return the number of digits in variables of types str, int, bool, and float and will no longer display the message "Could not get length of."


pekkaklarck:
The idea seems to be to change `Get Length` so that it works with integers so that e.g. `1234` would yield 4. I don't think that's a good idea for the following reasons:
- Most of the time you don't want numbers to have length. It's generally better to fail than return strange values.
- You can easily convert an integer to a string first and get length from it.
- Integers in Python actually have [a kind of a length](https://docs.python.org/3/library/stdtypes.html#int.bit_length), but it is not at all the same as proposed here.
- Getting a length as proposed is pretty ambiguous. What should be the length of `0`? What about negative numbers?
- Should also floats have a length?

Issue #4659, opened 2023-02-15
Performance regression when using `Run Keyword` and keyword name contains a variable


dinualx:
We are using RF 3.2.2 in production and we have decided to upgrade to the latest RF, 6.0.2.
But we encountered the situation that the execution time for our test cases increased extremely.

We started to downgrade incrementally until we reached RF 5.0.1, with which we no longer had this problem. So, starting with RF 6.0, this problem appears. With RF 5.0.1 the results are similar to what we have in production, RF 3.2.2, there are no differences.

As a concrete example, for a certain **testSetup** keyword, in production (RF 3.2.2) it takes about 1 minute, and with RF>6.0 it takes 14 minutes. In order to isolate the problem and see why this increase in time execution occurs, we moved all the keywords from the robot resource files that we were using (2 resource files), inside the test suite using the respective keywords without importing the robot resource files. In this way, the running time of this keyword decreased from 14 minutes to approximately 7 minutes. It's an obvious decrease, but it still hasn't reached the value we have in production, of 1 minute. We think that there's still another problem here and if we'll manage to isolate it further we will log a separate bug for it.

I added a screenshot showing these differences. We removed from the report the names of the respective keywords specific to our project for reasons of confidentiality.

![comparison](https://user-images.githubusercontent.com/101075348/219031978-444fdecb-5c0f-47c0-97ff-56dfd5d7790f.png)

dinualx:
The problem with the increased time seems to occur when we are performing operations which have as purpose adding the necessary elements for our config in the namespace. To populate the necessary objects we use some information stored in a dictionary, we create an instance in robot for a certain object, we populate that instance with all its properties using multiple python methods and we do this for a lot of elements that are needed in the test configuration.

To call these python methods and be able to apply them efficiently, we use different keywords with several arguments, and under these keywords it is necessary to use the run keyword syntax, they cannot be called otherwise. These keywords are placed in robot resource files which are imported into the test suites. Almost all these keywords are placed under one big keyword that includes them all, that of testSetup.

Keyword example of adding properties to an instance:

SetProperties
    [Arguments]  ${non_dict_property1}  ${name}  ${non_dict_property2}  ${propertyDict}  
    Run Keyword  ${name}.setmethod1 ${ non_dict_property1}
    Run Keyword  ${name}.setmethod2  ${propertyDict['${name}']}
    Run Keyword  ${name}.setmethod3  ${propertyDict['${name}']}
[...]

Issue #1595, opened 2014-06-30
Embedded arguments are not passed as objects when executed with `Run Keyword` or its variants


spooning:
> _Originally submitted to [Google Code](http://code.google.com/p/robotframework/issues/detail?id=1595) by @jussimalinen on 29 Nov 2013_

This does not work:

```robotframework
*** Test Cases ***
Example
    ${foo} =      Evaluate    {'a':42}
    Huu '${foo}'      # Works
    Run Keyword       Huu '${foo}'      # Does not work
    Run Keyword If    True    Huu '${foo}'      # Does not work either

*** Keywords ***
Huu '${foo}'
    Log    ${foo['a']}    WARN
```

This has never worked. Run Keyword variants accept the name of the keyword to execute to be specified as a variable, and thus the variable part is resolved too early. In the above example `Run Keyword` ends up running keyword `Huu '{'a': 42}'`, and thus the `${foo}` argument passed to the keyword will be string `{'a': 42}`.

pekkaklarck:
Implementing #2699 would be a workaround for this problem with lists and dictionaries (and a useful feature in general). It's currently targeted for RF 3.2, but I'm not sure do I have time to look at it so it may end up being descoped like it was from RF 3.1. Anyone else interested to implement it? It shouldn't cause backwards compatibility problems so could possibly be added already in RF 3.1.2.

pekkaklarck:
Issue #3278 has more potential than #2699 and would help with this problem as well. Neither of these issues would help with custom object but Python base types, including containers, ought to work.

MariuszJanczura:
I could look at it but I would need to have more details regarding that and some time I suppose :)

pekkaklarck:
I think the steps would be these:

1. Change `run_keyword` (and if we can make this work, all keywords that use it) to not resolve variables in arguments at all. That as easy as using `@run_keyword_variant(resolve=0)`.

2. Try finding a keyword without replacing variables in `name`. If that fails, try replacing variables.

3. Some care must be taken to handle `name` that is defined as `@{list}` variable (i.e. when it contains also arguments). Probably should just replace variables in that case.

pekkaklarck:
Some code examples (untested) that could be useful:
```python
match = search_variable(name)
if not match:
    # did not contain variables at all, act accordingly
elif match.is_list_variable:
    # list variable - need to replace it and use the first value as name
    args = self._variables.replace_list([name] + args, replace_until=1)
    name = args.pop(0)
[...]

PR #4899, opened 2023-10-13
libraries: BuiltIn: _accepts_embedded_arguments: do not recommend on …

totu:
…failure

We are not using this runner for anything here, so let's not gather recommendations on failure.

This is a pretty big performance hit for no apparent benefit.

We have to introduce `recommend_on_failure` variable to _ExecutionContext's get_runner(), but we can keep the default behaviour the same.

This might be fix for #4659 I found this for very similar performance issue. This came in 5.1b2 when the decoration for `run_keyword()` in BuiltIn was changed.

### DIFF
--- a/src/robot/libraries/BuiltIn.py
+++ b/src/robot/libraries/BuiltIn.py
@@ -1877,7 +1877,7 @@ class _RunKeyword(_BuiltInBase):
 
     def _accepts_embedded_arguments(self, name, ctx):
         if '{' in name:
-            runner = ctx.get_runner(name)
+            runner = ctx.get_runner(name, recommend_on_failure=False)
             return runner and hasattr(runner, 'embedded_args')
         return False
 
--- a/src/robot/running/context.py
+++ b/src/robot/running/context.py
@@ -304,8 +304,8 @@ class _ExecutionContext:
         method(data, result)
         self.steps.pop()
 
-    def get_runner(self, name):
-        return self.namespace.get_runner(name)
+    def get_runner(self, name, recommend_on_failure=True):
+        return self.namespace.get_runner(name, recommend_on_failure)
 
     def trace(self, message):
         self.output.trace(message)


pekkaklarck:
Could you clarify the problem you have and how this PR helps with that? If there are performance enhancements, do you have concrete numbers?

totu:
We have a lot of proprietary protocols implemented in Python which are then dynamically imported in Robot Framework. We are mainly calling into these libraries with the `Run Keyword` keyword from BuiltIn library ala:

`Run Keyword    ${lib}.Send Message     payload=${payload}
`

We have been running Robot Framework 4.1.2 for long while and now decided it was time to upgrade, but I noticed that these `Run Keyword` calls were taking 40-100ms to find the library after which the actual function only takes ~5ms to execute.

From my experiments the actual change was made in 5.1b2 where the decorator for `run_keyword()` function was change from `resolve=1` to `0`.  Changing this back also got the `Run Keyword` to execute in ball park of 1-5ms depending on test case. Most of the time was spent in `running/namespace.py` function `_raise_no_keyword_found()` specifically in line 283 where we are finding the recommended similar keyword, which at the end of the day we are discarding the in the `_accept_embedded_arguments()`.

An example test timings compared with release 6.1.1, 6.1.1 with the decorator changed back to 1, and my changes
```
Version          Test Run Time    Run Keyword Run Time
6.1              35sec            40-100ms
6.1-resolve=1    19.1sec          7ms    
6.1-my-change    19.4sec          7ms       
```
In all cases 5ms of that time is spec in our Python library.

pekkaklarck:
I was able to easily reproduce the performance regression with the following example. It runs in ~1.4s on my machine with the latest code and in ~0.7s with RF 5.0. I also tested that if I disable recommendations altogether, the execution time drops to ~0.8s. Still slower than with RF 5.0 but much closer! 

```robotframework
*** Variables ***
${KW}     No Operation

*** Test Cases ***
Example
    FOR    ${i}    IN RANGE    1000
        Run Keyword    ${KW}
    END
```

You are most likely right that this is the root cause for the performance regression reported in #4659. I had already earlier [analyzed](https://github.com/robotframework/robotframework/issues/4659#issuecomment-1448969663) that the problem was caused by a fix for #1595 that we didn't want to revert. I thought that slowness was caused only by needing to look for the keyword twice, and in the [following comment](https://github.com/robotframework/robotframework/issues/4659#issuecomment-1449078069) wrote that fixing isn't easy. It seems I was wrong and slowness was mostly caused by finding recommendations that were never even used for anything. Great that you found the real root cause!

The fix looks good and I'll merge this shortly. I'll also add #4659 to RF 7.0 scope and close it.

PS: You may want to consider using library search order instead of using `Run Keyword`. Depending on your usage, it will likely help cleaning up your tests and will also be faster without the extra overhead by `Run Keyword`. See the [aforementioned comment](https://github.com/robotframework/robotframework/issues/4659#issuecomment-1449078069) for more details.

PR #4911, opened 2023-10-24
performance optimization in visiting models

d-biehl:
as discussed in [slack](https://robotframework.slack.com/archives/C0K0240NL/p1698071160283019), here are some performance optimizations for the ModelVisitor and ModelTransformer.

I will make several commits, one for each improvement.

My local tests promise a speed increase of about factor 4.

This improvement has not necessarily influence on execution speed of Robot but all tools which use the ModelVisitor or the ModelTransformer, like RobotCode, robocop or tidy.

### DIFF
--- a/src/robot/parsing/model/statements.py
+++ b/src/robot/parsing/model/statements.py
@@ -47,7 +47,7 @@ class Node(ast.AST, ABC):
 
 
 class Statement(Node, ABC):
-    _fields = ('type', 'tokens')
+    _attributes = ('type', 'tokens') + Node._attributes
     type: str
     handles_types: 'ClassVar[tuple[str, ...]]' = ()
     statement_handlers: 'ClassVar[dict[str, Type[Statement]]]' = {}
--- a/src/robot/parsing/model/statements.py
+++ b/src/robot/parsing/model/statements.py
@@ -47,7 +47,7 @@ class Node(ast.AST, ABC):
 
 
 class Statement(Node, ABC):
-    _fields = ('type', 'tokens')
+    _attributes = (*Node._attributes, 'tokens', 'type')
     type: str
     handles_types: 'ClassVar[tuple[str, ...]]' = ()
     statement_handlers: 'ClassVar[dict[str, Type[Statement]]]' = {}
--- a/src/robot/parsing/model/statements.py
+++ b/src/robot/parsing/model/statements.py
@@ -47,7 +47,7 @@ class Node(ast.AST, ABC):
 
 
 class Statement(Node, ABC):
-    _attributes = (*Node._attributes, 'tokens', 'type')
+    _fields = ('type', 'tokens')
     type: str
     handles_types: 'ClassVar[tuple[str, ...]]' = ()
     statement_handlers: 'ClassVar[dict[str, Type[Statement]]]' = {}
--- a/src/robot/parsing/model/visitor.py
+++ b/src/robot/parsing/model/visitor.py
@@ -14,14 +14,14 @@
 #  limitations under the License.
 
 import ast
-from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, Type, Union
+from typing import Any, Callable, Iterator
 
 from ...utils.notset import NOT_SET, NotSet
 from .statements import Node
 
 
 class VisitorFinder:
-    __cls_finder_cache__: Dict[Type[Any], Union[Callable[..., Any], None, NotSet]]
+    __cls_finder_cache__: "dict[type[Any], Callable[..., Any]|None|NotSet]"
 
     def __new__(cls, *_args: Any, **_kwargs: Any):  # type: ignore[no-untyped-def]
         if not hasattr(cls, "__cls_finder_cache__"):
@@ -29,7 +29,7 @@ class VisitorFinder:
         return super().__new__(cls)
 
     @classmethod
-    def __find_visitor(cls, node_cls: Type[Any]) -> Optional[Callable[..., Any]]:
+    def __find_visitor(cls, node_cls: "type[Any]") -> "Callable[..., Any]|None":
         if node_cls is ast.AST:
             return None
         method_name = "visit_" + node_cls.__name__
@@ -47,14 +47,14 @@ class VisitorFinder:
         return None
 
     @classmethod
-    def _find_visitor_cached(cls, node_cls: Type[Any]) -> Optional[Callable[..., Any]]:
+    def _find_visitor_cached(cls, node_cls: "type[Any]") -> "Callable[..., Any]|None":
         result = cls.__cls_finder_cache__.get(node_cls, NOT_SET)
         if result is NOT_SET:
             result = cls.__cls_finder_cache__[node_cls] = cls.__find_visitor(node_cls)
         return result  # type: ignore[return-value]
 
 
-def _iter_field_values(node: Node) -> Iterator[Union[Node, List[Node], None]]:
+def _iter_field_values(node: Node) -> "Iterator[Node|list[Node]|None]":
     for field in node._fields:
         try:
             yield getattr(node, field)
@@ -62,7 +62,7 @@ def _iter_field_values(node: Node) -> Iterator[Union[Node, List[Node], None]]:
             pass
 
 
-def iter_fields(node: Node) -> Iterator[Tuple[str, Union[Node, List[Node], None]]]:
+def iter_fields(node: Node) -> "Iterator[tuple[str, Node|list[Node]|None]]":
     for field in node._fields:
         try:
             yield field, getattr(node, field)
@@ -106,11 +106,11 @@ class ModelTransformer(VisitorFinder):
     <https://docs.python.org/library/ast.html#ast.NodeTransformer>`__.
     """
 
-    def visit(self, node: Node) -> Union[Node, List[Node], None]:
+    def visit(self, node: Node) -> "Node|list[Node]|None":
         visitor = self._find_visitor_cached(type(node)) or self.__class__.generic_visit
         return visitor(self, node)
 
-    def generic_visit(self, node: Node) -> Union[Node, List[Node], None]:
+    def generic_visit(self, node: Node) -> "Node|list[Node]|None":
         for field, old_value in iter_fields(node):
             if old_value is None:
                 continue
--- a/src/robot/parsing/model/visitor.py
+++ b/src/robot/parsing/model/visitor.py
@@ -13,24 +13,29 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-import ast
-from typing import Any, Callable, Iterator
+from abc import ABC
+from ast import AST
+from collections import defaultdict
+from typing import Any, Callable, Generic, Iterator, TypeVar
 
 from ...utils.notset import NOT_SET, NotSet
 from .statements import Node
 
+TVisitorResult = TypeVar("TVisitorResult")
 
-class VisitorFinder:
-    __cls_finder_cache__: "dict[type[Any], Callable[..., Any]|None|NotSet]"
 
-    def __new__(cls, *_args: Any, **_kwargs: Any):  # type: ignore[no-untyped-def]
-        if not hasattr(cls, "__cls_finder_cache__"):
-            cls.__cls_finder_cache__ = {}
-        return super().__new__(cls)
+class VisitorFinder(ABC, Generic[TVisitorResult]):
+    __visitor_finder_cache: "dict[type[Any], Callable[[Node], Any]|None|NotSet]"
+    __default_visitor_method: "Callable[..., TVisitorResult]"
+
+    def __init_subclass__(cls, default_visitor_name: str = "generic_visit", **kwargs: Any) -> None:
+        super().__init_subclass__(**kwargs)
+        cls.__visitor_finder_cache = defaultdict(lambda: NOT_SET)
+        cls.__default_visitor_method = getattr(cls, default_visitor_name)
 
     @classmethod
-    def __find_visitor(cls, node_cls: "type[Any]") -> "Callable[..., Any]|None":
-        if node_cls is ast.AST:
+    def _find_visitor_class_method(cls, node_cls: "type[Any]") -> "Callable[..., TVisitorResult]|None":
+        if node_cls is AST:
             return None
         method_name = "visit_" + node_cls.__name__
         method = getattr(cls, method_name, None)
@@ -41,17 +46,17 @@ class VisitorFinder:
             if callable(method):
                 return method  # type: ignore[no-any-return]
         for base in node_cls.__bases__:
-            method = cls._find_visitor_cached(base)
+            method = cls._find_visitor_class_method(base)
             if method:
                 return method
         return None
 
     @classmethod
-    def _find_visitor_cached(cls, node_cls: "type[Any]") -> "Callable[..., Any]|None":
-        result = cls.__cls_finder_cache__.get(node_cls, NOT_SET)
+    def _find_visitor(cls, node_cls: "type[Any]") -> Callable[..., TVisitorResult]:
+        result = cls.__visitor_finder_cache[node_cls]
         if result is NOT_SET:
-            result = cls.__cls_finder_cache__[node_cls] = cls.__find_visitor(node_cls)
-        return result  # type: ignore[return-value]
+            result = cls.__visitor_finder_cache[node_cls] = cls._find_visitor_class_method(node_cls)
+        return result or cls.__default_visitor_method  # type: ignore[return-value]
 
 
 def _iter_field_values(node: Node) -> "Iterator[Node|list[Node]|None]":
@@ -70,7 +75,7 @@ def iter_fields(node: Node) -> "Iterator[tuple[str, Node|list[Node]|None]]":
             pass
 
 
-class ModelVisitor(VisitorFinder):
+class ModelVisitor(VisitorFinder[None]):
     """NodeVisitor that supports matching nodes based on their base classes.
 
     In other ways identical to the standard `ast.NodeVisitor
@@ -84,7 +89,7 @@ class ModelVisitor(VisitorFinder):
     """
 
     def visit(self, node: Node) -> None:
-        visitor = self._find_visitor_cached(type(node)) or self.__class__.generic_visit
+        visitor = self._find_visitor(type(node))
         visitor(self, node)
 
     def generic_visit(self, node: Node) -> None:
@@ -98,7 +103,7 @@ class ModelVisitor(VisitorFinder):
                 self.visit(value)
 
 
-class ModelTransformer(VisitorFinder):
+class ModelTransformer(VisitorFinder["Node|list[Node]|None"]):
     """NodeTransformer that supports matching nodes based on their base classes.
 
     See :class:`ModelVisitor` for explanation how this is different compared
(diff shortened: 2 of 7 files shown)


src/robot/parsing/model/statements.py
```diff
@@ -47,7 +47,7 @@ class Node(ast.AST, ABC):
 
 
 class Statement(Node, ABC):
-    _fields = ('type', 'tokens')
+    _attributes = (*Node._attributes, 'tokens', 'type')
```
pekkaklarck, line 50: We still need to double check this change.

src/robot/parsing/model/visitor.py
```diff
@@ -14,27 +14,49 @@
...
+            if callable(method):
+                return method  # type: ignore[no-any-return]
+        for base in node_cls.__bases__:
+            method = cls._find_visitor(base)
+            if method:
+                return method
         return None
 
+    @classmethod
+    def _find_visitor(cls, node_cls: Type[Any]) -> Optional[Callable[..., Any]]:
```
pekkaklarck, line 58: I don't think there's a need for both `_find_visitor` and `__find_visitor`. It ought to be enough to only have `_find_visitor` (with the current `__find_visitor` implementation) and then do cache checking and updating in `visit`.

src/robot/parsing/model/visitor.py
```diff
@@ -62,6 +84,6 @@ class ModelTransformer(ast.NodeTransformer, VisitorFinder):
     <https://docs.python.org/library/ast.html#ast.NodeTransformer>`__.
     """
 
-    def visit(self, node: Node):
-        visitor = self._find_visitor(type(node)) or self.generic_visit
-        return visitor(node)
+    def visit(self, node: Node) -> Node:
+        visitor = self._find_visitor(type(node)) or self.__class__.generic_visit
+        return visitor(self, node)
```
pekkaklarck, line 119: This method could be probably implemented as `return VisitorFinder.visit(self, node)` to avoid duplication. Possibly it would also be enough to just change the order of inherited classes.

pekkaklarck (commented):
Looks good in general. Things to be done:
1. The `_fields` -> `_attributes` change needs to be still investigated in a bit more detail.
2. Cache creation can be enhanced.
3. Probably updating the cache can be streamlined as well.

src/robot/parsing/model/visitor.py
```diff
@@ -14,27 +14,49 @@
...
 class VisitorFinder:
+    __NOT_SET = _NotSet()
+    __cls_cache: Dict[Type[Any], Union[Callable[..., Any], None, _NotSet]]
 
-    def _find_visitor(self, cls):
-        if cls is ast.AST:
+    def __new__(cls, *_args: Any, **_kwargs: Any):  # type: ignore[no-untyped-def]
+        # create cache on class level to avoid creating it for each instance
+        cls.__cls_cache = {}
+        return super().__new__(cls)
```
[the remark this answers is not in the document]
pekkaklarck: My comment above was written before the code changed. The new version looks even more complicated than the earlier. It seems like it would try to avoid recreating the cache with new instances that I mentioned above. In that case my proposed implementation is quite a bit simpler.

src/robot/parsing/model/visitor.py
```diff
@@ -14,27 +14,49 @@
 #  limitations under the License.
 
 import ast
+from typing import Any, Callable, Dict, Optional, Type, Union
 
 from .statements import Node
 
 
+class _NotSet:
+    pass
```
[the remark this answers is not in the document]
d-biehl: ok, this is new in RF 7 ;-) will change that

src/robot/parsing/model/visitor.py
pekkaklarck: Looking at this again, doing cache handling in `visit´ isn't probably a good idea. One method for actually finding the visitor and another for managing the cache is probably better. They just need to be named better.

src/robot/parsing/model/visitor.py
```diff
@@ -14,27 +14,49 @@
...
+    __cls_cache: Dict[Type[Any], Union[Callable[..., Any], None, _NotSet]]
 
-    def _find_visitor(self, cls):
-        if cls is ast.AST:
+    def __new__(cls, *_args: Any, **_kwargs: Any):  # type: ignore[no-untyped-def]
+        # create cache on class level to avoid creating it for each instance
+        cls.__cls_cache = {}
+        return super().__new__(cls)
+
+    @classmethod
```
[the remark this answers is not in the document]
d-biehl: because I want to cache the methods on class level and not on instance level, defining this as a classmethod save me a type() call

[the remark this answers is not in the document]
pekkaklarck: After looking at the code again, this being a classmethod looks like a good idea. We want to get the visitor method from the class, not the instance, anyway.

src/robot/parsing/model/visitor.py
```diff
@@ -14,27 +14,49 @@
...
+        for base in node_cls.__bases__:
+            method = cls._find_visitor(base)
+            if method:
+                return method
         return None
 
+    @classmethod
+    def _find_visitor(cls, node_cls: Type[Any]) -> Optional[Callable[..., Any]]:
+        result = cls.__cls_cache.get(node_cls, cls.__NOT_SET)
+        if result is cls.__NOT_SET:
```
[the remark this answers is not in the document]
d-biehl: no, this is `node_cls in cls.__cls_cache' and then `cls.__cls_cache[node_cls]` in one call this saves time because an `is` check is faster then a dictionary operation. 

NOT_SET is needed because if we did not found a method then we get None

src/robot/parsing/model/statements.py
d-biehl: yep, the attribute field is not used in RF in the `ast` module it is only used for duming the model.

src/robot/parsing/model/visitor.py
d-biehl: renamed it in the last commit

src/robot/parsing/model/visitor.py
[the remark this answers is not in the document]
d-biehl: simplify this again in my last commit

src/robot/parsing/model/visitor.py
```diff
@@ -14,27 +14,49 @@
 #  limitations under the License.
 
 import ast
+from typing import Any, Callable, Dict, Optional, Type, Union
```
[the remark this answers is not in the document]
d-biehl: saw both versions in the code ;-) 
will change it.

pekkaklarck:
I investigated the `_fields` -> `_attributes` change and it seems to affect at least logging/debugging. Changed can be seen with this code:
```python
import ast
import astpretty
from robot.parsing import get_model


kw = get_model('''\
*** Test Cases ***
Example
    Log    Hello, world!
''').sections[0].body[0].body[0]

print('[ast.dump]')
print(ast.dump(kw))
print()
print('[ast.dump w/ attributes]')
print(ast.dump(kw, include_attributes=True))
print()
print('[astpretty]')
astpretty.pprint(kw)
```
When I run the above code with the current code, I get this:
```
[ast.dump]
KeywordCall(type='KEYWORD', tokens=(Token(SEPARATOR, '    ', 3, 0), Token(KEYWORD, 'Log', 3, 4), Token(SEPARATOR, '    ', 3, 7), Token(ARGUMENT, 'Hello, world!', 3, 11), Token(EOL, '\n', 3, 24)))

[ast.dump w/ attributes]
KeywordCall(type='KEYWORD', tokens=(Token(SEPARATOR, '    ', 3, 0), Token(KEYWORD, 'Log', 3, 4), Token(SEPARATOR, '    ', 3, 7), Token(ARGUMENT, 'Hello, world!', 3, 11), Token(EOL, '\n', 3, 24)), lineno=3, col_offset=0, end_lineno=3, end_col_offset=25, errors=())

[astpretty]
KeywordCall(lineno=3, col_offset=0, end_lineno=3, end_col_offset=25, errors=(), type='KEYWORD', tokens=(Token(SEPARATOR, '    ', 3, 0), Token(KEYWORD, 'Log', 3, 4), Token(SEPARATOR, '    ', 3, 7), Token(ARGUMENT, 'Hello, world!', 3, 11), Token(EOL, '\n', 3, 24)))
```
As you can see, `type` and `tokens` are always logged, but `lineno` and other attributes aren't included when using plain `ast.dump()`. The other two print the same information but in slightly different orher.

When the current `_fields = ('type', 'tokens')` is changed to `_attributes = ('type', 'tokens') + Node._attributes`, the output changes to this:
```
[ast.dump]
KeywordCall()

[ast.dump w/ attributes]
KeywordCall(type='KEYWORD', tokens=(Token(SEPARATOR, '    ', 3, 0), Token(KEYWORD, 'Log', 3, 4), Token(SEPARATOR, '    ', 3, 7), Token(ARGUMENT, 'Hello, world!', 3, 11), Token(EOL, '\n', 3, 24)), lineno=3, col_offset=0, end_lineno=3, end_col_offset=25, errors=())

[astpretty]
KeywordCall(type='KEYWORD', tokens=(Token(SEPARATOR, '    ', 3, 0), Token(KEYWORD, 'Log', 3, 4), Token(SEPARATOR, '    ', 3, 7), Token(ARGUMENT, 'Hello, world!', 3, 11), Token(EOL, '\n', 3, 24)), lineno=3, col_offset=0, end_lineno=3, end_col_offset=25, errors=())
```

As you can see, plain `ast.dump()` doesn't anymore show much information. That's expected because now also `type` and `tokens` are attributes. The other two show the same information as earlier and now also they show it in the same order.

Based on this, I believe the change is fine. The `ast.dump()` output is rather complicated with bigger modelss so not showing tokens by default can be seen as a plus. As @d-biehl pointed out on Slack, `_fields` should contain names of the child nodes and tokens and even less type are such nodes.

The change can obviously cause issues if someone is inspecting `_fields` or `_attributes` explicitly, but I don't see why someone would be interested. It is nevertheless best to submit a separate issue about this change and mark it backwards incompatible so that we remember to mention it in the release notes. I'll do that.

d-biehl:
> The change can obviously cause issues if someone is inspecting _fields or _attributes explicitly, but I don't see why someone would be interested. It is nevertheless best to submit a separate issue about this change and mark it backwards incompatible so that we remember to mention it in the release notes. I'll do that.

the new generic_visitor method needs this change, so when then this should be done before integrating this PR

d-biehl:
Now:

before optimization
```
Found 1291 files
inner orig       min 203.0 ms max 235.0 ms avg 225.0 ms
common orig      min 218.0 ms max 235.0 ms avg 228.0 ms
```

after optimization
```
Found 1291 files
inner orig       min 46.00 ms max 63.00 ms avg 54.69 ms
common orig      min 46.99 ms max 78.00 ms avg 60.89 ms
```

pekkaklarck:
I submitted #4912 about the `_fields` -> `_attributes` change. I have done the change locally and run tests. Should I @d-biehl commit and push changes so that you can then rebase this PR? That change is uncontroversial and simple and getting it out of the way would be good.

src/robot/parsing/model/visitor.py
```diff
@@ -14,29 +14,63 @@
...
 
 class VisitorFinder:
+    __cls_finder_cache__: "dict[type[Any], Callable[..., Any]|None|NotSet]"
 
-    def _find_visitor(self, cls):
-        if cls is ast.AST:
+    def __new__(cls, *_args: Any, **_kwargs: Any):  # type: ignore[no-untyped-def]
+        if not hasattr(cls, "__cls_finder_cache__"):
+            cls.__cls_finder_cache__ = {}
+        return super().__new__(cls)
```
pekkaklarck, line 29: 1. Unless there's some concrete need to use `__new__`, this should be changed to `__init__`. It's less magical.
2. If `__new__` is needed, arguments it accepts should be passed forward to `super().__new__`.
3. `__dunder__` attributes shouldn't in general be used outside Python itself or specialized frameworks. We should just use `finder_cache`, `visitor_cache`, `visitor_method_cache`, or something like that. If you are afraid of attribute collision, using `__` prefix is fine.
4. `hasattr` could be avoided by initializing the cache to `None` on the class level. I guess it's a matter of taste does `not hasattr(foo, 'bar')` or `foo.bar is None` look better.

src/robot/parsing/model/visitor.py
```diff
@@ -14,29 +14,63 @@
...
+        for base in node_cls.__bases__:
+            method = cls._find_visitor_cached(base)
+            if method:
+                return method
         return None
 
+    @classmethod
+    def _find_visitor_cached(cls, node_cls: "type[Any]") -> "Callable[..., Any]|None":
+        result = cls.__cls_finder_cache__.get(node_cls, NOT_SET)
+        if result is NOT_SET:
```
pekkaklarck, line 52: What benefits does `NOT_SET` have here? Why isn't `node_cls in cache` enough?

src/robot/parsing/model/visitor.py
```diff
@@ -49,19 +83,53 @@ def visit_Statement(self, node):
             ...
     """
 
-    def visit(self, node: Node):
-        visitor = self._find_visitor(type(node)) or self.generic_visit
-        visitor(node)
+    def visit(self, node: Node) -> None:
+        visitor = self._find_visitor_cached(type(node)) or self.__class__.generic_visit
+        visitor(self, node)
```
pekkaklarck, line 88: 1. You used `_find_visitor_cached` and `__find_visitor` as method names. Possibly `_find_visitor` and `_find_fisitor_from_class` could be a bit better. Either way, names should use `_` or `__` prefix consistently. Probably `_` is better.
2. `_find_visitor` could possibly return `cls.generic_visit` instead of `None` to avoid `self.__class__.generic_visit` here.
3. If the above change doesn't make sense, `type(self).generic_visit` would be a bit nicer than `self.__class__.generic_visit`.

pekkaklarck (commented):
Changes to `generic_visit` look pretty complicated. I think it would be better to concentrate on listener method cache first.

You can ignore comments related to the code style if you want. I can fine-tune the code after merging.

d-biehl:
yes, you can commit an push and I will rebase this PR. But probably not before tomorrow.

src/robot/parsing/model/visitor.py
pekkaklarck: Yeah, looks like a perfect use case for `__init_subclass__`. I had heard about it earlier, but have never used it for anything.

src/robot/parsing/model/visitor.py
pekkaklarck: I noticed an older comment claiming that using `get` with `NOT_SET` would be faster than using `in`. That sounds a bit odd because both require a dictionary lookup but the former also requires one `is` usage. I decided to test it and at least with Python 3.11 on my machine `in` is a bit faster both when the item exists and when it doesn't. The code is also more idiomatic:
```python
>>> from timeit import timeit
>>>
>>> notset = object()
>>> 
>>> def t1(d, k):
...     v = d.get(k, notset)
...     if v is notset:
...        d[k] = 42
...     return d
... 
>>> def t2(d, k):
...     if k not in d:
...        d[k] = 42
...     return d
...
>>> timeit(lambda: t1({}, 1))
0.12673325699870475
>>> timeit(lambda: t2({}, 1))
0.10616556800232502
>>> timeit(lambda: t1({1: 2}, 1))
0.12745981800253503
>>> timeit(lambda: t2({1: 2}, 1))
0.10945483400428202
```

src/robot/parsing/model/visitor.py
d-biehl: ok, I changed it to __init_subclass__

src/robot/parsing/model/visitor.py
```diff
@@ -49,19 +88,53 @@ def visit_Statement(self, node):
...
+                        new_values.extend(new_value)
+                        continue
+                    new_values.append(new_value)
+                old_value[:] = new_values
+            else:
+                new_node = self.visit(old_value)
+                if new_node is None:
+                    delattr(node, field)
+                else:
+                    setattr(node, field, new_node)
```
pekkaklarck, line 138: This looks rather complicated and thus error prone. I strongly believe this PR should concentrate on visitor method caching and possible other enhancements left for future.

pekkaklarck (changes requested):
This PR is getting too complicated. As I wrote in a separate comment, I believe this PR should concentrate on visitor method caching and everything else left out. Caching is relatively simple, but the custom `generic_visit` implementation looks complicated and somewhat risky. I'd like to have simple stuff in first. After that more complicated changes are fine if they bring additional significant  performance benefits.

I'm also not that happy with the added typing because it makes the code pretty complicated. The main reason is that the syntax for typing callables in Python is pretty horrible. I also don't see any real need for making the class generic (which requites TypeVar). This isn't a public API and trying to type it fully is a no-goal. Could be done if types don't get in the way, but in this case they do.

Due to the budget situation with RF 7.0, I won't have much time to spend for this. If you consider the changes that I oppose important, it might be better that the code lives outside Robot core.

src/robot/parsing/model/visitor.py
d-biehl: In Python, each dictionary operation starts with:

- Generating the hash of the key
- searching for the hash value in the list of keys

and then do_something with the element....

this is not a problem with small dictionaries and with simple classes where the hash can be easily determined (i.e. int, str, float, type). But if you have big dictionaries and then with an own hash algorithm (the __hash__ method was overwritten) then it gets more complicated and then the method with `notset` is faster, but the one with the `defaultdict` a little bit more. `get` is a method and `[]` is an operator. In most variants an operator is faster.

So your simple test above is not really meaningful ;-)

Since in our case we are dealing with relatively small dictionaries and with a hashkey (type) handled by python, there is not much difference to measure.
That's why I adapted the code according to your wishes.

I can provide you with a simple script to measure visitor performance. I also have a more complex script to measure dictionary accesses, if you are interested ;-)

d-biehl:
```
>>> def t1(d, k):
...     v = d.get(k, notset)  # << dictionary operation
...     if v is notset:
...        v = d[k] = 42  # << dictionary operation
...     return v
... 
>>> def t2(d, k):
...     if k not in d:  # << dictionary operation
...        d[k] = 42  # << dictionary operation
...     return d[k]  # << dictionary operation
```

src/robot/parsing/model/visitor.py
d-biehl: i will remove this for now, although it also brings a little more performance. And the way we use it here breaks the Liskov substitution principle

d-biehl: We can also change it later if wanted. However, these are only slightly optimized forms from the original implementation of `NodeVisitor` and `NodeTransformer`.

src/robot/parsing/model/visitor.py
```diff
@@ -13,30 +13,69 @@
...
-    def _find_visitor(self, cls):
-        if cls is ast.AST:
+class VisitorFinder(ABC, Generic[TVisitorResult]):
+    __visitor_finder_cache: "dict[type[Any], Callable[[Node], Any]|None|NotSet]"
+    __default_visitor_method: "Callable[..., TVisitorResult]"
+
+    def __init_subclass__(cls, default_visitor_name: str = "generic_visit", **kwargs: Any) -> None:
+        super().__init_subclass__(**kwargs)
+        cls.__visitor_finder_cache = defaultdict(lambda: NOT_SET)
+        cls.__default_visitor_method = getattr(cls, default_visitor_name)
```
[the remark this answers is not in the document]
d-biehl: changed this

pekkaklarck:
The latest version looked good. I'll take another locally and may make some fine-tuning. I'll also add a note about the cache to ModelVisitor docstring.

pekkaklarck:
Now that this PR is in and will be part of RF 7.0 alpha 1, a separate PR can be submitted about enhancing `generic_visit`. If it's not too complicated or fragile (`NodeVisitor` may change in the future) and it enhances performance considerably, we can still get it before RF 7.0 final. Possible related PRs can be linked to issue #4934 that I submitted for tracking purposes.

Issue #4910, opened 2023-10-24
Make listener v3 the default listener API


pekkaklarck:
Currently when creating listeners, you need to explicitly specify are you using listener API version 2 or 3. Now that the v3 API is getting better (#3296), we can make it the default. In other words, if API version isn't specified, we assume the v3 API is used.

PR #4916, opened 2023-10-28
Listener version default - #4910

mitzeee:
PR to implement issue #4910.


atest/robot/output/listener_interface/unsupported_listener_version.robot
```diff
@@ -10,10 +10,6 @@ Unsupported version
     1    unsupported_listeners.InvalidVersionClassListener
     ...    Listener 'unsupported_listeners.InvalidVersionClassListener' uses unsupported API version 'kekkonen'.
 
-No version information
-    2    unsupported_listeners
-    ...    Listener 'unsupported_listeners' does not have mandatory 'ROBOT_LISTENER_API_VERSION' attribute.
-
```
pekkaklarck, line 16: A new test is needed to make sure using a listener without `ROBOT_LISTENER_API_VERSION` is possible.

src/robot/output/listeners.py
```diff
@@ -152,9 +152,8 @@ def _get_version(self, listener):
             if version not in (2, 3):
                 raise ValueError
         except AttributeError:
-            raise DataError("Listener '%s' does not have mandatory "
-                            "'ROBOT_LISTENER_API_VERSION' attribute."
-                            % self.name)
+            "Sicne 7.1, Listener API version is by default considered to be V3 unless explicitly mentioned as V2"
```
pekkaklarck, line 155: No need for this kind of a comment. If it was needed, it should be a normal comment starting with a `#`.

pekkaklarck (changes requested):
In addition to the changes mentioned in separate comments, also the User Guide needs to be updated.

pekkaklarck:
Thanks for a PR! As explained in my review, some changes are still needed. In addition to that, the listener code is currently in flux and there will be conflict. PR #4920 ought to contain all changes related to this, so it's a good idea to wait until it is merged (which should happen latest tomorrow).

pekkaklarck:
PR #4920 is now in and there's indeed a conflict. Let us know @mitzeee are you interested to continue with this PR. If not, we can make required changes also ourselves.

mitzeee:
@pekkaklarck Thanks for the comments. I will definitely make the changes and push.
* New test
* Comment correction
* User-guide update

pekkaklarck:
I need this functionality now and will implement it myself. Thanks anyway!

PR #4920, opened 2023-11-01
Listener refactoring

yanne:


src/robot/output/listeners.py
```diff
@@ -15,298 +15,591 @@
...
-    def discard_suite_scope(self):
-        self.listener.discard_suite_scope()
+class ListenerFacade(LoggerApi):
 
-    def message(self, message: 'model.Message'):
-        self.listener.message(message)
+    def __init__(self, listener, name, prefix=''):
+        self.listener = listener
+        self.name = name
+        self.prefix = prefix
```
pekkaklarck, line 214: We only use `prefix` when a leading underscore is allowed. Changing `prefix=''` to `allow_leading_underscore=False` could be a good idea.

src/robot/output/listeners.py
```diff
@@ -15,298 +15,591 @@
...
+                'patterns': list(result.patterns),
+                'pattern_type': result.pattern_type,
+                'variable': result.assign
+            })
+        return attrs
+
+    def _return_v2_attributes(self, data: 'running.Return', result: 'result.Return', is_end=False):
+        attrs = self._body_item_v2_attributes(data, result, is_end=is_end)
+        attrs['values'] = list(result.values)
+        return result._log_name, attrs
```
pekkaklarck, line 469: At least this method would be unnecessary if `_body_item_v2_attributes` accepted `**extra` (name could be something else as well) and then could be called like `values=result.values`. Some variants like `_try_v2_attributes` are conditional based on the actual type, but perhaps that conditional logic could be in the caller.

src/robot/output/listeners.py
```diff
@@ -15,298 +15,591 @@
...
-        self.listener.output_file(type_, path)
+    def log_message(self, message: 'model.Message'):
+        self._log_message(self._message_attributes(message))
 
-    def close(self):
-        self.listener.close()
+    def message(self, message: 'model.Message'):
+        self._message(self._message_attributes(message))
+
+    def _suite_v2_attributes(self, data, result, is_end=False):
```
pekkaklarck, line 383: `_v2` in the name is unnecessary in this context. Same thing with other similar methods.

src/robot/output/listeners.py
```diff
@@ -15,298 +15,591 @@
...
+    def _body_item_v2_attributes(self, data, result, is_end=False, is_keyword_like=False):
+        attrs = {
+            'doc': result.doc,
+            'lineno': data.lineno,
+            'type': result.type,
+            'status': result.status,
+            'starttime': result.starttime,
+            'source': str(data.source or '')
+        }
+        if is_keyword_like:
```
pekkaklarck, line 480: It would be more straightforward if keywords had a dedicated method for getting attributes. Related to that, control structures don't anymore have `doc` so with them it could be just `''`.

pekkaklarck (approved):
Awesome! So much annoying indirection and other bad code removed. I added few notes about minor things that could still be enhanced.

PR #4928, opened 2023-11-03
Correct a typo in CreatingTestSuites.rst

JeffLi01:
Correct a typo 'a test teardown' where 'a suite teardown' is intended.


pekkaklarck:
Good catch! Thanks for a PR.

PR #4932, opened 2023-11-07
Update v4.1.3 link in README.rst

Dor-bl:
I Believe it will be easier for users that want to use v4.1.3 is to redirect them to the release tag rather than the branch readme.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -53,7 +53,7 @@ Robot Framework requires Python 3.6 or newer and runs also on `PyPy <http://pypy
 If you need to use Python 2, `Jython <http://jython.org>`_ or
 `IronPython <http://ironpython.net>`_, you can use `Robot Framework 4.1.3`__.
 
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Could you clarify why https://github.com/robotframework/robotframework/releases/tag/v4.1.3 would be better than https://github.com/robotframework/robotframework/tree/v4.1.3#readme?

pekkaklarck:
You aren't supposed to download that version, it's just a package containing the whole repository content created by GitHub. The recommended approach to install v4.1.3 is `pip install robotframework==4.1.3`. If you want to download an installer, you should get the version we've [uploaded to PyPI](https://pypi.org/project/robotframework/4.1.3/#files).

The above information is missing from the [v4.1.3 branch](https://github.com/robotframework/robotframework/tree/v4.1.3) because it was left to the state it was when v5.0 development started.. Because it's a branch, the README there could be enhanced, though.

PR #4962, opened 2023-11-29
remove deprecated type on style tag

kkotenko:
per https://developer.mozilla.org/en-US/docs/Web/API/HTMLStyleElement/type and https://html.spec.whatwg.org/multipage/obsolete.html#attr-style-type

### DIFF
--- a/src/robot/htmldata/htmlfilewriter.py
+++ b/src/robot/htmldata/htmlfilewriter.py
@@ -113,4 +113,4 @@ class CssFileWriter(InliningWriter):
 
     def write(self, line: str):
         href, media = re.search('href="([^"]+)" media="([^"]+)"', line).groups()
-        self.inline_file(href, 'style', {'type': 'text/css', 'media': media})
+        self.inline_file(href, 'style', {'media': media})


pekkaklarck:
Does `type="text/css"` cause deprecation warnings or other issues with some browsers already? More importantly, are you certain that all modern browsers work without it?

kkotenko:
Testing with the following MWE:

```html
<!DOCTYPE html>
<html lang="en">
<head>
<title>Hello World!</title>
<style type="text/css">
span{color: red;}
</style>
</head>
<body>
<span>Hello World!</span>
</body>
</html>
```

It causes a deprecation warning in the built-in HTML validator in PyCharm/IntelliJ (which is how I found it). It also causes a warning in the W3C HTML vaildator:

> Warning: The type attribute for the style element is not needed and should be omitted.

Neither omitting nor leaving the attribute in seems to any issues in Firefox 120 (which is the most recent at time of writing), Firefox 115 (ESR), Chrome 119, Opera 105 (which seems to be using the Chrome developer tools), or Edge 119. (Or Safari 5.1.7, which came out 2011, but is the most recent version available for Windows.) 

I do not have easy access to other browsers right now. Are there any particular other ones you'd want me to look at to be sure? (See https://caniuse.com/mdn-html_elements_style_type for reference on current browsers.)

kkotenko:
Additional discussion here https://stackoverflow.com/questions/24411280/is-it-bad-practice-to-not-include-type-text-css-in-style-tags said back in 2014 that you don't need it in HTML 5. I think it's safe to say that if it was the prevailing opinion back then, it should be now.

PR #4990, opened 2023-12-21
implemented dark_light theme toggle button

Snooz82:
@pekkaklarck This implements a theme toggle butten at top left of log and report.

By default the system setting is used.
If toggle is clicked, it is stored in localStorage.

Additionally it is possible to add `?theme=dark` or `?theme=light` to the url and you directly get the requested theme.

maybe this can still be tried out in RC1 ?


pekkaklarck:
We have a custom [storage](https://github.com/robotframework/robotframework/blob/master/src/robot/htmldata/common/storage.js) functionality for accessing `localStorage` available for HTML outputs and [report](https://github.com/robotframework/robotframework/blob/master/src/robot/htmldata/rebot/report.html) already uses it (see `storage.init()`, `storage.get()` and `storage.set()`. The same solution should be used when accessing `localStorage` to get the theme for consistency reasons and also because the code handles code like `localStorage` not being available.

It would be nice, but not mandatory, if Libdoc would also get a toggle for selecting the theme. It then needed to play together with the `--theme` option Libdoc has. If this gets at all complicated, it's better to wait for RF 7.1.

I'll merge this now to get the functionality into the rc. Accessing `localStorage` should still be changed. Let's discuss on Slack who does it and should it be done before the rc.

PR #4991, opened 2023-12-22
String: Fix get substring example missing start index

cas--:
Without zero for the start index the end index would be parsed as the start index argument.


pekkaklarck:
Thanks for contribution. The old example was correct when data was in HTML tables and ought to still work with the pipe-separated format. In these cases the start index ends up being an empty string, and the keyword has special code to map that to `0`.  The example didn't make sense when using the normal plain text format, though, and including `0` makes sense also otherwise.

PR #4994, opened 2023-12-27
Miscellaneous type hint fixes

Serhiy1:
I was checking my contributions to the previous version of robotFramework were still functional in the RF 7.0 beta. I noticed some small issues.  

- Fixed a couple of instances of overzealous find-and-replace
- Fixed some missing imports
- Added some missing type arguments to Generic classes

### DIFF
--- a/src/robot/model/body.py
+++ b/src/robot/model/body.py
@@ -23,8 +23,7 @@ from .itemlist import ItemList
 from .modelobject import DataDict, full_name, ModelObject
 
 if TYPE_CHECKING:
-    from robot.result.model import ForIteration, WhileIteration
-    from robot.running.model import UserKeyword, ResourceFile
+    from robot.running.model import ResourceFile, UserKeyword
     from .control import (Break, Continue, Error, For, ForIteration, If, IfBranch,
                           Return, Try, TryBranch, Var, While, WhileIteration)
     from .keyword import Keyword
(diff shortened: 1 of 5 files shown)


src/robot/model/body.py
```diff
@@ -33,7 +32,7 @@
     from .testsuite import TestSuite
 
 
-BodyItemParent = Union['TestSuite', 'TestCase', 'UserKeyword', 'For', 'ForIteration',
+BodyItemParent = Union['TestSuite', 'TestCase', 'Keyword', 'For', 'ForIteration',
```
pekkaklarck, line 35: This change isn't totally correct. `Keyword` can be a parent only in the result model. `UserKeyword` only exists in the running model so having it here in the base may not be a good idea, though.

src/robot/running/model.py
```diff
@@ -57,7 +57,7 @@
 
 
 IT = TypeVar('IT', bound='IfBranch|TryBranch')
-BodyItemParent = Union['TestSuite', 'TestCase', 'UserKeyword', 'For', 'If', 'IfBranch',
+BodyItemParent = Union['TestSuite', 'TestCase', 'Keyword', 'For', 'If', 'IfBranch',
```
pekkaklarck, line 60: This change isn't correct. `Keyword` can be a parent only in the result model and here in the running model `UserKeyword` is a valid parent.

pekkaklarck:
Looks good now. Thanks for the fixes!

Issue #4929, opened 2023-11-05
Feature request: Ability to set log_level for Run Process keyword


richardstone:
I'd like to have the ability to set `log_level` for builtin `Run Process` keyword so that I can specify the loglevel of messages logged by this keyword.

pekkaklarck:
Sounds good to me. The option should also be added to Start Process, Wait For Process, etc. both for consistency and because Run Process uses them internally.

Some keywords currently log messages both with INFO and DEBUG. We needed to decide should we allow changing both or is hardcoded DEBUG ok.

richardstone:
In my opinion, hardcoded anything is not that great. But here, having some DEBUG messages that helps to understand what is going on in case of a failure is a good to have.

LowEQ:
if none is working on this can I give a try?

LowEQ:
#5204 here is the draft PR, need still those acceptance tests, and run existing tests to make sure works fine

oboehmer:
What is the use case to use different log level for messages logged by those keywords? And why only for these keywords specifically?

pekkaklarck:
That is actually a very good question @oboehmer. The logged messages aren't too big, so there shouldn't be problems with output file size. One use case that comes to mind is that arguments could contain confidential data, but now that we have Secret variables (#4537) and also Process library supports them (#5512), there's a better way to avoid such data being shown in the log files. 

After thinking this a bit more, I'm not sure is this a good in general. These are my concerns:
- As mentioned above, the actual use cases are not clear.
- The Secret variable supports means there's a built-in way to avoid sensitive data being shown in the log file.
- The `Set Log Level` keyword can be used to temporarily change the log level so that certain messages aren't included.
- If we add this support to Process, there could be requests to add it to other libraries/keywords as well. Configuring the log level can sometimes be useful, but I don't want to make all keywords logging something configurable. There need to be concrete benefits for that.

Due to the above reason I don't think it is a good idea to implement this in RF 7.4. Because there could be valid use cases, I don't close this issue as "not planned" at this point either. I'll move this to RF 7.5 scope and we can then make the final decision when the development of that version starts. I hope @richardstone and others who has requested this functionality could explain their use cases before that.

pekkaklarck:
No updates about use cases nor comments to concerns I raised in my previous comment. I'll close this as not planned.

PR #4996, opened 2023-12-31
#4929 Added global 'log' command for the 'Process' class

geco333:
My first contribution to the project please let me know if I am not following project conventions and/or guidelines.


kkotenko:
@geco333 It looks like the file you are adding is empty? I assume you forgot to add it before committing.

PR #5033, opened 2024-01-23
Creating a library to check various types of a variable

ferrarigian:
## Library Name: CheckType

The **CheckType** library is designed to provide a set of validation functions for checking the data types of Python objects. This library is particularly useful within the Robot Framework, offering a straightforward way to assert and validate the types of variables.

### Key Features:

1. **Type Checking Functions:**
   - The library includes a collection of functions to check if a given item is of a specific data type. Supported types include dictionaries, objects, integers, lists, booleans, floats, NoneType, tuples, and sets.

2. **Customizable Error Messages:**
   - Each type-checking function allows the user to include a custom error message. This feature enhances flexibility, enabling users to provide meaningful context in case of a type mismatch.

3. **Negation Functions:**
   - In addition to type-checking functions, the library provides corresponding functions to assert that an item is not of a certain type. This enables users to perform comprehensive type validation in their Robot Framework test cases.

### Example Usage:

```robot
*** Settings ***
Library    CheckType

*** Test Cases ***
Validate Dictionary Type
    ${data}=    Create Dictionary    key=value
    CheckType.Should Be Dictionary    ${data}    msg=Expected a dictionary, but got ${data}

Ensure Not a List
    ${value}=   Set Variable    SomeValue
    CheckType.Should Not Be List    ${value}    msg=${value} should not be a list

### DIFF
--- a/src/robot/libraries/CheckType.py
+++ b/src/robot/libraries/CheckType.py
@@ -1,0 +1,347 @@
+
+from robot.utils import type_name
+
+class CheckType:
+    def should_be_dictionary(self, item, msg=None):
+        """Fails if the given `item` is not a dictionary.
+
+        Args:
+            item: The item to be checked.
+
+        Raises:
+            AssertionError: If the `item` is not a dictionary.
+
+        Returns:
+            bool: True if the `item` is a dictionary.
+        
+        The default error message can be overridden with the optional ``msg`` argument.
+        """
+        if isinstance(item, dict):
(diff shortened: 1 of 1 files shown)


franzhaas:
Hi ferrarigian,

Have you considered putting this library with examples and tests onto the [marketquare](https://github.com/marketsquare)? and PyPI?

Do you think this could be a good alternative to merging it into core?

PR #5034, opened 2024-01-24
Use 'sphinx_rtd_theme' as theme

manykarim:
Since end of 2023, the default theme at ReadTheDocs has been changed to Alabaster.
This change will make sure the classic ReadTheDocs theme is used again.

If you want to check the document creation locally, you can do that my running
`TZ=UTC make html`
(`TZ=UTC` was needed on my system due to a timezone error)

When building locally, make sure to install the packages `sphinx_rtd_theme`.
via `pip install sphinx_rtd_theme`.
When building via ReadTheDocs, this is not needed as their standard image for building includes that package

### DIFF
--- a/doc/api/conf.py
+++ b/doc/api/conf.py
@@ -27,7 +27,7 @@ needs_sphinx = '1.3'
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['sphinx.ext.autodoc', 'sphinx.ext.viewcode']
+extensions = ['sphinx.ext.autodoc', 'sphinx.ext.viewcode', 'sphinx_rtd_theme']
 autoclass_content = 'both'
 autodoc_default_flags = ['members', 'undoc-members', 'inherited-members']
 autodoc_member_order = 'bysource'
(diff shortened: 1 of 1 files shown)


pekkaklarck:
I tried building API docs at Read The Docs but it failed with this error:
> Extension error:
> Could not import extension sphinx_rtd_theme (exception: No module named 'sphinx_rtd_theme')
The full traceback and some more additional info can be see [here](https://readthedocs.org/projects/robot-framework/builds/23257790/).

Could it be that this theme isn't available automatically?

manykarim:
Wondering if that's the root cause. They only mention that new projects
created on readthedocs since August 2023 won't have that theme installed by
default.

https://blog.readthedocs.com/python-core-requirements-changed/

Starting on August 7, 2023 all new projects imported on Read the Docs will
install only sphinx, mkdocs and readthedocs-sphinx-ext as “core
requirements”. The default behavior will be to install the latest version
of these requirements.

Note that previously Read the Docs was installing also jinja,
sphinx-rtd-theme, pillow, mock, alabaster, commonmark and recommonmark
specifying particular versions depending on different factors that were
confusing for users and hard to debug.


A possible solution is posted here:
https://stackoverflow.com/questions/76924291/error-on-sphinxs-generated-documentation-on-readthedocs


You need to install the Sphinx theme dependency on Read the Docs. To do so,
you have to create a requirements.txt file with the following content:

sphinx-rtd-theme==1.2.2

Then, modify your .readthedocs.yaml file to tell Read the Docs to install
these requirements:

python:
  install:
    - requirements: "requirements.txt

Pekka Klärck ***@***.***> schrieb am So., 28. Jan. 2024,
11:20:

PR #5047, opened 2024-02-04
fix argument name in docstring

touchedthecode:
changes 'included_tags' to 'include_tags' in TestSuite.configure() docstring


pekkaklarck:
Good catch!

PR #5064, opened 2024-02-23
Localization for Japanese language

5003:


pekkaklarck:
Thanks for contribution! @leeuwe, should these translations be added also to Crowdin?

leeuwe:
Ideally, yes. I'll see if I can add them back in there

Issue #4584, opened 2023-01-03
New `robot:flatten` tag for "flattening" keyword structures


pekkaklarck:
Introduction
------------

With nested keyword structures, especially with recursive keyword calls and with WHILE and FOR loops, the log file can get hard do understand with many different nesting levels. Such nested structures also increase output.xml size, because even a simple keyword like
```robotframework
*** Keywords ***
Keyword
    Log    Robot
    Log    Framework
```
creates this much content:
```xml
<kw name="Keyword">
<kw name="Log" library="BuiltIn">
<arg>Robot</arg>
<doc>Logs the given message with the given level.</doc>
<msg timestamp="20230103 20:06:36.663" level="INFO">Robot</msg>
<status status="PASS" starttime="20230103 20:06:36.663" endtime="20230103 20:06:36.663"/>
</kw>
<kw name="Log" library="BuiltIn">
<arg>Framework</arg>
<doc>Logs the given message with the given level.</doc>
<msg timestamp="20230103 20:06:36.663" level="INFO">Framework</msg>
<status status="PASS" starttime="20230103 20:06:36.663" endtime="20230103 20:06:36.664"/>
</kw>
<status status="PASS" starttime="20230103 20:06:36.663" endtime="20230103 20:06:36.664"/>
</kw>
```

We have had `--flattenkeywords` option for "flattening" such structures since RF 2.8.2 (#1551) and it works great. When a keyword is flattened, its child keywords and control structures are removed otherwise, but all their messages are preserved. It doesn't affect output.xml generated during execution, but flattening happens when output.xml files are parsed and can save huge amounts of memory. When `--flattenkeywords` is used with Rebot, it is possible to create a new flattened output.xml. For example, the above structure is converted into this if `Keyowrd` is flattened:
```
<kw name="Keyword">
<doc>_*Content flattened.*_</doc>
<msg timestamp="20230103 20:06:36.663" level="INFO">Robot</msg>
<msg timestamp="20230103 20:06:36.663" level="INFO">Framework</msg>
<status status="PASS" starttime="20230103 20:06:36.663" endtime="20230103 20:06:36.664"/>
</kw>
```

Proposal
--------

Flattening works based on keyword names and based on tags, but it needs to be activated separately from the command line. This issue proposes adding new built-in tag `robot:flatten` that activates this behavior automatically. Removing top level keywords from tests and leaving only their messages doesn't make sense, so `robot:flatten` should be usable only as a keyword tag.

[...]

PR #5084, opened 2024-03-18
new log level "user" for very short log files

test-fullautomation:
Dear community,

we are a team at Bosch who is since around 20 years in charge to provide test tool and infrastructure to our test departments.

We are happy to use Robot Framework.
One of the requirements we have is that tests need to run for weeks (e.g. hardware tests in climate chamber). The current Robot Framework core produces much too much log output for this purpose.

The new log level "USER" is transparent for Robot Framework users. Only if it is explicitly used and the trace level is set accordingly for the test execution, then the log output is reduced to only "USER" log entries. In "normal" log levels the new log level appears similar to already existing log outputs as per trace level.

Detailed documentation you find here:
robotframework/atest/robot/log_level/readme_log_level_test.txt

Origin of the change for this pull-reqeust ist here:
https://github.com/test-fullautomation/robotframework/pull/66

Changes:
new log level 'USER' introduced
amount of output reduced in debug log file and XML report file
self test in robotframework/atest/robot/log_level
documentation in robotframework/atest/robot/log_level/readme_log_level_test.txt
Based on robotframework-7.0.1.dev1-py3.9

"log level" self test extensions

IF condition and FOR loop added to log_level robot file and resource file
added log file comparison for debug log files
pattern files for log file comparison modified
documentation adapted


pekkaklarck:
Sorry for a delayed review due to my paternity leave. Some general comments first:

- When proposing big changes like this, it is generally better to start from an issue where you explain the enhancement, the design can be discussed, and so on.
- Starting discussion on the #devel channel on our Slack could be a good idea as well.
- The enhancement should be explained, at least in high level, directly in the issue (or PR), not in some file that's included in a PR.

Then about the enhancement itself:
- I don't like the idea to add a new log level named `USER`. It doesn't match the existing levels at all.
- I don't actually see any reason to add any new "hard coded" log level. The ones we have are adequate for most common usages, and adding one more for some usage could still leave gaps and we needed to add more.
- Making it possible for users to add custom log levels could be considered, though. Python's `logging` module supports something like that. 
- If I understood it correctly, the new `USER` log level would have a special functionality that it would surpress writong information about started keywords amd control structures as well. I understand the need for it, it reduces output.xml and log.html size considerably, but I don't think it is a good idea to combine that functionality with log levels. 
- Surpressing writing information about started keywords and control structures is already supported by Robot and it's called "flattening". It can be enabled from the command line with the `--flattenkeywords` option and with newer Robot versions also by using the special `robot:flatten` keyword tag. Is rhere a reason this functionality doesn't work for you? Could it be enhanced somehow?

test-fullautomation:
Dear @pekkaklarck ,
thank you for your detailed answer. No problem about the delay. I hope you had a nice time with your family.

Thank you also for your proposals at the beginning of your answer. We will follow this in future.

I think we have a misunderstanding regards the function we implement, we tried --flattenkeywords, 
but this unfortunately doesn't fulfill our needs at all.

Our need is to create a possiblity to **reduce the log file size** dramatically.
This is useful for people like us who run the tests for weeks, or even months. 
The file size is reduced dramatically, nevertheless we have still proper test results 
and the `USER` defined log entries (+WARN +ERROR).

Mainly only data will be logged which looks like this:
 `Log    ${log_message}    USER`
exectued with
 `mytests.robot --loglevel USER`
in addition `WARN` and `ERROR` messages will be logged into the files.

Finally we deactivate nearly all what Robotframework core would log, but still we allow
to put own log entries into the logfiles - an absolute minimum user controlled output.

If you execute our attached atest you can see how dramatically the size of the log files will be reduced.
The output of the selftest creates a file per log level in order to be able to compare the file size.

I apologize. Accidentally I didn't link to the rendered rst documentation.
Here is the full documentation:
https://github.com/test-fullautomation/robotframework/blob/HolQue/task/log_level_user_v7.0.1/atest/robot/log_level/readme_log_level_test.rst

Thank you,
 Thomas

pekkaklarck:
Sorry for another delay with the review. I've now taken a more detailed look at the code and also tested it locally. Although I think I understand your needs for this functionality, I don't think the proposed implementation would fit with rest of the Robot Framework features and I thus reject this pull request. I'll explain my reasoning in more detail below and also suggest alternative features that ought to provide same benefits for you.

## Problems with functionality

### Problems with new log level

As I already commented in my initial review, I don't think we should add any new hard-coded log level to Robot. The current levels have been adequate so far and, more importantly, adding a new level between INFO and WARN wouldn't help someone who'd like to have a new level, for example, between INFO and DEBUG. In addition to that, I consider name "USER" pretty strange as it could easily be used also by libraries. If we were to add a new level, it needed to use standard terminology similarly as the current levels.

Although I don't like to add any new hard-coded level, I would be fine with users adding custom levels. We could, for example, make it possible to specify a new level along with a priority relative to the existing levels. For example, something like `--custom-log-level USER:>INFO` could be used to specify a new level USER that has a priority higher than the existing INFO. You could then activate the new level normally with `--log-level USER`. If you think this would be a useful addition, please submit *an issue* about that and we can the discuss about the design in more detail before implementation.

I acknowledge it would be annoying to always use an option like `--custom-log-level` if you always want to have a custom level available. Current workarounds for that include using the `ROBOT_OPTS` environment variable or argument files, but a better solution would be adding support for configuration files where options like this could be defined. That would be a separate enhancement benefiting also various other usages. We've talked about that few times recently, but there isn't an issue that yet.

Another way to avoid unnecessary INFO level messages in the log file is removing them. You could, for example, have a convention that all messages that don't start with a word "USER" are automatically removed after execution using, for example, `--pre-rebot-modifier`. It would be more convenient to do that already during execution with a listener, but that's not possible at the moment. We were planning to support that so that if a listener sets message to an empty string, it is not logged at all (#5089). That idea was scrapped because there were valid usages for empty messages, but perhaps we reconsider this with a small adjustment that messages that are set to `None` are dropped. If you think that would work for you, please submit an issue.

### Problems with log level affecting "flattening"

The proposed USER level is *very* different to other levels as using `--log-level USER` also disables writing information about started keywords and control structures. That is problematic for various reasons, but the biggest one is that the same functionality is already supported by the Robot's "flattening" concept. I don't think it is a good idea to have two identical features with different implementations.

You wrote that "we tried --flattenkeywords, but this unfortunately doesn't fulfill our needs at all", but didn't clarify why the feature didn't work for you. Based on looking at your code and testing it locally, using `--log-level USER` and `--flatten-keywords name:*` seems to produce the same result from flattening perspective. The former obviously also drops all INFO level log messages, but that's not related to flattening.

Notice that flattening can also be enabled with the special `robot:flatten` tag (#4584). It's currently only supported as a keyword tag, but it wouldn't require much more than the following to allow using it also with tests:

```diff
--- a/src/robot/output/xmllogger.py
+++ b/src/robot/output/xmllogger.py
@@ -353,8 +353,13 @@ class XmlLogger(ResultVisitor):
     def start_test(self, test):
         self._writer.start('test', {'id': test.id, 'name': test.name,
                                     'line': str(test.lineno or '')})
+        if test.tags.robot('flatten'):
+            self.flatten_level = 1
+            self._writer = NullMarkupWriter()
 
     def end_test(self, test):
+        self.flatten_level = 0
+        self._writer = self._xml_writer
         self._writer.element('doc', test.doc)
```

If you have some concrete problems with flattening that your USER level implementation fixed, please clarify them. You can comment this PR or submit new issues about possible enhancements to the current flattening functionality.

### Problems with other features

Another problem related to features is that there were other changes that were not mentioned anywhere:

- The PR seems to make lot of changes to the debug-file functionality. I didn't look too closely what's changed, but whatever the changes are, they should have been proposed as a separate PR.

- Apparently the Log keyword got its own log level. I don't understand why, but looking at the code it most likely won't even work if the keyword is used like `level=DEBUG`.

When proposing a PR, you should limit it to one feature. More importantly, new features need to be explained.

## Other problems with the PR

There were various problems with the added code and tests as well. They should have fixed even if we had accepted these enhancements. That's not relevant now, but I list these problems here in case you want to propose new PRs in the future.

### Problems with code

- There's a lot of duplication with the added code.
- The code re-implements flattening functionality that already exists. It would have been a lot easier to just set `self.flatten_level` and `self._writer` similarly as in the above example demonstrating how to enable `robot:flatten` on test level.

### Problems with tests

- Old tests do not pass anymore. I run all acceptance tests locally and 28 failed. When proposing a PR, you **must** make sure you don't break anything. It could be that some tests just needed to be adjusted, but doing that is your work. If I'd accept this PR, I needed to do that.
- New tests don't use our existing acceptance tests system. You should read `atest/README.rst` to see how it works. It's a bit complicated, but we are happy to answer questions and help otherwise if needed.
- New tests contain a huge amount of data. Maintaining it would be _really_ big task for us in the future. I don't want to go into details, but in general comparing files directly isn't a good idea. Our aforementioned acceptance test system has much better functionality for that.
- Code related to new tests doesn't follow PEP-8 or project's coding conventions.
- Some new tests fail.

### Problems with documentation

- This isn't documented at all in the User Guide or in `--help` output.

test-fullautomation:
Dear @pekkaklarck ,

thank you for your spent efforts. We analyse your extensive response and if we still feel it's required we come back with a discussion in slack devel channel as you proposed.

Thank you again,
 Thomas

pekkaklarck:
Thanks! I understand the problems you have and believe they can be fixed in smaller pieces than originally proposed.

pekkaklarck:
I wrote earlier that we could enhance listeners so that they can remove messages by setting them to `None`. I submitted #5251 about that and it's now implemented.

I plan to also submit issues about configuration files and plugin API, that would both help you to extend Robot without changing its code. Implementing them is a bigger task and they cannot be done in RF 7.2, but they sound like good candidates for RF 7.3.

If there are enhancements to flattening that would benefit you, please submit issues or start a discussion on Slack.

test-fullautomation:
Thank you for the information. We will need some time to analyze your previous response. We will also consider your last response. Then we come back to you.

Issue #4577, opened 2022-12-26
Strip prefix from argument value if BDD keyword starts with embedded argument


allanlei:
When using `Given/When/Then/etc` to match keywords that start with a variable `${...}`, then `Give/When/Then` is captured into the variable value.

According to [Ignoring Given/When/Then/And/But prefixes](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#behavior-driven-style), 

> Prefixes Given, When, Then, And and But are dropped when matching keywords are searched, if no match with the full name is found

Is there a builtin method to reverse this logic such that it checks for the prefix-dropped version first?

```
*** Keywords ***
${url} is open
    Log To Console           Open "${url}"

${ele} is clicked
    Log To Console           Click "${ele}"

${ele} should be visible
    Log To Console           Is Visible "${ele}"


*** Variables ***
${url}          https://google.com
${btn_1}        xpath=//a/b/c
${modal_2}      xpath=//a/b/c


*** Test Cases ***
Valid Login
    Given ${url} is open
    When ${btn_1} is clicked
    Then ${modal_2} should be visible
```

Outputs
```
Open "Given https://google.com"
Click "When xpath=//a/b/c"
Is Visible "Then xpath=//a/b/c"
```

Wanted output
```
Open "https://google.com"
Click "xpath=//a/b/c"
Is Visible "xpath=//a/b/c"
```

---

I can get this behaviour by changing https://github.com/robotframework/robotframework/blob/master/src/robot/running/namespace.py#L292-L298 to 

```python
        runner = self._get_runner_from_suite_file(name)
        if not runner and '.' in name:
            runner = self._get_explicit_runner(name)
        if not runner:
            runner = self._get_bdd_style_runner(name)
        if not runner:
            runner = self._get_implicit_runner(name)
```

pekkaklarck:
There's PR #5096 implementing this functionality so let's include this issue into RF 7.1 scope. I won't have time to review the PR right now, but I return to it after RF 7.0.1 is out.

pekkaklarck:
There are some backwards incompatibility concerns due to behavior being changed. In my opinion risks are small enough so that this doesn't need to wait until a major version, but the change needs to be nevertheless be mentioned in release notes.

pekkaklarck:
Implemented in PR #5096 by @JFoederer. Thanks for the contribution!

PR #5096, opened 2024-03-27
Allow keywords starting with a variable in bdd style

JFoederer:
This is a fix for issue #4577.

From Specification by Example's best practices action steps are best written in third person form where the actor is named that performs the action. Oftentimes you want the actor to be a variable, because the actions of the different actors have different consequences.

For example:

> Given Alvin Admin is logged in on the portal
> and Betty Buyer is logged in on the portal
> then Alvin Admin sees Betty Buyer in the online users list
> but Betty Buyer cannot see Alvin Admin in the online users list

Before this fix, implementing a keyword like `${user} is logged in on the portal` would, in the given step, yield the faulty effect that `${user}` gets the value "_Given Alvin Admin_".

### DIFF
--- a/atest/testdata/keywords/optional_given_when_then.robot
+++ b/atest/testdata/keywords/optional_given_when_then.robot
@@ -48,6 +48,9 @@ Keyword can be used with and without prefix
     Then we are in Berlin city
     we are in Berlin city
 
+First word of a keyword can be a prefix
+    Given the prefix is part of the keyword
+
 First word in a keyword can be an argument
     Given we don't drink too many beers
     Then Pekka drinks lonkero instead
@@ -111,3 +114,6 @@ ${person} drinks water instead
 
 Multipart prefixes didn't work with RF 6.0
     No Operation
+
+Given the prefix is part of the keyword
+   No operation
--- a/atest/testdata/keywords/optional_given_when_then.robot
+++ b/atest/testdata/keywords/optional_given_when_then.robot
@@ -48,6 +48,11 @@ Keyword can be used with and without prefix
     Then we are in Berlin city
     we are in Berlin city
 
+Only single prefixes are a processed
+    [Documentation]  FAIL No keyword with name 'but then we are in Berlin city' found.
+    Given we are in Berlin city
+    but then we are in Berlin city
+
 First word of a keyword can be a prefix
     Given the prefix is part of the keyword
 
@@ -55,6 +60,7 @@ First word in a keyword can be an argument
     Given we don't drink too many beers
     Then Pekka drinks lonkero instead
     and Miikka drinks water instead
+    Étant donné Miikka drinks water instead
 
 Localized prefixes
     Oletetaan we don't drink too many beers
--- a/atest/testdata/keywords/optional_given_when_then.robot
+++ b/atest/testdata/keywords/optional_given_when_then.robot
@@ -48,6 +48,11 @@ Keyword can be used with and without prefix
     Then we are in Berlin city
     we are in Berlin city
 
+First word in a keyword can be an argument
+    Given we don't drink too many beers
+    Then Pekka drinks lonkero instead
+    and Miikka drinks water instead
+
 Localized prefixes
     Oletetaan we don't drink too many beers
     Kun we are in  museum  cafe
@@ -97,6 +102,10 @@ We ${x} This ${thing} Implemented
 
 We Go To ${somewhere}
     Should Be Equal  ${somewhere}  walking tour
+
+${person} drinks ${beverage} instead
+    Should not start with  ${person}  Then
+    Should not start with  ${person}  and
 
 Multipart prefixes didn't work with RF 6.0
     No Operation
--- a/atest/testdata/keywords/optional_given_when_then.robot
+++ b/atest/testdata/keywords/optional_given_when_then.robot
@@ -71,7 +71,7 @@ Prefix consisting of multiple words
     Fie ca multipart prefixes didn't work with RF 6.0
 
 Prefix must be followed by space
-    [Documentation]  FAIL
+    [Documentation]    FAIL
     ...    No keyword with name 'Givenwe don't drink too many beers' found. Did you mean:
     ...    ${SPACE*4}We don't drink too many beers
     Givenwe don't drink too many beers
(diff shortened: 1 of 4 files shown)


JFoederer:
To address any worries on compatibility. The only time you would notice a change, is when you have a keyword of which the first word is an argument and the data entered for that argument starts with one of the reserved bdd prefixes (_given, when, then_, etc.), followed by a space and another word. This is a rather unique situation.

For example:
When using the keyword “_${beverage} should be served at ${temp}°C_” in a non-bdd style context as “_But light should be served at 4°C_”, the word **But** would be matched as bdd prefix, but it will only be accepted as valid prefix, because “_light should be served at 4°C_” also still matches the keyword. You need the double worded ‘But light’ argument to still match the leading variable. This is an opposite example from the one mentioned in the PR description.

The available workarounds for when you do run into this are:
•	Making the keyword more specific: _The beverage But light should be served at 4°C_
•	Making the data explicit by adding quotes: _‘${beverage}’ should be served at ${temp}°C_

These are the same workarounds that one had to revert to in the old situation. Comparing them, the chance of the new situation occurring is way smaller (more unique) than the chance of the previous situation occurring, with identical penalty cost.

The new implementation still matches the documentation text. No updates needed there in my opinion.

atest/testdata/keywords/optional_given_when_then.robot
```diff
@@ -98,5 +103,9 @@ We ${x} This ${thing} Implemented
 We Go To ${somewhere}
     Should Be Equal  ${somewhere}  walking tour
 
+${person} drinks ${beverage} instead
+    Should not start with  ${person}  Then
+    Should not start with  ${person}  and
```
pekkaklarck, line 108: Validating the string with `Should Be Equal` would make the test more explicit and also a bit safer. The problem is then knowing who the `${person}` is, but perhaps there could just be two keywords `${person} drinks lonkero instead` and `${person} drinks water instead`. Not sure is that worth the effort, though.

Please use four spaces between arguments.

src/robot/running/namespace.py
```diff
@@ -297,21 +298,23 @@ def _get_runner(self, name):
             runner = self._get_explicit_runner(name)
         if not runner:
             runner = self._get_implicit_runner(name)
-        if not runner:
-            runner = self._get_bdd_style_runner(name, self.languages.bdd_prefixes)
+        # handle bdd prefixes, even when swallowed by a leading variable
+        if not runner or not search_variable(runner.keyword.name).before:
```
pekkaklarck, line 302: First getting a runner and then trying to get it again if the keyword name has a variable with something before it is a bit complicated way to implement this. It would be a lot simpler to just first try to find a BDD-style runner. I tested that this seems to work:
```diff
--- a/src/robot/running/namespace.py
+++ b/src/robot/running/namespace.py
@@ -286,13 +286,13 @@ class KeywordStore:
             raise DataError('Keyword name cannot be empty.')
         if not is_string(name):
             raise DataError('Keyword name must be a string.')
-        runner = self._get_runner_from_suite_file(name)
+        runner = self._get_bdd_style_runner(name, self.languages.bdd_prefixes)
+        if not runner:
+            runner = self._get_runner_from_suite_file(name)
         if not runner and '.' in name:
             runner = self._get_explicit_runner(name)
         if not runner:
             runner = self._get_implicit_runner(name)
-        if not runner:
-            runner = self._get_bdd_style_runner(name, self.languages.bdd_prefixes)
         return runner
```
The main benefit would be that the algorithm stays consistent. We always try to find a `runner` and if it fails, we try again using a different approach.

The only drawback I see is that the current approach to go through all possible BDD-prefixes in `_get_bdd_style_runner` can be a bit slow. Splitting the used keyword name from spaces and then trying all possible combinations one-by-one is a bit stupid when actual prefixes we have contain maximum two parts. I doubt that has a real performance effect, but I have some ideas how to easily enhance that.

pekkaklarck (changes requested):
Looks good but I think the implementation could be simplified. More importantly, this change needs to be mentioned in the User Guide.

atest/testdata/keywords/optional_given_when_then.robot
JFoederer: Spacing updated to 4 spaces, also for existing code.

I did indeed go for the solution with two keywords. There are other options but they make it more complicated.

src/robot/running/namespace.py
JFoederer: Good suggestion. So far I haven't found a downside to the simpler solution, so I took it exactly as suggested.

pekkaklarck (changes requested):
Very good! Two things to do still:
- The second `_get_bdd_style_runner` call is, unless I'm missing something, unnecessary, and should be removed.
- The explanation in the User Guide could still be enhanced a bit. Especially the subtle change to handling the prefix should be explained. That includes mentioning the version where it was changed.

pekkaklarck:
I wonder why this PR isn't linked to the issue properly. Perhaps if I say, fixes #4577, that will change.

src/robot/running/namespace.py
```diff
@@ -287,28 +287,26 @@ def _raise_no_keyword_found(self, name, recommend=True):
...
         return runner
 
-    def _get_bdd_style_runner(self, name, prefixes):
+    def _get_bdd_style_runner(self, name):
         parts = name.split()
         for index in range(1, len(parts)):
             prefix = ' '.join(parts[:index]).title()
-            if prefix in prefixes:
-                runner = self._get_runner(' '.join(parts[index:]))
+            if prefix in self.languages.bdd_prefixes:
```
pekkaklarck, line 308: Accessing `self.languages.bdd_prefixes` is a little slower than just accessing `prefixes`, but it's impossible to know without profiling does that matter in any real usage. This code is now used with all keyword calls, not only as the last resort as earlier, so performance is important. I have some ideas for performance improvements that I'll test later. Functionality-wise this ought to be ok so I'll merge this PR first.

pekkaklarck (approved):
Looks good to me. I'm slightly worried about the performance implications of always first looking for BDD style keywords, but I'll measure performance after merging this and try some ideas I have for making the BDD-prefix check a bit faster.

src/robot/running/namespace.py
JFoederer: Yes, and there was another thing still on my mind after changing the code after the first review, although I didn't realise this until writing the user documentation. It now states that a keyword that does not include a bdd prefix gets precedence over one that is written in full, including the prefix in the name. I do not expect this to be a practical issue in any way. If you start mixing styles you will be in trouble at one point anyway. For instance, when using the _and_ prefix. But, following the spirit of how we typically treat precedence in Robot, I would expect that the more specific keyword would get precedence over the less specific one.
However, the exact reason for this issue, considering variables as a first word, makes it impossible to find a solution without detecting the variable and/or the prefixes, which was discussed during the first review. So, without seeing any practical benefit, I dismissed the idea. Now, with your latest comment, it seems worth while to reconsider creating a prefix detection function, that can then also be optimised.

PR #5097, opened 2024-03-29
Fixes #5054: Enhance Should Contains for bytes

### LINKED ISSUE #5054: Enhance Should Contains for bytes
To use `Should Contains` with bytes, item that is searched needs to be first converted to bytes.
```robot
Test
    ${binaryVar} =    Convert To Bytes    \x00\x01\x02\xA0\xB0
    Should Contain    ${binary1Var}    \xA0
```
result:
```
TypeError: a bytes-like object is required, not 'str'
```

for int the value is converted automatically:
```robot
Test Should Contain int
    ${binary1Var} =    Convert To Bytes    \x00\x01\x02\xA0\xB0
    ${item} =    Convert To Bytes    \xC0

    Should Contain    ${binary1Var}    ${1000}
```
this fails with:
```
ValueError: byte must be in range(0, 256)
```
this is expected.

droeland:
Implemented as described in the issue.
Fixes #5054.

### DIFF
--- a/src/robot/libraries/BuiltIn.py
+++ b/src/robot/libraries/BuiltIn.py
@@ -1042,7 +1042,7 @@ class _Verify(_BuiltInBase):
                        ignore_case=False, strip_spaces=False, collapse_spaces=False):
         """Fails if ``container`` does not contain ``item`` one or more times.
 
-        Works with strings, lists, and anything that supports Python's ``in``
+        Works with strings, lists, bytes, and anything that supports Python's ``in``
         operator.
 
         See `Should Be Equal` for an explanation on how to override the default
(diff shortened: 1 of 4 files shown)


src/robot/libraries/BuiltIn.py
```diff
@@ -1072,6 +1072,11 @@ def should_contain(self, container, item, msg=None, values=True,
         in Robot Framework 4.1.
         """
         orig_container = container
+        if isinstance(container, bytes) and isinstance(item, str):
+            try:
+                item = item.encode('latin-1')
+            except UnicodeEncodeError:
+                raise ValueError(f"{item} can not be encoded into bytes.")
```
pekkaklarck, line 1079: You added a test where the item is `1000` and it fails for a ValueError. Should we actually validate that, if the item is an integer, it is in range 0-255 and provide a nicer error than the current `byte must be in range(0, 256)` that Python gives us? That would also avoid the issue with the error possibly changing in some future Python version.

pekkaklarck (changes requested):
Looks very good. See the inline comments for things that could be still enhanced. At least the documentation change needs to be done, other's are just "nice to have".

pekkaklarck:
This PR hasn't been updated, but I decided to merge it anyway. It is a good enough base and I can do required changes myself.

PR #5108, opened 2024-04-10
Screenshot changes to support more image types

chriscallan:
Modification of the Screenshot/Taker classes to support other image formats than just "jpg"
Implemented support for: jpg, png, tiff, and bmp; although other formats can be supported by extending Screenshot.IMAGE_FORMATS;
Extended test cases to prove out the support for non-jpg types;

### DIFF
--- a/atest/robot/standard_libraries/screenshot/take_screenshot.robot
+++ b/atest/robot/standard_libraries/screenshot/take_screenshot.robot
@@ -15,6 +15,14 @@ Basename May Be Defined
     ${tc}=  Check Test Case  ${TESTNAME}
     Check Embedding In Log  ${tc.kws[0].kws[0].msgs[1]}  foo_1.jpg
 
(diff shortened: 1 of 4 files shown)


chriscallan:
@pekkaklarck  looking at the Acceptance test results, the root error appears to be, "Make sure tests are run with a physical or virtual display."  Which I experienced on my local env when running tests originally, had to uncomment the "WaylandEnable=false" line inside /etc/gdm3/custom.conf (something about recent changes in Unbuntu).
Is there some other way to handle the above error?

pekkaklarck:
It's pretty much impossible to review the PR because the whole Screenshot.py file has changed. Have you possibly changed line endings?

chriscallan:
@pekkaklarck Sorry about the CRLFs.  I got those mixed up in my IDE.  Looks to be much cleaner now

PR #5109, opened 2024-04-11
generate_random_string_new_markers

### LINKED ISSUE #5093: Add Arabic and Polish support to `Generate Random String` keyword
can I use "Generate Random String" method under string library to generate arabic text and how I can do that?

menium878:
Added markers for ARABIC and POLISH characters.
closes #5093


pekkaklarck:
This PR would need tests.

pekkaklarck:
It would probably be a good idea to move the marker definition to module global scope to make updating it easier.

menium878:
@pekkaklarck Added tests, but if there is someone who is familiar with Arabic and could take a look would appreciate that - I'm not really familiar with language. 

Moved the marker to map, I hope this is what you meant by module global scope, if you would like me to move it somewhere else let me know.

src/robot/libraries/String.py
```diff
@@ -642,6 +642,8 @@ def generate_random_string(self, length=8, chars="[LETTERS][NUMBERS]"):
         | ``[UPPER]``   | Uppercase ASCII characters from ``A`` to ``Z``. |
         | ``[LETTERS]`` | Lowercase and uppercase ASCII characters.       |
         | ``[NUMBERS]`` | Numbers from 0 to 9.                            |
+        | ``[ARABIC]``  | Arabic characters from 0x0600 to 0x0700 unicode |
+        | ``[POLISH]``  | ASCII characters and Polish diacritical signs.  |
```
pekkaklarck, line 655: Few different things to fix/enhance:
- A note about these markers being new in RF 7.4 is needed. It probably should be at the end of the documentation where there's such a note already.
- The code uses a different range than 0x0600-0x07000 so this needs to be checked.
- Should the doc mention is the range inclusive or exclusive?
- `U+0600` could be a better to specify Unicode characters than `0x0600`.
- The correct spelling is "Unicode", not "unicode", and there should be `.` the end of a sentence.

src/robot/libraries/String.py
```diff
@@ -652,6 +654,21 @@ def generate_random_string(self, length=8, chars="[LETTERS][NUMBERS]"):
...
+            '[NUMBERS]': digits,
+            '[ARABIC]':  ''.join(chr(c) for c in range(0x0614, 0x0700)),
+            '[POLISH]': (
+                ascii_lowercase + ascii_uppercase +
+                    ''.join([
+                        "\u0105", "\u0107", "\u0119", "\u0142", "\u0144", "\u00F3", "\u015B", "\u017A", "\u017C",
+                        "\u0104", "\u0106", "\u0118", "\u0141", "\u0143", "\u00D3", "\u015A", "\u0179", "\u017B"
+                    ])
+            )
+        }
```
pekkaklarck: Few different things to look at again:
- Having markers in a dictionary is a good idea, but I believe it should be in the module scope with a name like `MARKERS`. Users could even extend it if they wanted to (but I don't think we need to document that possibility).

- The code creates Arabic characters using `range(0x0614, 0x0700)` but the documentation says the range starts from `0x0600`. Which one is correct?

- According to [Wikipedia](https://en.wikipedia.org/wiki/Arabic_script_in_Unicode) there are also other ranges with Arabic characters. I don't know enough about this to know should they be included or not so I'll ask about that in the issue. If we don't get answers, starting with the base range ought to be fine, especially when the range is documented, and other characters can be added in the future if needed.

- You can have the Polish characters as-is in the code, no need to use escape codes. You can also have a normal string like `"ąćęłńóśźżĄĆĘŁŃÓŚŹŻ"` instead of joining individual characters.

menium878: Sorry for the confusion about the range it was intended as 0x600 to 0x700.

menium878:
Thank you for the pointers hope now everything looks good. If something need more changes please let me know.

pekkaklarck:
Looks very good! Only thing that I noticed is that the `MARKERS` dictionary isn't formatted according to our code style. The easiest way to get that fixed is running `invoke format` (which needs various dependencies listed in `requirements-dev.txt`) and Contribution guidelines explain this more. I can obviously very easily do that myself too (and in the future we should automate formatting somehow).

pekkaklarck (approved):

pekkaklarck:
Hmm, I enabled executing tests against this PR and one of the runs failed like this:

> FAIL: Robot.Standard Libraries.String.Generate Random String.Generate Random String With [ARABIC]
Status of 'Generate Random String With [ARABIC]' should have been PASS but it was FAIL.
> 
> Error message:
> String 'ٚڮۥڰٻڹ۝أێ؋ڈڼلڸ؍ڞڴڶڈڝضۜ٧خڋًؒٽ؝۰؇ڰپٕؗؕؒېڠلڄہڀۄ٧ڤڔٿۀُؒڔِ٧ؚۜۑڐۙڝٔؕٶ٢؅؍ٗې٢ڒڀۥڻە۲؁ءٍ٠ڞٕۥٰڋف٪ٜ٣ڜُٯۗ؜ٞ۬ےڎصؤۜ' contains character '؜' which is not in allowed characters '؀؁؂؃؄؅؆؇؈؉؊؋،؍؎؏ؘؙؚؐؑؒؓؔؕؖؗ؛؝؞؟ؠءآأؤإئابةتثجحخدذرزسشصضطظعغػؼؽؾؿـفقكلمنهوىيًٌٍَُِّْٕٖٜٟٓٔٗ٘ٙٚٛٝٞ٠١٢٣٤٥٦٧٨٩٪٫٬٭ٮٯٰٱٲٳٴٵٶٷٸٹٺٻټٽپٿڀځڂڃڄڅچڇڈډڊڋڌڍڎڏڐڑڒړڔڕږڗژڙښڛڜڝڞڟڠڡڢڣڤڥڦڧڨکڪګڬڭڮگڰڱڲڳڴڵڶڷڸڹںڻڼڽھڿۀہۂۃۄۅۆۇۈۉۊۋیۍێۏېۑےۓ۔ەۖۗۘۙۚۛۜ۝۞ۣ۟۠ۡۢۤۥۦۧۨ۩۪ۭ۫۬ۮۯ۰۱۲۳۴۵۶۷۸۹ۺۻۼ۽۾ۿ؎'.: '؀؁؂؃؄؅؆؇؈؉؊؋،؍؎؏ؘؙؚؐؑؒؓؔؕؖؗ؛؝؞؟ؠءآأؤإئابةتثجحخدذرزسشصضطظعغػؼؽؾؿـفقكلمنهوىيًٌٍَُِّْٕٖٜٟٓٔٗ٘ٙٚٛٝٞ٠١٢٣٤٥٦٧٨٩٪٫٬٭ٮٯٰٱٲٳٴٵٶٷٸٹٺٻټٽپٿڀځڂڃڄڅچڇڈډڊڋڌڍڎڏڐڑڒړڔڕږڗژڙښڛڜڝڞڟڠڡڢڣڤڥڦڧڨکڪګڬڭڮگڰڱڲڳڴڵڶڷڸڹںڻڼڽھڿۀہۂۃۄۅۆۇۈۉۊۋیۍێۏېۑےۓ۔ەۖۗۘۙۚۛۜ۝۞ۣ۟۠ۡۢۤۥۦۧۨ۩۪ۭ۫۬ۮۯ۰۱۲۳۴۵۶۷۸۹ۺۻۼ۽۾ۿ؎' does not contain '؜'

This looks pretty odd and needs to be investigated. It could be that the issue doesn't occur in all runs, the result is random, or that it depends on the Python version.

menium878:
The issue is cause by data set for the test of Arabic language.

The data set for the test I had some problem with generating all the chars - if I wanted to generate them on the side and then copy it to have test date, not all chars would be included ( I think this is because Arabic chars, many of them are invisible etc. my editor is not handling them correctly ). 

Don’t really know how to handle that ( when I had the range from 0x614) it passed testes all the time, but I don’t think the tests should limit the range. 

Please let me know how you wanna handle that.

pekkaklarck:
I tested this locally and the problem seemed to be in the expected data i.e. in the value of the `${ARABIC}` variable. After changing it to the below that uses the inline Python evaluation syntax the problem vanished:
```
${ARABIC}         ${{[chr(i) for i in range(0x600, 0x700)]}}
```
I edited this PR (also removed an unnecessary BuiltIn import) and if tests now pass also on CI, I'll merge the changes.

PR #5111, opened 2024-04-11
replace WITH NAME by AS

kkotenko:


pekkaklarck:
Good catch! Thanks for a contribution.

PR #5136, opened 2024-05-30
feat: Add support for "week" as a valid unit of time in time strings

### LINKED ISSUE #5135: Add support for time strings containing "week" values
Currently, the largest unit of time supported by [time formats](https://robotframework.org/robotframework/latest/libraries/DateTime.html#Time%20string) in Robot Framework is `day`. This issue proposes adding `weeks`/`week`/`w` as supported units of time.

Our primary use case for this is that certain network devices will display uptime values with a maximum value of weeks. For example:
```
switch#show version
<snip>
Cisco IOS Software, C3560 Software (C3560-IPSERVICESK9-M), Version 12.2(55)SE11, RELEASE SOFTWARE (fc3)
Technical Support: http://www.cisco.com/techsupport
Copyright (c) 1986-2016 by Cisco Systems, Inc.
Compiled Wed 17-Aug-16 13:07 by prod_rel_team
Image text-base: 0x01000000, data-base: 0x02F00000

switch uptime is 11 weeks, 5 days, 3 hours, 7 minutes    <<<
System returned to ROM by power-on
System restarted at 13:16:45 EDT Sat Mar 9 2024
System image file is "flash:c3560-ipservicesk9-mz.122-55.SE11.bin"
```
As it stands right now, we must use a custom keyword to handle this time format. It'd be nice if a keyword like `Convert Time` could handle it effortlessly.

From our perspective, it doesn't make sense to add larger units of time (e.g. months, years, etc.) since the "real length" of those units could be debated and context-sensitive. For example, is a month 30 days or 31 days (or 28, in the case of February?) Is a year 365 days, or 364.25? etc. The "week" unit of time doesn't suffer from this issue, as it tends to be universally considered to be 7 days long.

ChristopherJHart:
This PR fixes #5135 and stands as a potential implementation for a "week" as a valid unit of time in a time string.

Documentation was updated to indicate that a "week" is now a valid unit of time. Furthermore, unit tests were also added to ensure backwards compatibility. I should call out that one of the unit tests ensuring invalid formats raise an exception needed to be modified from `w` to `y` in order to pass. I *think* the intent of this unit test was to ensure that units of time must be sorted (e.g. days, then hours, then minutes, etc.) and not reversed/arbitrary (minutes, then weeks, then days, etc.). However, if I misinterpreted that and need to modify the unit tests somehow, let me know!

### DIFF
--- a/atest/testdata/standard_libraries/datetime/convert_time_input_format.robot
+++ b/atest/testdata/standard_libraries/datetime/convert_time_input_format.robot
@@ -17,6 +17,12 @@ Time string           10 s                                  10
                       0.123456789 ms                        0.000123456789
                       123 μs                                0.000123
                       1 ns                                  1E-9
+                      1 week                                604800
+                      2 weeks                               1209600
+                      1 week 1 day 1 hour 1 min 1 sec 1 ms  694861.001
+                      2 weeks 1 day 1 hour 1 min 1 sec      1299661
+                      1w 1d 1h 1m 1s 1ms                    694861.001
+                      2w 1d 1h 1m 1s 1ms                    1299661.001
 
 Number as string      10                                    10
                       0.5                                   0.5
--- a/doc/userguide/src/Appendices/TimeFormat.rst
+++ b/doc/userguide/src/Appendices/TimeFormat.rst
@@ -54,7 +54,9 @@ Examples::
    1.5 minutes
    90 s
    1 day 2 hours 3 minutes 4 seconds 5 milliseconds 6 microseconds 7 nanoseconds
+   8 weeks 7 days 6 hours 5 minutes 4 seconds 3 milliseconds 2 microseconds 1 nanosecond
    1d 2h 3m 4s 5ms 6μs 7 ns
+   8w 7d 6h 5m 4s 3ms 2μs 1ns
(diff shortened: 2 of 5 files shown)


pekkaklarck (changes requested):
Changes look very good. Few things that should be done:

- It should be mentioned in the docs that `week` is new in RF 7.1. There could be a separate note about that or the list item with `weeks`, `week` and `w` could have something like `(New in Robot Framework 7.1)` at the end.
- The time format is explained also in a User Guide appendix. The new functionality should be explained there as well, along with a note about it being new in 7.1. The file to edit is `doc/userguide/src/Appendices/TimeFormat.rst`.
- It would be good to have this covered also be acceptance tests. Not sure are new tests needed, but some of the DateTime library tests could be modified to use `week`. For example at atest/testdata/standard_libraries/datetime/convert_time_input_format.robot.

ChristopherJHart:
Requested changes have been implemented! Sorry about not originally including the acceptance tests - I totally glossed over the fact the `atest` directory exists 😅

pekkaklarck (approved):

pekkaklarck:
Thanks for an excellent PR @ChristopherJHart!

PR #5145, opened 2024-06-09
Dutch language update

JFoederer:
Recently there has been a great effort to translate RIDE into different languages. Dutch being one of them. This pull request is there to bring Robot and RIDE translations in sync.

The branch with the new Dutch translations has been offered for review to the `rf-nl` and `rf-be` channels on Slack.


leeuwe:
@pekkaklarck: @JFoederer and I are discussing some things in Slack. Once we reach a desision I'll do a review here.
Because this is the first time there's a change in an initial translation, I would like to add something in the release notes about how to go from translations from version x to version y.

pekkaklarck:
Sounds good. Could someone submit an issue about this so that we can add it to RF 7.1 scope? We can discuss about the process of updating translations there as well.

leeuwe (approved):
Discussed, waited for further feedback from the community but none given

pekkaklarck:
Thanks for the new translations. The question is, should we still support the old ones or is it enough to just replace them? The latter is simpler, I just need to merge this PR, but also backwards incompatible. The former is safer, and would benefit also other languages, but obviously requires some work. It shouldn't require too much work, though, and I already have an idea how to implement that. Perhaps I experiment with my idea a bit and make a decision then.

Issue #5151, opened 2024-06-19
`Evaluate` keyword doesn't take attributes added into `builtins` module into account


mithunmithra:
The keyword fails for a particular usecase as follows :
helper file:
```
import builtins

class dummy:
    def __init__(self):
        self.name = 'abc'
        builtins.s = self

    
    def foo(self):
        print('in foo module')
    
    def initialize(self):
        print(self.name)


def dummy_initiliaze():
    dummy().initialize()
```

Testcase file:
```
*** Settings ***
Library           String
Library           Collections
Library           BuiltIn
Library           helper.py

*** Test Cases ***
TC6
   Dummy Initiliaze
   ${res} =    Evaluate    s.foo()
```

When I run this in Robot framework 3.1.2 , the function s.foo() is called and works
But in Robot framework 3.2.2, its erroring out with "Evaluating expression 's.foo()' failed: NameError: name 's' is not defined nor importable as module"
Have discussed the same in slack forum, thread : https://robotframework.slack.com/archives/C3C28F9DF/p1718686445500879

pekkaklarck:
I don't think modifying `builtins` like this is a good idea in general, but because fixing the issue seems to be easy, I don't see why it shouldn't be done either.

pekkaklarck:
Raised the priority slightly because this usage used to work so it being broken is a regression. It cannot be too bad a regression, though, because nobody had reported it earlier.

PR #5152, opened 2024-06-19
fixed the evaluate keyword #5151 

mithunmithra:
As per the slack thread : https://robotframework.slack.com/archives/C3C28F9DF/p1718686445500879 
I did the fix for the issue I reported #5151 
@pekkaklarck please review and merge this PR.

### DIFF
--- a/src/robot/variables/evaluation.py
+++ b/src/robot/variables/evaluation.py
@@ -149,6 +149,8 @@ class EvaluationNamespace(MutableMapping):
         if name in PYTHON_BUILTINS:
             raise KeyError
         try:
+            if hasattr(builtins,name):
+                return getattr(builtins, name)
             return __import__(name)
         except ImportError:
             raise NameError(f"name '{name}' is not defined nor importable as module")


src/robot/variables/evaluation.py
```diff
@@ -149,6 +149,8 @@ def _import_module(self, name):
         if name in PYTHON_BUILTINS:
             raise KeyError
         try:
+            if hasattr(builtins,name):
+                return getattr(builtins, name)
```
pekkaklarck, line 153: There's already a check for the `name` possibly being a built-in few lines above. The reason it doesn't work is that it uses the `PYTHON_BUILTINS` module level attribute that's not updated if `builtins` is changed. That check could just be replaced with `if hasattr(builtins, name)` and the whole constant removed. The constant feels unnecessary in general and I believe it has only been needed when we supported Python 2 and Python 3.

pekkaklarck (changes requested):
This fix needs a test both to validate that it works and to avoid future changes breaking it again. The code should also be enhanced as commented separately.

pekkaklarck:
As noted above, fixes like this need tests. If you don't know how Robot's acceptance test system works, see `atest/README.rst`. If you have questions after that, ping me on the #devel channel on our Slack.

pekkaklarck:
There were no updates to this PR so I decided to fix the issue myself. It was done in 25d64044b8429f36d9e62a6cb12c8c3a969080f6.

Issue #5089, opened 2024-03-22
Ignore empty log messages (including messages cleared by listeners)


pekkaklarck:
Listeners using the v3 API can modify the logged message, but they cannot remove it altogether. Even if they set the message to an empty string, that empty string is still logged. This adds unnecessary noise to the log file and increases the size of output.xml.

This is easiest to implement by ignoring log messages having an empty string as the value. Not writing them to output.xml is easy and that automatically takes care of removing them from the log file as well. They should not be included in the result model created during execution either, but I actually believe that model doesn't get log messages at all at the moment.

---

**UPDATE:** This change will obviously mean that all empty messages are ignored. Updated the title accordingly.

Meet2rohit99:
@pekkaklarck , i am new to this repo and would like to fix this issue. could you please assign this to me.

OukhellouFst:
@pekkaklarck, I'm currently working with Robotframework to do E2E testing for our application. I'll be glad to fix this simple issue

pekkaklarck:
Anyone is free to create a pull request with the fix and needed tests. Assuming we implement this simply by not writing empty messages to output.xml, the implementation is pretty easy. Creating a test isn't hard either, but requires being familiar with Robot's acceptance test system and designing where to add the test.

Do you @Meet2rohit99 and @OukhellouFst want to agree who tries to create a PR first or do you both want to do it? I'm happy to review PRs either way and can also help otherwise if needed.

pekkaklarck:
Is anyone working with this issue? If not, I can implement it myself as part of other listener enhancements.

Aadil-5122:
@pekkaklarck, I am new to this repo, and working on it.

Aadil-5122:
Here goes the Pull Request:
https://github.com/robotframework/robotframework/pull/5168

pekkaklarck:
The PR mentioned above only worked with messages logged by libraries using `robot.api.logger`, not if a message is cleared by a listener. It's easiest to do the latter by not writing empty messages to the generated output.xml. The aforementioned PR made me thinking that empty messages logged by libraries could be dropped already earlier, though.

pekkaklarck:
I got this implemented and when I run all tests there are some unexpected failures. The failures were due to empty messages logged with the `Log Many` keyword being ignored as well and that's actually not good at all. For example, if I use something like
```
Log Many    ${msg1}    ${msg2}    ${msg3}    ${msg4}
```
[...]

PR #5168, opened 2024-07-21
chore: Allow listeners to remove log messages altogether

Aadil-5122:
# Description

- This implements a check for an empty string as `msg` argument at `robots/api/loggers.py`. If the string is empty, it returns the function, thus no log is added.

# Test

- The test case for the same is also written in `utest/api/test_logging_api.py`. The test case passes an empty string as a message and asserts if any message is logged.

### DIFF
--- a/utest/api/test_logging_api.py
+++ b/utest/api/test_logging_api.py
@@ -79,6 +79,13 @@ class TestRedirectToPythonLogging(unittest.TestCase):
         logger.write("Doo", 'INFO')
         assert_equal(self.handler.messages, ['Foo', 'Boo', 'Goo', 'Doo'])
 
+    def test_empty_string_not_looged_to_python(self):
+        logger.info("")
+        logger.debug("")
+        logger.trace("")
+        logger.write("", 'INFO')
+        assert_equal(self.handler.messages, [])
+
     def test_logger_to_python_with_html(self):
         logger.info("Foo", html=True)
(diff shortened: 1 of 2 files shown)


utest/api/test_logging_api.py
```diff
@@ -79,6 +79,13 @@ def test_logged_to_python(self):
         logger.write("Doo", 'INFO')
         assert_equal(self.handler.messages, ['Foo', 'Boo', 'Goo', 'Doo'])
 
+    def test_empty_string_not_looged_to_python(self):
```
kkotenko, line 82: Could you fix the typo? (It says looged instead of logged)

Aadil-5122: Done!

pekkaklarck:
Thanks for the PR, but unfortunately this implementation doesn't work properly. It works when a library uses `robot.api.logger` with an empty message, but if a library uses the `logging` module or if a listener clears an existing message. The possibility for listeners to be able to remove messages altogether is the main point of #5089 so the PR cannot be merged in its current form.

We want to get RF 7.1 out ASAP and it's thus probably easiest if I implement this functionality. I'm sorry that due to my parental leave and holidays I didn't get this PR reviewed earlier to give you time to fix it. I hope you are still interested to contribute and find something else interesting to work on.

pekkaklarck:
In the end it there turned out to be too big problems with the change and I closed #5089 altogether.

PR #5176, opened 2024-08-08
Fixes #5054, Enhances the `should_contain` keyword in the BuiltIn lib…

### LINKED ISSUE #5054: Enhance Should Contains for bytes

Abhilash-du:
Changes:
- Updated the `should_contain` method to:
  - Automatically encode strings to bytes using 'latin-1' when the container is a bytes object.
  - Check if an integer item is within the range 0-255 before converting it to bytes.
  - Raise a `ValueError` with a clear message if the integer is out of range.


Abhilash-du:
Just wanted to follow up on this PR. It's been open for a couple of weeks, and I'd appreciate it if someone could take a look when they have a chance. The changes aim to enhance the should_contain method by adding better handling for byte objects and improving error messages for out-of-range integers.

@pekkaklarck, if you have some time, I would really appreciate your input on this as well. Let me know if there's anything else I can do to help get this reviewed.

pekkaklarck:
Thanks for the PR @Abhilash-du. It looks pretty good, but there seem to be smallish issues that needs to be fixed. Most importantly, the tests you added to `atest/testdata` side should be validated by tests on the `atest/robot` side. 

We already had PR #5097 fixing #5054, but also that has smallish issues. I want to get #5054 closed now to get RF 7.1 released soon, and I need to decide which one of these PRs to use as the base. I go with the earlier one simply because I had reviewed it in detail already earlier, so it's very fast for me to make the required changes.

Thanks again for the PR and sorry for closing this without giving you time to fix it. The delay in review was due to my paternity leave and holidays. I hope you are interested to contribute in the future!

PR #5178, opened 2024-08-10
Add Korean translation

rivercory:
https://github.com/robotframework/RIDE/pull/2848
Based on this PR, I added a Korean translation to the robotframework! Please review my PR and give me feedback!


pekkaklarck:
Thanks for the PR! Not knowing Korean I have no comments about translations themselves, but here are some general comments about the PR and otherwise:
1. You only need to modify the `src/robot/conf/languages.py` file. Everything in the User Guide is generated automatically.
2. There should be a note about when the language was added. You can see examples with other languages like Vietnamese.
3. There's an empty line missing at the end of the file.
4. Have you communicated about the terms you used with other Korean users? They can be changed in the future if needed, but that causes backwards compatibility issues and it would thus be better if there would be consensus about the terms to begin with.
5. We need an issue about this for tracking purposes. I can create one myself.
6. I hope we could get this still into RF 7.1 and I'll add the aforementioned issue into its scope. There's some hurry with it, though, because we are trying to get a release candidate out still this week. Adding translations is safe enough to be done also after that, but there's not too much time before the final release either.

pekkaklarck:
I tested these changes locally and User Guide modifications didn't seem to cause any issues so merging them with other changes ought to be fine. There's the "New in RF 7.1" note missing, but I can add that (and the missing newline at the end of the file) myself afterwards.

Issue #5189, opened 2024-08-29
Make result file paths hyperlinks on terminal


pekkaklarck:
Terminal emulators nowadays support hyperlinks pretty well.The [standard emerged in 2017](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) and the current list of [supporting terminal emulators](https://github.com/Alhadis/OSC8-Adoption/?tab=readme-ov-file) is pretty exhaustive. Based on a quick prototype, making the links to result files in the console after execution is easy and works very well.

The simple solution ought to work fine on all Linux and OSX terminals and the main problem is handling Windows. The traditional [Windows Console](https://en.wikipedia.org/wiki/Windows_Console) isn't a terminal emulator and doesn't support ANSI colors or hyperlinks. We have custom code for handling colors, but I don't think something like that is possible with links. There are, however, various other terminals for Windows and, for example, the Microsoft developed [Windows Terminal](https://en.wikipedia.org/wiki/Windows_Terminal) is a proper terminal emulator that supports both ANSI colors and hyperlinks.

The problem is that we don't currently have any code for detecting terminal capabilities. With colors we, by default, simply use ANSI outside Windows and on Windows use the aforementioned custom solution. Using the same approach with hyperlinks would be easy, but then Windows users with "proper" terminals would need to separately enable hyperlinks with  `--console-colors ansi` (or `-C ansi`). That's a bit annoying, but I believe it would be fine in the beginning.

I consider this so convenient feature that I tentatively add this to RF 7.1 scope and ask opinions from others on the #devel channel on our Slack. Because we want RF 7.1 out soon, there's no time for anything bigger, so my proposal is to support hyperlinks when ANSI colors are enabled and to keep them disabled on Windows by default. If we agree this is a good approach, I'll submit a separate issue about enhancing terminal capability testing in RF 7.2 to make the Windows support better.

pekkaklarck:
We have a highly related issue that proposed prefixing result file paths with `file:///` to make them hyperlinks on terminals that auto-detect URLs. The problem was that the prefix would have added some clutter and not that many terminal supported this functionality back in 2016 when the issue was submitted. The custom hyperlink support avoids the clutter and also the terminal support is good nowadays.

pekkaklarck:
I committed the changes that enabled hyperlinks using ANSI codes. As explained in the original description, we use ANSI codes by default outside Windows and on Windows it is possible to enable them using `-C ansi`. 

[...]

PR #5192, opened 2024-08-30
Allow WHILE limit to be modified in listener V3

### LINKED ISSUE #5194: Allow WHILE limit to be modified in listener V3
Currently it is impossible to change WHILE limit settings using the listener interface- listener object is created before control is passed to attached listeners, thus modifying while body parameters related to the limit is pointless.

It might be useful to allow listeners to change this- for example, an user might want to redefine default limit behaviour for WHILE loops in case it was not explicitly defined. For example, instead of having default 10k iterations as a default, no-configuration WHILE loop would abort after 10 seconds using this feature.

Blashaq:
resolves #5194

Currently it is impossible to change WHILE limit settings using the listener interface- listener object is created before control is passed to attached listeners, thus modifying _while_ body parameters related to the limit is pointless.

It might be useful to allow listeners to change this- for example, an user might want to redefine default limit behaviour for WHILE loops in case it was not explicitly defined. For example, instead of having default 10k iterations as a default, no-configuration WHILE loop would abort after 10 seconds using this feature.

This change moves creation of WhileLimit object a bit further - this allows listeners to change those settings.


pekkaklarck:
Sounds good to me and the code changes look simple as well. Looking at the code, it seems that reporting errors if the limit is invalid could change, but I'm not sure does that actually have effect. Did you run all tests to see did anything break?

Could you create an issue about this for tracking purposes? I can then assign it to RF 7.2 scope. If this is important for you, we can still think could we add it in RF 7.1. As you may have noticed, we just released RF 7.1 rc 1 and I wouldn't like to make any changes before the final release, but this looks simple and useful enough that an exception is possible.

Blashaq:
_Looking at the code, it seems that reporting errors if the limit is invalid could change, but I'm not sure does that actually have effect. Did you run all tests to see did anything break?_

Output is slightly different- before this change, bad limit arguments would log first iteration, however it would not run and was marked as [NOT RUN]. After this change, there is no mention of the iterations, just the error reported inside the while loop. One could argue that the log looks cleaner this way :)

This is how it looks before the change:
image

and after:
image

I will also attach output files before/after the change.

 _Did you run all tests to see did anything break?_
Yes, the approach to testing RF itself is very interesting. This change didn't break any of existing tests.
I didn't think about including this specific test (how bad limit arguments are reported and how log structure was changed), but I could add those if you wish.

_If this is important for you, we can still think could we add it in RF 7.1. As you may have noticed, we just released RF 7.1 rc 1 and I wouldn't like to make any changes before the final release, but this looks simple and useful enough that an exception is possible._

7.2 is good enough for me. Thank you for your time.

pekkaklarck:
Thanks for the update. I'll take a bit closer look at this later today or tomorrow. If there aren't issues, we can still included this to RF 7.1. We need rc 2 anyway due to small problems with #5189.

atest/testdata/output/listener_interface/body_items_v3/tests.robot
```diff
@@ -59,6 +59,19 @@ WHILE
         BREAK
     END
 
+WHILE with modified limit
+    [Documentation]    FAIL 
+    ...    WHILE loop was aborted because it did not finish within the limit of 1 iterations. Use the 'limit' argument to increase or remove the limit if needed.
```
pekkaklarck, line 64: This failure spec confused me, because it gives an impression that the test would fail with this message and in that case modifications wouldn't work. It turned out that the expected test status is set explicitly when this test is validated on `atest/robot` side, That means that this spec isn't used for anything and can be removed. I can do that myself and I'll consider changing how other tests handle specifying their expected status as well.

pekkaklarck (approved):
Looks great!

pekkaklarck:
In general we always run loops once even in the `NOT RUN` state to document what would have been run if the loop would have been executed. I don't think that's too relevant if the loop configuration is invalid, though. That ought to be pretty rare and also something that you notice and can fix immediately.

PR #5204, opened 2024-09-10
Process: Support configuring log level keywords use

LowEQ:
initial implementation, still need add `atest`  probably to `atest\testdata` and `atest\robot`


src/robot/libraries/Process.py
```diff
@@ -323,7 +323,7 @@ def __init__(self):
         self._processes = ConnectionCache('No active process.')
         self._results = {}
 
-    def run_process(self, command, *arguments, **configuration):
+    def run_process(self, command, log_level=None, *arguments, **configuration):
```
pekkaklarck, line 326: `log_level` cannot be a positional argument because it would break pretty much all current usages of this keyword. It should be a keyword-only argument, but I'm not sure is it better to be a separate argument or should it be passed in in `**configuration` like other such options. The reason we've used `**configuration` instead of real keyword-only arguments is that this code was written originally for Python 2 that didn't even have keyword-only arguments. It probably should be refactored now, but that should be done separately from this PR. In the current situation including it in `**configuration` is probably easiest and most consistent.

src/robot/libraries/Process.py
```diff
@@ -349,26 +349,38 @@ def run_process(self, command, *arguments, **configuration):
...
         | ${result} = | Run Process | ${command} | stdout=${CURDIR}/stdout.txt | stderr=STDOUT |
         | ${result} = | Run Process | ${command} | timeout=1min | on_timeout=continue |
         | ${result} = | Run Process | java -Dname\\=value Example | shell=True | cwd=${EXAMPLE} |
+        
+        ``log_level`` examples:
+        | ${result} = | Run Process | python | 'INFO' | -c | print('Hello, world!') |
+        | Should Be Equal | ${result.stdout} | Hello, world! |
+        | ${result} = | Run Process | ${command} | 'debug' | stdout=${CURDIR}/stdout.txt | stderr=STDOUT |
+        | ${result} = | Run Process | ${command} | 'Warn' | timeout=1min | on_timeout=continue |
+        | ${result} = | Run Process | java -Dname\\=value Example | 'Debug' | shell=True | cwd=${EXAMPLE} |
```
pekkaklarck, line 368: There's no need to duplicate all examples with `log_level`. Adding a new example with it, or possibly just adding it to one of the existing examples, is enough.

src/robot/libraries/Process.py
LowEQ: I am not sure, if it is properly done  by just adding `log_level` through `ProcessConfiguration` `__init__`.

LowEQ:
tried run acceptance tests, all failed, rolling back this branch changes, and try to add the log_level arg into `**configuration`

LowEQ:
added atests

franzhaas:
this interacts with this one https://github.com/robotframework/robotframework/pull/5302. 

It is likely that there will be some coordination necessary.

LowEQ:
> this interacts with this one #5302.
> 
> It is likely that there will be some coordination necessary.

sorry, I was in an long vacation, will try to merge these changes together

LowEQ:
Should be able to review again. when running locally one of the Unit test failed for some reason, but after pushing the change the unit tests passes all?

pekkaklarck:
Now that RF 7.4 is getting ready, I finally had a possibility to look at this PR. Unfortunately there are lot of conflicts, most likely due to #5412, that need to be resolved before this could be merged. I did, however, take a quick look at the changes. They look good in general, but there are various bigger and smaller things that would need to be fixed.

Are you @LowEQ still around and interested on this PR? If you are, I can do a proper review for the current changes later this week. In addition to addressing the review comments, you then needed to resolve the conflicts.

LowEQ:
> Now that RF 7.4 is getting ready, I finally had a possibility to look at this PR. Unfortunately there are lot of conflicts, most likely due to #5412, that need to be resolved before this could be merged. I did, however, take a quick look at the changes. They look good in general, but there are various bigger and smaller things that would need to be fixed.
> 
> Are you @LowEQ still around and interested on this PR? If you are, I can do a proper review for the current changes later this week. In addition to addressing the review comments, you then needed to resolve the conflicts.

yeah, just got time read email this week in detail, Sure, I got time to look at this PR and do needed resolves and changes, as soon as I have done it. I can again request an review and following fix based on the review.

LowEQ:
Ok, finally passed all test + changed some doc, could you review it @pekkaklarck. might be still need changes. 
`wait_for_process` and `terminate_process` still contains `log_level` in parameter, otherwise the `log_level` config happens in `env_extra` dictionary.

oboehmer:
> Ok, finally passed all test + changed some doc, could you review it @pekkaklarck. might be still need changes. `wait_for_process` and `terminate_process` still contains `log_level` in parameter, otherwise the `log_level` config happens in `env_extra` dictionary.

Adding this option via `env_extra` is counter-intuitive, the environment should not have any influence on the actual keyword execution, only on the process.

pekkaklarck:
I agree with @oboehmer that `env_extra` should only be used for configuring environment variables of the started process, not the log level keywords or other such configuration. It would be easiest if all keywords had a separate `log_level` argument, but there's a problem that some keywords log with both INFO and DEBUG levels so a single argument accepting single value wouldn't work well.

A bigger problem with all this is that the use case for the whole functionality is somewhat unclear. As I just [commented](https://github.com/robotframework/robotframework/issues/4929#issuecomment-3480955092) the original issue, the new Secret variable support and the existing `Set Log Level` keyword can solve many/most/all of the possible usages. In this situation implementing this in RF 7.4 isn't a good idea, so I moved the issue to RF 7.5 scope and we can decide about this later.

If it turns out this won't be implemented, I'm sorry that your hard work @LowEQ with this will be wasted. In that case I hope you are interested to contribute something else later.

pekkaklarck:
Issue #4929 was closed as not planned so this PR needs to closed as well. There are also conflicts due to somewhat recently added type hints that needed to be resolved, but that doesn't really matter at this point.

PR #5211, opened 2024-09-20
fix  sys.__stderr__.write

miaopass-future:
Using pyinstaller packaging, an error is reported when the parameter console configures False

 sys.__stderr__.write(console_encode(stderr, stream=sys.__stderr__))
AttributeError: 'NoneType' object has no attribute 'write'

### DIFF
--- a/src/robot/running/outputcapture.py
+++ b/src/robot/running/outputcapture.py
@@ -52,7 +52,8 @@ class OutputCapturer:
             LOGGER.log_output(stdout)
         if stderr:
             LOGGER.log_output(stderr)
-            sys.__stderr__.write(console_encode(stderr, stream=sys.__stderr__))
+            if sys.__stderr__:
+                sys.__stderr__.write(console_encode(stderr, stream=sys.__stderr__))
 
     def _release(self):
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Could you submit an issue about this problem? We use issues for tracking purposes and, for example, need to decide is this something we should fix in RF 7.1.1 or is waiting for RF 7.2 enough.

Someone setting `sys.__stderr__` to `None` sounds pretty odd. If you have more information about this, please clarify it in the issue.

miaopass-future:
> Could you submit an issue about this problem? We use issues for tracking purposes and, for example, need to decide is this something we should fix in RF 7.1.1 or is waiting for RF 7.2 enough.
> 
> Someone setting `sys.__stderr__` to `None` sounds pretty odd. If you have more information about this, please clarify it in the issue.

https://github.com/robotframework/robotframework/issues/5212

pekkaklarck:
This wasn't the only place where `sys.__strderr__` was accessed without checking is it `None` or not. I went all those case through and fixed them in 5efbcdeb20ca3cbd5e084e84cfd8140285191224.

PR #5220, opened 2024-09-25
Dictionary Should Contain Sub Dictionary is missing ignore_value_order

### LINKED ISSUE #5007: Collections: Support ignoring order in values when comparing dictionaries
Currently https://github.com/robotframework/robotframework/blob/master/src/robot/libraries/Collections.py#L839 we are missing out on ignore_order already done for 
```
    def lists_should_be_equal(self, list1, list2, msg=None, values=True,
                              names=None, ignore_order=False, ignore_case=False):
```
Ref : https://github.com/ngoan1608/robotframework/commit/1d852208255d1245cce33a34c700478455cc7298#diff-0bb57664ba32d4e2b91c6435cf0b867d630c2db282af83004073303030775bd1 

Adding as well ignore_key would be much appreciated since we might deal with ID's and timestamps, and in such cases those keys are to be ignored.

MarcinGmurczyk:
fix #5007


pekkaklarck:
Looks good on a mobile screen. I'll add the related issue to RF 7.2 scope and review this properly later.

MarcinGmurczyk:
Acceptance tests failed. I might not fully understand how to properly add them to the pipelines.

MarcinGmurczyk:
Although code is working fine shouldn't better approach be to extend Collections.Normalize.normalize_dict method?

MarcinGmurczyk:
Going back to this because frankly speaking I'm using patched version of Collection.py at work everyday.
I've found out a new case that occured in our test runs.
When compared dicts has a key that has a value of a list of another dictionaries
> {
	&emsp;"a": 5,
> &emsp;"b": [
> &emsp;&emsp;{"a": 3, "b": 4},
> &emsp;&emsp;{"a": 1, "b": 2} 
> &emsp;]
> }

using [current solution here](https://github.com/robotframework/robotframework/pull/5220/files#diff-0bb57664ba32d4e2b91c6435cf0b867d630c2db282af83004073303030775bd1R1179) will throw an exception
`TypeError: '<' not supported between instances of 'dict' and 'dict'` so sorting them by casting to `str` should work fine.

atest/testdata/standard_libraries/collections/dictionaries_should_be_equal.robot
```diff
@@ -134,3 +140,15 @@ Different values and custom error message with values
     ...    Dictionary {'a': 1, 'A': 2} contains multiple keys that are normalized to 'a'. \
     ...    Try normalizing only dictionary values like 'ignore_case=values'.
     Dictionaries Should Be Equal    ${{{'a': 1, 'A': 2}}}    ${{{'a': 2, 'A': 1}}}    ignore_case=True
+
+`ignore_value_order` set to True
+    Dictionaries Should Be Equal    ${D10A}    ${D10B}    ignore_value_order=True
+    
+`ignore_value_order` set to True when dictionaries has key with value of a list of a dictionaries
+    Dictionaries Should Be Equal    ${D10C}    ${D10D}    ignore_value_order=True
```
pekkaklarck, line 148: Instead of using variables created in the Variables section just for these tests, they could inline Python expressions similarly as in the test above. I believe most of the existing variables predate the inline expression syntax, and in cases where variable is reused creating it in Variables is a good idea as well.

The above said, this approach is fine as well. You can leave it like it is if you prefer it that way.

atest/testdata/standard_libraries/collections/dictionary_should_contain.robot
```diff
@@ -4,8 +4,10 @@ Library           CollectionsHelperLibrary.py
 
 *** Variables ***
 @{LIST}          a    B
+@{LIST_2}          B    a
 ${TUPLE}         ${{'a', 'B'}}
 &{D}             a=x    B=Y    c=${3}   ${4}=E    ß=Straße    list=${LIST}    ${TUPLE}=tuple
+&{D_2}             a=x    B=Y    list=${LIST_2}    ${TUPLE}=tuple
```
pekkaklarck, line 10: Items should be aligned so that first items start from the same column.

src/robot/libraries/Collections.py
```diff
@@ -783,7 +783,7 @@ def dictionary_should_not_contain_value(self, dictionary, value, msg=None,
         )
 
     def dictionaries_should_be_equal(self, dict1, dict2, msg=None, values=True,
-                                     ignore_keys=None, ignore_case=False):
+                                     ignore_keys=None, ignore_case=False, ignore_value_order=False):
```
pekkaklarck, line 786: This row seems to be over our maximum 88 character recommendation. Please change it to something like below and go through also other long lines.
```python
ignore_keys=None, ignore_case=False,
ignore_value_order=False):
```

src/robot/libraries/Collections.py
```diff
@@ -1165,7 +1175,10 @@ def normalize_string(self, value):
     def normalize_list(self, value):
         cls = type(value)
         if self.ignore_order:
-            value = sorted(value)
+            try:
+                value = sorted(value)
+            except TypeError:
+                value = sorted(value, key=str)
```
pekkaklarck, line 1181: I don't know is this a good idea for the following reasons:
1. Documentation already says that list items need to be sortable. Sorting failing in other cases is expected.
2. Sorting heterogeneous lists would actually be nice, but  I don't think it makes sense to sort something like `['A', 'X', None]` as `['A', None, 'X']`.

I think it's best to drop this change from this PR. If you think sorting heterogeneous lists is valuable, please submit a separate issue (and a PR) about that. A better way to sort such lists would be something like this:
```python
>>> sorted(['A', 'X', None], key=lambda i: (type(i).__name__, i))
[None, 'A', 'X']
```
The above key function mimics the sorting behavior under Python 2, but also it fails if items aren't sortable with each others:
```python 
>>> sorted([1j, 2j], key=lambda i: (type(i).__name__, i))
  File "<python-input-16>", line 1, in <module>
    sorted([1j, 2j], key=lambda i: (type(i).__name__, i))
    ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'complex' and 'complex'
```
I don't know would we need to care about the above limitation, but discussing it is anyway out of the scope here. Please submit a separate issue as I proposed and we can discuss about alternatives there.

pekkaklarck (changes requested):
Looks very good! See the individual comments for details.

pekkaklarck:
I noticed your comment about sorting lists containing dictionaries only after reviewing changes. I understand now why you used `sorted(value, key=str)`, but still believe that change doesn't belong to this PR. As I already mentioned, we could support sorting heterogeneous lists and could also think how to handle items that aren't themselves sortable, but that requires a separate issue.

Related to sorting (and being off topic here), I though my earlier `key=lambda i: (type(i).__name__, i)` trick would mimic Python 2 behavior, but apparently there have been other changes as well. In Python 2 you could sort lists containing dictionaries, but sorting complex numbers was explicitly prevented:

```python
>>> sorted([{'a': 1}, {'a': 2}])
[{'a': 1}, {'a': 2}]
>>> sorted([1j, 2j])
TypeError: no ordering relation is defined for complex numbers
```

As I already demonstrated, the aforementioned trick didn't support sorting complex numbers, but it also fails the same with dicts:
```python
>>> sorted([{'a': 1}, {'a': 2}], key=lambda i: (type(i).__name__, i))
  File "<python-input-1>", line 1, in <module>
    sorted([{'a': 1}, {'a': 2}], key=lambda i: (type(i).__name__, i))
    ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'dict' and 'dict'
```

A solution to this would be a key function that handles arguments we care about specifically. For example, this handles both dictionaries and complex numbers:
```python
def sort_everything_key(item):
    type_name = type(item).__name__
    if isinstance(item, Mapping):
        item = tuple(item.items())    # Pprobably we should also sort items here
    elif isinstance(item, complex):
        item = (item.real, item.imag)    # In case we want to support also complex numbers
    elif ...:
        ...    # Handle possible other unsortable items
    return type_name, item
```

atest/testdata/standard_libraries/collections/dictionaries_should_be_equal.robot
MarcinGmurczyk: leaving this as is

src/robot/libraries/Collections.py
MarcinGmurczyk: resolved

src/robot/libraries/Collections.py
MarcinGmurczyk: will propose a new PR for those changes

MarcinGmurczyk:
I will open a new PR this issue https://github.com/robotframework/robotframework/pull/5220#issuecomment-2457742608 
and all others issues are resolved.

pekkaklarck (approved):

pekkaklarck:
This looks very good! I'll merge it now and we can continue discussion about sorting separately.

pekkaklarck:
I noticed that my earlier `sort_everything_key` doesn't handle numbers correctly as it sorts, for example, ints and floats separately: ```python
>>> sorted([1, -2.1, 3.14], key=sort_everything_key)
[-2.1, 3.14, 1]
```

The reason is that all number types get a different `type_name`. The same problem occurs also with different mappings. This ought to fix both of these problems:

A fix would be something like this:
```python
>>> from collections.abc import Mapping
>>> from numbers import Complex, Number
>>>
>>> def sort_everything_key(item):
...     type_name = type(item).__name__
...     if isinstance(item, Complex):
...         type_name = 'numbers.Complex'
...         item = (item.real, item.imag)
...     elif isinstance(item, Number):
...         type_name = 'numbers.Number'
...     elif isinstance(item, Mapping):
...         type_name = 'collections.abc.Mapping'
...         item = tuple(item.items())    # sort these
...     elif ...:
...         ...
...     return type_name, item
...
>>> sorted([1, -2.1, 3.14], key=sort_everything_key)
[-2.1, 1, 3.14]
```

Now I'll stop commenting about sorting in this unrelated PR. Let's continue discussing it after you submit a separate issue.

Issue #5210, opened 2024-09-17
Support warnings, not only errors, in parsing model


pekkaklarck:
The parsing model supports setting errors for individual tokens, but there are two problems with the current approach:

1. Errors (typically syntax errors) aren't separated from warnings (typically deprecations). This caused a problem when deprecating singular section headers (#4432) and the code reporting the warning as part of execution needed to know [what errors can occur in different contexts](https://github.com/robotframework/robotframework/blob/d007855fea4957031cb952c4157fa36dfb3b7228/src/robot/running/builder/transformers.py#L516) to handle the problem. With the planned deprecation of the pipe-separated format (#5203) this is likely even a bigger problem, because there's no single context where pipes can be used.
2. It isn't easy to recognize if the same error/warning occurs multiple times in same file. This requires comparing messages, but because they typically contain some variables parts, they cannot be compared directly. Searching for a certain sub-string is also problematic, because error messages can change.

My proposal for solving both of the above problems is adding new `errno` attribute to to the `Token` objects with the following semantics:
1. `errno` is an integer representing an error code.
2. Each error and warning has a different code.
3. Codes have constant names like `ERR_SINGULAR_HEADER_DEPRECATED` stored as attributes of the `Token` class.
4. Errors and warnings have different ranges so that it is easy to separate them. We could, for example, use negative values with warnings and positive with errors. Zero would mean no error/warning.

mikeleppane:
Hey! If it is okay, I would be interested in looking at this.

mikeleppane:
After looking into this, I wondered if it would make sense to introduce a data model for errors/warnings. 
```python
@dataclass
class TokenError:
    type: TokenErrorType
    code: TokenErrorCode
    message: str | None = None
```
This way, we don't need any specific ranges for errors/warnings.

mikeleppane:
This is just playing around with the code. Trying to find the best solution. Here's a one proposal:
```python
visit_SectionHeader(self, node):
        token: 'Token|None' = node.get_token(*Token.HEADER_TOKENS)
        if token is None or token.error is None:
            return
        token.dump_error_or_raise()
```
It allows the Token type to handle its reporting. I think this `raise_on_invalid_header` does not belong to the ` ErrorReporter` type.

mikeleppane:
I've had a working implementation ready locally for some time now. Should I create a pull request, or should we discuss the implementation approach?

pekkaklarck:
[...]

Issue #5203, opened 2024-09-09
Deprecate pipe-separated format


pekkaklarck:
The pipe-separated format is very little used and for the reasons explained in #5199 we have decided to remove it. Removing it is a backwards incompatible change and this issue covers deprecating it.

pekkaklarck:
An open design decision related to the deprecation is when to emit warnings and how to mark deprecation in the parsing model. The pipe-separated format is recognized line-by-line, but it certainly isn't a good idea to emit a warning for each line. A warning per file would probably be good, but a warning only once per execution can be considered as well.

A technical challenge related to this is separating deprecation warnings from errors in the parsing model. At the moment there's no such separation. Everything is reported as an error and then the code handling errors when the execution model is built just knows based on context are errors actually warnings or not. That may not work too well in this case and we probably should add some better mechanism.

d-biehl:
I would report it for every file, but I'd say if it's possible to report block-wise, if the entire file is pipe-delimited, then it's one block, and it only reported once, but I've seen robot files where only a few lines of a file are pipe-separated to mimic the behavior of cucumber/gherkin example tables.

Snooz82:
maybe add a custom parser that continues to support Pipes.
As i  #5199

pekkaklarck:
Some users on Slack have told they really like the pipe-separated format and have lot of tests using it. To avoid breaking them, we could consider having the pipe format available optionally and in that case we shouldn't deprecate it. We want to get RF 7.2 out soon, so let's move this issue to RF 7.3 and decide about it later.

test-fullautomation:
Would it be good to deactivate the pipe-separated format by default. This avoids that new projects / users use it.
A commandline / config parameter could allow to activate it for those users who have already existing test suites.
This procedure would support the transition.

PR #5241, opened 2024-10-17
♻️ Enhance reporting errors and warnings in parsing model token

mikeleppane:
This change introduces a new type, `InvalidTokenError` to hold and represent invalid tokens while parsing source files. This improves the separation of concerns and allows better control and isolation of how invalid tokens should be handled.

Error codes are represented by an enum (surprisingly, as a type `ErrorCode`). The `InvalidTokenError` type holds a kind parameter to describe the token's error level, which is either a warning or an error. 

Note: The import order is sorted because IDE does that automatically. I assume this is okay. 

[Related Issue](https://github.com/robotframework/robotframework/issues/5210)

### DIFF
--- a/src/robot/parsing/__init__.py
+++ b/src/robot/parsing/__init__.py
@@ -21,8 +21,21 @@ this module directly, issue should be submitted about exposing it explicitly via
 :mod:`robot.api.parsing`.
 """
 
-from .lexer import get_tokens, get_resource_tokens, get_init_tokens, Token
+from .lexer import (
+    ErrorCode,
+    ErrorKind,
+    InvalidTokenError,
+    Token,
+    get_init_tokens,
+    get_resource_tokens,
+    get_tokens,
+)
 from .model import File, ModelTransformer, ModelVisitor
-from .parser import get_model, get_resource_model, get_init_model
-from .suitestructure import (SuiteFile, SuiteDirectory, SuiteStructure,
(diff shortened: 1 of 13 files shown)


pekkaklarck:
Sorry for not reviewing this in timely manner @mikeleppane due to there being more important issues in RF 7.2 scope. In the end the deprecation of the pipe-separated format (#5203) what required this enhancement was also postponed to RF 7.3. There are currently no other benefits from this change, so I believe it's best to move #5210 (and this PR) to RF 7.3 as well.

There are some conflicts, but they seem to be easy to fix.

mikeleppane:
> Sorry for not reviewing this in timely manner @mikeleppane due to there being more important issues in RF 7.2 scope. In the end the deprecation of the pipe-separated format (#5203) what required this enhancement was also postponed to RF 7.3. There are currently no other benefits from this change, so I believe it's best to move #5210 (and this PR) to RF 7.3 as well.
> 
> There are some conflicts, but they seem to be easy to fix.

No problem. Okay, let's move it to RF 7.3. I will fix those conflicts when I have time. 

While working on this and experimenting with feature proposals, I realized that there is significant room for improvement in error messages and diagnostics to enhance the developer experience (DX). This goes beyond just addressing issue #5203. Consequently, I started a side project to rewrite the parsing logic in Rust with the aim of creating a tool similar to tidy for Robot. My two main objectives are to provide great error messages and achieve high performance.

PR #5265, opened 2024-11-14
Restore root logging level

### LINKED ISSUE #5262: `logging` module log level is not restored after execution
Hi,

It seems like that the robot handler is changing the root logger log level via ``set_level`` function (``robot.output.pyloggingconf``) but the original root logger level is not restored back after the end of the ``robot.running.model.TestSuite.run`` method or ``robot.run`` module.

The original context manager:

```python
@contextmanager
def robot_handler_enabled(level):
    root = logging.getLogger()
    if any(isinstance(h, RobotHandler) for h in root.handlers):
        yield
        return
    handler = RobotHandler()
    old_raise = logging.raiseExceptions
    root.addHandler(handler)
    logging.raiseExceptions = False
    set_level(level)
    try:
        yield
    finally:
        root.removeHandler(handler)
        logging.raiseExceptions = old_raise
```

Would it be necessary to restore the log level after changing it, in case the test script or any other third-party tool has already modified it for any reason?

```python
@contextmanager
def robot_handler_enabled(level):
    root = logging.getLogger()
    if any(isinstance(h, RobotHandler) for h in root.handlers):
        yield
        return
    handler = RobotHandler()
    old_raise = logging.raiseExceptions
    * -> old_level = logging.getLevelName(root.level)
    root.addHandler(handler)
    logging.raiseExceptions = False
    set_level(level)
    try:
        yield
    finally:
        root.removeHandler(handler)
        logging.raiseExceptions = old_raise
        * -> set_level(old_level)
```

theodore86:
Fixes #5262

### DIFF
--- a/src/robot/output/pyloggingconf.py
+++ b/src/robot/output/pyloggingconf.py
@@ -36,6 +36,7 @@ def robot_handler_enabled(level):
         return
     handler = RobotHandler()
     old_raise = logging.raiseExceptions
+    old_level = root.level
     root.addHandler(handler)
     logging.raiseExceptions = False
     set_level(level)
@@ -43,6 +44,7 @@ def robot_handler_enabled(level):
         yield
     finally:
         root.removeHandler(handler)
+        root.setLevel(old_level)
         logging.raiseExceptions = old_raise
 
 


theodore86:
@pekkaklarck just a gentle reminder to take a look at this PR when you have a moment. Let me know if there's anything I can clarify!

PR #5279, opened 2024-11-28
added body tree traverser to test model

Snooz82:


atest/resources/TestCheckerLibrary.py
```diff
@@ -14,48 +14,61 @@
 from robot.result.model import Body, Iterations
 from robot.utils.asserts import assert_equal
 
-
-class NoSlotsKeyword(Keyword):
+class BodyGetter:
```
pekkaklarck, line 17: `BodyGetter` is an ok name, but perhaps something like `WithDirectBodyAccess` could be a bit better. We've used `With` prefix also elsewhere with mixins and it looks good also when used like `NoSlotsFor(For, WithDirecBodyAccess)`.

Related to the above, a better name for `NoSlotsWhatever` would be good too. That `NoSlotsPrefix has never been that good, but now when these objects have also other functionality than not having `__slots__` it's even worse. Would an `Atest` prefix like `AtestFor` be fine?

atest/resources/TestCheckerLibrary.py
```diff
@@ -14,48 +14,61 @@
...
+        if isinstance(index, (int, slice)):
+            return self.body[index]
+        if isinstance(index, tuple):
+            item = self
+            for i in index:
+                item = item.body[int(i)]
+            return item
+        raise TypeError(...)
+
+class NoSlotsKeyword(Keyword, BodyGetter):
```
pekkaklarck, line 31: - Even though that `TypeError` should never happen, it would be good to have _some_ message with it. Perhaps `type(index).__name__` so that it shows what type the index had or just `repr(index)` to show the value in a format that often indicates also the type.
- An empty line missing before class.

atest/robot/cli/dryrun/if.robot
```diff
@@ -7,17 +7,17 @@ Resource          dryrun_resource.robot
 IF will not recurse in dry run
     ${tc}=    Check Test Case    ${TESTNAME}
     Check Branch Statuses    ${tc.body[0]}                            Recursive if         PASS
-    Check Branch Statuses    ${tc.body[0].body[0].body[0].body[0]}    Recursive if         NOT RUN
+    Check Branch Statuses    ${tc[0,0,0,0]}    Recursive if         NOT RUN
```
pekkaklarck, line 10: I believe that `0, 0, 0, 0` would be better than `0,0,0,0` but don't feel too strongly about that. We can discuss on Slack.

pekkaklarck (changes requested):
Looks great! Things to do:
- Let's agree should we use `0, 1, 0` or `0,1,1` style. We can use Slack.
- I added some ideas for possible better class names.
- Let's replace all `kws` occurrences with the direct body access as well. I believe it should be done separately after this PR is in, though.
- Let's replace all `msgs` usages either with the direct body access or with `messages` as well. This should be done at the same time as removing `kws`.

atest/resources/TestCheckerLibrary.py
Snooz82: i took repr(index) because it it would be a structure of something, it may be interesting what the types are.

pekkaklarck (approved):
Excellent!

Issue #4409, opened 2022-07-15
Test metadata


tomaspekarovic:
It would be great If we have keyword that will be able to set `test metadata` to test report.  This is not issue, when you are running tests using robot, but when you're running tests using pabot the `suite metadata` is not enough to cover all the information that should be visible at the end of the run.

Not mentioning that  rebot merge will mess  suite metadata  when they are added dynamically.

I'm testing insurance systems and when i'm running tests I get a lot of different URLs where insurances are created. Each testcase have its own url and I'm looking for way how to `expose` that url in the report.
For now I have custom keyword that logs this url to report, but it is not sufficient as you need to open test steps, open keyword and look for url there. So it should be really cool, if there are any  `test metadata`  that I can use to extract information and show them at on place in report

pekkaklarck:
I assume you mean name-value pairs like with suite metadata? Tests already have tags and documentation and I don't think them getting such metadata would add too much value. Could you clarify what benefits it would give over using tags and doc?

What we can consider is enhancing tags so that you could use them as name value pairs like `name: value`, `name = value`, or `name-value`. We could show such tags differently than normal tags in logs and reports and possibly add new functionality that operates only with tag names or values. We basically needed to decide what separator to use or should we make the separator configurable. I believe having a common separator would be best and like `name: value` the most. Anyway, this is something that should probably get a separate issue.

The main reason we added metadata, not tags, to suites was that we wanted to make it explicit that tags are always related to tests and metadata to suites. We have, however, already long time ago added keyword tags so nowadays it's not only tests that have tags. Another reason to add metadata for suites was that it's not clear that `Force Tags` and `Default Tags` settings are related to tests. We have recently made a decision to replace them with more explicit `Test Tags` so also that reason isn't so relevant anymore. I've been thinking we could thus change suites to also have tags, not metadata, in the future. That would require better support for using tags as name-value pairs as well as a very long deprecation period.

tomaspekarovic:
Hi, thanks for reply.
`Tags` - for me, tags are something that groups test cases into some sort of 'folders' (regression, quick, en/pl/ro tests ... ) 
Something that I can use for separate or select subset of tests I need to run. So from my point of view storing information should not be part of this. (but i can be wrong). This also makes report much more difficult to read, because if I use this, I will have 200 tags per full suite run.

`Documentation` is something descriptive about what test is doing. For example we're using it to **document which part of application is covered** with this tests and **little description** of test, if the name is not descriptive enough. 

[...]

PR #5281, opened 2024-11-29
Custom Test Settings a.k.a. Test Metadata

DanielSeim:
Hello Pekka,
here is the first PR regarding the ticket #4409 custom test settings a.k.a. test metadata.
Please have a look at it. Tests will follow soon for the metadata setting.
After that i will implement the custom settings: See also https://github.com/robotframework/robotframework/issues/4409#issuecomment-2338407700 bullet points 5 and 6.

Best regards,
Daniel Seim


pekkaklarck:
Thanks for the PR. In order to get RF 7.2 out still this year, we decided to move this functionality to RF 7.3 that we start developing right after the new year. Because I need to concentrate on 7.2, I won't have time to properly review this until it is in the release candidate phase. I took a quick look at the changes, though, and here are quick comments based on that:

- In general changes look good.
- The functionality to specify allowed metadata/setting names is missing. It is probably a good idea to implement it separately after the basic functionality is done, but it needs to be taken into account. It would be good to design it soon, but that can be discussion in the issue or on Slack.
- Plenty of tests are needed. Parser changes can be tested using unit tests, but otherwise this is easiest to test on the acceptance tests level.
- `metadata` should be after `tags` in `TestCase.__init__` to avoid breaking usages where tags are passed as a positional argument. It needed to be at the end of the argument list to be even safer, but then the position would be too inconsistent.
- XML and JSON schemas need to be updated.
- Documentation is needed. We are going to recreate the User Guide as part of RF 7.3, so perhaps this should wait until that's done.

LucianCrainic:
Hello @DanielSeim is there a way i can help you out with this functionality ? Do you need a hand on anything regarding ?

DanielSeim:
FYI i will change the loading of the custom settings to cli options to get rid of the toml dependency as clarified at robocon.

Noordsestern:
The PR has merge conflicts that must be resolved before further evaluation.

Noordsestern:
I think, this PR had been refactored after review in #5495

This PR can probably be closed.

Issue #5186, opened 2024-08-22
Expand environment variables in argumentfile


simonmeggle:
My request is to enable the expansion of environment variables inside of the argumentfile. 

**Background**: when RF reads additional path-like arguments from an argumentfile, arguments like `--pythonpath`  or `--output` are hardcoded. There is currently no way to make them dynamic. 

Writing this 

    --output $TESTBASEDIR/somefolder/output.xml

inside of an argumentfile would be awesome.
Thanks!

pekkaklarck:
This would probably be pretty easy to implement using [os.path.expandvars](https://docs.python.org/3/library/os.path.html#os.path.expandvars). As explained in its documentation, it works like this:
- `$name` and `${name}` formats are supported universally and on Windows also `%name%` works. This is convenient, because it's possible to use `$` variants to create OS-agnostic argument files, but on Windows the native format works too.
- Non-existing or malformed variables are left unchanged. This is good for backwards compatibility. Otherwise all current argument files containing `$` characters would be broken, but with these semantics there are problems only if there are accidental matches to existing environment variables.

Some examples assuming that `$HOME` contains `/usr/robot`:
```python
>>> os.path.expandvars('$HOME/path')
'/usr/robot/path'
>>> os.path.expandvars('$HOMER/path')
'$HOMER/path'
>>> os.path.expandvars('${HOME}R/path')
'/usr/robotR/path'
```
There are some design decisions to be solved still:
1. Do we need to care that they apparently is no way to escape variables so that they wouldn't be expanded? At least the docs didn't mention about that and with both `r'\$HOME/path'` and `$$HOME/path` the `$HOME` variable is replaced. If we needed to care about this, and there's no undocumented escaping functionality, we needed to implement something ourselves. That's likely a lot bigger task than this whole issue is otherwise.
2. Do we want to replace variables only in option values or should we do it everywhere? I was thinking the former would be better, but to my small surprise the shell replaces variables everywhere and, for example, if I set `$x` to `version` then `robot --$x` works fine. Based on that replacing variables everywhere would be fine and it certainly would simplify implementation even further.

Snooz82:
One concern is that escaping seems not to be possible.

pekkaklarck:
I also mentioned the lack of escaping in `os.path.expandvars` above. It's unlikely to cause issues often, but the problem is that there's no workaround if it causes them. For example, Bash supports escaping with `\` and by using single quotes:
```bash
$ echo \$HOME
$HOME
[...]

PR #5297, opened 2024-12-24
issue #5186 : native expanding env var for argumentfile

hassineabd:
added a cross platform solution to expand environment variables ( %VAR% on Windows, $VAR on Linux/macOS) using string.Template.safe_substitute

- %VAR% (windows) are converted to $VAR internally
- ($$) for escaping mechanism
- missing variables visibly unexpanded (not replaced with empty strings)
- temporarily includes verbose logging in _expand_arguments to facilitate debugging and testing

---> currently prints each variable's expansion or missing status for development and testing purpose. once the approach is confirmed, I will to refactor and reduce logging based on the review. so i have question: 

- logging strategy: should we log variables that are expanded? what about those who are escaped? is there a function that logs with configurable verbosity ? 

**Design Recap (since my comment was not accurate and done before getting involved**
- no flag (always expand: was a suggestion but abandoned)
- no tilde expansion : tilde expansion (~) is postponed ([slack discussion](https://robotframework.slack.com/archives/C3C28F9DF/p1734386590457609))
- caase normalization : robot.utils.NormalizedDict has been used to handle environment variables in a case-insensitive way on windows , but maintaining case sensitivity on linux/macOS
- escaping mechanism using  string.Template :  $$ for producing a literal $ (removing the need for a separate escaping system)
- os.path.expandvars has not been used at all: 
    * missing variables are silently dropped
    * no built in mechanism exists for escaping $

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from setuptools import find_packages, setup
 
 # Version number typically updated by running `invoke set-version <version>`.
 # Run `invoke --help set-version` or see tasks.py for details.
-VERSION = '7.2rc1'
+VERSION = '7.2rc2.dev1'
 with open(join(dirname(abspath(__file__)), 'README.rst')) as f:
     LONG_DESCRIPTION = f.read()
     base_url = 'https://github.com/robotframework/robotframework/blob/master'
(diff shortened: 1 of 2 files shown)


hassineabd:
anyone can review this PR please 
@pekkaklarck could you please help reviewing this draft

PR #5299, opened 2024-12-30
Remove python 3.8

franzhaas:
Python 3.8 is now end of life, and sooner or later we will drop it from the supported python versions.

This PR removes the 3.8 compatibility code and makes 3.9 the minimum python version supported.

Especially the hasattr of the types, and zipapp/importlib related code.

Removing 3.8 probably does not make sense for the  7.2 release, but I believe it will be soon enough for this PR to make sense.


franzhaas:
Sounds like a good plan to me.

pekkaklarck:
There are already some conflicts and I believe there are more of them before we actually remove Python 3.8 support. We'll also drop Python 3.9 at the same time, so the CI configuration done here needs to be updated. I don't think it makes sense to keep this PR open and try to maintain it. Let's look at this again later.

Issue #5122, opened 2024-04-26
Enhance documentation related to importing same library multiple times


oetzus:
According to the [user guide](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#library-scope), "If a library is imported multiple times with different arguments, a new instance is created every time regardless the scope."
But that does not seem to work for me.

Here is the example test case I use: 
``` robotframework
*** Settings ***
Library          testlib.py    num=1    AS    Test
Library          testlib.py    num=2    AS    Test
Library          testlib.py    num=3    AS    Test

*** Test Cases ***
Example Test
    Log Params
    BuiltIn.Import Library    ${CURDIR}/testlib.py    num=4    AS    Test
    Log Params
```
with the following example library:
``` python
from robot.api import logger

class testlib():

    ROBOT_LIBRARY_SCOPE = 'GLOBAL'

    def __init__(self, num):
        self.num = num

    def log_params(self):
        logger.console(f'num: {self.num}')
```
As I import the library each time with a different argument, I would expect, that a new instance of the library is created that overwrites the current instance. Instead, the very first instance of the library seems to be used as the output of both calls of `Log Params` is 1.

Did I understand the note in the user guide wrong or is this an unintended behaviour?

franzhaas:
When you try this you will see that there is a new instance created every time.

```python

from robot.api import logger

class testlib():

    ROBOT_LIBRARY_SCOPE = 'GLOBAL'

    def __init__(self, num):
        self.num = num
        logger.error(f"loading {num}")

    def log_params(self):
        logger.console(f'num: {self.num}')
```

Together with this robot file you can achieve what you are aiming for.:

```robot
*** Settings ***
Library          testlib.py    num=1      AS  test1
Library          testlib.py    num=2      AS  test2
Library          testlib.py    num=3      AS  test3    

*** Test Cases ***
Example Test
    test3.Log Params
    BuiltIn.Import Library    ${CURDIR}/testlib.py    num=4        AS  test4
    test4.Log Params
```

The documentation does not state explicitly which of those loaded library instances will be accessible via the name/be the global instance.

franzhaas:
@pekkaklarck can we put this in the 7.3 scope? I failed to do that myself... Sry

pekkaklarck:
We can try enhancing docs related to this in RF 7.3. Behavioral changes need to wait for RF 8.0.

franzhaas:
[...]

PR #5301, opened 2025-01-01
Improve test and doc repeated imports

franzhaas:
[this issue](https://github.com/robotframework/robotframework/issues/5122) reminded me that there is a wired situation that is a bit hard to understand and might benefit from a more verbose documentation

When you load the same library repeated times on the same "AS" name.
All of them get performed from top to button
But the initial one is accessible under the name.
The same applies when using Import Library

Doing this is in my point of view not a good idea. However, documenting this property would have helped me when analyzing/debugging in the past.

There is a risk by documenting this to make this behavior a requirement hindering future developments. In that case it would be a reasonable alternative to declare the behaviour undefined. The problem for me is that the sentence.:

If a library is imported multiple times with different arguments, a new instance is created every time regardless the scope.

can be misunderstood to imply that the created instance is also assigned.


franzhaas:
I noticed the added test was not run at all at CI. I fixed that. I also moved the test to the core tests.

doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
```diff
@@ -224,7 +224,9 @@ a string and it can have the following three values:
   modules are always global.
 
 .. note:: If a library is imported multiple times with different arguments__,
-          a new instance is created every time regardless the scope.
+          a new instance is created every time regardless the scope. If a 
+          library is imported multiple times using the same destination
+          the initial instance remains available unaffected by the new imports.
```
pekkaklarck, line 229: The proposed wording, especially "same destination" sounds a bit odd for me. Could it something like this be better:
> If a library is imported multiple times using same arguments, subsequent imports are ignored.

pekkaklarck (changes requested):
Are you sure the current behavior isn't already covered by tests? I would be surprised if it wasn't, but I can certainly be wrong.

See also my comment about the documentation change.

doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst
franzhaas: The problem is that subsequent imports do not get ignored. If they would get ignored, and ideally a warning generated, the situation would be much better.

They do happen, and the `__init__` code gets executed, which leaves robotframework in a Frankenstein situation where the global state of the last, and the object of the first import attempt to work together, which is confusing at least.

It gets worse when while debugging someone reads the logs buttom up. The newest `__init__` call is found first, and both global and object state are logged in a consistent state, adding to the confusion.

I think this is unintuitive, and I believe it should be clearly documented. 

If multiple destinations are used, the situation becomes much less confusing, as there are multiple objects, one of which is consistent with global state, the others are not. This bug behaves in robotframework the same way as it would in python, much easier to deal with.

If no global state is involved with multiple destinations, the behaviour is actually the way I would expect it to be.

franzhaas:
> Are you sure the current behavior isn't already covered by tests? I would be surprised if it wasn't, but I can certainly be wrong.
> 

I do not know. I did not spend a lot of time looking for preexisitng tests for this topic,  but I did not stumble on them.

Issue #4173, opened 2021-12-17
Process: Avoid deadlock when standard streams are not redirected to files


jonilam:
One of our test cases started to hang after switching from Run And Return Rc And Output-keyword to a Run Process-keyword.  We debugged the issue and noticed that one of the threads of our test application was stuck in fflush(). Issue is caused by the usage of  Popen.wait() + read in Process-library. There is a note about this in a python subprocess-documentation:

"This will deadlock when using stdout=PIPE or stderr=PIPE and the child process generates enough output to a pipe such that it blocks waiting for the OS pipe buffer to accept more data. Use Popen.communicate() when using pipes to avoid that."
https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

As stated by the note above, this happens only when pipes are used, and can be avoided by writing stdout/stderr to a file. I verified that indeed using Popen.communicate() instead of Popen.wait() fixes the problem.

Noticed the issue with RF 3.1.1 but tested that the issue is still there in version 4.1.3.

pekkaklarck:
I don't remember why exactly we use `wait` and not `communicate`. We possibly could change that if the user hasn't provided `stdout` and `stderr`. That's the default so it could avoid hanging also in other cases.

Did you test does redirecting `stdout/err` to a file fox the problem? If it does, that's a good workaround.

jonilam:
Yes, we are currently using stdout/err redirecting to a file as a workaround and it is working without issues.

pekkaklarck:
Good to know that redirecting to files works. It would still be good to investigate could we use `communicate` when `stdout` and `stderr` are PIPEs (which is the default) to avoid deadlocking in cases like this. A problem is that although we call `wait()`, we only do that after elsewhere calling `poll()` to verify that the process has stopped and my assumption is that it hangs already there. It's thus possible we needed to restructure the code a bit more. Good thing is that we have lot of tests so even bigger changes can be done safely.

Unfortunately I'm super busy and don't have time to look at this in the near future. I leave this issue open in case someone else is interested to look at it.

pekkaklarck:
PR #5302 implements this.

franzhaas:
> PR #5302 implements this.

@pekkaklarck can we put this into the 7.3 scope?

I am sorry but I failed to do this myself

franzhaas:
this is also related to this issue.: https://github.com/robotframework/robotframework/issues/5345

pekkaklarck:
[...]

PR #5302, opened 2025-01-01
Popen move from wait to communicate to avoid deadlock

franzhaas:
[see here](https://github.com/robotframework/robotframework/issues/4173)


franzhaas:
I do not understand why the second last run failed, or why the last changes fixed it...

Thus said, I believe this is good enough for review now.

pekkaklarck:
Switching from `wait` to `communicate` is great. Some quick comments:
- The added test is currently not run at all, there needs to be a test on the `atest/robot` side for running it. It probably could be added to some existing suite and also the test under `atest/testdata` moved to some existing suite. See `atest/README.rst` for details and ask help on #devel if needed.
- This can be added earliest in RF 7.3. I'll add #4173 to its scope.
- The code needs to be cleaned up. I'll review it properly once the test is added and we start RF 7.3 development.

franzhaas:
I added the test and made the change more invasive, leading to clearer code. I documented the two spots with the most wiredness...

src/robot/libraries/Process.py
```diff
@@ -533,7 +533,15 @@ def _wait(self, process):
         # wait function was used.
         _stdin = process.stdin
         process.stdin = process.stdin if (process.stdin and not process.stdin.closed) else io.StringIO("")
-        (_stdout, _stderr,) = process.communicate()
+        while True:
+            try:
```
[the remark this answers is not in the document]
franzhaas: donne

franzhaas:
@pekkaklarck can we up-prioritise this PR? It solves 2 issues, the deadlock and the timeout issue.

I tried to balance the intrusiveness of the patch with code quality, but maybe some more aggressive usage of "communicate" can simplify this code for the cost of an invasive changeset. Maybe this is welcome in the long run, maybe for the 8.0 target?

pekkaklarck:
I tested the code and it fixes both #4173 and #5345 which is awesome. The code is rather complicated, though, and there are some style issues like accessing private attributes of the `ExecutionResult` object. While testing the implementation locally, I was able to simplify the code a lot. Now the options are:

1. I write a code review so that you can fix the issues.
2. I update the PR with my changes and merge it.
3. I simply commit my changes and close this PR.

Because the fix is based on your solution, you'd be acknowledged in the release notes in all these cases.

src/robot/libraries/Process.py
```diff
@@ -524,8 +526,37 @@ def _manage_process_timeout(self, handle, on_timeout):
 
     def _wait(self, process):
         result = self._results[process]
-        result.rc = process.wait() or 0
-        result.close_streams()
+
+        _take_stdout = (process.stdout and (not process.stdout.closed))
+        _take_stderr = (process.stderr and (not process.stderr.closed))
+
```
pekkaklarck, line 532: Style:
- We don't generally use empty line inside methods. If methods get so long that they need to be grouped by empty lines, we prefer introducing helper methods. In this particular case empty lines could just be removed.
- We don't use `_` prefix with local variables.

src/robot/libraries/Process.py
```diff
@@ -524,8 +526,37 @@ def _manage_process_timeout(self, handle, on_timeout):
...
+        # This is needed to handle stdin closes identically to how it was when the
+        # wait function was used.
+        _stdin = process.stdin
+        process.stdin = process.stdin if (process.stdin and not process.stdin.closed) else io.StringIO("")
```
pekkaklarck, line 536: It's true that `Popen.communicate()` doesn't like `stdin` to be a closed PIPE, but there can also be problems if `stdout` and `stderr` are closed PIPEs. How they are all handled is pretty inconsistent and I submitted a bug about that:
https://github.com/python/cpython/issues/131064

To be consistent with our old `Popen.wait()` based code, we need to handle all streams. Closed streams can be replaced with `None` instead `io.StringIO("")`, though.

src/robot/libraries/Process.py
```diff
@@ -524,8 +526,37 @@ def _manage_process_timeout(self, handle, on_timeout):
...
+        while True:
+            with contextlib.suppress(subprocess.TimeoutExpired):
+                # this is to allow the process to be aborted when on windows
+                # an exception is used to mange the test timeout situation
+                (_stdout, _stderr,) = process.communicate(timeout=0.1)
+                break
```
pekkaklarck, line 542: This isn't related to #4173 anymore and should rather be submitted as a separate PR.

In my opinion `try/except` would make the code easier to understand that `contextlib.suppress`.

src/robot/libraries/Process.py
```diff
@@ -524,8 +526,37 @@ def _manage_process_timeout(self, handle, on_timeout):
...
+        # wait function was used.
+        if _stdout:
+            result._stdout = result._format_output(_stdout if _take_stdout else "")
+        
+        if _stderr:
+            result._stderr = result._format_output(_stderr if _take_stderr else "")
+
+        for f in [process.stdout, process.stderr] + result._custom_streams:
+            with contextlib.suppress(AttributeError):
+                f.close()
```
pekkaklarck, line 558: Accessing `_private` members of the `result` object like the above code does is not ok. Making `result.stdout` property settable is a better solution.

pekkaklarck (changes requested):

pekkaklarck:
I looked at this again, and noticed that the change broke a case where only `stdout` or `stderr` is a `PIPE` and it is closed externally. For example, the following test succeeds with the old code but now with this one:
```robotframework
*** Test Cases ***
Example
    ${p}=    Start process    python    -V    stderr=DEVNULL
    Call method    ${p.stdout}    close
    ${r}=    Wait for process
    Log to console    ${r.stdout}
```
I have fixed the problem locally and added a test for it. I've also fixed other issues I mentioned in the review, and it would probably be easiest if I committed my changes and closed this PR. I'll wait until tomorrow with that in case you want to fix the issues yourself.

Regardless how we proceed, big thanks for the PR. It proved that the deadlock with PIPEs can be avoided.

franzhaas:
Sorry, I was unavailable the last few days. Sure, go ahead and commit + push if you already have the changes available.

Most of the things mentioned here I have no strong opinion about, except for the 'close stdout' issue, which should be fixed.

pekkaklarck:
I committed my changes in f2f2f99ad43825162109ae83e7e6886bcdfa60ba. Notice that those changes don't contain the fix for timeouts (#5345) but I'll fix that separately.

PR #5307, opened 2025-01-06
Fix confusing error message when rebot with empty suite results

jnhyperion:
* empty robot file: `test.robot`
* run `robot --runemptysuite test.robot`
* Got "output.xml"
* run `rebot output.xml`

Before fix:

[ ERROR ] Suite 'Test' contains no tests after model modifiers.

Try --help for usage information.

After fix:
[ ERROR ] Suite 'Test' contains no tests.

Try --help for usage information.

Because user doesn't input any rebot pre run modifiers, the error message ` Suite 'Test' contains no tests after model modifiers.` is very confusing, and the pre run modifier should not be called as well.

### DIFF
--- a/src/robot/reporting/resultwriter.py
+++ b/src/robot/reporting/resultwriter.py
@@ -115,10 +115,11 @@ class Results:
                                            *self._sources)
             if self._settings.rpa is None:
                 self._settings.rpa = self._result.rpa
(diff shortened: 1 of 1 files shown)


pekkaklarck:
The change makes sense but there should also be a test for it. We also want to have issues for tracking purposes unless changes are trivial like typo fixes. I'll handle both of these myself this time to get this into RF 7.2.

PR #5315, opened 2025-01-11
Debugfile available for threads/processes/async

franzhaas:
This is possible by using a single process which accepts write requests via a queue.

This lock and thread free design brings the following advantages over threads.:

- threads, fork, async, subprocess and in the future also subinterpreter based parallelism is possible.

As soon as a single thread is started, Fork and fork based subprocesses are no longer an option.

Whole classes of bugs are not aplicable
- life lock
- dead lock

Outlook.:
subinterpreter and subprocesses on windows are on the todo list.


franzhaas:
It has to be noted that.:

spawn based subprocessing and subinterpreter based concurrency is not (yet) supported.
asyncio is supported starting with python 3.10 and newer, if needed I probably could make it compatible with 3.9, but I believe this is not productive.

I did check if isubprocesses work with the forkserver (expected 3.14 default subprocess start method for linux) start method, and it did work for me.

pekkaklarck:
This looks _really_ complicated. If you want this to be merged, there need to _very_ big benefits. If I got it right, the benefit is allowing logging to the debugfile whem using threads, and I'm not convinced it is a big benefit enough.

It would be good if you'd create an issue where you describe the problem and explain the proposed fix in high level. We can then discuss the design and possible other alternatives.

franzhaas:
@pekkaklarck 

Thanks for the feedback. I agree that discussion is needed, and I will make a ticket for that.

It's not just threads. It also works with async and subprocesses (currently Linux only), Subinterpreters are also on the todo list.

PR #5324, opened 2025-01-23
Fix `Lists should be equal` with `ignore_case` and `ignore_order`

### LINKED ISSUE #5321: `Lists Should Be Equal` does not work as expected with `ignore_case` and `ignore_order` arguments
Looks like 'Lists Should Be Equal' does not work in one of the scenarios as mentioned below

```
${list1}    Create List    AbC    AEF
${list2}    Create List    AEF    ABC
Lists Should Be Equal    ${list1}    ${list2}    ignore_case=${True}    ignore_order=${True}
```
We are expecting this to be passed since the lists are same as we are ignoring the order and the case. However, It is failing. 

Can someone explain reason behind it please? Does it have something to do with ASCII codes

I would be glad to provide additional details if required. Thanks

m-usmani:
Fixes #5321 
Moved the step to normalize the case before sorting the list.

### DIFF
--- a/atest/testdata/standard_libraries/collections/list.robot
+++ b/atest/testdata/standard_libraries/collections/list.robot
@@ -672,6 +672,12 @@ List Should Not Contain Duplicates With Ignore Case
 List Should Contain Value With Ignore Case And Nested List and Dictionary
     [Setup]    Create Lists For Testing Ignore Case
     List Should Contain Value  ${L4}  value=d    ignore_case=${True}
+    
+Lists Should be equal with Ignore Case and Order
+    [Setup]    Create Lists For Testing Ignore Case
+    [Template]    Lists Should Be Equal
+    list1=${L7}    list2=${L8}    ignore_order=${True}    ignore_case=${True}
+    list1=${L9}    list2=${L10}    ignore_order=${True}    ignore_case=${True}
 
 *** Keywords ***
 Validate invalid argument error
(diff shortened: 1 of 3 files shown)


atest/testdata/standard_libraries/collections/list.robot
```diff
@@ -672,6 +672,12 @@ List Should Not Contain Duplicates With Ignore Case
 List Should Contain Value With Ignore Case And Nested List and Dictionary
     [Setup]    Create Lists For Testing Ignore Case
     List Should Contain Value  ${L4}  value=d    ignore_case=${True}
+    
+Lists Should be equal with Ignore Case and Order
+    [Setup]    Create Lists For Testing Ignore Case
+    [Template]    Lists Should Be Equal
+    list1=${L7}    list2=${L8}    ignore_order=${True}    ignore_case=${True}
+    list1=${L9}    list2=${L10}    ignore_order=${True}    ignore_case=${True}
```
pekkaklarck, line 680: Two comments mostly related to style.

1. Instead of creating these lists that are used just once in a separate keyword, it would probably be simpler to create them here using inline Python evaluation like `${{['apple', 'Banana', 'cherry']}}`. That syntax is somewhat complicated, but here it would allow seeing compared values directly.

2. There's no need to use e.g. `ignore_order=${True}` when simpler `ignore_order=True` works. 

Both of the above notes apply also to many other tests and changing them as part of this PR isn't a good idea. I think I'll merge this now and we can consider cleaning up all tests separately later.

pekkaklarck (approved):

PR #5333, opened 2025-02-07
Styling dialogs

Alpha-Centauri-00:
i made some new styling for Dialogs library.. 
i was trying to get the same style as robotframework.org .. and i think it needs more marging for some inputs but that should not be a big thing.. i "hard coded" colors in init to change it later if nessessery .. 

i did not change any functionality of the keywords.. only the colors and buttons ... i test it on windows and linux (debian) only. i wish its gonna work with mac.. 

image

image

image

image


franzhaas:
while not related on a technical level, it still touches the same code base and requires the same manual tests.:

https://github.com/robotframework/robotframework/issues/5335

Alpha-Centauri-00:
the mac issue is solved now and its working after my last commit. but still looks very wird on mac as you can see in the photo below. 

we'll see how we can make a work around with Andre on mac ...

image

Alpha-Centauri-00:
Due to compatibility issues with macOS, the canvas element is not functioning as expected for styling buttons. To address this, alternative technologies can be employed to achieve the desired styling. If left unchanged, the default appearance of the button remains decent, and also the color modification to the button on macOS is not possible, unless using ttk or customtkinter libraries. 

for now i think mac is okay

Alpha-Centauri-00:
Styling apply now on windows and linux. mac is going to show the old default styling. a big thanks to @AndreMochinin for helping testing this on mac.

pekkaklarck:
I finally had time to look at this properly. I tested the changes locally and also looked the code quickly. Below is an example showing the same dialog first with the current styles and then with the styles from this PR.

Kuvakaappaus 2025-03-22 11-11-09

As I've written earlier, in general I like the idea of enhancing the styles and using the official Robot Framework colors. In my opinion this gets a bit too far, though, and there are also some technical problems (bugs) as well as . I'll clarify my concerns below.

## Style issues

1. I don't understand why the normal buttons for minimizing, resizing and closing dialogs are removed. Are the some concrete benefits from that? I can see them being useful in some cases.
2. Dialog title not being centered makes it harder to separate it from the dialog text.
3. Input widget not using all the horizontal space is odd.
4. Buttons are unnecessary small.
5. Overall I'd prefer the dialogs to look more like platform native dialogs. 

## Bugs

1. Dialogs aren't closed properly at least on Linux. The dialog itself goes away, but the entry in the panel where open applications are shown stays until the execution ends. The panel gets _very_ crowded if you use multiple dialogs.
2. The aforementioned panel entry only shows `tk` as the text when it used to show `Robot Framework`.
3. The dialog disappears when I move it down or right from the original position. It reappears when I release the mouse, vanishing while dragging is annoying. The dialog can also flash.
4. Dialogs not working the same on all platforms isn't acceptable. I don't like it in general, but the bigger problems is that further changes need to take different platforms into account.

## Code issues

The changes look pretty complicated and that's likely the reason for some of the above bugs. Even if the bugs were fixed, complicated code is a huge problem from maintainability point of view.

pekkaklarck:
Due to the problems explained my previous comment, I don't think this big changes make sense. Even if the bugs and style issues would be fixed, I don't consider the benefits big enough compared to the maintenance issues caused by the added complex code. The code not working on OSX also makes me worried that further changes to tkinter can break it also on other platforms.

I'd still be happy with the dialogs being enhanced a bit, but changes should have the following characteristics:
- Must work on all our supported platforms.
- No platform specific conditional logic unless the benefits are really big.
- No complex logic unless benefits are extremely big. This avoids the code being broken if tkinter changes and eases maintenance. As an example, I don't consider rounded corners important enough to warrant much extra code.
- Bugs mentioned in the previous comment fixed.

Now that I think about this, very simple changes like using a custom color in the dialog title background could be enough. If you want fancier dialogs, you could also create an external project for that. It could then have additional dependencies, didn't need to support all OSes, and so on.

pekkaklarck:
The complexity and the problems caused by it aren't worth the benefits and I'll close this PR. Simpler enhancements are still possible. I [commented](https://github.com/robotframework/robotframework/issues/5334#issuecomment-2747965906) the issue #5334 about things that I think can be changed.

PR #5341, opened 2025-02-16
Update languages.py

### LINKED ISSUE #5150: Enhance BDD support (GIVEN/WHEN/THEN) for French language
Hi,

According to the Gherkin Syntax, there's a lot of ways to translate the BDD keywords in French.
Cf: https://cucumber.io/docs/gherkin/languages/

As an exemple, GIVEN => 16 possibilities

image

Currently, only one seems implemented in Robot Framework:  'Étant donné'

This has impacts:
- emphasis on capital letter is not easy to type
- this involves sometimes to midy the name of the step so that the general meaning of the sentence is correct

To resolves this, will it be possible to implements all the possibilities ?
Or at least  "Etant donné",  "Etant donné que", "Étant donné que".

Regards

### LINKED ISSUE #5340: BDD prefixes with same beginning are not handled properly
**Description**:
When defining custom language prefixes for BDD steps in the Robot Framework (for example, in a French language module), I observed that adding multiple prefixes with overlapping substrings (e.g., "Sachant que" and "Sachant") leads to intermittent failures in matching the intended BDD step.

**Steps to Reproduce:**

Create a custom language (e.g., a French language class) inheriting from Language.
Define given_prefixes with overlapping entries such as:

`given_prefixes = ['Étant donné', 'Soit', 'Sachant que', 'Sachant']`

Run tests that rely on detecting BDD step prefixes.
Occasionally, tests fail with errors indicating that the keyword "Sachant que" cannot be found, even though it is defined.
Expected Behavior:
The regex built from the BDD prefixes should consistently match the longer, more specific prefix ("Sachant que") instead of prematurely matching the shorter prefix ("Sachant").

**Observed Behavior:**
Due to the unordered nature of Python sets, the bdd_prefixes property (which is constructed using a set) produces a regex with the alternatives in an arbitrary order. This sometimes results in the shorter prefix ("Sachant") matching before the longer one ("Sachant que"), causing intermittent failures.

**Potential Impact:**

Unpredictable test failures when multiple BDD prefixes with overlapping substrings are used.
Difficulty in debugging issues due to the non-deterministic nature of the problem.
Suggested Fix:
Sort the BDD prefixes by descending length before constructing the regular expression. For example:

```
@property
def bdd_prefix_regexp(self):
    if not self._bdd_prefix_regexp:
        # Sort prefixes by descending length so that the longest ones are matched first
        prefixes = sorted(self.bdd_prefixes, key=len, reverse=True)
        pattern = '|'.join(prefix.replace(' ', r'\s') for prefix in prefixes).lower()
        self._bdd_prefix_regexp = re.compile(rf'({pattern})\s', re.IGNORECASE)
    return self._bdd_prefix_regexp

```
This change would ensure that longer, more specific prefixes are matched before their shorter substrings, eliminating the intermittent failure.

**Environment:**

Robot Framework version: 7.1.1
Python version: 3.11.5

**Additional Notes:**
This bug appears to be caused by the unordered nature of sets in Python, which is used in the bdd_prefixes property to combine all prefixes. The issue might not manifest when there are only three prefixes or when there is no overlapping substring scenario.

I hope this detailed report helps in reproducing and resolving the issue.

orenault:
Fix bug on BDD prefix matching failure and add some Gherkin prefix for french localization.

Fixes #5150   (add French Gherkin Localization)
Fixes #5340  (fix bug on bdd prefix matching)

### DIFF
--- a/src/robot/conf/languages.py
+++ b/src/robot/conf/languages.py
@@ -63,8 +63,9 @@ class Languages:
     @property
     def bdd_prefix_regexp(self):
         if not self._bdd_prefix_regexp:
-            prefixes = '|'.join(self.bdd_prefixes).replace(' ', r'\s').lower()
-            self._bdd_prefix_regexp = re.compile(rf'({prefixes})\s', re.IGNORECASE)
+            prefixes = sorted(self.bdd_prefixes, key=len, reverse=True)
+            pattern = '|'.join(prefix.replace(' ', r'\s') for prefix in prefixes).lower()
+            self._bdd_prefix_regexp = re.compile(rf'({pattern})\s', re.IGNORECASE)
         return self._bdd_prefix_regexp
 
     def reset(self, languages: Iterable[LanguageLike] = (), add_english: bool = True):
@@ -556,11 +557,11 @@ class Fr(Language):
     template_setting = 'Modèle'
     timeout_setting = "Délai d'attente"
     arguments_setting = 'Arguments'
-    given_prefixes = ['Étant donné']
-    when_prefixes = ['Lorsque']
-    then_prefixes = ['Alors']
-    and_prefixes = ['Et']
-    but_prefixes = ['Mais']
+    given_prefixes = ['Étant donné', 'Étant donné que', "Étant donné qu'", 'Soit', 'Sachant que', "Sachant qu'", 'Sachant', 'Etant donné', 'Etant donné que', "Etant donné qu'", 'Etant donnée', 'Etant données']
+    when_prefixes = ['Lorsque', 'Quand', "Lorsqu'"]
+    then_prefixes = ['Alors', 'Donc']
+    and_prefixes = ['Et', 'Et que', "Et qu'"]
+    but_prefixes = ['Mais', 'Mais que', "Mais qu'"]
     true_strings = ['Vrai', 'Oui', 'Actif']
     false_strings = ['Faux', 'Non', 'Désactivé', 'Aucun']
 


src/robot/conf/languages.py
```diff
@@ -556,11 +557,11 @@ class Fr(Language):
     template_setting = 'Modèle'
     timeout_setting = "Délai d'attente"
     arguments_setting = 'Arguments'
-    given_prefixes = ['Étant donné']
-    when_prefixes = ['Lorsque']
-    then_prefixes = ['Alors']
-    and_prefixes = ['Et']
-    but_prefixes = ['Mais']
+    given_prefixes = ['Étant donné', 'Étant donné que', "Étant donné qu'", 'Soit', 'Sachant que', "Sachant qu'", 'Sachant', 'Etant donné', 'Etant donné que', "Etant donné qu'", 'Etant donnée', 'Etant données']
```
pekkaklarck, line 560: Please split this to max 88 chars.

src/robot/conf/languages.py
```diff
@@ -63,8 +63,9 @@ def __init__(self, languages: 'Iterable[LanguageLike]|LanguageLike|None' = (),
     @property
     def bdd_prefix_regexp(self):
         if not self._bdd_prefix_regexp:
-            prefixes = '|'.join(self.bdd_prefixes).replace(' ', r'\s').lower()
-            self._bdd_prefix_regexp = re.compile(rf'({prefixes})\s', re.IGNORECASE)
+            prefixes = sorted(self.bdd_prefixes, key=len, reverse=True)
+            pattern = '|'.join(prefix.replace(' ', r'\s') for prefix in prefixes).lower()
+            self._bdd_prefix_regexp = re.compile(rf'({pattern})\s', re.IGNORECASE)
```
pekkaklarck, line 68: It would be good to add a unit test for this fix in `utest/api/test_languages.py `.

pekkaklarck (approved):
Looks very good. Changes aren't exactly related, so it would be better to have separate PRs. Because changes are so simple, splitting the PR now doesn't have any real benefits, though. 

See the inline comments for few other notes. I can fix those issues myself as well if needed.

Issue #5335, opened 2025-02-10
tkinter thread issues on windows


franzhaas:
TkInter on windows is not thread safe, when starting robotframework in a nontypical thread (For example in its own dedicated thread) wired things happen which do not happen on linux.

I want to use multiprocessing.Process to shift the tkinter calls into their process getting the threading topic in this instance completely out of robot framework.

By limiting multiprocessing to only use forkserver or spawn, we are safe from deadlock side effects introduced by the combination of threads and fork.

I am willing to provide a PR for this issue.

The drawback is an increase in library load time, and a slight overhead when using the dialogs, however, compared to human interaction speeds this is not an issue.

this would also solve for example this issue.: https://github.com/test-fullautomation/robotframework/issues/110

pekkaklarck:
This is a tk issue, right? If there's a simple workaround, we certainly can consider it, but using multiprocessing sounds pretty complicated.

franzhaas:
Well in order to use multiprocessing we need to do a view steps, but these steps are quite simple.

For this example to work the implementation of pause_execution needs to be moved to dialogs_py.py

@pekkaklarck what do you think about this approach?
 - Not  a single control structure added
 - All added controll flow is linear
 - New function average length 3 Lines
 - heavily uses the standard library

Maybe we can use dynamic keyword generation to reduce the boilerplate?

```python
import importlib
import multiprocessing as mp
import concurrent.futures
from functools import wraps

__version__ = get_version()
__all__ = ['execute_manual_step', 'get_value_from_user',
           'get_selection_from_user', 'pause_execution', 'get_selections_from_user']

_ctx = mp.get_context("spawn") # spawn is thread safe (forkserver would also work, but is not available on windows, fork is not thread safe)
_pool = concurrent.futures.ProcessPoolExecutor(max_workers=1, mp_context=_ctx) # only one worker, this makes calls sequential, eventhough parallel calls would be possible

def _process_worker(fun_name, args, kwargs):
    m = importlib.import_module("robot.libraries.dialogs_py") 
    # dialogs_py is only loaded in target process, 
    # tkinter is not loaded before the function is
    # executed.
    #
    # This avoids the problem of tkinter not beeing pickleable
    return getattr(m, fun_name)(*args, **kwargs)

def run_in_process(fun):
    @wraps(fun)
    def wrapper(*args, **kwargs):
        return _pool.submit(_process_worker, fun.__name__, args, kwargs).result()
    return wrapper

@run_in_process
[...]

PR #5343, opened 2025-02-18
Tkinter any thread

franzhaas:
A suggestion to make Dialogs operational independent of the thread robotframework runs out of.

This is important in this use case.:

I have a gui using tkInter, running in "main thread"
I have a file handle used by this gui, and also going to be used in robotframework (otherwise we could start a new process)
I have a robotframework task which uses this file handle and Dialogs

The reasonable solution here would be to run robotframework in another thread, while "main thread" continues to run tkinter. However tkinter is not multithread capable on windows. This PR changes that. Due to the usage of spawn this solution is multithread safe.

see also https://github.com/robotframework/robotframework/issues/5335

### DIFF
--- a/src/robot/libraries/Dialogs.py
+++ b/src/robot/libraries/Dialogs.py
@@ -51,8 +51,8 @@ def run_in_process(fun):
     return wrapper
 
 # the acutall functions are implemented here
-_dialogs_py = importlib.import_module('robot.libraries._Dialogs')
+_Dialogs = importlib.import_module('robot.libraries._Dialogs')
 
 
 for name in __all__:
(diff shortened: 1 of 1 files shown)


franzhaas:
The latest changes make sure that.:

 tkinter never gets loaded into memory by robotframework in the "main" process/thread _no interference with other tk users is possible_.

The price for this is.:
- a lot of code got moved
- some modules get loaded using importlib instead of import 
  - harming static code analysis
  - introduces a lot of small code changes
- a view very compact linear functions got added
- there is a small change in load time, faster on my machine
  - no loading of tkinter (~20ms saved)
  - instantiation of a processpool (~9ms spent)

PR #5356, opened 2025-03-05
add arabic language 

hassineabd:
- added arabic language support


src/robot/conf/languages.py
```diff
@@ -1179,6 +1179,7 @@ class It(Language):
     and_prefixes = ['E']
     but_prefixes = ['Ma']
     true_strings = ['Vero', 'Sì', 'On']
+    teardown_setting = 'Hậu thiết lập'
```
pekkaklarck, line 1182: This addition, and the removal of the same line from the `Vi` class seems to be a mistake. Please revert.

src/robot/conf/languages.py
```diff
@@ -1356,3 +1356,47 @@ class Ko(Language):
     but_prefixes = ['하지만']
     true_strings = ['참', '네', '켜기']
     false_strings = ['거짓', '아니오', '끄기']
+
+class Ar(Language):
+    """Arabic
+
+    New in Robot Framework X.Y.
```
pekkaklarck, line 1363: You can change `X.Y` this to `7.3` because that's the next release (unless surprises happen).

Please also add second empty line before the `class Ar(Language):` line to follow our coding conventions. Adding a newline at the end of the file would be nice too.

utest/api/test_languages.py
```diff
@@ -53,7 +53,7 @@ def test_standard_languages_have_code_and_name(self):
             assert cls.name
 
     def test_standard_language_doc_formatting(self):
-        added_in_rf60 = {'bg', 'bs', 'cs', 'de', 'en', 'es', 'fi', 'fr', 'hi',
+        added_in_rf60 = {'ar', 'bg', 'bs', 'cs', 'de', 'en', 'es', 'fi', 'fr', 'hi',
```
pekkaklarck, line 56: Please remove `ar` here. This set contains the languages that were added in RF 6.0 when the localization support was implemented. The test verifies that their docstring contains only the language name and that all the others contain the `New in Robot Framework ...` note.

pekkaklarck (changes requested):
Looks very good but there are few small things to fix.

pekkaklarck:
Thanks for the contribution! I'll create an issue to track adding Arabic and possible other new languages in RF 7.3.

Arabic is the first right-to-left language to be added. How does it work in general? Are there other changes we could do to make the experiment better?

pekkaklarck:
@leeuwe, should Arabic be added to Crowdin as well?

leeuwe:
Arabic was already in Crowdin, but in 2022, we decided not to merge, since there are a few other issues:
https://github.com/robotframework/robotframework/pull/4507#issuecomment-1278177727 
I believe the left to right support for Log.html and Report.html is not resolved yet - maybe there isn't even an issue about it.

pekkaklarck:
Although we didn't add Arabic earlier due to right-to-left issues, I believe could add it now and then try to fix the remaining issues. At least I see on harm with that. How do you @hassineabd feel about that?

Can you @leeuwe provide a link to the translations in Crowdin? @hassineabd could possibly check what has been proposed earlier.

hassineabd:
@pekkaklarck  

My point of view of adoption of as many language as we can (including arabic): 

1. When non-native english speakers must process technical concepts in English, they experience additional cognitive burden that native speakers don't face. This can slow learning and adoption, and the impact is very significant, especially for beginners who are trying to get started (there are many research papers on this topic). Robot Framework's core value is being beginner-friendly, simple, and accessible, so I see that this aligns well.

2. I think that all arab speaking people/ countries have English as a second or third language, but Arabic is the only official language (as far as I know). In educational settings, government institutions, and enterprises, the officially accepted documents are in Arabic. For governmental projects ( there is many of them who adopt robot framework ) , having the ability to ,natively code and present test results and documentation in Arabic significantly improves stakeholder buy-in and prevents the need for an additional layer of translation.


On a deeper level, I'd like to share a broader context that helps in understanding why this contribution matters:
There are many initiatives currently working on making science and technology popular by simplifying concepts and explaining complex things in an easy way with accessible language. There are many such initiatives in the Arabic-speaking world (I Believe in Science, El Academia Post, Syrian Researchers, Arageek, etc.)
Yassin Khadraoui (PhD and renowned Wiki editor who is part of many initiatives, and who helped proofread the translations as well) and I believe in this. We believe also that localization and accurate translation are at the core of it and cannot make science and technology popular without it.

hassineabd:
This how it looks like on the vscode and on the logs 

image

image


I believe that it can be good first step for supporting RTL languages , it is understandable and common sense display ( since that mixing RTL and LTR is common behavior or already for RTL speakers as i know ) 

further support could be adding RTL ability in the HTML logs and reports

PR #5363, opened 2025-03-12
Keywords in process

franzhaas:
This is a proposal for a decorator which shifts the execution of a keyword into a dedicated process

This brings the advantage that processes _CAN_ be killed, which allows timeouts to work in situations where they previously did not, and also multiple code entities run as "the main thread".

The keywords running in this process environment are currently limited to not communicate with other libraries, and if they want to communicate with robotframework it is currently limited to the logging API of robot.api.

Examples where this is used at the moment are.:
 - OperatingSystem get file
 - Dialogs
 
There is currently the questionable feature that _IF_ the process in which the keyword is running gets killed (for example due to timeout), the children of this process do not necessarily get killed as well.

There are limitations on what can be passed in and out of those keywords, only pickleable data can be exchanged. For example files can not be shared.

The processes get started using "spawn" to avoid conflicts with threading.

### DIFF
--- a/.github/workflows/acceptance_tests_cpython.yml
+++ b/.github/workflows/acceptance_tests_cpython.yml
@@ -6,6 +6,7 @@ on:
       - main
       - master
       - v*-maintenance
+      - keywords_in_process
     paths:
       - '.github/workflows/**'
       - 'src/**'
--- a/atest/robot/standard_libraries/operating_system/get_file.robot
+++ b/atest/robot/standard_libraries/operating_system/get_file.robot
@@ -135,3 +135,6 @@ Grep File With Windows line endings
 
 Path as `pathlib.Path`
     Check Test Case    ${TESTNAME}
+
+read from fifo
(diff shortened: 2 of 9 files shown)


pekkaklarck:
Using `multiprocessing` to run functions so that they can be stopped by Robot's timeouts is an interesting idea and this implementation is pretty impressive, but I see also plenty of problems with it:

- The code is _very_ complicated. Adding something this complex to the core, where it would need to be maintained going forward, would need to have really big benefits.
- Code that cannot be stopped by Robot's timeouts is pretty rare. In most cases it's better to handle that code in some context specific way and I don't think a generic solution is commonly useful.
- The added decorator is only used with the Dialogs library. In that particular case using `multiprocessing` may be warranted due to Tkinter issues with threads, but I'm sure there could be a lot simpler solution that is only used in that context.
- There's some overhead with running code on external processes. That's not a problem with Dialogs, but it could cause issues if this decorator was used in other contexts.
- Debugging code that runs in an external process is complicated.

Due to the above reasons I don't think a generic decorator for running keywords in external processes is worth the added complexity. The `multiprocessing` approach could work really well with Dialogs, though.

PR #5379, opened 2025-03-21
Type converter to Robot data

### LINKED ISSUE #3278: Variable type conversion

aaltat:
Fixes #3278

### DIFF
--- a/atest/robot/variables/variable_types.robot
+++ b/atest/robot/variables/variable_types.robot
@@ -1,0 +1,166 @@
+*** Settings ***
+Suite Setup       Run Tests    ${EMPTY}    variables/variable_types.robot
+Resource          atest_resource.robot
+
+*** Test Cases ***
+Variable section
+    Check Test Case    ${TESTNAME}
+
+Variable section: list
+    Check Test Case    ${TESTNAME}
+
+Variable section: dictionary
+    Check Test Case    ${TESTNAME}
+
+Variable section: with invalid values or types
+    Check Test Case    ${TESTNAME}
+
+Variable section: parings errors
+    Error In File
+    ...    2    variables/variable_types.robot
+    ...    17    Setting variable '\${BAD_TYPE: hahaa}' failed: Unrecognized type 'hahaa'.
+    Error In File
+    ...    3    variables/variable_types.robot    19
+    ...    Setting variable '\@{BAD_LIST_TYPE: xxxxx}' failed: Unrecognized type 'xxxxx'.
+    Error In File
+    ...    4    variables/variable_types.robot    21
+    ...    Setting variable '\&{BAD_DICT_TYPE: aa=bb}' failed: Unrecognized type 'aa'.
+    Error In File
+    ...    5    variables/variable_types.robot    22
+    ...    Setting variable '\&{INVALID_DICT_TYPE1: int=list[int}' failed:
+    ...    Parsing type 'dict[int, list[int]' failed:
+    ...    Error at end: Closing ']' missing.
+    ...    pattern=False
+    Error In File
+    ...    6    variables/variable_types.robot    23
+    ...    Setting variable '\&{INVALID_DICT_TYPE2: int=listint]}' failed:
+    ...    Parsing type 'dict[int, listint]]' failed: Error at index 18:
+    ...    Extra content after 'dict[int, listint]'.
+    ...    pattern=False
+    Error In File
+    ...    7    variables/variable_types.robot    20
+    ...    Setting variable '\&{BAD_DICT_VALUE: str=int}' failed:
+    ...    Value '{'x': 'a', 'y': 'b'}' (DotDict) cannot be converted to dict[str, int]:
+    ...    Item 'x' got value 'a' that cannot be converted to integer.
+    ...    pattern=False
+    Error In File
+    ...    8    variables/variable_types.robot    18
+    ...    Setting variable '\@{BAD_LIST_VALUE: int}' failed:
+    ...    Value '['1', 'hahaa']' (list) cannot be converted to list[int]:
+    ...    Item '1' got value 'hahaa' that cannot be converted to integer.
+    ...    pattern=False
+    Error In File
+    ...    9    variables/variable_types.robot    16
+    ...    Setting variable '\${BAD_VALUE: int}' failed: Value 'not int' cannot be converted to integer.
+    ...    pattern=False
+
+VAR syntax
+    Check Test Case    ${TESTNAME}
+
+VAR syntax: list
+    Check Test Case    ${TESTNAME}
+
+VAR syntax: dictionary
+    Check Test Case    ${TESTNAME}
+
+VAR syntax: invalid scalar value
+    Check Test Case    ${TESTNAME}
+
+VAR syntax: Invalid scalar type
+    Check Test Case    ${TESTNAME}
+
+VAR syntax: type can not be set as variable
+    Check Test Case    ${TESTNAME}
+
+VAR syntax: type syntax is not resolved from variable
+    Check Test Case    ${TESTNAME}
+
+Vvariable assignment
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: list
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: dictionary
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: invalid value
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: invalid type
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: Invalid variable type for list
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: Invalid type for list
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: Invalid variable type for dictionary
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: multiple
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: multiple list and scalars
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: Invalid type for list in multiple variable assignment
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: type can not be set as variable
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: type syntax is not resolved from variable
+    Check Test Case    ${TESTNAME}
+
+Variable assignment: extended
+    Check Test Case    ${TESTNAME}
+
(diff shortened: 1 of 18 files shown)


atest/robot/variables/variable_types.robot
```diff
@@ -0,0 +1,164 @@
...
+    Error In File
+    ...    3    variables/variable_types.robot    19
+    ...    Setting variable '\@{BAD_LIST_TYPE: xxxxx}' failed: Unrecognized type 'xxxxx'.
+    Error In File
+    ...    4    variables/variable_types.robot    21
+    ...    Setting variable '\&{BAD_DICT_TYPE: aa=bb}' failed: Unrecognized type 'aa'.
+    Error In File
+    ...    5    variables/variable_types.robot    22
+    ...    Setting variable '\&{INVALID_DICT_TYPE1: int=list[int}' failed: Parsing type
+    ...    'dict[int, list[int]' failed: Error at end: Closing ']' missing.
```
pekkaklarck, line 31: In cases like this, splitting after the `:` would be a bit nice. No need to split after every `:`, though. This could look like below, but two last rows could also be merged.
```
...    Setting variable '\&{INVALID_DICT_TYPE1: int=list[int}' failed:
...    Parsing type 'dict[int, list[int]' failed:
...    Error at end: Closing ']' missing.
```

aaltat: Done

src/robot/running/arguments/embedded.py
```diff
@@ -14,9 +14,10 @@
 #  limitations under the License.
 
 import re
-from typing import Any, Mapping, Sequence
+from typing import Any, List, Mapping, Sequence
 
 from robot.errors import DataError
+from robot.running.arguments.typeinfo import TypeInfo
```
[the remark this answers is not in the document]
aaltat: Done

src/robot/running/arguments/embedded.py
```diff
@@ -45,8 +48,20 @@ def matches(self, name: str) -> bool:
...
+            try:
+                converted_args.append(info.convert(arg))
+            except TypeError:
+                raise DataError(f"Unrecognized type '{info.name}'.")
+            except ValueError:
+                raise DataError(f"Invalid value '{arg}' for type '{info.name}'.")
         if not placeholders:
-            return args
+            return tuple(converted_args)
         return tuple([self._replace_placeholders(a, placeholders) for a in args])
```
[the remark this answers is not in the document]
aaltat: Moved code to `map` as discussed.

src/robot/running/arguments/typeinfo.py
```diff
@@ -276,6 +280,42 @@ def from_sequence(cls, sequence: 'tuple|list') -> 'TypeInfo':
...
 
+    @classmethod
+    def from_variable(cls, variable: 'str|VariableMatch',
+                      handle_list_and_dict: bool = True) -> 'TypeInfo|None':
+        """Construct a ``TypeInfo`` based on a variable.
+
+        TODO: Write better doc later
+        """
+        if isinstance(variable, tuple):
+            variable = variable[0]
```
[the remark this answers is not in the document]
aaltat: Good catch. I think it happened before the changes in argumentparser.py `_validate_arg` sometimes returning tuple and sometimes not. I think this is not anymore needed.

atest/testdata/variables/variable_types.robot
```diff
@@ -0,0 +1,352 @@
...
+    Kwargs does not support key=value type syntax
+
+Embedded arguments
+    [Tags]    kala
+    Embedded 1 and 2
+    Embedded type 1 and no type 2
+    Embedded type with custom regular expression 111
+    VAR    ${x}    1
+    VAR    ${y}    2
+    Embedded ${x} and ${y}
```
pekkaklarck, line 267: Embedded args with variables would deserve its own test. More importantly, variable that's not a string should be tested.

pekkaklarck, line 267: Embedded args with variables would deserve its own test. More importantly, variable that's not a string should be tested.

src/robot/running/arguments/typeinfo.py
```diff
@@ -20,6 +20,8 @@
 from enum import Enum
 from pathlib import Path
 from typing import Any, ForwardRef, get_args, get_origin, get_type_hints, Literal, Union
+
+from robot.variables.search import VariableMatch, search_variable
```
pekkaklarck, line 24: Should be imported from `robot.variables`.

src/robot/running/arguments/typeinfo.py
```diff
@@ -276,6 +280,37 @@ def from_sequence(cls, sequence: 'tuple|list') -> 'TypeInfo':
             return infos[0]
         return cls('Union', nested=infos)
 
+    @classmethod
+    def from_variable(cls, variable: 'str|VariableMatch',
+                      handle_list_and_dict: bool = True) -> 'TypeInfo|None':
+        """Construct a ``TypeInfo`` based on a variable."""
+        if isinstance(variable, str):
+            variable = search_variable(variable, parse_type=True)
```
pekkaklarck, line 288: It would be a little bit more clear to change the condition to `if not isinstance(variable, VariableMatch)`. It's not obvious to known that `search_variable` returns `VariableMatch` by looking just the code.

src/robot/running/arguments/typeinfo.py
```diff
@@ -276,6 +280,37 @@ def from_sequence(cls, sequence: 'tuple|list') -> 'TypeInfo':
...
+
+    @classmethod
+    def _validate_var_type(cls, info):
+        if info.type is None:
+            raise DataError(f"Unrecognized type '{info.name}'.")
+        if info.nested and info.type is not Literal:
+            for nested in info.nested:
+                cls._validate_var_type(nested)
+
+
```
pekkaklarck, line 313: Extra empty line.

src/robot/variables/assigner.py
```diff
@@ -205,45 +205,65 @@ def _normal_assign(self, name, value, variables):
...
+            return ScalarsAndListReturnValueResolver(assignment)
+        return ScalarsOnlyReturnValueResolver(assignment)
 
-class NoReturnValueResolver:
+    def resolve(self, return_value):
+        raise NotImplementedError
+
+    def _split_assignment(self, assignment, handle_list_and_dict=True):
+        match: VariableMatch = search_variable(assignment, parse_type=True)
+        from robot.running import TypeInfo
```
pekkaklarck, line 225: - Import within a method should be on the first line.
- `match` shouldn't need a type hint because `search_variable` has a return type.

src/robot/variables/tablesetter.py
```diff
@@ -75,15 +88,26 @@ def resolve(self, variables) -> Any:
...
+                self.name = self.name[:2] + variables.replace_string(self.name[2:-1]) + '}'
             self.resolved = True
         return self.value
 
     def _replace_variables(self, variables) -> Any:
         raise NotImplementedError
 
+    def _convert(self, value, type_):
+        from robot.running import TypeInfo
+        info = TypeInfo.from_type_hint(type_)
```
pekkaklarck, line 105: Using `TypeInfo.from_variable` would avoid the need to handle `@` and `&` variables in `ListVariableResolver` and `DictVariableResolver`, but because we already have a type here, and needed to construct a variable, that might be more complicated anyway.

src/robot/running/arguments/typeinfo.py
```diff
@@ -276,6 +280,37 @@ def from_sequence(cls, sequence: 'tuple|list') -> 'TypeInfo':
             return infos[0]
         return cls('Union', nested=infos)
 
+    @classmethod
+    def from_variable(cls, variable: 'str|VariableMatch',
+                      handle_list_and_dict: bool = True) -> 'TypeInfo|None':
+        """Construct a ``TypeInfo`` based on a variable."""
```
pekkaklarck, line 286: Here could be an example about the variable syntax and should be a note about this being new in RF 7.3.

pekkaklarck (approved):
Looks great! There are few things still to be enhanced, see the inline comments for details, but I can take care of them.

src/robot/running/arguments/typeinfo.py
pekkaklarck: Turned out `VariableMatch` wasn't exposed via `robot.variables`. It should be.

Issue #5334, opened 2025-02-07
Dialogs: Enhance look and feel


Alpha-Centauri-00:
The Dialogs library in Robot Framework, which is based on Tkinter, plays a key role in creating interactive tests. However, the styling and design options are currently quite limited.

LucianCrainic:
Hello @pekkaklarck, what do you think about the idea of switching from Tkinter to Flet (https://flet.dev/) which provides a better styling and design options ?

pekkaklarck:
I don't think it's a good idea with the version distributed with Robot. It would be fine if you'd create an external library for this purpose.

LucianCrainic:
Thanks for your feedback! Could you clarify what you mean by creating an external library for this purpose? Are you suggesting a standalone package that integrates with Robot Framework, or something else? I'd appreciate any guidance on the best approach since i would love to start contributing to the project after the wonderful experience at the Robocon 2025.

pekkaklarck:
PR #5333 implements this.

pekkaklarck:
@LucianCrainic Yes, I meant creating an external library. It could have same keywords as the standard Dialogs to make switching to it easy, but it could also get new keywords if there are needs. New keywords could then possibly be ported to Dialogs if they are generally useful and its technically feasible.

PR #5381, opened 2025-03-24
ttk theming for Dialogs library

d-biehl:
Had to play around a bit today to distract myself from the other important things I should actually be doing, just to clear my head a bit… 😉

Regarding issue #5334 and today’s Slack discussion, I remembered reading that Tkinter can be themed. So I experimented a little, refreshed my TCL skills (with a bit of help from Claude 😉), and made the Dialogs library themeable.

What we have now:

- A first version of the Dialogs library supporting Dark and Light modes  
- Automatically follows the system theme (Dark/Light) on Windows/Mac/Linux (Gnome/GTK)  
- A touch of turquoise for that Robot Framework vibe  
- A theme file to adjust all aspects of Tkinter dialogs/controls without changing the code

What’s still missing:

- Option to set the default theme  
- Support for custom themes

Take a look — maybe it’ll be useful; if not, at least I procrastinated 😄

### DIFF
--- a/src/robot/libraries/dialogs_py.py
+++ b/src/robot/libraries/dialogs_py.py
@@ -13,10 +13,12 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+
+import os
 import sys
 from threading import current_thread
-from tkinter import (BOTH, Button, END, Entry, Frame, Label, LEFT, Listbox, Tk,
-                     Toplevel, W)
+from tkinter import BOTH, END, LEFT, Tk, Toplevel, W
+from tkinter import ttk
 from typing import Any, Union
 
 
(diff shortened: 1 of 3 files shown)


pekkaklarck:
I tested this locally. Comments based on that and otherwise:
- On the Slack poll the clear majority preferred using the system default colors. I'm thus not sure does it make sense to change them. I'm also not sure do people recognize "Robot colors" if they are used only with buttons.
- Using light/dark theme depending on user preference is nice, but I'm not sure is it worth all the extra code. I also tested the current code on an OSX laptop with a dark theme today, and dialogs were black. Could it be that OSes can handle the theme automatically? In that case I think it's best we don't touch colors too much.
- I don't like the reduced padding or smaller font size compared to the current styles. I increased them on purpose as part of #5334 earlier.
- List selection widget is broken on Linux at least on Linux. I'll attach a screenshot.

PR #5388, opened 2025-03-27
Mention logging with ERROR level when using --exitonerror

gohierf:
Seems like this is not mentioned anywhere else. I scratched my head on it for a while.

It appears that the term "ERROR" is used for different purpose in TRY/EXCEPT (i.e. exception or failure) or logging (i.e. level).
Maybe it would be valuable to also mention it in the TRY/EXCEPT section :

```rst
.. note:: TRY/EXCEPT do not catch "ERROR" level logs, therefore they will not prevent exiting when :option:`--exitonerror` is used.
```

### DIFF
--- a/doc/userguide/src/ExecutingTestCases/TestExecution.rst
+++ b/doc/userguide/src/ExecutingTestCases/TestExecution.rst
@@ -692,6 +692,11 @@ fatal and execution stopped so that remaining tests are marked failed. With
 parsing errors encountered before execution even starts, this means that no
 tests are actually run.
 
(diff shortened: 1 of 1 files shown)


pekkaklarck:
I agree it's a good idea to mention that using the ERROR log level stops execution if `--exit-on-error` is used. I submitted issue #5424 about this. The reason TRY/EXCEPT doesn't catch these errors is that it doesn't in general catch any error that stops the whole execution. Finally, it's true that the term "error" is used in many different places meaning slightly different things.

The added documentation looks good in general, but there are few things to be adjusted. I'll just merge this PR and do cleanup afterwards.

Thanks for you contribution!

Issue #5384, opened 2025-03-25
deprecate / remove "pythonpathsetter"


franzhaas:
pythonpathsetter is a module that modifies the `sys.path` in a way that it is possible to run robot-related scripts without having robot installed in your environment.

This makes robot fail in very wired ways IF.:

 - you use a zipapp/frozen app which _runs_ a script or the content of a script without importing robot first
 - you use a `scripts=["src/robot/run.py"]` instead of an entry point in setup.py
 - you copy code into your repl for experimentation

While it is possible to work around this for zipapp and frozen app usage, it is the main source of effort. The problem tends only to manifest when the frozen executable is built, making debugging hard and experimentation slow. It is hard to comprehend how seemingly innocent changes (run main / call main) break and fix things, and at least for me, I might be using imprecise language and end up mixing run vs call in communication.

In most circumstances, I believe that this concept is the better solution.:
```bash
PYTHONPATH=/path/to/your/robotdir/src/robot/ python src/robot/libdoc.py
```
It makes the productive robot code simpler, it makes it more visible that the `sys.modules` gets tinkered with.

Many solutions to support running robot-related scripts outside of an installed robot are possible, but it is not the use case I use, so I would like to leave this up to the real users.

This change might break the workflow of some people, so depricating first and collecting feedback should be the right step.

pekkaklarck:
The reason why `pythonpathsetter` was created, is that back then it wasn't as convenient to run Python modules as scripts as it is nowadays. Although I somewhat often run `src/robot/libdoc.py`, which requires the logic provided by `pythonpathsetter`,  during development, I tend to agree that the problems are bigger than benefits nowadays.

The proposed alternative to use
```
PYTHONPATH=src/robot python src/robot/libdoc.py
```
instead once `pythonpathsetter` is gone isn't that good, though, because having `src/robot` in `sys.path` can cause (and has caused) strange problems. The proper way to do that is:

pekkaklarck:
Although I doubt `pythonpathsetter` functionality is widely needed, it's best to deprecate it first and only remove it in a major version. We could deprecate it in RF 7.3 (or 7.4) and then remove it RF 8. On the code level deprecation shouldn't need much more than adding `warnings.warn(message)` to the `pythonpathsetter` module itself. In addition to that, the User Guide needs to be studied to see is using the old approach still recommended somewhere. Are you @franzhaas interested to do this?

pekkaklarck:
[...]

PR #5390, opened 2025-03-28
No pythonpathsetter3

franzhaas:
This attempts to deal with this issue.: https://github.com/robotframework/robotframework/issues/5384


franzhaas:
I changed from running the script to calling the module.

This is necessary as the way the sys.path is constructed in python is a bit confusing.

With this PR the [a|u]test/run.py need the working directory to be tested also installed editably, or the PYTHONPATH needs to be adjusted with alternative means.

python src/robot/run.py still works without an installation, but it creates a deprication warning

@pekkaklarck the deprication warning is written to the console, before robot logging is ready... is this ok? Where do you think this shall be moved to?

franzhaas:
This PR is in conflict with the goal to run the robot scripts from a source tree. (https://github.com/robotframework/robotframework/issues/5384)

PR #5391, opened 2025-04-02
Feat: libdoc dynamic languages in help

yanne:


src/robot/libdocpkg/languages.py
```diff
@@ -0,0 +1,22 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+
+# This is modified by invoke, do not edit by hand
+LANGUAGES = ['EN', 'FI', 'FR', 'IT', 'NL', 'PT-BR', 'PT-PT']
```
pekkaklarck, line 18: It would be good to format this like below (incl. the trailing comma). The main reason is that there's a plan to format the code using ruff, and this format it wouldn't touch even if new languages were added. A smaller benefit is that this way it's easier to see what exact language has been added in a diff.
```python
LANGUAGES = [
    'EN',
    'FI',
    'FR',
    'IT',
    'NL',
    'PT-BR',
    'PT-PT',
]
```

src/robot/libdocpkg/languages.py
```diff
@@ -0,0 +1,22 @@
...
+
+def format_languages():
+    indent = 26 * ' '
+    return '\n'.join(f'{indent}- {lang}' for lang in LANGUAGES)
```
pekkaklarck, line 22: Newline missing at the end.

tasks.py
```diff
@@ -147,6 +149,34 @@ def release_notes(ctx, version=None, username=None, password=None, write=False):
...
+    web project and updates the languages that are used in the libdoc command line
+    tool for help and language validation.
+
+    This task needs to be run if there are any changes to libdoc.
+    """
+    subprocess.run(['npm', 'run', 'build', '--prefix', 'src/web/'])
+
+    src_path = Path(__file__).parent / "src" / "web" / "libdoc" / "i18n" / "translations.json"
+    data = json.loads(open(src_path).read())
+    keys = sorted([key.upper() for key in data.keys()])
```
pekkaklarck, line 168: - No need to use `Path(__file__).parent` here. The `assert` at the top of the module makes sure this script can run only at the project root, so relative paths are fine.
- When using `Path` objects, `path / 'x/y/z'` is more convenient than `path / 'x' / 'y' / 'z'`.
- `open(path).read()` -> `path.read_text(encoding='utf-8')`
- `for key in data.keys()` -> `for key in data` in general.
- In this case `languages = sorted(data, key=str.upper)` is even better than the above.

pekkaklarck (approved):
Looks good, but there's some cleanup that could be done.

tasks.py
yanne: fixed all but the last, since I want the keys in the final list to be uppercased (they are not in the original dictionary)

PR #5421, opened 2025-05-04
auto_keywords=None suppresses 'library contains no keywords' warning

JFoederer:


JFoederer:
Hi @pekkaklarck, can you please review this PR?

I basically have 2 questions:

- What is your opinion on suppressing this warning?
- Is the low-impact way of '_abusing_' the existing `auto_keywords` option the preferred way?

The alternative would be to have a separate option, but I feel this might be overkill for such a niche option.

pekkaklarck:
`auto_keywords=None` is both clever and hacky. It's certainly better than a separate parameter for this niche use case. Related to that, could you explain your use case for this behavior? Why would you have a library without any keywords? Instead of explaining that here, you could submit an issue because we'd anyway need one for tracking purposes.

Notice also that there's another way to avoid this warning: adding a listener to the library. Listeners don't need to have any listener methods and nowadays don't even need to `ROBOT_LISTENER_API_VERSION` attribute, so you could probably use just `listener=object()`.

JFoederer:
I agree it is hacky, but to be honest, turning it into a fake listener sounds even more hacky than `auto_keywords=None`. It would do the trick, though. But I must add that I wouldn't have thought of this option, even after reading the documentation and seeing the source code. I hadn't realised that empty unregistered listeners would work.

I will submit the issue and then we have three options: accept it as-is, reject it and keep using the listener or find a _pretty_ solution.

s-wozny:
> I agree it is hacky, but to be honest, turning it into a fake listener sounds even more hacky than `auto_keywords=None`

Hi, I totally agree with you!
In my use case, the library is used to enforce a default User-Agent header on all outbound HTTP requests made with the requests library by "patching" the request function. Unfortunately, I couldn't find any global option or config to handle this specific need. So, I don't actually need a keyword in this case. :)

Does anyone knows if auto_keywords=None will be available, and when?
Using listener=object() works for me as well now, but I’m not a big fan of this “hacky” workaround.

pekkaklarck:
@s-wozny, could you explain your use case in issue #5429 that's used for tracking this enhancement? Having more than one use case proves the feature would be generally useful.

s-wozny:
[All done](https://github.com/robotframework/robotframework/issues/5429#issuecomment-3236281533), thanks for the quick response !

PR #5430, opened 2025-05-13
Fix variable types for statements

aaltat:


pekkaklarck:
FOR loop variable conversion was implemented in 448fb072fcd5a24976d82346e237f0498037f224. Tests it contain are based on the tests in this PR but implementation is different.

Issue #5322, opened 2025-01-22
Log: Toggle icon is stuck to `[+]` after toggling failed or skipped test


jyoti-arora1991:
Description:

When a test is marked as FAILED or SKIPPED, the expand button in the generated log.html report always displays the + sign, regardless of whether the user has clicked it to expand or not.

This behavior is inconsistent and can lead to confusion, as the button should toggle between + (collapsed) and - (expanded) correctly.

Steps to Reproduce:

Run a test suite with tests that result in FAILED or SKIPPED status.
Open the generated log.html report.
Expand a failed or skipped test.
Notice that the expand button remains as + even when expanded.
Expected Behavior:
The expand button should display - when the test is expanded and toggle back to + when collapsed.

pekkaklarck:
Failed and skipped tests are expanded automatically by design. I just run some tests to verify that and in my case those expanded tests had `[-]` icon as expected.

Could you clarify what exactly is the issue you see? If you wouldn't want failed/skipped tests to be expanded, that's not something we plan to change at least as the default behavior. If the problem is that icons of expanded tests are wrong, please include a screenshot and tell what browser you are using.

jyoti-arora1991:
Try this test and click on `check report issue` , try 2 3 times, you will see the issue

*** Test Cases ***
Check Report Issue
     Log to console    hi
     Should be true    False

Attaching screenshot where you can see test is expanded but still showing plus sign.

<img width="625" alt="Image" src="https://github.com/user-attachments/assets/ebec4c4a-0694-452f-a72c-4148fa7d8949" />


 pip freeze|grep robot
robotframework==7.2.2
robotframework-hub==1.0.1
robotframework-sshlibrary==3.8.0
sphinxcontrib-robotframework==0.8.1

pekkaklarck:
I hadn't understood the issue earlier. The initial toggle icon state is right (`[+]` or `[-]`), but after toggling a failed or skipped test, the icon was stuck to `[+]`. Luckily @psaikkonen had understood the problem properly and has provided PR #5445 to fix it.

pekkaklarck:
Notice that the toggling functionality itself worked fine, this was just a visual glitch.

PR #5445, opened 2025-06-07
Fix expand icons for failed and skipped tests (#5322)

psaikkonen:
Fixes an issue with log where expand icon of failed and skipped tests got stuck as `+` while toggling tests.
This was caused by expandFailed first removing the `closed` class from failed/skipped tests, only to be immediately re-inserted by toggleElement.

Now toggleElement is called only if test is not expanded with expandFailed (status != skip/fail), or the test element is currently not closed. In the latter case expandFailed does not affect the toggle, as it only removes `closed` class if it exists.

https://github.com/user-attachments/assets/b5f2c70b-5d55-445c-b3f6-b6a9c41cffbf

### DIFF
--- a/src/robot/htmldata/rebot/log.js
+++ b/src/robot/htmldata/rebot/log.js
@@ -5,10 +5,15 @@ function toggleSuite(suiteId) {
 }
 
 function toggleTest(testId) {
-    toggleElement(testId, ['keyword']);
     var test = window.testdata.findLoaded(testId);
-    if (test.status == "FAIL" || test.status == "SKIP")
+    var autoExpand = test.status == "FAIL" || test.status == "SKIP";
+    var closed = $('#' + testId).children('.element-header').hasClass('closed');
+
+    if (autoExpand)
         expandFailed(test);
+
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Thanks for the PR! The analysis and the fix look good.

Now that the code checks is the element open, expanding failed (and skipped) keywords when the element is going to be closed feels unnecessary. Perhaps the code should be changed from
```js
if (autoExpand)
    expandFailed(test);
if (!autoExpand || !closed)
    toggleElement(testId, ['keyword']);
```
to
```js
if (autoExpand && closed)
    expandFailed(test);
else
    toggleElement(testId, ['keyword']);
```
I feel the latter easier to understand as well. I need to test this locally to make sure I didn't make any Boolean logic errors, though, and I anyway merge this PR first.

To get this fix listed in the release notes, we'll need an issue in addition to this PR. I want to get this into RF 7.3.1 so I'll create the issue myself.

pekkaklarck:
Ooops, there's already issue #5322 about this. It's also mentioned in the PR title, but I didn't notice it there.

Issue #5373, opened 2025-03-19
Add type hints to standard libraries


robinmackaij:
As discussed [here](https://github.com/robotframework/robotframework/issues/5361).

This issue is the parent issue to track progress on all the standard libraries (sub-issues).

Overall implementation decisions can be discussed / recorded here.

The Python typing best practices can be found here:
https://typing.python.org/en/latest/reference/best_practices.html

twbecker:
Came here looking for this. I'd been experimenting with custom argument converters and was disappointed to see that they aren't invoked for standard keywords because there are no type hints. Am I correct in assuming that this would correct that?

pekkaklarck:
@twbecker Standard libraries predate all automatic argument conversion support so they needed to implement their own argument conversion. All that custom code can be removed when we add type hints, but that probably doesn't bring any concrete benefits for normal users. Libdoc documentation getting better is nice, though.

pekkaklarck:
In order to get RF 7.3 out soon, I move this issue to RF 7.4 scope. It would still be possible to add type hints to _some_ library or libraries in RF 7.3, though. My plan is to get RF 7.3 release candidate out early next week and that's a deadline for all code changes (which includes annotations). Are there any libraries you'd like to work on in that time frame @robinmackaij? If there are, please submit separate issues for each library and I can add them to RF 7.3 scope.

pekkaklarck:
Some keywords accepting configuration parameters as `**config` instead of named-only arguments makes adding type hints to them annoying. The reason to use `**config` is Python 2 support, so these keywords can be updated. I submitted #5412 about that and try to get it done still in RF 7.3 timeline.

pekkaklarck:
Also return type hints should be added where applicable.

silentw0lf:
@robinmackaij  I would be interested to work on it too. 

Any updates for coordination?

silentw0lf:
Libraries would be (vias docs):
(x) BuiltIn.py
(x) Collections.py
(x) DateTime.py
(x) Dialogs.py
(x) OperatingSystem.py
(x) Process.py
(x) Screenshot.py
(x) String.py
(x) Telnet.py
(x) XML.py


I do not know how to create sub-issues here (probably no permissions). 
I will start working on 1 library for the beginning:  **DateTime.py** .

pekkaklarck:
[...]

PR #5453, opened 2025-06-17
Add return type information to all standard libraries

joshuafeld:
This pull request adds return types to all of the standard libraries. This will enhance the compatibility with libdoc as it can now report the returntype field for these libraries.

### DIFF
--- a/src/robot/libraries/BuiltIn.py
+++ b/src/robot/libraries/BuiltIn.py
@@ -1545,7 +1545,9 @@ class _Verify(_BuiltInBase):
                 self._get_string_msg(string, pattern, msg, values, "does not match")
             )
 
-    def should_match_regexp(self, string, pattern, msg=None, values=True, flags=None) -> str:
+    def should_match_regexp(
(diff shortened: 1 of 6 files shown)


pekkaklarck:
Thanks for the PR. Unfortunately I don't think it can be accepted in the current format for the following reasons:
- I don't think it makes sense to add only return type information. When going through the keywords, it's better to add also argument types. We have issue #5373 about that and I just added a comment there about return type hints.
- This PR is rather big. It would be a lot easier to review a PR covering only a single library. This is especially true if/when there are also argument type hints.
- The PR seems to be using `Any` in places where the return type isn't known, but my understanding is that `object` should be used in such cases instead. In some cases there probably could be a more concrete type like `str`. Going all these through in a big PR like this is not feasible.

I'll close this PR now, but I hope you are interested to continue and create PR or PRs adding both argument and return types. @robinmackaij has shown interest towards adding type hints as well, so it would probably be a good idea to coordinate the efforts. You can use issue #5373 or the #devel channel on our Slack for that.

PR #5462, opened 2025-06-25
Support not printing errors and warnings to console

adombeck:
There are use cases for printing a warning or error message to the log but not to the console.

This PR adds an optional `also_console` argument to `logger.write`, `logger.warn`, and `logger.error`. Its value is stored in the message object and based on it the loggers in `src/robot/output/console/` print the message or not.

### DIFF
--- a/src/robot/api/logger.py
+++ b/src/robot/api/logger.py
@@ -74,7 +74,12 @@ from robot.running.context import EXECUTION_CONTEXTS
 LOGLEVEL = Literal["TRACE", "DEBUG", "INFO", "CONSOLE", "HTML", "WARN", "ERROR"]
 
 
-def write(msg: str, level: LOGLEVEL = "INFO", html: bool = False):
+def write(
+    msg: str,
+    level: LOGLEVEL = "INFO",
+    html: bool = False,
+    also_console: bool | None = None,
+):
     """Writes the message to the log file using the given level.
 
     Valid log levels are ``TRACE``, ``DEBUG``, ``INFO`` (default), ``WARN``,
@@ -88,8 +93,11 @@ def write(msg: str, level: LOGLEVEL = "INFO", html: bool = False):
     specific methods such as ``info`` and ``debug`` that have separate
     ``html`` argument to control the message format.
     """
+    if also_console is None:
+        also_console = level in ("WARN", "ERROR")
+
     if EXECUTION_CONTEXTS.current is not None:
-        librarylogger.write(msg, level, html)
+        librarylogger.write(msg, level, html, also_console)
     else:
         logger = logging.getLogger("RobotFramework")
         level_int = {
@@ -120,19 +128,17 @@ def info(msg: str, html: bool = False, also_console: bool = False):
     If ``also_console`` argument is set to ``True``, the message is
     written both to the log file and to the console.
     """
-    write(msg, "INFO", html)
-    if also_console:
-        console(msg)
+    write(msg, "INFO", html, also_console)
 
 
-def warn(msg: str, html: bool = False):
+def warn(msg: str, html: bool = False, also_console: bool = True):
     """Writes the message to the log file using the ``WARN`` level."""
-    write(msg, "WARN", html)
+    write(msg, "WARN", html, also_console)
 
 
-def error(msg: str, html: bool = False):
+def error(msg: str, html: bool = False, also_console: bool = True):
     """Writes the message to the log file using the ``ERROR`` level."""
-    write(msg, "ERROR", html)
+    write(msg, "ERROR", html, also_console)
 
 
 def console(
--- a/src/robot/model/message.py
+++ b/src/robot/model/message.py
@@ -32,7 +32,7 @@ class Message(BodyItem):
 
     type = BodyItem.MESSAGE
     repr_args = ("message", "level")
-    __slots__ = ("message", "level", "html", "_timestamp")
+    __slots__ = ("message", "level", "html", "also_console", "_timestamp")
 
     def __init__(
         self,
@@ -40,12 +40,14 @@ class Message(BodyItem):
         level: MessageLevel = "INFO",
         html: bool = False,
         timestamp: "datetime|str|None" = None,
+        also_console: bool = False,
         parent: "BodyItem|None" = None,
     ):
         self.message = message
         self.level = level
         self.html = html
         self.timestamp = timestamp
+        self.also_console = also_console
         self.parent = parent
 
     @setter
--- a/src/robot/output/console/highlighting.py
+++ b/src/robot/output/console/highlighting.py
@@ -124,8 +124,9 @@ class HighlightingStream:
             "ERROR": highlighter.red,
             "WARN": highlighter.yellow,
             "SKIP": highlighter.yellow,
-        }[status]
-        start()
+        }.get(status)
+        if start:
+            start()
         try:
             yield
         finally:
--- a/src/robot/output/logger.py
+++ b/src/robot/output/logger.py
@@ -208,7 +208,7 @@ class Logger(AbstractLogger):
             logger.log_message(msg)
         if self._log_message_parents and self._output_file.is_logged(msg):
             self._log_message_parents[-1].body.append(msg)
-        if msg.level in ("WARN", "ERROR"):
+        if msg.also_console or msg.level in ("WARN", "ERROR"):
             self.message(msg)
 
     def log_output(self, output):
--- a/src/robot/output/console/dotted.py
+++ b/src/robot/output/console/dotted.py
@@ -62,7 +62,7 @@ class DottedOutput(LoggerApi):
             self.stdout.write("\n")
 
     def message(self, msg):
-        if msg.level in ("WARN", "ERROR"):
+        if msg.also_console:
             self.stderr.error(msg.message, msg.level)
 
     def result_file(self, kind, path):
--- a/src/robot/output/console/quiet.py
+++ b/src/robot/output/console/quiet.py
@@ -25,7 +25,7 @@ class QuietOutput(LoggerApi):
         self._stderr = HighlightingStream(stderr or sys.__stderr__, colors)
 
     def message(self, msg):
-        if msg.level in ("WARN", "ERROR"):
+        if msg.also_console:
             self._stderr.error(msg.message, msg.level)
 
 
--- a/src/robot/output/console/verbose.py
+++ b/src/robot/output/console/verbose.py
@@ -70,7 +70,7 @@ class VerboseOutput(LoggerApi):
             self.writer.keyword_marker(result.status)
 
     def message(self, msg):
-        if msg.level in ("WARN", "ERROR"):
+        if msg.also_console:
             self.writer.error(msg.message, msg.level, clear=self.running_test)
 
     def result_file(self, kind, path):
--- a/src/robot/output/librarylogger.py
+++ b/src/robot/output/librarylogger.py
@@ -32,7 +32,7 @@ from .loggerhelper import Message, write_to_console
 LOGGING_THREADS = ["MainThread", "RobotFrameworkTimeoutThread"]
 
 
-def write(msg: Any, level: str, html: bool = False):
+def write(msg: Any, level: str, html: bool = False, also_console: bool = False):
     if not isinstance(msg, str):
         msg = safe_str(msg)
     if level.upper() not in ("TRACE", "DEBUG", "INFO", "HTML", "WARN", "ERROR"):
@@ -42,7 +42,7 @@ def write(msg: Any, level: str, html: bool = False):
         else:
             raise RuntimeError(f"Invalid log level '{level}'.")
     if current_thread().name in LOGGING_THREADS:
-        LOGGER.log_message(Message(msg, level, html))
+        LOGGER.log_message(Message(msg, level, html, also_console=also_console))
 
 
 def trace(msg, html=False):
--- a/src/robot/output/loggerhelper.py
+++ b/src/robot/output/loggerhelper.py
@@ -98,9 +98,12 @@ class Message(BaseMessage):
         level: "MessageLevel|PseudoLevel" = "INFO",
         html: bool = False,
         timestamp: "datetime|str|None" = None,
+        also_console: bool = False,
     ):
         level, html = self._get_level_and_html(level, html)
-        super().__init__(message, level, html, timestamp or datetime.now())
+        super().__init__(
+            message, level, html, timestamp or datetime.now(), also_console
+        )
 
     def _get_level_and_html(self, level, html) -> "tuple[MessageLevel, bool]":
         level = level.upper()


adombeck:
@pekkaklarck it required more changes than I expected. I tested it by running `robot --listener Listener.py ./test.robot` with this listener:

```
$ cat Listener.py
from robot.api import logger

class Listener:
    def end_test(self, data, result):
        logger.info("XXX 1: end_test: INFO (should not be printed to console)")
        logger.info("XXX 2: end_test: INFO, also_console=True", also_console=True)
        logger.error("XXX 3 : end_test: ERROR (should be printed to console automatically)")
        logger.error("XXX 4: end_test: ERROR, also_console=False", also_console=False)
        logger.write("XXX 5: end_test: write, also_console=True", also_console=True)

    def end_suite(self, data, result):
        logger.info("XXX 6: end_suite: INFO (should not be printed to console)")
        logger.info("XXX 7: end_suite: INFO, also_console=True", also_console=True)
        logger.error("XXX 8: end_suite: ERROR (should be printed to console automatically)")
        logger.error("XXX 9: end_suite: ERROR, also_console=False", also_console=False)
        logger.write("XXX 10: end_suite: write, also_console=True", also_console=True)
```

and this suite file:

```
$ cat test.robot
*** Test Cases ***
My Test Case
    My Keyword

*** Keywords ***
My Keyword
    Fail
```

It prints the expected messages to the console (2, 3, 5, 7, 8, 10) and to the log.html file (all except 6, 7 and  10, which is due to the issue that info messages logged by the `end_suite` method do not end up in the log.html file).

adombeck:
@pekkaklarck any estimate when this could be reviewed?

src/robot/api/logger.py
```diff
@@ -74,7 +74,12 @@ def my_keyword(arg):
 LOGLEVEL = Literal["TRACE", "DEBUG", "INFO", "CONSOLE", "HTML", "WARN", "ERROR"]
 
 
-def write(msg: str, level: LOGLEVEL = "INFO", html: bool = False):
+def write(
+    msg: str,
+    level: LOGLEVEL = "INFO",
+    html: bool = False,
+    also_console: bool | None = None,
```
pekkaklarck: This module is part of the public API, so the new argument should be mentioned in the docstring. All other arguments apparently aren't covered either, but they are a bit more self-explanatory. It would be best to explain all args, but that's not really related to this PR and we are also going to change API doc style in the near future. In the current situation adding a paragraph like below would probably be the easiest solution.

> The ``also_console`` parameter controls writing the message to the console in addition to the log file. By default, messages with the ``WARN`` and ``ERROR`` level are logged to the console and others are not. This parameter is new in Robot Framework 7.4.

src/robot/api/logger.py
```diff
@@ -88,8 +93,11 @@ def write(msg: str, level: LOGLEVEL = "INFO", html: bool = False):
     specific methods such as ``info`` and ``debug`` that have separate
     ``html`` argument to control the message format.
     """
+    if also_console is None:
+        also_console = level in ("WARN", "ERROR")
+
```
pekkaklarck, line 118: We support log levels case-insensitively, so the above should be changed to `level.upper() in (...)` or there should be a line with `level = level.upper()` earlier. Yet another solution would be passing `also_console` as-is to `librarylogger.write` and handle it only later.

Please remove the unnecessary empty line inside the function.. Our coding style, which is explained in the Contribution guidelines, is not to use them except in special cases.

src/robot/api/logger.py
```diff
@@ -120,19 +128,17 @@ def info(msg: str, html: bool = False, also_console: bool = False):
...
     """
-    write(msg, "INFO", html)
-    if also_console:
-        console(msg)
+    write(msg, "INFO", html, also_console)
 
 
-def warn(msg: str, html: bool = False):
+def warn(msg: str, html: bool = False, also_console: bool = True):
     """Writes the message to the log file using the ``WARN`` level."""
```
pekkaklarck: `also_console` should be mentioned in the docstring similarly as with the `info` method. It should also be added to `error`.

src/robot/model/message.py
```diff
@@ -32,20 +32,22 @@ class Message(BodyItem):
...
         html: bool = False,
         timestamp: "datetime|str|None" = None,
+        also_console: bool = False,
         parent: "BodyItem|None" = None,
     ):
         self.message = message
         self.level = level
         self.html = html
         self.timestamp = timestamp
+        self.also_console = also_console
```
pekkaklarck: Adding the `also_console` attribute to the message objects is a good idea, but I don't think this base class for all messages should get it. This information isn't preserved in the results (and there's no need to do that either), so the attribute has no use in the result model. I believe only the `robot.output.loggerhelper.Message` class actually needs this information. Those objects are the ones that are used during execution.

The aforementioned `robot.output.loggerhelper.Message` also normalizes log level. I believe it could also accept `also_console: bool | None = None` and handle the `None` case itself. This would simplify `robot.api.logger.write` (see my comment there for details).

Another thing to consider is the name of this attribute. Using `console` instead of `also_console` on this level would feel a bit better to me, but I don't have a strong opinion on that. A problem with `console` would be that it's inconsistent with `also_console` that still needs to be used in `robot.api.logger` due to it being used there already earlier. You can decide either way and I can also think about this again after this PR is merged.

src/robot/output/console/highlighting.py
```diff
@@ -124,8 +124,9 @@ def _highlighting(self, status):
             "ERROR": highlighter.red,
             "WARN": highlighter.yellow,
             "SKIP": highlighter.yellow,
-        }[status]
-        start()
+        }.get(status)
+        if start:
+            start()
```
pekkaklarck, line 129: Why is this change needed?

pekkaklarck:
Sorry for the delay with the review @adombeck. Life got to the way, but the review is now done. I'll be working more actively in the near future also otherwise and try to get RF 7.4 finished somewhat soon.

pekkaklarck:
Are you @adombeck still around and do you have time and interest to look at this PR? I'd like to get RF 7.4 out pretty soon, so there's not too much time to address the review comments. I could possibly do that myself as well, but RF 7.4 is already quite a bit late so its more likely I'll just descope this feature from RF 7.4.

adombeck:
Thanks for the reminder. I'll try to find time for it in the next few days.

src/robot/output/console/highlighting.py
adombeck: Because the `level` that's passed to `highlight` in line 115 can now have other values than `ERROR` or `WARN`. For example if it's `INFO`, then 

```python
        start = {
            "PASS": highlighter.green,
            "FAIL": highlighter.red,
            "ERROR": highlighter.red,
            "WARN": highlighter.yellow,
            "SKIP": highlighter.yellow,
        }[status]
```

fails with

```
KeyError: 'INFO'
```

This change ignores levels for which no highlighting is configured.

src/robot/model/message.py
adombeck: I removed the `also_console` attribute from the base class, renamed it to `console` in `robot.output.loggerhelper.Message`, and handle the `console is None` case there instead of in `robot.api.logger.write`.

src/robot/api/logger.py
adombeck: done

src/robot/api/logger.py
adombeck: done

src/robot/api/logger.py
adombeck: done

adombeck:
> Tests are needed. This is the biggest problem at the moment. If you are unsure how/where to add tests, let me know and I can help.

Yes, it would help me if you could tell where you would like me to add tests.

pekkaklarck:
Quick instructions on creating tests:
1. First of all, read [atest/README.rst](https://github.com/robotframework/robotframework/blob/master/atest/README.rst) to understand how the acceptance test system works in general.
2. [atest/testdata/test_libraries/logging_api.robot](https://github.com/robotframework/robotframework/blob/master/atest/testdata/test_libraries/logging_api.robot) seems to be a good place for a new test. The test will need new keyword in the library that's used by that suite.
3. [atest/robot/test_libraries/logging_api.robot](https://github.com/robotframework/robotframework/blob/master/atest/robot/test_libraries/logging_api.robot) then needs to be updated to validate that the added test behaves as expected.

See the existing tests for examples and ask questions here or on the #devel channel on our Slack if needed.

adombeck:
The `Write messages to console` test in `atest/robot/test_libraries/logging_api.robot` fails on this branch and I'm having some trouble figuring out why. The call to `logger.info("To log and console", also_console=True)` doesn't print the message to the console in the test but it does when I try it with the steps I described in https://github.com/robotframework/robotframework/pull/5462#issuecomment-3004168796. There, message 2 is printed to the console via `logger.info` with `also_console=True`.

pekkaklarck:
I don't have time to investigate this locally now, but I enabled CI run to see the actual error message. Turned out that unit tests failed on Python 3.8 due to typing like `str | None`. We still support Python < 3.10 so unions needs to be stringified like `"str | None"`. See the Contribution guidelines for more info.

I'll be pretty busy this week and may not have time to look at this, but I hope you can investigate the issue yourself.

adombeck:
I now understand why the `Write messages to console` test fails on this branch. On `master`, `logger.info` with `also_console=True` used `librarylogger.console` which prints to stdout. On this branch, we print the message to stderr instead, because we use the same code that was already used to log warnings and error to now print all messages which have `Message.console` set to `True`.

Can we accept that new behavior and change the existing test to check stderr instead of stdout?

adombeck:
If we want to avoid changing the existing behavior, we could keep using `librarylogger.console` when `logger.info` is called with `also_console=True` and only use the new code path for `logger.warn` and `logger.error`. It could be a bit surprising that the logger functions print to a different file descriptor though.

pekkaklarck:
This looks pretty good to me now, bu I need to take one more look with a bit fresher eyes.

The only thing that bothers me is the name of the `also_console` argument, especially when used like `also_console=False`. I know the reason for the name is that `info` already used it and one function using `also_console` and others `console` would be even worse than using `also_console` consistently. Using `console` with all these would require deprecating `also_console` with `info` which is some extra work, but I believe that's still the best solution. I think I'm responsible on the bad `also_console` name initially, so I can also take care of deprecating it after this PR has been merged.

A bigger problem is that the following two tests failed when I enabled test run on GitHub Actions:
```
FAIL: Robot.Test Libraries.Logging When Run On Thread.logging
Wrong log message: Error message to log and console != warning message
----------------------------------------------------------------------------------------------------
FAIL: Robot.Test Libraries.Logging With Logging.Suppress errors from logging module
'' contains 'Traceback' 0 times, not 1 time.
```

Did you @adombeck run tests locally and see these? It could be that tests just need to be updated, but it's possible that they also found some problems. If you don't understand why the tests are failing, I can investigate this locally as well.

adombeck:
> Did you @adombeck run tests locally and see these? 

I ran the tests but I must have missed those failures in the list of other atest failures which also fail on master when I run them locally.

> ```
> FAIL: Robot.Test Libraries.Logging When Run On Thread.logging
> Wrong log message: Error message to log and console != warning message
> ```

That was easy to fix, I just had to update the index of the `$ERRORS` which the test is checking.

> ```
> FAIL: Robot.Test Libraries.Logging With Logging.Suppress errors from logging module
> '' contains 'Traceback' 0 times, not 1 time.
> ```

Here we have an actual problem that's uncovered by the test: Messages logged via the Python `logging` module don't end up on stderr anymore. That's because we only determine if messages should be logged to the console in `robot.output.loggerhelper.Message` (as suggested in https://github.com/robotframework/robotframework/pull/5462#discussion_r2374941576) which doesn't seem to be used when logging via the `logging` module. Moving that functionality back to `robot.model.Message` seems to fix it. Should we do that or do you see a better solution?

pekkaklarck:
`robot.model.Message` is still a wrong model object to contain the `console` attribute. It's only needed by the runtime object, not by others.

I tested this locally and believe I found the reason for the failure. Messages logged by the `logging` module are forwarded via `pyloggingconf` to `librarylogger` and its `write` method has `also_console=False`. This means that messages logged using it aren't logged to the console by default, regardless the level. Changing it to `also_console=None` fixed the failing test and no other test broke.

Not directly related to the above change, it seems that `also_console` handling in `robot.api.logger` could be removed and the given value could be passed to `librarylogger.write` and let the `Message` object handle it.

adombeck:
> I tested this locally and believe I found the reason for the failure. Messages logged by the `logging` module are forwarded via `pyloggingconf` to `librarylogger` and its `write` method has `also_console=False`. This means that messages logged using it aren't logged to the console by default, regardless the level. Changing it to `also_console=None` fixed the failing test and no other test broke.

Great! I can confirm that changing it to `also_console=None` fixes the test. Thanks for looking into it. I only had very limited time this week and didn't come to the right conclusions about that issue.
 
I pushed a commit with that change.

> Not directly related to the above change, it seems that `also_console` handling in `robot.api.logger` could be removed and the given value could be passed to `librarylogger.write` and let the `Message` object handle it.

Agreed. I also a pushed commit which does that.

I ran all unit and acceptance tests again and they all pass now, except for the  screenshot tests which also fail on master when I execute them locally.

Is there anything else blocking the merge of this PR? If not, I would like to rebase this on master and squash the fixup commits before the merge.

pekkaklarck:
I didn't look at rhe latest changes yet, but if tests niw pass this should be ready to be merged. GitHub by default squashes commits and rebases, but if there's still something you want to fine-tune I can wait a little time.

Notice that you can exclude Screenshot tests with something like `--exclude require-screenshot` if installing the needed dependencies is complicated or if those tests fail anyway. See `atest/README.rst` for details.

adombeck:
I rebased and squashed the fixup commits now. I made sure that the diff between the commit before the force push and after is empty (see `git diff e6f5675 ba1af11`).

src/robot/output/logger.py
```diff
@@ -202,7 +202,7 @@ def _log_message(self, msg, no_cache=False):
             and self._output_file.is_logged(msg)
         ):
             self._log_message_parents[-1].body.append(msg)
-        if msg.level in ("WARN", "ERROR"):
+        if msg.console or msg.level in ("WARN", "ERROR"):
```
pekkaklarck, line 205: This looks odd. Shouldn't it be enough to use `msg.console` also here? Could it be that this was needed when `librarylogger.write` still had `also_console=False`? I can test that locally after merging the PR.

pekkaklarck:
I did a little cleanup locally, including changing `also_console` to `console` and noticed a somewhat big problem with this PR. Normal messages, not WARN or ERROR messages, logged to console were handled differently depending on how they were logged. In some cases they were printed to `stdout` the same way as earlier, but in others they were written to `stderr` so that they got `[ INFO ]` prefix. The prefix makes it impossible to log a bigger message in pieces and the inconsistency was not good in general. I've been able to fix this so that all "normal" messages are logged to `stdout` and there's no prefix. In other words, the behavior is same earlier in this regard. I'll commit changes shortly.

PR #5465, opened 2025-06-28
Visualization of the robotframework

ivanmilevtues:
Hey, I have generated a diagram representation for the robotframework codebase. You can see how it renders here: 
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/robotframework/on_boarding.md

The idea of these diagrams is to be a codemap for new contributors - giving them high-level representation of the main components and now having context on how to start working on the component that they are interested. Seems like there are lots of people playing with your codebase (2.5k+ forks!).

I saw you have a custom docs generation process, we have introduced a free github action to keep the generations up-to-date, so I'd love to integrate that for you if you like the diagram first documentation!

Full transparency: we’re exploring this idea as a potential startup, but we’re still early and figuring out what’s actually useful to developers.

### DIFF
--- a/.codeboarding/Execution_Core.md
+++ b/.codeboarding/Execution_Core.md
@@ -1,0 +1,291 @@
+```mermaid
+
+graph LR
+
+    SuiteRunner["SuiteRunner"]
+
+    BodyRunner["BodyRunner"]
+
+    Namespace["Namespace"]
+
+    VariableManager["VariableManager"]
+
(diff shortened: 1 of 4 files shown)


ivanmilevtues:
A quick update on our side, we've just released our engine as an open-source repo: https://github.com/CodeBoarding/CodeBoarding

If you are interested on how exactly the whole generation works, you can check it out!

PR #5479, opened 2025-08-07
extract untrusted github.head_ref and github.ref_name inputs to environment variables

gcanlin:
Hi! GitHub treated `github.head_ref` and `GitHub.ref_name` as [the untrusted inputs](https://securitylab.github.com/resources/github-actions-untrusted-input/). Extracting them to env var will prevent some potential script injection even if `subprocess.check_call` seems could prevent command injection. But don't know whether there exists any unknown injections for it. So I think that extracting will be safer.

### DIFF
--- a/.github/workflows/acceptance_tests_cpython.yml
+++ b/.github/workflows/acceptance_tests_cpython.yml
@@ -94,6 +94,7 @@ jobs:
         if: failure()
         env:
           RFLOGS_API_KEY: ${{ secrets.RFLOGS_API_KEY }}
+          BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
         working-directory: atest/results
         shell: python
         run: |
@@ -116,7 +117,7 @@ jobs:
                   "--tag", f"workflow:${{ github.workflow }}",
                   "--tag", f"os:${{ runner.os }}",
                   "--tag", f"python-version:${{ matrix.python-version }}",
-                  "--tag", f"branch:${{ github.head_ref || github.ref_name }}",
+                  "--tag", f"branch:{os.environ['BRANCH_NAME']}",
                   result_dir
               ]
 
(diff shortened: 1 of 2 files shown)


gcanlin:
@pekkaklarck Could you please take a look for this PR? Thanks very much!

pekkaklarck:
Could you clarify how the current action code is vulnerable? I read the referenced post and understood that the problems can be related to disclosing secrets and command injection.`f"branch:{__import__('os').system(...)}"` 

The change is essentially this:

```diff
- "--tag", f"branch:${{ github.head_ref || github.ref_name }}",
+ "--tag", f"branch:{os.environ['BRANCH_NAME']}",
```
This apparently hides the ref values, but I don't think they are actually secret information. Is the risk that someone could construct a ref like `{__import__('os').system(...)}` that in the original case would be evaluated to `f"branch:{__import__('os').system(...)}"` by GA and then later evaluated as an f-string? That shouldn't happen in the second case because the supplied values is replaced as part of f-string evaluation, not by GA itself, so the end result would be a normal string `"branch:{__import__('os').system(...)}"`. If this is the case, wouldn't it be safer to just make those strings normal strings instead of f-strings to begin with? It seems that none of the `--tag` values actually need to be f-strings.

If you consider the environment variable approach better, I guess we can use that as well. It would nevertheless be a good idea to make strings that don't need to be f-strings normal ones to avoid potential issues with malicious f-string evaluation.

gcanlin:
Thanks for replying.
I think that there's no actual vulnerability in the current code. The concern from GitHub's guidance seems to be a general best practice about handling untrusted input in commands, rather than a specific f-string evaluation risk here.
In more details——As you said, the f-string itself indeed poses no risk of command injection. However, from the perspective of the entire pipeline, our initial concern was that a malicious input, once injected, could be used in other commands later and lead to malicious attacks. For example, if `subprocess.check_call(cmd, shell=True)` were used instead of `subprocess.check_call(list)`, it could result in the execution of malicious commands, even if the f-string is safe. In this case, since `subprocess.check_call(list)` is used, there is actually no vulnerability.

Nevertheless, I would still recommend using environment variables. If, in the future, other commands are added that use `github.head_ref` without proper safeguards against injection attacks, vulnerabilities could be introduced unintentionally.

pekkaklarck:
I guess using environment variables is a bit safer and works as an example for the future. I still believe Unnecessary f-strings should be removed, but I can do that later myself.

PR #5488, opened 2025-08-19
5373 type hints datetime

silentw0lf:


pekkaklarck:
One of the DateTime library tests fails with this error message:
> Expected error 'ValueError: Unsupported input 'None'.' but got 'ValueError: Invalid timestamp 'None'.'.

My expectation is that the keyword is used with Python `None` object and earlier the type has been validated internally and rejected as unsupported type. Now that the keyword has a type hint defined as a union with `str` as one of the accepted values, the given `None` is converted to a string and now the type isn't anymore invalid but the value itself is. I consider the new behavior fine so the fix is updating the test.

I recommend you to run at least DateTime library tests locally before committing changes.

pekkaklarck:
Notice that I reviewed this PR at https://github.com/silentw0lf/robotframework/pull/1/. I assume the code is same there so comments apply also to this PR. The other should be closed now.

silentw0lf:
> One of the DateTime library tests fails with this error message:
> 
> > Expected error 'ValueError: Unsupported input 'None'.' but got 'ValueError: Invalid timestamp 'None'.'.
> 
> My expectation is that the keyword is used with Python `None` object and earlier the type has been validated internally and rejected as unsupported type. Now that the keyword has a type hint defined as a union with `str` as one of the accepted values, the given `None` is converted to a string and now the type isn't anymore invalid but the value itself is. I consider the new behavior fine so the fix is updating the test.
> 
> I recommend you to run at least DateTime library tests locally before committing changes.

You are correct. The Python `None` object is no procosessed as a `str` type  within the python function/keyword `convert_time` (right from entering the the function, no conversion within).

src/robot/libraries/DateTime.py
```diff
@@ -304,9 +304,15 @@
 |     # ...
 """
 
+# can be removed when using Python 3.13+
+# needed for typing to use | operator together with forward referenceing
+# see: def __sub__(self, other: Date | Time) -> Date | Time:
+from __future__ import annotations
```
pekkaklarck: This future import should be removed. Instead of relaying to this future import that in practice stringifies type hints, we have stringified them manually when needed. That approach is also used in this PR in most cases and the remaining cases should be converted to use `"Date | Time"` format as well.

src/robot/libraries/DateTime.py
```diff
@@ -325,13 +331,20 @@
     "subtract_time_from_time",
 ]
 
+FloatOrString = Union[float, str]
+FloatOrStringOrInt = Union[FloatOrString, int]
```
pekkaklarck: These type aliases seem to be unnecessary as they are only used by the following type aliases. It's simpler and even shorter to create them using the base types directly.

src/robot/libraries/DateTime.py
```diff
@@ -325,13 +331,20 @@
...
+DatetimeMixed = Union[datetime.datetime, FloatOrString]
+DateTimeExpanded = Union[datetime.datetime, datetime.date, FloatOrStringOrInt]
+TimedeltaExpanded = Union[datetime.timedelta, FloatOrStringOrInt]
+TimedeltaMixed = Union[datetime.timedelta, FloatOrString]
```
pekkaklarck: "Mixed" and "Expanded" aren't that clear names. Looking at the changes, it seems that "Expanded" variants are used with arguments and "Mixed" with return types. I'd say that terms "Input" and "Output" would thus work better.

This library has a defined meaning for terms "date" and "time" explained in the documentation. In practice "date" maps to Python's `datetime` and "time" maps to `timedelta`. These generic "date" and "time" terms should also be used in type alias names.

Combining the above points, I believe these type aliases should be named "DateInput", "DateOutput", "TimeInput" and "TimeOutput".

src/robot/libraries/DateTime.py
```diff
@@ -549,16 +576,23 @@ def subtract_time_from_time(time1, time2, result_format="number", exclude_millis
...
 class Date:
-
-    def __init__(self, date, input_format=None):
-        self.datetime = self._convert_to_datetime(date, input_format)
+    def __init__(
+        self,
+        date: DateTimeExpanded,
+        input_format: "str|None" = None,
+    ):
+        self.datetime: datetime.datetime = self._convert_to_datetime(date, input_format)
```
pekkaklarck: 1. Please don't remove the empty line after the class declaration. Our coding style is to include it.
2. There's no need to annotate `self.datetime` because `self._convert_to_datetime` is annotated to return `datetime.datetime`.

src/robot/libraries/DateTime.py
```diff
@@ -637,48 +675,47 @@ def __sub__(self, other):
 
 
 class Time:
+    def __init__(self, time: TimedeltaExpanded):
+        self.seconds: float = self._convert_time_to_seconds(time)
```
pekkaklarck: See the comment for Date about the empty line after the class declaration and the need to annotate `self.seconds`.

src/robot/libraries/DateTime.py
silentw0lf: Done.

src/robot/libraries/DateTime.py
silentw0lf: Done.

src/robot/libraries/DateTime.py
silentw0lf: Done.

src/robot/libraries/DateTime.py
silentw0lf: Done.

src/robot/libraries/DateTime.py
silentw0lf: Done.

silentw0lf:
I did the Style changes. (also passed the atests)

I also answered to your comments here (the old wrong PR, that I will delete after merging this one here.)
https://github.com/silentw0lf/robotframework/pull/1

Do we need to adapt something like a changelog for now?


Otherwise, I can continue with String.py, followed by the other libs.

pekkaklarck:
Great, I hopefully have time to review the changes tomorrow. We don't keep a separate CHANGELOG file but instead create release notes template based on issues and then manually add more information to it.

src/robot/libraries/DateTime.py
```diff
@@ -325,13 +326,18 @@
     "subtract_time_from_time",
 ]
 
+DateOutput = Union[datetime.datetime, float, str]
+DateInput = Union[datetime.datetime, datetime.date, float, int, str]
+TimeInput = Union[datetime.timedelta, float, int, str]
```
pekkaklarck, line 331: `int` is most likely unnecessary here, because accepting `float` means that values like string `"2"` and integer `3` are accepted and will be converted to floats `2.0` and `3.0`, respectively. I'll test that locally after merging this PR and also look how this shows up in Libdoc output.

pekkaklarck (approved):
Great work!

src/robot/libraries/DateTime.py
pekkaklarck: I tested this locally and `int` is unnecessary. In the generated documentation listing it along with `float` makes it more explicit that integers are accepted, though, so I guess we can leave it there.

PR #5497, opened 2025-09-10
5373 type hints lib: [string]

silentw0lf:


pekkaklarck:
This has been replaced by #5554.

PR #5501, opened 2025-09-24
Fix #5500: event loop cannot re-initialize after finishing a suite

### LINKED ISSUE #5500: Robot API `run` fails after the first run
```python
from robot import run
def main():
    # Run first time, works fine
    exit_code = run("tests/example.robot")
    print(f"Robot Framework exited with code: {exit_code}")
    
    # Run second time, fails with "Event loop is closed" and "RuntimeWarning: coroutine 'MyLibrary.my_async_keyword' was never awaited" 
    exit_code = run("tests/example.robot")
    print(f"Robot Framework exited with code: {exit_code}")
```

maisonsmd:
This PR fixes #5500

## Issue

After `robot.run()` ends, `close_loop()` is called, closing the active event loop, but not resetting _loop_ref, so when the next call, it cannot reinitialize the event loop again

## Solution
reset _loop_ref to None after closing the loop

### DIFF
--- a/src/robot/running/context.py
+++ b/src/robot/running/context.py
@@ -35,6 +35,7 @@ class Asynchronous:
     def close_loop(self):
         if self._loop_ref:
             self._loop_ref.close()
+            self._loop_ref = None
 
     def run_until_complete(self, coroutine):
         task = self.event_loop.create_task(coroutine)


pekkaklarck:
The fix looks good but needs a test. Adding one can be pretty complicated and I can take a lot ok at that myself as well.

Issue #3845, opened 2021-02-06
Support removing keywords with `NOT RUN` status using `--removekeywords`


adiralashiva8:
As a user I want to hide/remove `NOT_RUN` status keywords from `log.html` to avoid confussion for other teammates

Points Discussed in slack (#devel)
 - `--removekeywords` to support removing `NOT_RUN` status keywords similar to `PASSED`
 - `IF/ ELSE/ ELSE IF` should not be excluded while using `--removekeywords`

pekkaklarck:
I don't consider this important enough to be added in already late RF 4.0. Can be included in RF 4.1 if there are concrete needs.

pekkaklarck:
Being able to remove `NOT RUN` keywords using `--removekeywords` could be handy. Not super important but if someone would be willing to provide a pull request the functionality could certainly be included. I'll update the title accordingly. Possible other way to exclude these keywords and control structures can get their own issues.

pekkaklarck:
For the reasoning why we have keywords with the `NOT RUN` status see #3842.

LowEQ:
let me try to implement this

GauravT-07:
Hi @pekkaklarck! I'm interested in working on this issue as my first contribution to Robot Framework. Is this issue still available? If so, I'd be happy to contribute.

PR #5507, opened 2025-10-02
Added NotRun Keyword Remover functionality

Begleco:
Expanded functionality of "--removekeywords" to also remove keywords with the "NOT RUN" status as per the request in issue #3845.

### DIFF
--- a/src/robot/result/keywordremover.py
+++ b/src/robot/result/keywordremover.py
@@ -40,11 +40,12 @@ class KeywordRemover(SuiteVisitor, ABC):
                 "FOR": ForLoopItemsRemover,
                 "WHILE": WhileLoopItemsRemover,
                 "WUKS": WaitUntilKeywordSucceedsRemover,
+                "NOT_RUN": NotRunKeywordRemover,
             }[upper]()
         except KeyError:
             raise DataError(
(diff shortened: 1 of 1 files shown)


src/robot/result/keywordremover.py
```diff
@@ -172,6 +173,13 @@ def _remove_keywords(self, body):
                     body.remove(kw)
 
 
+class NotRunKeywordRemover(KeywordRemover):
+    def visit_test(self, test):
+        before = len(test.body)
+        test.body = [kw for kw in test.body if kw.status != "NOT RUN"]
+        self.removal_message.set_to_if_removed(test, before)
```
pekkaklarck, line 180: This only removes keywords from test body, not from child keywords or control structures. The easiest way to extend the functionality would be implementing `start_test` instead of `visit_test` (that way also body is visited automatically) and implementing also `start_body_item` that's, by default, called with all keywords and control structures.

pekkaklarck (changes requested):
Few things to be done:

- As I mentioned in the separate comment, the current implementation removes keywords only from test body, not from child keywords and control structures.
- Tests are needed. `atest/README.rst` explains the test system and I can help here or on the #devel channel on our Slack if you have questions.
- User Guide documentation is needed. I can help with that as well if needed.

PR #5511, opened 2025-10-09
fix minor error in get_variable_value docstring

oboehmer:
Fixing minor docstring error in `BuiltIn.get_variable_value`

don't think it's worth filing an issue for this, @pekkaklarck


pekkaklarck:
Thanks for the fix! I don't think this deserves a separate issue.

Issue #5512, opened 2025-10-10
`Secret` support to OperatingSystem and Process library keywords


Nilsty:
I have been trying out the [new secrets variable feature](https://github.com/robotframework/robotframework/blob/master/doc/releasenotes/rf-7.4b1.rst#secret-variables) in Robot Framework 7.4 beta 1.

I'm running into the issue that I need to write the contents of the secret variable into a file, but I'm getting the error `TypeError: write() argument must be str, not Secret`

My use case is that I have a google cloud credentials json string, which I need to write into a file to use it with a gcloud command.
e.g.
```
Setup Gcloud
    VAR    ${GCLOUD_CRED_FILE}    ${TEMPDIR}${/}gcp-qa-sa.json
    Create File    ${GCLOUD_CRED_FILE}    ${GCLOUD_CRED_FILE_CONTENT}
    ${gloud_info}    Run
    ...    gcloud auth activate-service-account ${GCLOUD_SERVICE_ACCOUNT} --key-file=${GCLOUD_CRED_FILE}
```

The credentials file content is injected via an environment variable like this:
```
*** Variables ***
${GCLOUD_CRED_FILE_CONTENT: Secret}     %{GCLOUD_CRED_FILE_CONTENT}
```

[Related thread in RobotFramework Slack](https://robotframework.slack.com/archives/C3C28F9DF/p1760026656659189)

pekkaklarck:
Very good point that Create File and probably also various other keywords should accept Secret values! Ping @aaltat.

pekkaklarck:
As I already commented, we should add `Secret` support to `Create File` and to various other keywords. Adding `Secret` support to _all_ standard library keywords probably wouldn't make sense and would be too big a task for RF 7.4 anyway, but let's use this issue to collect keywords that would actually benefit from the `Secret` support. I hope community members can help with that!

pekkaklarck:
Adding the `Secret` support is pretty easy, but needs to be done separately with each keyword. It consists of the following steps:
1. Add test or tests.
2. Handle `Secret` values in code with something like `if isinstance(value, Secret): value = value.value`.
3. Make absolutely sure secret values are not logged! Adjusts tests to validate this.
4. Mention the `Secret` support in keyword documentation if it makes sense.

I can do this myself for keywords considered important to have the `Secret` support, but pull requests are appreciated as well.

oboehmer:
[...]

PR #5516, opened 2025-10-17
add support for Secret variables to selected OperatingSystem lib keywords

oboehmer:
relates to #5512  

Added support for `Secret` variables to the following keywords in `OperatingSystem`

```
Create File
Append to File
Set Environment Variable
Append to Environment Variable
```

### DIFF
--- a/src/robot/libraries/OperatingSystem.py
+++ b/src/robot/libraries/OperatingSystem.py
@@ -581,7 +581,7 @@ class OperatingSystem:
 
     # Creating and removing files and directory
 
-    def create_file(self, path: "str | Path", content: "str | Secret" = "", encoding: str = "UTF-8"):
+    def create_file(self, path: "Path", content: "str | Secret" = "", encoding: str = "UTF-8"):
         """Creates a file with the given content and encoding.
 
         If the directory where the file is created does not exist, it is
@@ -646,7 +646,7 @@ class OperatingSystem:
         path = self._write_to_file(path, content, mode="wb")
         self._link("Created binary file '%s'.", path)
 
-    def append_to_file(self, path: "str | Path", content: "str | Secret" = "", encoding: str = "UTF-8"):
+    def append_to_file(self, path: "Path", content: "str | Secret" = "", encoding: str = "UTF-8"):
         """Appends the given content to the specified file.
 
         If the file exists, the given text is written to its end. If the file


src/robot/libraries/OperatingSystem.py
```diff
@@ -602,7 +603,12 @@ def create_file(self, path, content="", encoding="UTF-8"):
         and `Create Binary File` if you need to write bytes without encoding.
         `File Should Not Exist` can be used to avoid overwriting existing
         files.
+
+        This keyword supports passing a `Secret` variable as content (new in
+        RobotFramework 7.4)
```
pekkaklarck: I have been thinking should we add this not to all keywords accepting `Secret`. It kind of feels unnecessary, but seeing this not "live" I believe adding it makes sense. There are some changes to be done, though:
- There should be double backticks around `Secret` because the single backtick version is used for internal linking.
- There should be a `.` at then of the sentence and "Robot Framework" is two words (at least in our documentation).
- We haven't in general used parenthesis around these "New in RF X.Y." notes.

In practice something like this would be fine:
```
This keyword supports passing a ``Secret`` variable as content. New in Robot Framework 7.4.
```

We could also consider adding a section about this topic to the general library documentation and then use the internal linking syntax to refer to it. I don't think that's worth the effort, though, especially when we are planning to add type hints to standard library documentation possibly already in RF 7.4 (#5373) and that gives us `Secret` documentation automatically.

pekkaklarck: After thinking this a bit more, having `Secret` in type hints and the automatic `Secret` documentation link it gives us makes the whole note about the `Secret` support mostly redundant. We have three options:
1. Add the note now and possibly remove it later when type hints are added.
2. Don't add the note, Just wait for the type hints to be added later.
3. Don't add the note. Add type hints when adding `Secret` support in general.

I think option 3. would be best, because then there's no extra work afterwards and the `Secret` support is automatically adequately documented. There are few things to take into account if that's done, though:
- Type hints should only be added to keywords that get the `Secret` support. Adding other type hints is covered by #5373. This may leave libraries into an inconsistent state if type hints aren't added in RF 7.4, but I don't think that would be a big problem.
- It would good to add type hints to all arguments to the keyword that gets the `Secret` support to make individual keywords consistent. If this gets tricky, type hints can be omitted with other arguments.

src/robot/libraries/OperatingSystem.py
```diff
@@ -1003,12 +1019,18 @@ def append_to_environment_variable(self, name, *values, separator=os.pathsep):
...
         """
         sentinel = object()
         initial = self.get_environment_variable(name, sentinel)
         if initial is not sentinel:
             values = (initial, *values)
-        self.set_environment_variable(name, separator.join(values))
+        self.set_environment_variable(
+            name,
+            separator.join([v.value if isinstance(v, Secret) else v for v in values]),
+        )
```
pekkaklarck: This gets a bit complicated. Perhaps adding
```python
values = [v.value if isinstance(v, Secret) else v for v in values]
```
above the existing
```python
self.set_environment_variable(name, separator.join(values))
```
line would be simpler.

atest/resources/atest_resource.robot
```diff
@@ -430,3 +430,13 @@ Traceback Should Be
     # Remove '~~~^^^' lines.
     ${msg.message} =    Evaluate    '\\n'.join(line for line in $msg.message.splitlines() if line.strip('~^ ') or not line)
     Check Log Message    ${msg}    ${exp}\n${error}    DEBUG    pattern=True    traceback=True
+
+Check for Secret Value Not in Log Messages
+    [Arguments]    ${tc}    ${value}
+    FOR    ${kw}    IN    @{tc.body}
+        FOR   ${i}    ${msg}    IN ENUMERATE    @{kw.messages}
```
pekkaklarck: There are two problems with this keyword:
- Model objects representing control structures like `FOR` don't have the `messages` attribute.
- This doesn't recurse to child keywords.

Neither of the issues would be an issue in a local keyword when the used data is known. A generic keyword in `atest_resource.robot` should be more generic, though.

atest/resources/atest_resource.robot
```diff
@@ -430,3 +430,13 @@ Traceback Should Be
...
+            Should Not Contain     ${value}     ${msg.message}
+            ...    msg=Keyword "${kw.name}" logged the secret (log index ${i})
+            ...    values=${FALSE}
```
pekkaklarck: `values=False` works as well, no need to use `${FALSE}`.

atest/resources/atest_resource.robot
oboehmer: Agreed, will add it back into the atest/robot test suite.. can I add it there as a keyword to avoid code duplication?

src/robot/libraries/OperatingSystem.py
oboehmer: I would also opt for (3). Do you have a preference on how to add type hints for standard library keywords? Using `typing.Optional`, ` | None`, other preferences?

src/robot/libraries/OperatingSystem.py
oboehmer: done

atest/resources/atest_resource.robot
oboehmer: ack (thanks to `is_falsy`, I love this helper!).  what do you prefer?

src/robot/libraries/OperatingSystem.py
pekkaklarck: Conventions ought to be documented in the Contribution guidelines. In practice this should be `content: "str | Secret"`. If an argument can have `None` as a value, then add `| None`.

atest/resources/atest_resource.robot
pekkaklarck: This generic resource file exists to avoid duplication. In some contexts there are separate resource files like `libdoc_resource.robot` (that should all be renamed to so that `_resource.robot` is replaced with `.resource`). If this keyword is used widely, this is the right place, but the above things need to be taken into account.

Recursion is as easy as calling the keyword itself with each child item. Handling non-keyword items most likely requires an additional condition like `IF    hasattr($item, "messages")`. Changing `@{item.messages}` to `@{{getattr($item, "messages", [])}}` would probably work as well. It would be somewhat complex itself, but it would avoid the IF.

atest/resources/atest_resource.robot
pekkaklarck: This conversion is actually done by automatic argument conversion based on the default value. If `is_falsy` is still used by this keyword, it ought to be redundant. I'd prefer `False` over `${False}` but it's not a biggie.

src/robot/libraries/OperatingSystem.py
oboehmer: ok, I gave it a shot, please review..

oboehmer: not sure if you prefer `pathlib.Path` over `Path` in the type hints.. nowadays `Path` is so well known and widely used, hence I used this.. please advise..

pekkaklarck: I prefer using just `Path` over `parhlib.Path`. Not sure why the code used the latter earlier, but changing it is certainly fine.

Could you still try using just `Path`, not `str | Path` as the type hint? That way argument conversion converts strings to `Path` objects automatically. We then could switch to using `Path` instead if `os.path` in general, but that's out of the scope of this PR.

oboehmer: done, thanks.. if you are happy, please merge, so I can finalize #5519

pekkaklarck:
Looks good to me. Some final comments:
1. I'm not sure does it make sense in tests to go through all log messages to see that the secret value hasn't leaked. None of these keywords log the value in the first place, so there shouldn't be any risk with that. Secret values not being exposed by Robot when tests are run on the TRACE level ought to be tested elsewhere and there's no need to test that here again. The only benefit I see is that we could in the future modify keywords so that they blindly log the value, but I don't think that's too likely. There's also a problem that tests is validating that a hard coded string doesn't exist and that test will never fail if the actual secret value is changed.
2. We prefer keywords after tests, not before them, in suite files.
3. Some of the added code lines look rather long. Did you run `invoke format`? See Contribution guidelines for details.

After thinking this a bit, I believe going through all messages to see are secrets exposed isn't worth the effort. Because I didn't comment about that earlier, I'll make the related changes myself after merging this. That also fixes the second point above and I can also run `invoke format` before my commit. Just take these comments into account with the next PR.

pekkaklarck:
Hmm, `invoke format` reformatted code in the added library to this:
```python
    assert (
        SECRET_VALUE in content
    ), "The secret value is not present in the file's content"
```
That looks odd to me. I'll refactor this to use `if condition: raise AssertionError` instead of `assert`.

pekkaklarck:
While making the cleanup I mentioned above, I noticed there's actually a _very_ bad bug in the `set_environment_variable` code. The secret value is actually logged!
```python
if isinstance(value, Secret):
    value = value.value
set_env_var(name, value)
self._info(f"Environment variable '{name}' set to value '{value}'.")
```
Apparently the keyword that was supposed look for this didn't work correctly. This proves my earlier comment that the keyword was somewhat fragile. I noticed this when I changed tests to validate the actual logged value and to my surprise the secret was there.

I don't understand how I missed this in my earlier reviews. Need to be more careful in the future!

oboehmer:
> While making the cleanup I mentioned above, I noticed there's actually a _very_ bad bug in the `set_environment_variable` code. The secret value is actually logged!
> 
> ```python
> if isinstance(value, Secret):
>     value = value.value
> set_env_var(name, value)
> self._info(f"Environment variable '{name}' set to value '{value}'.")
> ```
> 
> Apparently the keyword that was supposed look for this didn't work correctly. This proves my earlier comment that the keyword was somewhat fragile. I noticed this when I changed tests to validate the actual logged value and to my surprise the secret was there.
> 
> I don't understand how I missed this in my earlier reviews. Need to be more careful in the future!

Oh, bummer. I am sorry for missing this one.

PR #5524, opened 2025-10-21
Add Secret argument support for `Run Process` and `Start Process` keywords

oboehmer:
`Secret` arguments are supported in `Run Process` and `Start Process` for 

- regular arguments
- `stdin` argument
- variables used in `env` and `env:<name>` arguments

Part of #5512

### DIFF
--- a/atest/testdata/standard_libraries/process/secrets.robot
+++ b/atest/testdata/standard_libraries/process/secrets.robot
@@ -3,7 +3,8 @@ Suite Setup       Set Environment Variable    v1    system
 Resource          process_resource.robot
 
 *** Variables ***
-@{COMMAND}        python    -c    import os; print(' '.join([os.getenv('v1', '-'), os.getenv('v2', '-'), os.getenv('v3', '-')]))
+@{COMMAND}           python    -c    import os; print(' '.join([os.getenv('v1', '-'), os.getenv('v2', '-'), os.getenv('v3', '-')]))
+${SECRET: Secret}    %{=This is secret!}
 
 *** Test Cases ***
 Run Process with Secret Argument
@@ -18,6 +19,15 @@ Start Process with Secret Argument
     ${handle} =    Start Process    python    ${SCRIPT}    ${SECRET}
     ${result} =    Wait For Process    ${handle}
     Script result should equal    ${result}    stdout=This is secret!
+
+Stdin as Secret
+    ${result} =    Run Process    python    -c    import sys; print(sys.stdin.read())    stdin=${SECRET}
+    Should Be Equal    ${result.stdout}    This is secret!
+
+Start Process with Stdin as Secret
+    ${handle} =    Start Process    python    -c    import sys; print(sys.stdin.read())    stdin=${SECRET}
+    ${result} =    Wait For Process    ${handle}
+    Should Be Equal    ${result.stdout}    This is secret!
 
 Secret in environment variable via env Dict
     ${env} =    Create environ    v1    ${SECRET}
--- a/atest/testdata/standard_libraries/process/secrets.robot
+++ b/atest/testdata/standard_libraries/process/secrets.robot
@@ -17,7 +17,7 @@ Run Process with Mixed Arguments Including Secret
 
 Run Process with Stdin as Secret
     ${result} =    Run Process    python    -c    import sys; print(sys.stdin.read())    stdin=${SECRET}    cwd=%{TEMPDIR}
-    Should Be Equal    ${result.stdout}    This is secret!
+    Script result should equal    ${result}    stdout=This is secret!    stderr=
 
 Start Process with Secret Argument
     ${handle} =    Start Process    python    ${SCRIPT}    ${SECRET}
@@ -27,17 +27,17 @@ Start Process with Secret Argument
 Start Process with Stdin as Secret
     ${handle} =    Start Process    python    -c    import sys; print(sys.stdin.read())    stdin=${SECRET}    cwd=%{TEMPDIR}
     ${result} =    Wait For Process    ${handle}
-    Result should equal    ${result}    stdout=This is secret!
+    Script result should equal    ${result}    stdout=This is secret!    stderr=
 
 Secret in environment variable via env Dict
     ${env} =    Create environ    v1=${SECRET}
     ${result} =    Run Process    @{COMMAND}    env=${env}
-    Result should equal    ${result}    stdout=This is secret! - -
+    Script result should equal    ${result}    stdout=This is secret! - -    stderr=
 
 Secret in environment variable via env:name Syntax
     ${result} =    Run Process    @{COMMAND}    env:v2=${SECRET}
-    Result should equal    ${result}    stdout=system This is secret! -
+    Script result should equal    ${result}    stdout=system This is secret! -    stderr=
 
 Multiple Secrets in environment variables
     ${result} =    Run Process    @{COMMAND}    env:v1=${SECRET}    env:v2=XX    env:v3=${SECRET}
-    Result should equal    ${result}    stdout=This is secret! XX This is secret!
+    Script result should equal    ${result}    stdout=This is secret! XX This is secret!    stderr=
--- a/atest/testdata/standard_libraries/process/secrets.robot
+++ b/atest/testdata/standard_libraries/process/secrets.robot
@@ -15,17 +15,17 @@ Run Process with Mixed Arguments Including Secret
     ${result} =    Run Process    python    ${SCRIPT}    public_arg    ${SECRET}
     Script result should equal    ${result}    stdout=public_arg    stderr=This is secret!
 
+Run Process with Stdin as Secret
+    ${result} =    Run Process    python    -c    import sys; print(sys.stdin.read())    stdin=${SECRET}    cwd=%{TEMPDIR}
+    Should Be Equal    ${result.stdout}    This is secret!
+
 Start Process with Secret Argument
     ${handle} =    Start Process    python    ${SCRIPT}    ${SECRET}
     ${result} =    Wait For Process    ${handle}
     Script result should equal    ${result}    stdout=This is secret!
 
-Stdin as Secret
-    ${result} =    Run Process    python    -c    import sys; print(sys.stdin.read())    stdin=${SECRET}
-    Should Be Equal    ${result.stdout}    This is secret!
-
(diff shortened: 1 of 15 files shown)


src/robot/libraries/Process.py
```diff
@@ -333,19 +335,19 @@ def __init__(self):
...
-        timeout=None,
-        on_timeout="terminate",
-        env=None,
-        **env_extra,
+        command: str,
+        *arguments: "str | Secret",
+        cwd: "str | None" = None,
+        shell: bool = False,
+        stdout: "str | None" = None,
+        stderr: "str | None" = None,
```
pekkaklarck, line 343: `stdout/stderr` probably should be `Path`, not `str`. It may require code changes as well, though, and that's out of the scope of this PR. I'd be fine with not adding type hints to arguments that don't need `Secret` in this case as well.

src/robot/libraries/Process.py
```diff
@@ -333,19 +335,19 @@ def __init__(self):
...
+        stdin: "str | Path | Secret | Any" = None,
```
pekkaklarck: I believe `stdin` could be `int` and then `Any` wouldn't be needed (but `None` would also need to be added).

src/robot/libraries/Process.py
```diff
@@ -455,6 +457,15 @@ def start_process(
         Earlier versions returned a generic handle and getting the process object
         required using `Get Process Object` separately.
         """
+        # Check for secrets in arguments and unwrap them
+        secret = any(isinstance(arg, Secret) for arg in arguments)
+        if secret:
```
pekkaklarck: I don't think the comment is needed, the code is clear enough without it. `has_secrets` could also be a bit better name than `secret`.

src/robot/libraries/Process.py
```diff
@@ -455,6 +457,15 @@ def start_process(
...
+            masked_arguments = [
+                "<redacted>" if isinstance(arg, Secret) else arg for arg in arguments
```
pekkaklarck: The `Secret` class itself uses `<secret>` in place of the actual value and for consistency it should be used here as well. An alternative would be using `str(arg)` and use the string representation of the Secret class, but that's currently `Secret(value=<secret>)` which doesn't look good to me in this context. As I just commented on Slack, we could consider changing the string representation, though.

src/robot/libraries/Process.py
```diff
@@ -466,9 +477,13 @@ def start_process(
             env=env,
             **env_extra,
         )
-        command = conf.get_command(command, list(arguments))
-        self._log_start(command, conf)
-        process = subprocess.Popen(command, **conf.popen_config)
+        actual_command = conf.get_command(command, list(arguments))
```
pekkaklarck: This variable could still be `command`, it's clear enough if the other is `masked_command`. Using `actual_command` is fine as well if you prefer that. The `list()` call seems to be totally unnecessary. Not sure why the original code had it.

src/robot/libraries/Process.py
```diff
@@ -1075,18 +1093,32 @@ def _construct_env(self, env, extra):
...
     def _get_initial_env(self, env, extra):
         if env:
-            return {system_encode(k): system_encode(env[k]) for k in env}
+            result = {}
+            for name, value in env.items():
+                name = system_encode(name)
+                if isinstance(value, Secret):
+                    result[name] = system_encode(value.value)
+                    # remember secrets
+                    self.env_secret_keys.append(name)
```
pekkaklarck: The comment is unnecessary. Perhaps `secret_env_keys` instead of `env_secret_keys`?

src/robot/libraries/Process.py
```diff
@@ -1125,14 +1157,20 @@ def result_config(self):
         }
 
     def __str__(self):
+        if self.env is None:
+            printable_env = None
+        else:
+            printable_env = self.env.copy()
+            for k in self.env_secret_keys:
+                printable_env[k] = "<redacted>"
```
pekkaklarck: `<redacted>` -> `<secret>`

pekkaklarck (changes requested):
Excellent work! The only thing that really should be changed is using `<secret>` instead of `<redacted>` for consistency reasons. Other comments are mostly about style and some got pretty close to nitpicking...

src/robot/libraries/Process.py
oboehmer: Windows passes a path name (instead of an int), noticed this when I wanted to check stdin process config
```
Process configuration:
cwd:     C:\Users\RUNNER~1\AppData\Local\Temp\robotatest\Python-3.13.8-Windows
shell:   False
stdout:  PIPE
stderr:  PIPE
stdin:   C:\Users\RUNNER~1\AppData\Local\Temp\tmp9h3hy1kx
alias:   None
env:     None
```

pekkaklarck: `stdin` certainly can be a string. I just would like to avoid using `Any` and believe it could be removed if `int` and `None` were added. `int` will be used when `stdin` is `PIPE` or a file descriptor. If this gets complicated, leave `Any` and I look at this myself when adding type hints to this library later.

src/robot/libraries/Process.py
oboehmer: Well, both stdout and stderr also get discrete string values like `DEVNULL` or `STDOUT`, while there is no provision to convert path.. I will add `Path`, and handle `Path` being passed.

src/robot/libraries/Process.py
oboehmer: changed to 
```
        stdin: "str | Path | Secret | int | None" = None,
```

src/robot/libraries/Process.py
oboehmer: replacing this by `<secret>` for the time being, as discussed on Slack

src/robot/libraries/Process.py
pekkaklarck: Very good point that we support strings like `DEVNULL` and `STDOUT` so we cannot use just `Path` and that the code needs to specially handle paths passes as strings. Perhaps having `Path` there anyway makes sense to point out that you can use a path as a value.

I think `suprocess.Popen` would also support integer values (`suprocess.PIPE`, file descriptors), but I don't think our code handles them anyway. If no tests fail with these changes, no need to add `int`.

pekkaklarck: The code ought to already support using `Path`. If not, it probably should be fixed separately and in that case it's not a good idea to add `Path` typing.

oboehmer: agreed, will not add `Path` to stdin/stderr as part of this PR.

src/robot/libraries/Process.py
oboehmer: ack, will adjust.

src/robot/libraries/Process.py
oboehmer: Can it? as I reference the original value two lines down:
```
masked_command = conf.get_command(command, list(masked_arguments))
```
and I guess I want to use the original command, not the result from `conf.get_command(command, ...)`? I don't think I fully understand the subtleties of this handling ;-) I'll leave actual_command, and remove the `list()`

src/robot/libraries/Process.py
oboehmer: ack

src/robot/libraries/Process.py
oboehmer: Actually adding `int` will cause stdin.robot test to fail with
```
<_IO.BUFFEREDREADER NAME=5> != HELLO, WORLD!
``` 
any suggestion on how to fix this?

oboehmer: fixed with using `io.IOBase` hint

src/robot/libraries/Process.py
pekkaklarck: I may have read the code wrong. It's certainly fine as it is.

atest/testdata/standard_libraries/process/secrets.robot
```diff
@@ -0,0 +1,51 @@
...
+    ${result} =    Run Process    @{COMMAND}    env:v1=${SECRET}    env:v2=XX    env:v3=${SECRET}
+    Should Be Equal    ${result.stdout}    This is secret! XX This is secret!
+
+*** Keywords ***
+Create environ
+    [Arguments]    @{environ}
+    ${path} =    Get Environment Variable    PATH    default=.
+    ${systemroot} =    Get Environment Variable    SYSTEMROOT    default=.
+    ${environ} =    Create Dictionary    @{environ}    PATH=${path}    SYSTEMROOT=${SYSTEMROOT}
+    RETURN    ${environ}
```
pekkaklarck: This seems to be copied from `env_config.robot`. To avoid duplication, it would be better to move it `process_resource.robot` that both of these files use.

The keyword could also be changed to accept `&{environ}` instead of `@{environ}` so that its usage would be like `Create environ    name=value` instead of `Create environ    name    value`. That would require making changes to `env_config.robot` but changes would be trivial,

atest/testdata/standard_libraries/process/secrets.robot
```diff
@@ -0,0 +1,51 @@
...
+    ${result} =    Run Process    python    ${SCRIPT}    ${SECRET}
+    Script result should equal    ${result}    stdout=This is secret!
+
+Run Process with Mixed Arguments Including Secret
+    ${result} =    Run Process    python    ${SCRIPT}    public_arg    ${SECRET}
+    Script result should equal    ${result}    stdout=public_arg    stderr=This is secret!
+
+Run Process with Stdin as Secret
+    ${result} =    Run Process    python    -c    import sys; print(sys.stdin.read())    stdin=${SECRET}    cwd=%{TEMPDIR}
+    Should Be Equal    ${result.stdout}    This is secret!
```
pekkaklarck: Using `Result Should Equal` would make sure also stderr and rc are validated.

atest/testdata/standard_libraries/process/secrets.robot
oboehmer: ack, done..

atest/testdata/standard_libraries/process/secrets.robot
oboehmer: done, made all changes

pekkaklarck:
Looks great! I don't see anything to be changed when looking this on GitHub, but I'll test this locally after merge and may make small adjustments.

PR #5526, opened 2025-10-27
Fix: Handle empty string in `_convert_to_index`

ramesh2rao:
The `_convert_to_index` helper method in the `String` library incorrectly converted an empty string to `0`, causing slicing operations like `get_substring` and `split_to_lines` to fail when an empty string was used as the `end` index.

This commit changes `_convert_to_index` to return `None` for an empty string, which correctly slices to the end of the sequence.

Added unit tests for `get_substring` and `split_to_lines` to verify the fix.

### DIFF
--- a/src/robot/libraries/String.py
+++ b/src/robot/libraries/String.py
@@ -820,7 +820,7 @@ class String:
 
     def _convert_to_index(self, value, name):
         if value == "":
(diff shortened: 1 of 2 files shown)


ramesh2rao:
pull request approved

Issue #5527, opened 2025-10-29
Add annotations to Collections.py


robinmackaij:
This issue is part of the annotation effort on the standard libraries.

The linked PR highlights some points of attention that need a decision before finalizing the PR.

Main points of attention:

### Keyword signature consistency & automatic type conversion
The keywords in Collections mainly deal with lists and dictionaries. However, narrow typing in combination with automatic type conversion may result in undesirable behavior.

For example: `Append To List` is currently annotated as `list[object]`. If another type of `Sequence` (or perhaps even `Iterable` is passed as an argument, automatic type conversion could convert the argument to a list and then in-place modify that list.

Note that this module also has the `Normalizer` that has methods that explicitly try to retain the original type of the arguments during normalization. An example of a keyword that uses this normalization is `List Should Contain Value` that currently is annotated to accept a `list[object]`. If automatic type conversion is applied, it seems to me that trying to retain type during normalization is not needed.

Since signatures are used to generate the documentation for our users, I can see a preference for narrow typing, but I see a risk with automatic type conversion that might not be desirable for all keywords.

### List item and dictionary key/value annotations
During the annotation effort, I found that some keywords require more narrow item types than others.
Most noticeably, any keyword that relies on normalization requires the values to be sortable. So we have some keywords that work with items / values that are not sortable while other keywords do not.

### Dict key annotation
Python requires the keys of a dict to be `Hashable`. This might be a bit too technical for our uers, but using `str` annotation might be problematic if automatic type conversions changes e..g. `int` keys to their string representation.

> Note: I added "formal" annotations (no pyright errors with just 4 suppressions for issues that cannot be addressed before Python 3.10) resulting in a number of TypeVars / generics. Considering the annotated code is unlikely to be called from Python, the value of those generics is limited but they do highlight / reveal the issues mentioned above.

pekkaklarck:
First of all, I made this a sub-issue of #5373. Never used that functionality earlier, but it seems to work fine. I guess it then doesn't make sense to add labels or milestone, because in the release notes we only want to show the parent issue and we don't want the tool that generates the release notes template to "see" this. Alternatively the tool could be enhanced to automatically skip sub-issues. Ping @aaltat.

pekkaklarck:
[...]

PR #5528, opened 2025-10-29
Add annotations to collections

robinmackaij:
See #5527

### DIFF
--- a/src/robot/libraries/Collections.py
+++ b/src/robot/libraries/Collections.py
@@ -18,7 +18,7 @@ from collections.abc import Hashable
 from itertools import chain
 from typing import (
     Any, Iterable, Iterator, Literal, Mapping, MutableMapping, MutableSequence,
-    NoReturn, overload, Sequence
+    NoReturn, overload, Sequence, Union
 )
 
 from robot.api import logger
@@ -32,6 +32,7 @@ NOT_SET = NotSet()
 
 IgnoreKeyCase = Literal["KEY", "KEYS"]
 IgnoreValueCase = Literal["VALUE", "VALUES"]
+ListLike = Union[Sequence, Mapping, set]
 
 
 class _List:
@@ -82,7 +83,7 @@ class _List:
         """
         list_.insert(index, value)
 
-    def combine_lists(self, *lists: Sequence) -> list:
+    def combine_lists(self, *lists: ListLike) -> list:
         """Combines the given ``lists`` together and returns the result.
 
         The given lists are not altered by this keyword.
@@ -156,7 +157,7 @@ class _List:
         except IndexError:
             self._index_error(list_, index)
 
-    def remove_duplicates(self, list_: Sequence) -> list:
+    def remove_duplicates(self, list_: ListLike) -> list:
         """Returns a list without duplicates based on the given ``list``.
 
         Creates and returns a new list that contains all items in the given
@@ -317,7 +318,7 @@ class _List:
 
     def list_should_contain_value(
         self,
-        list_: Sequence,
+        list_: ListLike,
         value: object,
         msg: "str | None" = None,
         ignore_case: bool = False,
@@ -339,7 +340,7 @@ class _List:
 
     def list_should_not_contain_value(
         self,
-        list_: Sequence,
+        list_: ListLike,
         value: object,
         msg: "str | None" = None,
         ignore_case: bool = False,
@@ -392,8 +393,8 @@ class _List:
 
     def lists_should_be_equal(
         self,
-        list1: Sequence,
-        list2: Sequence,
+        list1: ListLike,
+        list2: ListLike,
         msg: "str | None" = None,
         values: bool = True,
         names: "Mapping[int, str] | Sequence[str] | None" = None,
@@ -487,8 +488,8 @@ class _List:
 
     def list_should_contain_sub_list(
         self,
-        list1: Sequence,
-        list2: Sequence,
+        list1: ListLike,
+        list2: ListLike,
         msg: "str | None" = None,
         values: bool = True,
         ignore_case: bool = False,
@@ -1115,7 +1116,7 @@ class Collections(_List, _Dictionary):
 
     def should_contain_match(
         self,
-        list: Sequence,
+        list: ListLike,
         pattern: str,
         msg: "str | None" = None,
         case_insensitive: "bool | None" = None,
@@ -1176,7 +1177,7 @@ class Collections(_List, _Dictionary):
 
     def should_not_contain_match(
         self,
-        list: Sequence,
+        list: ListLike,
         pattern: str,
         msg: "str | None" = None,
         case_insensitive: "bool | None" = None,
@@ -1202,7 +1203,7 @@ class Collections(_List, _Dictionary):
 
     def get_matches(
         self,
-        list: Sequence,
+        list: ListLike,
         pattern: str,
         case_insensitive: "bool | None" = None,
         whitespace_insensitive: "bool | None" = None,
@@ -1230,7 +1231,7 @@ class Collections(_List, _Dictionary):
 
     def get_match_count(
         self,
-        list: Sequence,
+        list: ListLike,
         pattern: str,
         case_insensitive: "bool | None" = None,
         whitespace_insensitive: "bool | None" = None,
@@ -1349,6 +1350,9 @@ class Normalizer:
     def normalize(self, value: Sequence) -> Sequence: ...
 
     @overload
+    def normalize(self, value: ListLike) -> ListLike: ...
+
+    @overload
     def normalize(self, value: object) -> object: ...
 
     def normalize(self, value: Any) -> Any:
@@ -1365,7 +1369,7 @@ class Normalizer:
     def normalize_string(self, value: str) -> str:
         return value.casefold() if self.ignore_case else value
 
-    def normalize_list(self, value: Sequence) -> Sequence:
+    def normalize_list(self, value: ListLike) -> ListLike:
         cls = type(value)
         value_list = [self.normalize(v) for v in value]
         if self.ignore_order:
--- a/atest/testdata/standard_libraries/collections/list.robot
+++ b/atest/testdata/standard_libraries/collections/list.robot
@@ -619,34 +619,34 @@ List Should Not Contain Value, Value Found And Own Error Message Glob
 Check List Error
     [Template]    Validate invalid argument error
     VAR    ${invalid_arg}    I am a string. Not a list.
-    Append to list                        xyz
-    Combine Lists                         ${invalid_arg}    arg_name=lists
-    Combine Lists                         ${L0}   ${invalid_arg}   arg_name=lists    invalid_argument=${invalid_arg}
-    Combine Lists                         ${invalid_arg}    ${L0}   arg_name=lists
-    Copy list
-    Count values in list                  ${invalid_arg}    xyz
-    Get from list                         ${invalid_arg}    0
-    Get Index From List                   ${invalid_arg}    a
-    Get Match Count                       ${invalid_arg}    abc    arg_name=list
-    Get Matches                           ${invalid_arg}    abc    arg_name=list
-    Get slice from list
-    Insert into list                      ${invalid_arg}    0    a
-    List Should Contain Sub List          ${invalid_arg}    ${L0}    arg_name=list1
-    List Should Contain Sub List          ${L0}    ${invalid_arg}    arg_name=list2    invalid_argument=${invalid_arg}
+    Append to list                        xyz                         annotation=Sequence: Invalid expression
+    Combine Lists                         ${invalid_arg}              arg_name=lists
+    Combine Lists                         ${L0}   ${invalid_arg}      arg_name=lists    invalid_argument=${invalid_arg}
+    Combine Lists                         ${invalid_arg}    ${L0}     arg_name=lists
+    Copy list                                                         annotation=Sequence: Invalid expression
+    Count values in list                  ${invalid_arg}    xyz       annotation=Sequence: Invalid expression
+    Get from list                         ${invalid_arg}    0         annotation=Sequence: Invalid expression
+    Get Index From List                   ${invalid_arg}    a         annotation=Sequence: Invalid expression
+    Get Match Count                       ${invalid_arg}    abc       arg_name=list
+    Get Matches                           ${invalid_arg}    abc       arg_name=list
+    Get slice from list                                               annotation=Sequence: Invalid expression
+    Insert into list                      ${invalid_arg}    0    a    annotation=Sequence: Invalid expression
+    List Should Contain Sub List          ${invalid_arg}    ${L0}     arg_name=list1
+    List Should Contain Sub List          ${L0}    ${invalid_arg}     arg_name=list2    invalid_argument=${invalid_arg}
     List should contain value             ${invalid_arg}    a
-    List Should Not Contain Duplicates    xyz
+    List Should Not Contain Duplicates    xyz                         annotation=Sequence: Invalid expression
     List Should Not Contain Value         ${invalid_arg}    x
-    Lists Should Be Equal                 ${invalid_arg}    ${L0}    arg_name=list1
-    Lists Should Be Equal                 ${L0}    ${invalid_arg}    arg_name=list2    invalid_argument=${invalid_arg}
-    Log List
+    Lists Should Be Equal                 ${invalid_arg}    ${L0}     arg_name=list1
+    Lists Should Be Equal                 ${L0}    ${invalid_arg}     arg_name=list2    invalid_argument=${invalid_arg}
+    Log List                                                          annotation=Sequence: Invalid expression
     Remove Duplicates
-    Remove From List                      ${invalid_arg}    0
-    Remove Values From List               ${invalid_arg}    a
-    Reverse List
-    Set List Value                        ${invalid_arg}    0    a
-    Should Contain Match                  ${invalid_arg}    a    arg_name=list
-    Should Not Contain Match              ${invalid_arg}    xyz    arg_name=list
-    Sort List                             annotation=list
+    Remove From List                      ${invalid_arg}    0         annotation=Sequence: Invalid expression
+    Remove Values From List               ${invalid_arg}    a         annotation=Sequence: Invalid expression
+    Reverse List                                                      annotation=Sequence: Invalid expression
+    Set List Value                        ${invalid_arg}    0    a    annotation=Sequence: Invalid expression
+    Should Contain Match                  ${invalid_arg}    a         arg_name=list
+    Should Not Contain Match              ${invalid_arg}    xyz       arg_name=list
+    Sort List                                                         annotation=list: Invalid expression
 
 Lists Should Be Equal With Ignore Case
     [Template]  Lists Should Be Equal
@@ -684,12 +684,12 @@ Lists Should be equal with Ignore Case and Order
 
 *** Keywords ***
 Validate invalid argument error
-    [Arguments]    ${keyword}    ${argument}=I'm not a list, I'm a string.    @{args}    ${arg_name}=list_    ${annotation}=Sequence    ${invalid_argument}=${NONE}
+    [Arguments]    ${keyword}    ${argument}=I'm not a list, I'm a string.    @{args}    ${arg_name}=list_    ${annotation}=Sequence, Mapping or set    ${invalid_argument}=${NONE}
     IF    not $invalid_argument
         VAR    ${invalid_argument}    ${argument}
     END
     Run keyword and expect error
-    ...    ValueError: Argument '${arg_name}' got value '${invalid_argument}' that cannot be converted to ${annotation}: Invalid expression.
+    ...    ValueError: Argument '${arg_name}' got value '${invalid_argument}' that cannot be converted to ${annotation}.
     ...    ${keyword}    ${argument}    @{args}
 
 Create Lists For The Tests


robinmackaij:
PR updated with changes as discussed on Slack.

- `list` / `dict` annotations used where possible (i.e. not breaking current behavior). These types should be clear to our users.
    - Note: `_validate_list()` / `_validate_dictionary()` calls removed where automatic argument conversions now ensures this.
- `MutableSequence` / `MutableMapping` annotations used and `NOTE` comments added to keywords with side effect.
    - Note: While these annotations may be unclear to many of our users, we could perhaps address this in the documentation generation (e.g. document as `list` / `dict` in the docs). It's important to realize that any automatic type conversion will result in "Do Nothing" behavior for these keywords.
- `_index_to_int()` calls removed where no longer relevant due to automatic argument conversion.
- `copy.copy()` used in keywords where `copy.deepcopy()` to allow these keywords to be less restrictive in supported argument types.
- `Hashable` annotation for `Mapping` keys where needed to prevent automatic argument conversion from e.g. `tuple` keys to their string representation.

src/robot/libraries/Collections.py
```diff
@@ -27,18 +32,33 @@
 
 NOT_SET = NotSet()
 
+I_ = TypeVar("I_", bound=Iterable)
+M_ = TypeVar("M_", bound=Mapping)
```
pekkaklarck: I'm not sure are TypeVars needed in the first place, I'll take a look at the overloads using them separately, but if they are, please drop the `_` suffix. We haven't used that elsewhere and consistency would be good. You may have used it because linters don't like `I` as a variable name, but those warnings could either be ignored or some other character used instead. I believe the correct bound would anyway be `Sequence` and then the TypeVar could be `S`.

src/robot/libraries/Collections.py
```diff
@@ -27,18 +32,33 @@
...
+IC = Union[
```
pekkaklarck: My understanding is that type aliases should be named like classes and that's the style we've used elsewhere. This should be renamed to `IgnoreCase`.

src/robot/libraries/Collections.py
```diff
@@ -27,18 +32,33 @@
...
+    bool, Literal["key", "KEY", "keys", "KEYS", "value", "VALUE", "values", "VALUES"]
```
pekkaklarck: Please don't list all different case combinations. Our literal converter accepts values case-insensitively and converts them to the format used by `Literal`. This should be `Literal["key", "keys", "value", "values"]`. Alternatively values could be all caps, but we've so far used lower case values when using literals with standard libraries.

As mentioned already, the converter accepts values case-insensitively so the used case doesn't really matter, but whatever is used here will be shown in Libdoc outputs and can be considered the preferred style. The decision thus boils down to deciding do we prefer `ignore_case=values` or `ignore_case=VALUES`. I don't have a strong preference (other than that I prefer consistency).

src/robot/libraries/Collections.py
```diff
@@ -27,18 +32,33 @@
...
 
-    def convert_to_list(self, item):
+    # NOTE: Annotation with list would cause the keyword to fail on string arguments.
+    # This would break current behavior. Annotation with "list | str" would however
+    # cause weird behvior with non-iterable arguments since their string representation
+    # would be converted to a list and returned. For this reason, object is used as an
+    # annotation, effectively preventing automatic argument conversion.
+    # In the future support for string arugments will be dropped;
+    # 'Split String To Characters' from the String library can be used instead.
+    # This change in behavior does however require a deprecation period.
```
pekkaklarck: I don't like this kind of comments in the source code. If we want to do something to this in the future, following things should be done instead:
1. Make sure there are tests that validate the current behavior. That way anyone changing the typing will notice that the change breaks something. That is much more effective than having a comment.
2. Update documentation to mention that `Split String To Characters` is preferred when converting strings to list. The comment could also mention that the current behavior may be deprecated in the future.
3. Submit an issue about the deprecation. This can be omitted at this point.

src/robot/libraries/Collections.py
```diff
@@ -27,18 +32,33 @@
...
         """Converts the given ``item`` to a Python ``list`` type.
 
         Mainly useful for converting tuples and other iterable to lists.
         Use `Create List` from the BuiltIn library for constructing new lists.
         """
-        return list(item)
+        return list(item)  # type: ignore
 
-    def append_to_list(self, list_, *values):
+    # NOTE: https://github.com/robotframework/robotframework/issues/5536
```
pekkaklarck: We have the issue so this comment is unnecessary. The plan is to implement the issue as part of RF 7.4 and the comment can be removed at that point, but I still think the comment is unnecessary.

src/robot/libraries/Collections.py
```diff
@@ -48,11 +68,11 @@ def append_to_list(self, list_, *values):
         | ${L1} = ['a', 'xxx']
         | ${L2} = ['a', 'b', 'x', 'y', 'z']
         """
-        self._validate_list(list_)
         for value in values:
             list_.append(value)
```
pekkaklarck: Not related to this PR, but couldn't this be `list_.extend(value)`?

src/robot/libraries/Collections.py
```diff
@@ -73,10 +93,9 @@ def insert_into_list(self, list_, index, value):
         | ${L1} = ['xxx', 'a']
         | ${L2} = ['a', 'xxx', 'b']
         """
-        self._validate_list(list_)
-        list_.insert(self._index_to_int(index), value)
+        list_.insert(index, value)
 
-    def combine_lists(self, *lists):
+    def combine_lists(self, *lists: list) -> list:
```
pekkaklarck: Input type could be `Sequence`. It doesn't matter in practice because `chain.from_iterable` works the same way with both types, but conceptually there's no need to require lists. If we think about possible Python usage, the current typing disallows `combine_lists([1, 2], (3, 4))` which is actually supported.

src/robot/libraries/Collections.py
```diff
@@ -148,21 +167,19 @@ def remove_from_list(self, list_, index):
...
         """
-        self._validate_list(list_)
         try:
-            return list_.pop(self._index_to_int(index))
+            return list_.pop(index)
         except IndexError:
             self._index_error(list_, index)
 
-    def remove_duplicates(self, list_):
+    def remove_duplicates(self, list_: list) -> list:
```
pekkaklarck: Input type could be `Sequence`. Doesn't matter in practice, but is in my opinion more correct.

src/robot/libraries/Collections.py
```diff
@@ -171,7 +188,7 @@ def remove_duplicates(self, list_):
         logger.info(f"{removed} duplicate{s(removed)} removed.")
         return ret
 
-    def get_from_list(self, list_, index):
+    def get_from_list(self, list_: list, index: int) -> object:
```
pekkaklarck: Input type must be `Sequence`. This unnecessarily converts custom sequences to lists which may affect how item access works. A test for that behavior would be nice, but it requires creating a custom sequence. I can add it later if needed.

src/robot/libraries/Collections.py
```diff
@@ -190,13 +207,18 @@ def get_from_list(self, list_, index):
...
-            return list_[self._index_to_int(index)]
+            return list_[index]
         except IndexError:
             self._index_error(list_, index)
 
-    def get_slice_from_list(self, list_, start=0, end=None):
+    # NOTE: start is annotated with int | Literal[''] for backwards compatibility.
+    def get_slice_from_list(
+        self,
+        list_: list,
```
pekkaklarck: Type must be `Sequence`. We shouldn't change the type unnecessarily. This can be tested by using a tuple and validating that the return value is a tuple as well:
```
${slice} =    Get Slice From List    ${{(1, 2, 3)}}    1
Should Be Equal    ${slice}    (2, 3)    type=tuple
```

src/robot/libraries/Collections.py
```diff
@@ -190,13 +207,18 @@ def get_from_list(self, list_, index):
...
         """
-        self._validate_list(list_)
         try:
-            return list_[self._index_to_int(index)]
+            return list_[index]
         except IndexError:
             self._index_error(list_, index)
 
-    def get_slice_from_list(self, list_, start=0, end=None):
+    # NOTE: start is annotated with int | Literal[''] for backwards compatibility.
```
pekkaklarck: As I've wrote few times already, I don't like notes like this in code. It would be better to mention the special empty string handling in documentation. As discussed on Slack, that handling should also be deprecated and the docs should mention that as well. Deprecation isn't part of this PR and the comment can be removed when that's done, but I still consider it unnecessary.

src/robot/libraries/Collections.py
```diff
@@ -220,13 +242,17 @@ def get_slice_from_list(self, list_, start=0, end=None):
...
-        if end is not None:
-            end = self._index_to_int(end)
+        if not start:
+            start = 0
         return list_[start:end]
 
-    def count_values_in_list(self, list_, value, start=0, end=None):
+    def count_values_in_list(
+        self,
+        list_: list,
```
pekkaklarck: Type much be `Sequence`. Custom sequences may handle `count` differently than `list` and we want to honor them.

src/robot/libraries/Collections.py
robinmackaij: https://github.com/robotframework/robotframework/issues/5538 created to track this.

src/robot/libraries/Collections.py
robinmackaij: ruff indeed failed on `I` (since it's almost the same as `l` and `1` depending on font).

I've updated the overloads to just use `Sequence` and `Mapping` since we don't annotate container content, so the TypeVars have been removed entirely.

src/robot/libraries/Collections.py
robinmackaij: The convention that type checkers apparently check for are either CamelCase or UPPERCASE.

A search through the codebase for TypeVar shows that almost all TypeVars are one or two uppercase characters so I believe `IC` is in line with the rest of the codebase.

src/robot/libraries/Collections.py
robinmackaij: I've kept the uppercase versions. Other literals (e.g. `INFO`, `NO VALUES`) used in this library are also uppercase so this seems the most consistent. I've updated the library docstring to be consistent with this.

src/robot/libraries/Collections.py
robinmackaij: Removed the links to the issue and added a comment to the issue to help identify the affected keywords.

src/robot/libraries/Collections.py
robinmackaij: I've updated `list` and `dict` annotations for arguments to use `Sequence` and `Mapping` (where possible).

In many places, `Iterable` would actually be a more fitting annotation, but since we don't have a converter for it (yet?) `Sequence` is the best fit at the moment.

src/robot/libraries/Collections.py
robinmackaij: I've updated the annotations to `Sequence`. New tests would be good (as discussed). Especially the behavior of the keywords for non-Sequence Iterables would be good to cover.

src/robot/libraries/Collections.py
pekkaklarck: This is isn't a TypeVar, though, this is a type alias. They seem to be named in CamelCase:
https://typing.python.org/en/latest/spec/aliases.html

src/robot/libraries/Collections.py
robinmackaij: This implementation does pass the current tests:
```
    def append_to_list(self, list_: MutableSequence, *values: object):
        """Adds ``values`` to the end of ``list``.
        ...
        """
        list_.extend(values)
```
I didn't dig into the tests to see if there's corner cases that might fail, but since `append` takes `object` as argument and `extend` takes `Iterable[object]`, it makes sense that the implementations are interchangeable.

If you prefer this implementation and feel confident about the change, I can commit this.

robinmackaij:
You're right, I'll update it.

________________________________

src/robot/libraries/Collections.py
pekkaklarck: If tests pass, changing this is fine but not required.

src/robot/libraries/Collections.py
```diff
@@ -288,10 +303,9 @@ def reverse_list(self, list_):
         =>
         | ${L3} = ['c', 'b', 'a']
         """
-        self._validate_list(list_)
         list_.reverse()
 
-    def sort_list(self, list_):
+    def sort_list(self, list_: list):
```
pekkaklarck, line 307: It's annoying that we cannot use `MutableSequence` because it doesn't have `sort()`. If we wanted to support custom sequences that have it, typing could be `MutableSequence` and then code could do something like:
```python
if not hasattr(list_, 'sort'):
    list_ = list(list_)
```
but that's probably not worth the effort at least now.

src/robot/libraries/Collections.py
```diff
@@ -301,10 +315,15 @@ def sort_list(self, list_):
...
-        self._validate_list(list_)
         list_.sort()
 
-    def list_should_contain_value(self, list_, value, msg=None, ignore_case=False):
+    def list_should_contain_value(
+        self,
+        list_: Sequence,
+        value: object,
+        msg: "str | None" = None,
+        ignore_case: IgnoreCase = False,
```
pekkaklarck: Shouldn't `ignore_case` be just `bool` with list related keywords? In that case, `IgnoreCase` should probably be renamed to `IgnoreDictCase` or something like that.

src/robot/libraries/Collections.py
```diff
@@ -444,14 +470,23 @@ def lists_should_be_equal(
...
+        names: "Mapping[int, str] | Iterable[str] | None",
+        list_length: int,
+    ) -> "dict[int, str]":
         if not names:
             return {}
         if is_dict_like(names):
-            return {int(index): names[index] for index in names}
-        return dict(zip(range(list_length), names))
+            return {int(index): names[index] for index in names}  # type: ignore
+        return dict(zip(range(list_length), names))  # type: ignore
```
pekkaklarck: Some minor things:
- `is_dict_like` could be replaced with `isinstance(names, Mapping)` or `isinstance(names, dict)`. Not really related to this PR, but I've been replacing these utils with `isinstance` when possible also otherwise. One reason is that type checkers understands `isinstance` but not our utils (until we can make them type guards). This should avoid the need to have the latter `# type: ignore`.
- If `names` is a dictionary, we know its keys are integers so there's no need for `int(call)`. The value can actually be used as-is. This avoids the first `# type: ignore`.
- This helper function is used just ones and it becomes pretty simple. It could possibly be inlined. It would only require replacing the `_get_list_index_name_mapping` with something like this:
   ```python
   if not names:
       names = {}
   elif isinstance(names, list):
       names = dict(zip(range(len1), names))
   ```

src/robot/libraries/Collections.py
```diff
@@ -444,14 +470,23 @@ def lists_should_be_equal(
...
 
-    def _yield_list_diffs(self, list1, list2, names):
+    def _yield_list_diffs(
+        self,
+        list1: Iterable,
+        list2: Iterable,
+        names: "dict[int, str]",
+    ) -> Generator[str, None, None]:
```
pekkaklarck: - The code always passes here sequences, so typing should use `Sequence` instead of `Iterable`.
- `Generator[str, None, None]` is technically correct, but `Iterator[str]` would be simpler and we've used in elsewhere with `yield`. See the docs here for details:
https://docs.python.org/3/library/typing.html#annotating-generators-and-coroutines

src/robot/libraries/Collections.py
```diff
@@ -491,18 +525,17 @@ def list_should_contain_sub_list(
...
         Valid levels are TRACE, DEBUG, INFO (default), and WARN.
 
         If you only want to the length, use keyword `Get Length` from
         the BuiltIn library.
         """
-        self._validate_list(list_)
         logger.write("\n".join(self._log_list(list_)), level)
 
-    def _log_list(self, list_):
+    def _log_list(self, list_: Sequence) -> Generator[str, None, None]:
```
pekkaklarck: Use `Iterator[str]` instaed.

src/robot/libraries/Collections.py
```diff
@@ -512,32 +545,13 @@ def _log_list(self, list_):
...
-
-    def _validate_lists(self, *lists):
-        for index, item in enumerate(lists, start=1):
-            self._validate_list(item, index)
-
 
 class _Dictionary:
 
-    def convert_to_dictionary(self, item):
+    def convert_to_dictionary(self, item: Mapping) -> dict:
```
pekkaklarck: Should use `object` for same reasons as `convert_to_list`. I'm not sure has anyone done so with this keyword, but it's possible to convert also other types than mappings to dicts. For example, `[('k1', 1), ('k2', 2)]` and just `[]` work.

src/robot/libraries/Collections.py
```diff
@@ -661,15 +692,18 @@ def get_dictionary_keys(self, dictionary, sort_keys=True):
...
             except TypeError:
                 pass
         return list(dictionary)
 
-    def get_dictionary_values(self, dictionary, sort_keys=True):
+    def get_dictionary_values(
+        self,
+        dictionary: Mapping,
+        sort_keys: bool = True,
+    ) -> list:
```
pekkaklarck, line 698: This possibly could be `list[object]`. It doesn't really matter here, but the `items` variant below would benefit from more detailed typing.

src/robot/libraries/Collections.py
```diff
@@ -682,11 +716,14 @@ def get_dictionary_values(self, dictionary, sort_keys=True):
...
-        self._validate_dictionary(dictionary)
         keys = self.get_dictionary_keys(dictionary, sort_keys=sort_keys)
         return [dictionary[k] for k in keys]
 
-    def get_dictionary_items(self, dictionary, sort_keys=True):
+    def get_dictionary_items(
+        self,
+        dictionary: Mapping,
+        sort_keys: bool = True,
+    ) -> list:
```
pekkaklarck: This keyword returns a list of tuples and the return type should show that. `list[tuple[object, object]]` would probably work. The first `object` could be replaced with `Hashable`, but I believe it would mainly add confusion.

src/robot/libraries/Collections.py
```diff
@@ -643,12 +671,15 @@ def copy_dictionary(self, dictionary, deepcopy=False):
...
             return copy.deepcopy(dictionary)
-        return dictionary.copy()
+        return copy.copy(dictionary)
 
-    def get_dictionary_keys(self, dictionary, sort_keys=True):
+    def get_dictionary_keys(
+        self,
+        dictionary: Mapping,
+        sort_keys: bool = True,
+    ) -> list:
```
pekkaklarck, line 674: This could be `list[object]`. It doesn't really matter here, but the `items` variant below would benefit from more detailed typing.

`list[Hashable]` would be technically correct as well, but I believe the simpler variant is better.

src/robot/libraries/Collections.py
```diff
@@ -733,10 +773,10 @@ def get_from_dictionary(self, dictionary, key, default=NOT_SET):
...
     def dictionary_should_contain_key(
         self,
-        dictionary,
-        key,
-        msg=None,
-        ignore_case=False,
+        dictionary: Mapping,
+        key: Hashable,
+        msg: "str | None" = None,
+        ignore_case: IgnoreCase = False,
```
pekkaklarck: Hmm, in this usage `ignore_case` should only accept `key/keys`, not `value/values`. Should we remove the generic type alias and just use `Literal` with relevant values where needed? Or should we have type aliases `IgnoreKeyCase` and `IgnoreValueCase` that could be used separately and like `bool | IgnoreKeyCase | IgnoreValueCase`?

src/robot/libraries/Collections.py
```diff
@@ -851,13 +886,13 @@ def dictionary_should_not_contain_value(
...
-        msg=None,
-        values=True,
-        ignore_keys=None,
-        ignore_case=False,
-        ignore_value_order=False,
+        dict1: Mapping,
+        dict2: Mapping,
+        msg: "str | None" = None,
+        values: bool = True,
+        ignore_keys: "Sequence[Hashable] | str | None" = None,
```
pekkaklarck: This could probably be `list[Hashable] | None` or `Sequence[Hashable] | None` if you prefer less strict type. Either way, list literals are converted to lists. Both also accept any sequence so there's no practical difference.

src/robot/libraries/Collections.py
```diff
@@ -899,7 +933,14 @@ def dictionaries_should_be_equal(
         self._should_have_same_keys(dict1, dict2, msg, values)
         self._should_have_same_values(dict1, dict2, msg, values)
 
-    def _should_have_same_keys(self, dict1, dict2, message, values, validate_both=True):
+    def _should_have_same_keys(
+        self,
+        dict1: Mapping,
+        dict2: Mapping,
+        message: "str | None",
+        values: "bool| str",
```
pekkaklarck: `bool|` -> `bool |`

src/robot/libraries/Collections.py
```diff
@@ -955,35 +1001,33 @@ def dictionary_should_contain_sub_dictionary(
...
         the BuiltIn library.
         """
-        self._validate_dictionary(dictionary)
         logger.write("\n".join(self._log_dictionary(dictionary)), level)
 
-    def _log_dictionary(self, dictionary):
+    def _log_dictionary(
+        self,
+        dictionary: Mapping,
+    ) -> Generator[str, None, None]:
```
pekkaklarck: -> `Iterator[str]`

src/robot/libraries/Collections.py
```diff
@@ -1262,12 +1302,21 @@ def _get_matches(
         ]
 
 
-def _verify_condition(condition, default_message, message, values=False):
+def _verify_condition(
+    condition: object,
```
pekkaklarck: This should probably be `bool`. There may be cases where objects are passed and type checkers likely complain, but such usages should be changed to explicitly call `bool` or otherwise handle Boolean conversion.

src/robot/libraries/Collections.py
```diff
@@ -1277,17 +1326,25 @@ def _report_error(default_message, message, values=False):
 
 class Normalizer:
 
-    def __init__(self, ignore_case=False, ignore_order=False, ignore_keys=None):
-        self.ignore_case = ignore_case
+    def __init__(
+        self,
+        ignore_case: IgnoreCase = False,
+        ignore_order: bool = False,
+        ignore_keys: "Iterable[Hashable] | str | None" = None,
```
pekkaklarck: Change to `Sequence[Hashable] | None`, `list[Hashable] | None`.

src/robot/libraries/Collections.py
```diff
@@ -1277,17 +1326,25 @@ def _report_error(default_message, message, values=False):
...
+    ):
         if isinstance(ignore_case, str):
             self.ignore_key_case = ignore_case.upper() not in ("VALUE", "VALUES")
             self.ignore_value_case = ignore_case.upper() not in ("KEY", "KEYS")
```
pekkaklarck, line 1328: `upper()` calls aren't needed anymore. Preserving them in case someone calls these keywords programmatically is fine, though.

src/robot/libraries/Collections.py
```diff
@@ -1299,9 +1356,21 @@ def _parse_ignored_keys(self, ignore_keys):
             raise ValueError(
                 f"'ignore_keys' value '{ignore_keys}' cannot be converted to a list."
             )
```
pekkaklarck: It should be possible to remove this `try/except` altogether. Argument conversion ought to guarantee that we got a list (or a sequence, depending on the used typing) here.

src/robot/libraries/Collections.py
```diff
@@ -1312,25 +1381,28 @@ def normalize(self, value):
...
         cls = type(value)
-        value = [self.normalize(v) for v in value]
+        value_list = [self.normalize(v) for v in value]
         if self.ignore_order:
-            value = sorted(value)
-        return self._try_to_preserve_type(value, cls)
+            try:
+                value_list = sorted(value_list)  # type: ignore
+            except TypeError:
+                pass
```
pekkaklarck: I don't think ignoring the error if values aren't comparable a good idea. If user has requested to get items sorted, returning them in the original order can cause problems. I believe the error is better. If this was to be changed, it would need a separate issue and tests.

A more realistic change would be enhancing sorting so that it would never fail. We could emulate the Python 2 style where objects with different types were sorted separately. There are some annoying corner cases with that, though, so it's not that easy. Anyway, all this is out of the scope of this PR.

src/robot/libraries/Collections.py
robinmackaij: Done. Bit clearer and possibly a tiny bit faster.

src/robot/libraries/Collections.py
robinmackaij: I've added a comment / suggesting related to this to #5536

src/robot/libraries/Collections.py
robinmackaij: Good catch, fixed.

src/robot/libraries/Collections.py
robinmackaij: I've removed the import of `is_dict_like` since it's indeed just a 1-line isinstance check.

I also removed `_get_list_index_name_mapping` since the argument conversion ensures a dict / Mapping passed to the `names` argument. That makes the signature for `Lists Should Be Equal` somewhat more complex that the other keywords in Collections, but I believe it's worth the benefits.

src/robot/libraries/Collections.py
robinmackaij: Is this really the same situation as `Convert To List`? The reason we use `object` there is because there's no possible annotation that supports both the intended use as well as the `Split String Into Characters` behavior.

I added tests for the `iterable` overload of the `dict` constructor and those pass with the annotation `item: "Mapping | Sequence")`.
Of course, `Sequence` isn't the completely correct annotation:
> Otherwise, the positional argument must be an [iterable](https://docs.python.org/3/glossary.html#term-iterable) object. Each item in the iterable must itself be an iterable with exactly two elements. The first element of each item becomes a key in the new dictionary, and the second element the corresponding value. If a key occurs more than once, the last value for that key becomes the corresponding value in the new dictionary.

So the most correct annotation I can think of would be `item: "Mapping[Hashable, object] | Iterable[Tuple[Hashable, object]]"` (since `Tuple` is, as far as I know, the only way to annotate an iterable with a fixed length and variable content types).

Now I'm also pretty sure we don't want to inflict annotations like that on our users (outside the fact that we'd also have to support `Iterable` and `Tuple` first).

While I don't think this is really the same as `Convert To List`, I would be inclined to go with `object` as annotation though, simply because the alternative would be `"Mapping | Sequence"` that I consider too confusing for the intended audience.

Edit: an argument in favor of `"item: Mapping | Sequence"` is that there's no type error, while annotation with `object` violates the signature for `dict(item)`.

src/robot/libraries/Collections.py
robinmackaij: While `list[Hashable]` would be technically correct, it'd also introduce a typing error since `Mapping` does not have it's items annotated, which would mean changing the dictionary annotation to `Mapping[Hashable, object]`. That would of course be technically correct pretty much everywhere but I don't feel doing that would really help our users.

For that reason, I'm going to leave the return type annotations for `list` and `dict` if they are either `object` or `Hashable` to keep things simple for the "common cases".

src/robot/libraries/Collections.py
robinmackaij: I've updated this one and went with `"list[tuple[Hashable, object]]"`. `Hashable` is already been used in some other places and I feel that if we do add annotations for container items, we might as well do so as correct as possible.

src/robot/libraries/Collections.py
robinmackaij: Done, only 1 call to `_verify_condition` was passing an object (list).

src/robot/libraries/Collections.py
robinmackaij: Would that mean the `ignore_keys` annotation should be changed to not include `str` anymore?

pekkaklarck: Yes. I think I mentioned it separately somewhere. Also docs could be updated, because the automatic type doc explains how list literals work. It's also kind of wrong to say in keyword doc that it accepts list literals when it actually doesn't.

src/robot/libraries/Collections.py
robinmackaij: I made this change since the type checker picked up on it and `Get Dictionary Keys` handles unsortable keys in the same way (keyword implementation unchanged in this PR):
```
    if sort_keys:
        try:
            return sorted(dictionary)
        except TypeError:
            pass
    return list(dictionary)
```
To me, this is the same issue that's handled differently within the same module. In both cases, there's an extra requirement on the items in the container; `SupportsRichComparison`.

Now I know about the xkcd-effect, but if this TypeError is raised, it would cause an assertion keyword to fail. Such an assertion in a test suite would be very questionable; you'd have an always failing tests because the data you provide cannot be asserted by the keyword being used.

src/robot/libraries/Collections.py
robinmackaij: Updating the relevant keyword annotations and removing the try/catch did change the expected error messages for unsupported literals, but otherwise tests passing and another type: ignore cleared. I've also updated the keyword docs as indicated.

src/robot/libraries/Collections.py
robinmackaij: I went with the latter...using the Literals in the signatures would be really spammy.

src/robot/libraries/Collections.py
pekkaklarck: The keyword documentation explicitly states that items must be sortable. I don't think hiding errors is a good idea in this case. This should be changed back. Catching the exception and reporting it better could be considered, but it isn't really related to this PR.

Situation is different when normalizing for comparison. If both compared values fail in sorting, the result may nevertheless be equal.

src/robot/libraries/Collections.py
pekkaklarck: The point of this keyword is to convert anything to a `dict`. In that regard it's best if argument conversion doesn't touch the original value at all. Alternatively would be typing all possible convertible values, but as you noted that gets _really_ complicated.

In this kind if cases it's fine to ignore type checker warnings. If you want to add an explicit `type: ignore`, that's fine. Catching and reraising the error with a better error could also be done, but that's not really part of this PR.

src/robot/libraries/Collections.py
robinmackaij: Test with custom sequences etc. will be added later / separately, as discussed.

src/robot/libraries/Collections.py
robinmackaij: Annotated with `object` and added `type: ignore`, same as with `Convert To List`.

src/robot/libraries/Collections.py
robinmackaij: Reverted the try/catch. I guess with annotations / type conversions this type of behavior can be improved in the future, but that's indeed out of score for this PR.

src/robot/libraries/Collections.py
```diff
@@ -148,21 +152,19 @@ def remove_from_list(self, list_, index):
...
         """
-        self._validate_list(list_)
         try:
-            return list_.pop(self._index_to_int(index))
+            return list_.pop(index)
         except IndexError:
             self._index_error(list_, index)
 
-    def remove_duplicates(self, list_):
+    def remove_duplicates(self, list_: ListLike) -> list:
```
pekkaklarck, line 160: I think this should be `MutableSequence` or `Sequence`. Implementation doesn't require that, but `Mapping` or `set` cannot even have duplicates. For consistency reasons I think modifying keywords should only accept sequences, validation in general work also with others.

src/robot/libraries/Collections.py
```diff
@@ -288,10 +302,9 @@ def reverse_list(self, list_):
         =>
         | ${L3} = ['c', 'b', 'a']
         """
-        self._validate_list(list_)
         list_.reverse()
 
-    def sort_list(self, list_):
+    def sort_list(self, list_: list):
```
pekkaklarck, line 307: This could be `MutableSequence`. It needs a small adjustment to the implementation, though.

src/robot/libraries/Collections.py
```diff
@@ -444,14 +473,12 @@ def lists_should_be_equal(
...
-            return {}
-        if is_dict_like(names):
-            return {int(index): names[index] for index in names}
-        return dict(zip(range(list_length), names))
-
-    def _yield_list_diffs(self, list1, list2, names):
+    def _yield_list_diffs(
+        self,
+        list1: Iterable,
+        list2: Iterable,
```
pekkaklarck, line 479: Iterable -> Sequence

src/robot/libraries/Collections.py
```diff
@@ -643,12 +663,15 @@ def copy_dictionary(self, dictionary, deepcopy=False):
...
             return copy.deepcopy(dictionary)
-        return dictionary.copy()
+        return copy.copy(dictionary)
 
-    def get_dictionary_keys(self, dictionary, sort_keys=True):
+    def get_dictionary_keys(
+        self,
+        dictionary: Mapping,
+        sort_keys: bool = True,
+    ) -> list:
```
pekkaklarck, line 674: For consistency with the `items` variant, add parameter to this and the `values` variant.

src/robot/libraries/Collections.py
```diff
@@ -1277,60 +1317,74 @@ def _report_error(default_message, message, values=False):
...
         if is_list_like(value):
             return self.normalize_list(value)
         return value
 
-    def normalize_string(self, value):
+    def normalize_string(self, value: str) -> str:
         return value.casefold() if self.ignore_case else value
 
-    def normalize_list(self, value):
+    def normalize_list(self, value: ListLike) -> ListLike:
```
pekkaklarck, line 1372: This should be `Sequence` or possibly `Sequence | Set`. Execution gets here via `isinstance` check and thus this cannot be called with a `Mapping`.

src/robot/libraries/Collections.py
```diff
@@ -1277,60 +1317,74 @@ def _report_error(default_message, message, values=False):
...
+    def normalize(self, value: str) -> str: ...
+
+    @overload
+    def normalize(self, value: Mapping) -> Mapping: ...
+
+    @overload
+    def normalize(self, value: Sequence) -> Sequence: ...
+
+    @overload
+    def normalize(self, value: ListLike) -> ListLike: ...
```
pekkaklarck, line 1353: This seems unnecessary. Mapping and Sequence convert 2/3 of ListLike, the final part would be Set.

src/robot/libraries/Collections.py
```diff
@@ -1277,60 +1317,74 @@ def _report_error(default_message, message, values=False):
...
+
+    def normalize(self, value: Any) -> Any:
         if not self:
             return value
         if isinstance(value, str):
             return self.normalize_string(value)
-        if is_dict_like(value):
+        if isinstance(value, Mapping):
             return self.normalize_dict(value)
         if is_list_like(value):
```
pekkaklarck, line 1365: Probably should check for `Sequence` here.

src/robot/libraries/Collections.py
```diff
@@ -14,31 +14,41 @@
...
-    type_name
+    is_list_like, Matcher, NotSet, plural_or_not as s, seq2str, seq2str2, type_name
 )
 from robot.utils.asserts import assert_equal
 from robot.version import get_version
 
 NOT_SET = NotSet()
 
+IgnoreKeyCase = Literal["KEY", "KEYS"]
+IgnoreValueCase = Literal["VALUE", "VALUES"]
```
pekkaklarck, line 34: I realized we actually shouldn't have separate key and value literals but should always use a literal with both. The reason is that normalization is recursive and, for example, a list can have dicts that should be normalized. This probably should get tests.

pekkaklarck (approved):
Looks good. There are minor things that can still be fine-tuned, but I can do that myself.

PR #5539, opened 2025-11-09
feat: deprecate string usage in Convert To List with warning

### LINKED ISSUE #5538: Deprecate possibility to use Convert To List (from Collections) to do the same as Split String To Characters (from String)
Currently, `Convert To List` and `Split String To Character` share the same implementation. Adding annotations (and thus enabling automatic type conversion) will change their behavior so they are actually two keywords to do two different things.

However, the `Convert To List` keyword cannot be annotated as desired since arbitrary strings (i.e. strings not resulting in an Iterable after literal_eval) would no longer be accepted unless an annotation such as `Sequence | str` would be used. That would however result in undesirable behavior for non-iterable arguments; passing e.g. the `int` 42 to `Convert To List` would then result in `["4", "2"]` being returned (instead of a type conversion error).

`Convert To List` should emit a deprecation warning when a `str` argument is passed to it, pointing the user to `Split String To Character` from the String library. Then after the deprecation period the annotation can be changed to `item: Sequence` (or `item: Iterable` if a converter for that annotation is implemented in the future).

CommitToday:
## Overview
This PR addresses issue #5538 by deprecating the usage of strings with the `Convert To List` keyword in the Collections library. A deprecation warning is now issued when a string is passed to this keyword, directing users to use `Split String To Characters` from the String library instead.

## Checklist
- [x] Code changes implemented
- [ ] Tests updated (pending)
- [x] Deprecation warning added

## Proof
The changes introduce a deprecation warning when the `Convert To List` keyword receives a string input. This will guide users to use the more appropriate `Split String To Characters` keyword from the String library, improving clarity and preventing potential misuse.

## Closes
#5538

### DIFF
--- a/logs/.e0b4d3cecc61067bf318824d38b6444bf2fbbe27-audit.json
+++ b/logs/.e0b4d3cecc61067bf318824d38b6444bf2fbbe27-audit.json
@@ -1,0 +1,15 @@
+{
+    "keep": {
+        "days": true,
+        "amount": 14
+    },
+    "auditLog": "/Users/soulsniper/robotframework/logs/.e0b4d3cecc61067bf318824d38b6444bf2fbbe27-audit.json",
+    "files": [
+        {
+            "date": 1762656359615,
+            "name": "/Users/soulsniper/robotframework/logs/mcp-puppeteer-2025-11-08.log",
+            "hash": "fd7fa9fade75e7b4aaff6aa1c47bc16a40c323abecd2641bb4497f1d27a8aeeb"
+        }
+    ],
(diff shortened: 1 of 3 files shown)


pekkaklarck:
Thanks for the PR. As you can read from the issue, we haven't yet decided do we want to do this and my current thinking is no. In that case this PR will be closed. Otherwise tests are needed and the included log files should be removed.

pekkaklarck:
#5538 that proposed this change was closed as not planned. I'll close this as well. Hopefully you find something else you can work on!

PR #5542, opened 2025-11-11
Add annotations to xml

### LINKED ISSUE #5541: Add annotations to XML.py
This issue is part of the annotation effort on the standard libraries.

The linked PR highlights some points of attention that need a decision before finalizing the PR.

#### Use of Protocols
I've added Protocols to annotate the relevant etree classes / methods.
The reason for this is the nature of the `xml.etree` module and the support for `lxml`. While the `xml.etree` implementation is a reference interface that other tools should follow, no Protocols are available in the `xml.etree` module. While using the actual classes from the `xml.etree` module could be used for annotations, there is the risk that any type checks would fail at runtime if `lxml` is used.
In addition, both `xml.etree` and `lxml` are not particularly well annotated (if at all). The addition of Protocols with proper annotations should allow us control over any automatic argument conversion and / or documentation generation.

#### `"bytes | str"` usage
In many places, the `"bytes | str"` annotation is used and there's a number of unresolved pyright/mypy errors due to this. The core of this issue is that fact that most properties of an `Element` can be `str` or `bytes` (https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element):
> The element name, attribute names, and attribute values can be either bytestrings or Unicode strings. tag is the element name. attrib is an optional dictionary, containing element attributes. extra contains additional attributes, given as keyword arguments

Now for a given keyword call / `Element` it is to be expected that either `str` or `bytes` are used, but there is no way to enforce this.

#### `source: "Element | Unparsed"` annotation
This is the most common annotation within this library and I'd like to centralize / alias it, but this does not seem possible without access to `TypeAlias`. It appears that a type alias using `Union` does not preserve / respect the order of the types and that causes problems with automatic argument conversion; if the `str` annotation (the last part of the `Unparsed` alias) is found before `Element`, the argument conversion will convert the `Element` to its string representation.

robinmackaij:
closes #5541

### DIFF
--- a/src/robot/libraries/XML.py
+++ b/src/robot/libraries/XML.py
@@ -17,7 +17,6 @@ import copy
 import os
 import re
 from collections.abc import Iterator
-from io import BytesIO
 from pathlib import Path
 from typing import Any, Callable, NoReturn
 from xml.etree import ElementTree as ET
@@ -596,7 +595,7 @@ class XML:
         | ${children} =    | Get Elements | ${XML} | first/child |
         | Should Be Empty  |  ${children} |        |             |
         """
-        if isinstance(source, (str, bytes, bytearray, Path, BytesIO)):
+        if isinstance(source, (str, bytes, os.PathLike)):
             source = self.parse_xml(source)
         finder = ElementFinder(self.etree, self.modern_etree, self.lxml_etree)
         return finder.find_all(source, xpath)
@@ -842,7 +841,7 @@ class XML:
         self,
         source: Any,
         xpath: str = ".",
-    ) -> "dict[bytes | str, bytes | str]":
+    ) -> "dict[str, str]":
         """Returns all attributes of the specified element.
 
         The element whose attributes to return is specified using ``source`` and
@@ -1598,7 +1597,7 @@ class NameSpaceStripper:
         self,
         elem: Element,
         preserve: bool = True,
-        current_ns: "bytes | str | None" = None,
+        current_ns: "str | None" = None,
         top: bool = True,
     ):
         if elem.tag.startswith("{") and "}" in elem.tag:
@@ -1617,7 +1616,7 @@ class NameSpaceStripper:
     def unstrip(
         self,
         elem: Element,
-        current_ns: "bytes | str | None" = None,
+        current_ns: "str | None" = None,
         copied: bool = False,
     ) -> Element:
         if not copied:


robinmackaij:
Updated annotation for `source` to be 3.8/3.9 and 3.10+ compatible.

While trying some things, I found that `Union` versions for `Unparsed` and `Element` file on all versions, while the stringified `source: "Element | Unparsed"` works for 3.8/3.9 but fails for 3.10+. So there's somehow a difference in the behavior of `Union[Element, Unparsed]` and `"Element | Unparsed"`.

While the current `source: Source` with `Source = Union[Element, Any]` (where the `source` can be an Element of Unparsed) isn't great, I'm inclined to say improving this is better done in / after the release where we drop 3.8 and 3.9 support so we get access to `TypeAlias` and non-stringified `|` usage. We'd also have access to `typing.Annotated` that could be helpful in case we can't get rid of the `Any` usage; using `Annotated` metadata we could get nice libdoc output even if the internal annotations aren't  complete.

src/robot/libraries/XML.py
```diff
@@ -991,25 +1141,27 @@ def elements_should_match(
...
-    def _sort_children(self, element):
+    def _sort_children(self, element: Element):
         tails = [child.tail for child in element]
         element[:] = sorted(element, key=lambda child: child.tag)
         for child, tail in zip(element, tails):
             child.tail = tail
 
-    def set_element_tag(self, source, tag, xpath="."):
+    def set_element_tag(
+        self, source: Source, tag: str, xpath: str = "."
```
robinmackaij: `tag` should be annotated `"str | bytes"`. This applies to other keywords too.

src/robot/libraries/XML.py
```diff
@@ -43,6 +47,113 @@
 should_match = BuiltIn().should_match
 
 
+@runtime_checkable
+class Element(Protocol):
+    tag: "bytes | str"
+
+    text: "None | str"
```
robinmackaij: @pekkaklarck `text` and `tail` can (per the `xml.etree.ElementTree.Element` documentation) contain arbitrary objects:
> These attributes can be used to hold additional data associated with the element. Their values are usually strings but may be any application-specific object.
> ...
> Applications may store arbitrary objects in these attributes.

This means that technically, the annotations here (and in all keywords related to `text` or `tail`) would have to be 'object'. At the same time, no tests fail on the current `"None | str"` annotations which indicates to me that we have no tests that check on handling "application-specific / arbitrary objects" in keywords. I also don't really see how working with arbitrary data for this would actually work. What do we want to do here?

pekkaklarck:
Let's not make this complicated with Protocols. We are adding typing mainly for documentation purposes and partly to help Robot editord, the idea isn't, at least now, to do type checking. Complex types also don't help with the documentation goal, they just look confusing for non-technical users.

I want to get RF 7.4 out soon. To keep this simple and to avoid wasting time, there are only two ways I see `source` can be annottated:

1. As `Any`. This is also the way I believe it should be annotated in the beginning to get all the other annotations in.
2. As a custom `Source` class with a custom converter having documentation that explains supported values. This would be good considering the main goal is enhancing documentation. Type checkers may not like it, but type checking isn't the goal so that's not a big problem.

If you want, you can experiment with Protocols and other ways to make typing better after the option 1. above is ready. If that brings concrete benefits, doesn't complicated code and doesn't make Libdoc documentation complicated, I can consider including it. I anyway want to be extremely clear that *most likely such a PR will not be merged* at least in RF 7.4 timeframe.

My proposal how to move forward:
1. You create a PR with `source: Any` (or with `source: Source` where `Source` is a type alias `Source = Any`).
2. I merge that PR.
3. You continue experimenting with better typing if you have interest and don't mind if such enhancement may not be accepted.
4. I test how using a custom class with custom docs looks like.
5. I make a decision between `Any`, custom class and your possible PR with enhanced typing.

robinmackaij:
Ok, let's do that. I do believe that at least exploring Protocol with
the @runtime_checkable is worth some of our time since I think that if we
can support that, it'll be useful for library authors.

I'll see if I can update the PR tomorrow but I have a busy weekend ahead so
computer time may be very limited.

robinmackaij:
Subclassing `Any` is not possible / allowed. Also, docstrings for custom classes are not reflected in the libdoc annotations, so I've used 2 "empty" classes `Element` and `ElementOrUnparsed`.

This means mypy / pylance don't report anything meaningful, but I believe this is the most useful / clear for the users of the library / keyword docs.

src/robot/libraries/XML.py
```diff
@@ -30,18 +33,27 @@
...
 
+Unparsed = Union[BytesIO, bytes, bytearray, Path, str]
+
+
+class Element:
+    """An XML Element"""
+
+
+class ElementOrUnparsed:
+    """Element | Unparsed"""
```
pekkaklarck: I thought I was explicit [when I wrote](https://github.com/robotframework/robotframework/pull/5542#issuecomment-3528632502) that I want `source` to be initially typed as `Any` or as `source: Source` where `Source` is a type alias `Source = Any`. We can then experiment with better typing, but I'm fairly sure using something complicated like `BytesIO | bytes | bytearray | Path | str` or something like `ElementOrUnparsed` that's not explained anywhere isn't a good idea.

If you absolutely don't want to use `Any` initially, that's fine. I'll then just change it myself after merging this.

src/robot/libraries/XML.py
```diff
@@ -512,17 +529,17 @@ def parse_xml(self, source, keep_clark_notation=False, strip_namespaces=False):
         """
         if isinstance(source, os.PathLike):
             source = str(source)
-        with ETSource(source) as source:
-            tree = self.etree.parse(source)
+        with ETSource(source) as source_:
+            tree = self.etree.parse(source_)
```
pekkaklarck, line 527: Why this change?

src/robot/libraries/XML.py
```diff
@@ -512,17 +529,17 @@ def parse_xml(self, source, keep_clark_notation=False, strip_namespaces=False):
...
-            lxml_etree.strip_elements(tree, *strip, with_tail=False)
+            strip = (lxml_etree.Comment, lxml_etree.ProcessingInstruction)  # type: ignore
+            lxml_etree.strip_elements(tree, *strip, with_tail=False)  # type: ignore
         root = tree.getroot()
         if not keep_clark_notation:
             self._ns_stripper.strip(root, preserve=not strip_namespaces)
         return root
 
-    def get_element(self, source, xpath="."):
+    def get_element(self, source: ElementOrUnparsed, xpath: str = ".") -> Element:
```
pekkaklarck: As I already commented, `ElementOrUnparsed` should be changed to `Any` or, probably better, to `Source` that is a type alias. Using `Element` as the return type looks good to me, though. For consistency it could be just a type alias initially, but using a concrete class is fine too. After the basic stuff is merged, we can experiment adding type documentation to it.

src/robot/libraries/XML.py
```diff
@@ -582,12 +601,14 @@ def get_elements(self, source, xpath):
...
-        if isinstance(source, (str, bytes, os.PathLike)):
+        if isinstance(source, (str, bytes, bytearray, Path, BytesIO)):
             source = self.parse_xml(source)
         finder = ElementFinder(self.etree, self.modern_etree, self.lxml_etree)
         return finder.find_all(source, xpath)
 
-    def get_child_elements(self, source, xpath="."):
+    def get_child_elements(
+        self, source: ElementOrUnparsed, xpath: str = "."
+    ) -> "list[Element]":
```
pekkaklarck: As explained in the Contribution guidelines, I find signature formatting like
```python
def method(
    self, arg1: type1, arg2: type2, arg3: type3 = "default"
):
```
hard to read and prefer
```python
def method(
    self, 
    arg1: type1, 
    arg2: type2, 
    arg3: type3 = "default",
):
```
that you get automatically if you add the trailing comma. Not a big issue, I can add the comma and reformat the code, and in some cases the above is fine.

src/robot/libraries/XML.py
```diff
@@ -684,18 +720,23 @@ def get_element_text(self, source, xpath=".", normalize_whitespace=False):
             text = self._normalize_whitespace(text)
         return text
 
-    def _yield_texts(self, element, top=True):
+    def _yield_texts(self, element: Element, top: bool = True) -> "Iterator[str]":
```
pekkaklarck, line 715: Nit: `Iterator` has been imported from `typing` and in that case `Iterator[str]` works without quotes regardless the Python version. `typing.Iterable` is [deprecated](https://docs.python.org/3/library/typing.html#typing.Iterable), though, so it would actually be better to import it from `collections.abc` and preserve quoting until we drop Python 3.8 support. Latest then we also need to make sure we import ABCs like this only from `collections.abc`, not from `typing`.

src/robot/libraries/XML.py
```diff
@@ -776,8 +817,13 @@ def element_text_should_match(
         text = self.get_element_text(source, xpath, normalize_whitespace)
         should_match(text, pattern, message, values=False)
 
-    @keyword(types=None)
-    def get_element_attribute(self, source, name, xpath=".", default=None):
+    def get_element_attribute(
+        self,
+        source: ElementOrUnparsed,
+        name: "bytes | str",
```
pekkaklarck: Do bytes actually work here? My understanding is that you can create an `Element` based on a bytes source, but even then you should use strings when working with it. This experiment seems to prove that:
```python
>>> from xml.etree.ElementTree import XML
>>> e = XML(b'<elem a="1"/>')
>>> e.get(b'a')
>>> e.get('a')
'1'
>>> e.attrib
{'a': '1'}
```

If you are sure `bytes` as the name actually works, please add some tests for it as well. Otherwise simply change the typing to `str`.

src/robot/libraries/XML.py
```diff
@@ -799,7 +845,9 @@ def get_element_attribute(self, source, name, xpath=".", default=None):
         """
         return self.get_element(source, xpath).get(name, default)
 
-    def get_element_attributes(self, source, xpath="."):
+    def get_element_attributes(
+        self, source: ElementOrUnparsed, xpath: str = "."
+    ) -> "dict[bytes | str, bytes | str]":
```
pekkaklarck: I don't think attribute names or values can be bytes. See the previous comment for details.

src/robot/libraries/XML.py
```diff
@@ -821,11 +869,11 @@ def get_element_attributes(self, source, xpath="."):
...
     def element_attribute_should_be(
         self,
-        source,
-        name,
-        expected,
-        xpath=".",
-        message=None,
+        source: ElementOrUnparsed,
+        name: "bytes | str",
+        expected: "None | bytes | str",
```
pekkaklarck: `bytes` most likely isn't needed and is actually wrong. See the previous comments for more info.

I'd prefer `expected` typing to be `str | None` instead of `None | str` because `str` is the more common type. The order shouldn't affect argument conversion in this case.

src/robot/libraries/XML.py
```diff
@@ -852,11 +900,11 @@ def element_attribute_should_be(
 
     def element_attribute_should_match(
         self,
-        source,
-        name,
-        pattern,
-        xpath=".",
-        message=None,
+        source: ElementOrUnparsed,
+        name: "bytes | str",
```
pekkaklarck: Remove `bytes`.

src/robot/libraries/XML.py
```diff
@@ -877,7 +925,13 @@ def element_attribute_should_match(
             raise AssertionError(f"Attribute '{name}' does not exist.")
         should_match(attr, pattern, message, values=False)
 
-    def element_should_not_have_attribute(self, source, name, xpath=".", message=None):
+    def element_should_not_have_attribute(
+        self,
+        source: ElementOrUnparsed,
+        name: "bytes | str",
```
pekkaklarck: Remove `bytes`.

src/robot/libraries/XML.py
```diff
@@ -991,25 +1045,27 @@ def elements_should_match(
...
-    def _sort_children(self, element):
+    def _sort_children(self, element: Element):
         tails = [child.tail for child in element]
         element[:] = sorted(element, key=lambda child: child.tag)
         for child, tail in zip(element, tails):
             child.tail = tail
 
-    def set_element_tag(self, source, tag, xpath="."):
+    def set_element_tag(
+        self, source: ElementOrUnparsed, tag: "bytes | str", xpath: str = "."
```
pekkaklarck: I don't think `bytes` works here either:
```python
>>> from xml.etree.ElementTree import XML, tostring
>>> e = XML(b'<elem a="1"/>')
>>> e.tag
'elem'
>>> e.tag = b'new'
>>> tostring(e)
  [snip]
TypeError: cannot serialize b'new' (type bytes)
```

src/robot/libraries/XML.py
```diff
@@ -1031,7 +1087,9 @@ def set_element_tag(self, source, tag, xpath="."):
         self.get_element(source, xpath).tag = tag
         return source
 
-    def set_elements_tag(self, source, tag, xpath="."):
+    def set_elements_tag(
+        self, source: ElementOrUnparsed, tag: "bytes | str", xpath: str = "."
```
pekkaklarck: Remove `bytes`.

src/robot/libraries/XML.py
```diff
@@ -1074,8 +1137,13 @@ def set_element_text(self, source, text=None, tail=None, xpath="."):
             element.tail = tail
         return source
 
-    @keyword(types=None)
-    def set_elements_text(self, source, text=None, tail=None, xpath="."):
+    def set_elements_text(
+        self,
+        source: ElementOrUnparsed,
+        text: "None | str" = None,
+        tail: "None | str" = None,
```
pekkaklarck: Use `str | None`, not `None | str`.

src/robot/libraries/XML.py
```diff
@@ -1086,7 +1154,13 @@ def set_elements_text(self, source, text=None, tail=None, xpath="."):
             self.set_element_text(elem, text, tail)
         return source
 
-    def set_element_attribute(self, source, name, value, xpath="."):
+    def set_element_attribute(
+        self,
+        source: ElementOrUnparsed,
+        name: "bytes | str",
+        value: "bytes | str",
```
pekkaklarck: Remove `bytes`.

src/robot/libraries/XML.py
```diff
@@ -1113,7 +1187,13 @@ def set_element_attribute(self, source, name, value, xpath="."):
         self.get_element(source, xpath).attrib[name] = value
         return source
 
-    def set_elements_attribute(self, source, name, value, xpath="."):
+    def set_elements_attribute(
+        self,
+        source: ElementOrUnparsed,
+        name: "bytes | str",
+        value: "bytes | str",
```
pekkaklarck: Remove `bytes`.

src/robot/libraries/XML.py
```diff
@@ -1124,7 +1204,9 @@ def set_elements_attribute(self, source, name, value, xpath="."):
             self.set_element_attribute(elem, name, value)
         return source
 
-    def remove_element_attribute(self, source, name, xpath="."):
+    def remove_element_attribute(
+        self, source: ElementOrUnparsed, name: "bytes | str", xpath: str = "."
```
pekkaklarck: Remove `bytes`. Consider adding `,` to the end of the argument list to get nicer formatting.

src/robot/libraries/XML.py
```diff
@@ -1149,7 +1231,9 @@ def remove_element_attribute(self, source, name, xpath="."):
             attrib.pop(name)
         return source
 
-    def remove_elements_attribute(self, source, name, xpath="."):
+    def remove_elements_attribute(
+        self, source: ElementOrUnparsed, name: "bytes | str", xpath: str = "."
```
pekkaklarck: Remove `bytes`. Consider adding `,` to the end of the argument list to get nicer formatting.

src/robot/libraries/XML.py
```diff
@@ -1356,7 +1458,22 @@ def copy_element(self, source, xpath="."):
...
 
-    def element_to_string(self, source, xpath=".", encoding=None):
+    @overload
+    def element_to_string(
+        self, source: ElementOrUnparsed, xpath: str = ".", encoding: None = None
+    ) -> str: ...
+
+    @overload
+    def element_to_string(
+        self, source: ElementOrUnparsed, xpath: str, encoding: str
```
pekkaklarck: Shouldn't `xpath` have a default here as well?

src/robot/libraries/XML.py
```diff
@@ -1378,7 +1495,9 @@ def element_to_string(self, source, xpath=".", encoding=None):
             string = string.encode(encoding)
         return string
 
-    def log_element(self, source, level="INFO", xpath="."):
+    def log_element(
+        self, source, level: logger.LogLevel = "INFO", xpath: str = "."
```
pekkaklarck: `source` lacks typing.

src/robot/libraries/XML.py
```diff
@@ -1391,7 +1510,12 @@ def log_element(self, source, level="INFO", xpath="."):
         logger.write(string, level)
         return string
 
-    def save_xml(self, source, path, encoding="UTF-8"):
+    def save_xml(
+        self,
+        source: ElementOrUnparsed,
+        path: "Path | bytes | str",
```
pekkaklarck: Looking at the implementation, giving `path` as `bytes` doesn't work. There's `path.replace("/", os.sep)` that fails for `TypeError` in that case. If you add typing like this, please also test that it actually works!

It would be best to just `path: Path` and let argument conversion handle conversion. It would make it possible to simplify `path` handling as well, but the existing code ought to work also without changes.

src/robot/libraries/XML.py
```diff
@@ -1477,9 +1609,14 @@ def strip(self, elem, preserve=True, current_ns=None, top=True):
...
             self.strip(child, preserve, current_ns, top=False)
         if top and not preserve and self.lxml_tree:
-            self.etree.cleanup_namespaces(elem)
+            self.etree.cleanup_namespaces(elem)  # type: ignore
 
-    def unstrip(self, elem, current_ns=None, copied=False):
+    def unstrip(
+        self,
+        elem: Element,
+        current_ns: "None | bytes | str" = None,
```
pekkaklarck: `bytes` most likely is wrong here as well and additionally `None` should be last. In other words, make the typing `str | None`.

src/robot/libraries/XML.py
```diff
@@ -1456,16 +1582,22 @@ def evaluate_xpath(self, source, expression, context="."):
...
+    def __init__(self, etree, lxml_etree: bool = False):
         self.etree = etree
         self.lxml_tree = lxml_etree
 
-    def strip(self, elem, preserve=True, current_ns=None, top=True):
+    def strip(
+        self,
+        elem: Element,
+        preserve: bool = True,
+        current_ns: "None | bytes | str" = None,
```
pekkaklarck: `bytes` most likely is wrong here as well and additionally `None` should be last. In other words, make the typing `str | None`.

src/robot/libraries/XML.py
```diff
@@ -1523,21 +1665,39 @@ def _get_xpath(self, xpath):
             return xpath
 
 
+class Location:
+
+    def __init__(self, path: "bytes | str", is_root: bool = True):
```
pekkaklarck: Are you sure about `bytes` here and elsewhere in this class?

src/robot/libraries/XML.py
```diff
@@ -1523,21 +1665,39 @@ def _get_xpath(self, xpath):
...
+        exclude_children: bool = False,
     ):
         self.comparator = comparator
         self.normalizer = normalizer or (lambda text: text)
         self.child_sorter = child_sorter
         self.exclude_children = exclude_children
 
-    def compare(self, actual, expected, location=None):
+    def compare(
+        self, actual: Element, expected: Element, location: "None | Location" = None
```
pekkaklarck: Use `Location | None` instead of `None | Location`. Consider adding `,` at the end to get nicer formatting.

src/robot/libraries/XML.py
```diff
@@ -1523,21 +1665,39 @@ def _get_xpath(self, xpath):
...
+        return Location(f"{self.path}/{tag}", is_root=False)
+
+
 class ElementComparator:
 
     def __init__(
         self,
         comparator,
         normalizer=None,
         child_sorter=None,
```
pekkaklarck: I know typing callables in Python is annoying, but I think it it would be good to add _some_ typing here anyway. Using just `Callable` would be fine for me, but I think these would be the correct versions and they aren't even overly complicated:
```python
comparator: Callable[[object, object, str], None],
normalizer: "Callable[[str], str] | None" = None,
child_sorter: "Callable[[Element], None] | None" = None,
```

src/robot/libraries/XML.py
```diff
@@ -1557,14 +1717,23 @@ def _compare_tags(self, actual, expected, location):
...
         )
 
-    def _compare(self, actual, expected, message, location, comparator=None):
+    def _compare(
+        self,
+        actual: object,
+        expected: object,
+        message,
+        location: Location,
+        comparator=None,
```
pekkaklarck: I don't personally care too much are helper methods typed or not, but if they are, typing should be consistent. With `message` the typing is just `str` and with `comparator` it ought to be `Callable[[object, object, str], None] | None`.

src/robot/utils/etreewrapper.py
```diff
@@ -15,29 +15,31 @@
...
+from os import fsdecode, PathLike
 from pathlib import Path
 
 
 class ETSource:
 
     def __init__(self, source):
-        self._source = source
-        self._opened = None
+        self._source: "BytesIO | bytes | bytearray | Path | str" = source
```
pekkaklarck: Why is this typing here and not in `__init__` signature? `source` can also be an opened file object so the typing should use something like `IOBase` instead of `BytesIO`. This same typing is used also elsewhere, so creating a type alias covering all cases could be a good idea.

Alternatively all this could be untyped. The reason why `ETSource` was initially created was working around compatibility issues with Python and ElementTree versions and it may be totally or at least mostly useless nowadays.

src/robot/utils/etreewrapper.py
```diff
@@ -15,29 +15,31 @@
...
+        self._opened: "BytesIO | None" = None
 
-    def __enter__(self):
+    def __enter__(self) -> "BytesIO | PathLike":
```
pekkaklarck, line 29: `BytesIO` isn't correct here either.

src/robot/utils/etreewrapper.py
```diff
@@ -15,29 +15,31 @@
...
         self._opened = self._open_if_necessary(self._source)
         return self._opened or self._source
 
-    def _open_if_necessary(self, source):
+    def _open_if_necessary(
+        self, source: "BytesIO | bytes | bytearray | Path | str"
+    ) -> "BytesIO | None":
```
pekkaklarck: `BytesIO` is wrong in arguments but correct as the return type.

src/robot/libraries/XML.py
robinmackaij: The type-checker complained about the assignment; the types that source can be before and after the context manager are different. The `source` passed into the context manager can be e.g. `str` or `bytes` but what the context manager returns (`as source`) can never be `str` or `bytes`. So it's putting the same label on something different.

src/robot/utils/etreewrapper.py
robinmackaij: I indeed overlooked the `__init__`.

These are the types that are used in `isinstance` checks within the methods of this class. I can see using higher level base classes like `IOBase` instead, but I'm somewhat reluctant to make the judgement call as to where that'd be safe since I simply don't know the codebase / xml / etree well enough.

If elsewhere in the codebase we have similar annotations, centralized / uniform annotations sounds like a good idea.

src/robot/utils/etreewrapper.py
robinmackaij: What would it need to be? Considering `_open_if_necessary` what the `__enter__` returns is either a new `BytesIO` instance (from `_open_if_necesssay`) or something that is `PathLike`.

I'd say `BytesIO` is a concrete type that's created within the method so it's correctly annotated.

src/robot/libraries/XML.py
robinmackaij: I based the annotations for the attributes of the `ElementTree.Element`. The (Python 3.8) docstring for that class is this:
>An XML element.
>
>   This class is the reference implementation of the Element interface.
>
>    An element's length is its number of subelements.  That means if you
>    want to check if an element is truly empty, you should check BOTH
>    its length AND its text attribute.
>
>    **The element tag, attribute names, and attribute values can be either
>    bytes or strings.**
>
>    ***tag* is the element name.**  *attrib* is an optional dictionary containing
>    element attributes. *extra* are additional element attributes given as
>    keyword arguments.
>
>    Example form:
>        <tag attrib>text<child/>...</tag>tail

So `tag` (which is the element `name`) and both the keys and the values of the `attrib` dict (so, all element `attributes`) can be either `bytes` or `strings`.

Note that this does not apply to `text` and `tail`. They both have this stipulation in their docstring:
> This is either a string or the value None.

Note that the official documentation is subtly different (`bytes or strings` vs `bytestrings or Unicode strings`):
> https://docs.python.org/3.8/library/xml.etree.elementtree.html#element-objects
> The element name, attribute names, and attribute values can be either bytestrings or Unicode strings. tag is the element name. attrib is an optional dictionary, containing element attributes. extra contains additional attributes, given as keyword arguments.

Also (as noted in a previous comment), the docs for `text` and `tail` suggest that arbitrary objects are acceptable values:
> These attributes can be used to hold additional data associated with the element. Their values are usually strings but may be any application-specific object.

So for all the `bytes | str` annotations, I based that on the Python class docstring / official documentation.

I did not check the tests in detail to see if / to what extend `bytes` interaction with XML is covered. To minimize the risk of making a backwards incompatible change, I followed the Python docs.

I'm not **sure** `bytes` usage is as documented, but I do see the risk if we annotate with just `str`, it turns out to be a backwards incompatible change (the xkcd thing).

pekkaklarck: Baded on my experiments those docs are just wrong. `bytes` works  as the original source, but querying or modifying elements only works with `str`.

Let's go with `str` only and fix issues if someone actually has managed to use `bytes`. At that pont we then can add tests fo those parts that work.

pekkaklarck: Actually, the Python docs aren't wrong, they are just a bit confusing. The doc you quoted is for the `Element.__init__` and it's true you can use bytes when constructing these elements. The docs should be more clear that when working with `Element` instances, only strings work.

pekkaklarck: And after actually testing this with `Element`, not only with `XML` that I used when I was experimenting with this yesterday, I believe the docs really are wrong:
```python
>>> from xml.etree.ElementTree import Element, tostring, XML
>>> 
>>> e = Element(b'elem')
>>> e.tag
b'elem'
>>> tostring(e)
  ...
TypeError: cannot serialize b'elem' (type bytes)
>>> 
>>> e = XML(b'<elem/>')
>>> type(e)
<class 'xml.etree.ElementTree.Element'>
>>> e.tag
'elem'
>>> tostring(e)
b'<elem />'
>>> tostring(e, encoding='unicode')
'<elem />'
```
So, you can create an `Element` instance directly with bytes, but the resulting element cannot be used even with ElementTree's own functions. If you create an element using the `XML` function, then bytes works, but in the resulting elements `tag` and other attributes are actually strings. Interestingly `tostring` the yields bytes again unless you specify `encoding='unicode'` which itself look really odd.

I believe most/all of this is due to the code originally being written during Python 2 times when `str` actually was bytes. I'll submit a bug report about the confusing/broken docs if I can find time or you want to do that instead.

Based on all this, I very strongly believe we can use `str` as the typing with all keywords inspecting or modifying elements.

src/robot/libraries/XML.py
robinmackaij: Technically, yes, but since the overload does not work if `encoding` has a default (passing the `encoding` is what determines whether the return is `str` or `bytes`) and arguments with defaults must come after other arguments, the order of arguments in the overload would have to be different, which then is a signature issue.

I've removed the overloads altogether since it only solved a type error in `Log Element`.

src/robot/utils/etreewrapper.py
robinmackaij: The argument is typed the same `source` from the `__init__`.

When taking another look at this code, there's actually no reason for the (private) methods `_open_if_necessary`, `_is_path`, `_is_already_open` and `_find_encoding` to take any arguments at all; they all get `self._source` passed as `source` argument when you trace the method calls back from the `__enter__`.

My focus / intend when looking at this wasn't to refactor / rewrite this class, but I believe it could be refactored to this:
```
class ETSource:

    def __init__(self, source: "BytesIO | bytes | bytearray | Path | str"):
        self._source = source
        self._opened: "BytesIO | None" = None

    def __enter__(self) -> "BytesIO | PathLike":
        self._opened = self._open_if_necessary()
        return self._opened or self._source

    def _open_if_necessary(self) -> "BytesIO | None":
        if self._is_path or self._is_already_open:
            return None

        source = self._source
        if isinstance(source, (bytes, bytearray)):
            return BytesIO(source)
        return BytesIO(source.encode(self._encoding))

    @property
    def _is_path(self) -> bool:
        source = self._source
        if isinstance(source, Path):
            return True
        if isinstance(source, str):
            return not source.lstrip().startswith("<")
        if isinstance(source, bytes):
            return not source.lstrip().startswith(b"<")
        return False

    @property
    def _is_already_open(self) -> bool:
        return not isinstance(self._source, (str, bytes, bytearray))

    @property
    def _encoding(self) -> str:
        match = re.match(r"\s*<\?xml .*encoding=(['\"])(.*?)\1.*\?>", self._source)
        return match.group(2) if match else "UTF-8"

    def __exit__(self, exc_type, exc_value, exc_trace):
        if self._opened:
            self._opened.close()

    def __str__(self) -> str:
        source = self._source
        if self._is_path:
            return self._path_to_string(source)
        if hasattr(source, "name"):
            return self._path_to_string(source.name)
        return "<in-memory file>"

    def _path_to_string(self, path) -> str:
        if isinstance(path, Path):
            return str(path)
        if isinstance(path, bytes):
            return fsdecode(path)
        return path
```
A bit out of scope of this PR, but it makes the argument / annotation flow more clear.

src/robot/libraries/XML.py
pekkaklarck: Type checkers are annoying. The code is worse this way in my opinion and I may change this back.

src/robot/libraries/XML.py
```diff
@@ -582,12 +596,12 @@ def get_elements(self, source, xpath):
         | ${children} =    | Get Elements | ${XML} | first/child |
         | Should Be Empty  |  ${children} |        |             |
         """
-        if isinstance(source, (str, bytes, os.PathLike)):
+        if isinstance(source, (str, bytes, bytearray, Path, BytesIO)):
```
pekkaklarck: This changes keyword behavior and I don't think the change makes sense. Nobody has requested support for `bytearray` or `BytesIO`. The latter would also be really strange type to support as it would only support in-memory file-like objects, not actual opened files. This kind of changes should also be tested.

This needs to be changed so that `bytearray` and `BytesIO` are removed.

src/robot/libraries/XML.py
```diff
@@ -512,17 +524,17 @@ def parse_xml(self, source, keep_clark_notation=False, strip_namespaces=False):
...
-            lxml_etree.strip_elements(tree, *strip, with_tail=False)
+            strip = (lxml_etree.Comment, lxml_etree.ProcessingInstruction)  # type: ignore
+            lxml_etree.strip_elements(tree, *strip, with_tail=False)  # type: ignore
         root = tree.getroot()
         if not keep_clark_notation:
             self._ns_stripper.strip(root, preserve=not strip_namespaces)
         return root
 
-    def get_element(self, source, xpath="."):
+    def get_element(self, source: Any, xpath: str = ".") -> Element:
```
pekkaklarck, line 536: Using `Source` type alias instead of `Any` directly would have made future experiments easier. Not a biggie, can be changed later.

src/robot/libraries/XML.py
```diff
@@ -799,7 +838,11 @@ def get_element_attribute(self, source, name, xpath=".", default=None):
         """
         return self.get_element(source, xpath).get(name, default)
 
-    def get_element_attributes(self, source, xpath="."):
+    def get_element_attributes(
+        self,
+        source: Any,
+        xpath: str = ".",
+    ) -> "dict[bytes | str, bytes | str]":
```
pekkaklarck: This should be `dict[str, str]`.

src/robot/libraries/XML.py
```diff
@@ -1378,7 +1492,12 @@ def element_to_string(self, source, xpath=".", encoding=None):
             string = string.encode(encoding)
         return string
 
-    def log_element(self, source, level="INFO", xpath="."):
+    def log_element(
+        self,
+        source: Any,
+        level: logger.LogLevel = "INFO",
+        xpath: str = ".",
+    ) -> str:
```
pekkaklarck, line 1499: Looking at the code, this method may actually return the value as `bytes`. Need to check that and either change implementation or typing.

src/robot/libraries/XML.py
```diff
@@ -1456,16 +1585,22 @@ def evaluate_xpath(self, source, expression, context="."):
...
+    def __init__(self, etree, lxml_etree: bool = False):
         self.etree = etree
         self.lxml_tree = lxml_etree
 
-    def strip(self, elem, preserve=True, current_ns=None, top=True):
+    def strip(
+        self,
+        elem: Element,
+        preserve: bool = True,
+        current_ns: "bytes | str | None" = None,
```
pekkaklarck: Remove `bytes`.

src/robot/libraries/XML.py
```diff
@@ -1477,9 +1612,14 @@ def strip(self, elem, preserve=True, current_ns=None, top=True):
...
             self.strip(child, preserve, current_ns, top=False)
         if top and not preserve and self.lxml_tree:
-            self.etree.cleanup_namespaces(elem)
+            self.etree.cleanup_namespaces(elem)  # type: ignore
 
-    def unstrip(self, elem, current_ns=None, copied=False):
+    def unstrip(
+        self,
+        elem: Element,
+        current_ns: "bytes | str | None" = None,
```
pekkaklarck: Remove `bytes`.

src/robot/utils/etreewrapper.py
pekkaklarck: Your mistake here is that `ETSource` can be used with open file objects. One such use case is the [ExecutionResult](https://github.com/robotframework/robotframework/blob/3bda0950fd3cf6250509b88e3bfab97e2e95af81/src/robot/result/resultbuilder.py#L30) factory method that uses `ETSource` internally and accepts inputs in various formats. Thus the typing of `__init__` and all methods that accept `source` is wrong.

I don't care is this module typed at least right now. I do care that added typing is correct, though, and consider untyped code better than incorrectly typed code. I'll see if I can merge this PR so that this module is ignore. You can continue with it later, but I don't want it to delay RF 7.4.

pekkaklarck (changes requested):
There are still some things to be changed.

pekkaklarck:
Quick summary of the latest review:
- Few small changes still needed to the XML library itself. 
- Typing `ETSource` properly seems to be somewhat tricky. Getting it typed isn't part of getting the XML library typed so I believe it would be best to just forget about it now. As I commented above, incorrect typing is worse than no typing. If you want to give it one more go as part of this PR that's fine, though.
- I can make the final changes myself. I would have actually already done them if that would have been easy via the web UI.

src/robot/libraries/XML.py
pekkaklarck: I was wrong, the keyword returns a string. Typing is correct and no other changes needed either.

robinmackaij: This is why I earlier had an overload for `element_to_string` but that was a bit ugly overload due to parameters without defaults having to go before parameters with defaults, leading to a signature clash in the overload (since defaulting is the behavioral switch here). Since this is the only affected place (from a typing perspective) I decided to remove the overload.

src/robot/utils/etreewrapper.py
robinmackaij: I've annotated the `ETSource` `source` as `Any` for now (as per Slack message).

pekkaklarck (approved):

Issue #5548, opened 2025-11-17
Explicit and enhanced `bytes` support to BuiltIn keywords that support bytes and strings


pekkaklarck:
Various validation keywords in BuiltIn currently work with `bytes`, but that support is not explicitly documented or tested. There's generally no `bytes` compatibility code, but `bytes` just has lot of the same methods as `str`:
```python
>>> b'FOO'.lower().startswith(b'f')
True
```

Now that we are adding typing to BuiltIn (#5544), this `bytes` compatibility will be broken if we type arguments as `str`. That would be bad, because it is likely that these keywords are used with `bytes` even though the `bytes` support isn't advertised. A good example is #5054 that added `bytes` related enhancement to `Should Contain`. The problem described in this issue wouldn't occur with `Should Contain`, though, because it accepts also other values than strings (and bytes) and will likely be typed as `object`. Keywords like `Should Start With` that only make sense with strings (and bytes) are in bigger risk.

This is what we need to do:
1. Make sure keywords that work with `bytes` get typing `str | bytes` instead of just `str`. This preserves the current functionality.
2. Add tests. This makes sure current functionality isn't accidentally broken later.
3. Fix possible API compatibility issues between `bytes` and `str` when feasible. For example, case-insensitivity is implemented with `str.casefold`, but `bytes` don't have that method and `bytes.lower` should be used instead. This particular problem is easy to fix with a helper method, in others we may just document that certain feature doesn't work with bytes.
4. Consider enhancing `bytes` support when feasible. For example, `Should Start With` currently works if both the validated value and the expected start value are `bytes`, but validating `bytes` so that the expected value is `str` fails with a `TypeError`. The aforementioned #5054 enhanced `Should Contain` in this regard and doing same enhancement with other `bytes` compatible keywords would be nice.
5. Update documentation. Using `str | bytes` type hint already documents that `bytes` can be used, but adding some documentation about converting expected values could still be good. Concrete enhancements also require "New in RF 7.4." note.

pekkaklarck:
It probably makes sense to first add typing to BuiltIn (#5544) and afterwards implement this. The commit adding typing will be big even without these changes. We can use `str` as the type initially, but using `str | bytes` would be a bit better.

pekkaklarck:
[...]

PR #5554, opened 2025-11-20
5373 type hints string

silentw0lf:


silentw0lf:
Rebased upstream/master

src/robot/libraries/String.py
```diff
@@ -32,6 +32,13 @@
     "[ARABIC]": "".join(chr(c) for c in range(0x0600, 0x0700)),
     "[POLISH]": ascii_lowercase + ascii_uppercase + "ąćęłńóśźżĄĆĘŁŃÓŚŹŻ",
 }
+Error = Union[Literal["strict", "ignore", "replace"], str]
```
pekkaklarck: I'm not sure is it a good idea to list some of the [possible error handlers](https://docs.python.org/3/library/codecs.html#codec-base-classes) separate. Using just `str` as the type would be simpler.

src/robot/libraries/String.py
```diff
@@ -89,8 +98,13 @@ def convert_to_upper_case(self, string):
         """
         return string.upper()
 
-    @keyword(types=None)
-    def convert_to_title_case(self, string, exclude=None):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
+    # TODO(@pekkaklarck): there are some tests indicating that it should only work with str/ not bytes (at least in the past) e.g. Should Be Title Case Does Not Work With ASCII Bytes
```
pekkaklarck: The current explicit check for bytes is due to the keyword otherwise anyway failing with them. For example, `re.split(r"(\s+)", string)` fails if `string` is actually `bytes` because the pattern is `str`. It could be fixed by using `br"(\s+)"` instead with `bytes`, but there are also other places that needed to be adjusted. If the signature is changed to `string: str`, then automatic argument conversion will convert bytes to `str` and the keyword will kind of work. It would return a `str` instead of `bytes`, though, which in my opinion would be wrong.

We have several options how to handle this:
1. Just use `str` as the type.
2. Use `str | bytes` as the type, but explicitly reject bytes like nowadays. It would be a bit odd if the signature would say that bytes are fine but the keyword would fail with them, but at least the behavior would be explicit.
3. Enhance the keyword so that it actually works with bytes. Looking at the code, it shouldn't be overly complicated as `str` and `bytes` APIs are mostly the same. We already did this with various BuiltIn and Collections keywords (#5548) so this would be a logic change also in this case.

Considering that options 1. and 2. have some drawbacks, I believe the option 3. is the way to go. This is what I propose we do in practice:
1. Use typing `str | bytes` as part of this PR. Preserve the current `TypeError` when used with `bytes`.
2. Re-open #5548 with a note that `Convert To Title Case` needs to be updated as well. If there are other similar keywords that don't actually work with bytes in String, they can be mentioned in that issue as well.
3. Merge this PR when it's otherwise ready.
4. Enhance `Convert To Title Case`, and possible other keywords, separately to get #5548 closed again. I can take care of this.

src/robot/libraries/String.py
```diff
@@ -190,11 +212,16 @@ def decode_bytes_to_string(self, bytes, encoding, errors="strict"):
         and `Convert To String` in ``BuiltIn`` if you need to
         convert arbitrary objects to strings.
         """
-        if isinstance(bytes, str):
-            raise TypeError("Cannot decode strings.")
         return bytes.decode(encoding, errors)
 
-    def format_string(self, template, /, *positional, **named):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
```
pekkaklarck: This is similar to `Convert To Title Case` where I think enhancing the keyword so that it actually works with bytes would make sense. In this case that would be harder, though, because there's no `bytes.format` method. I guess we could to something like this to support `bytes`, though:
```python
if isinstance(template, bytes):
    template = template.decode('latin-1')
    return_bytes = True
else:
    return_bytes = False
# current implementation
return result if not return_bytes else result.decode('latin-1')
```
I'm not sure is this really worth the effort, but the alternatives have the same problems as with `Convert To Title Case`. In this case I believe we should do this:
1. Use just `str` as the typing initially.
2. Mention this in #5548 and decide later should bytes support be added or not.

src/robot/libraries/String.py
```diff
@@ -229,13 +256,19 @@ def format_string(self, template, /, *positional, **named):
...
 
-    def get_line_count(self, string):
+    def get_line_count(self, string: "str | bytes") -> int:
         """Returns and logs the number of lines in the given string."""
         count = len(string.splitlines())
         logger.info(f"{count} lines.")
         return count
 
-    def split_to_lines(self, string, start=0, end=None):
+    # does not work with bytes: ${{b'hello\nworld'}} -> Length of '["b'hello\\nworld'"]' should be 2 but is 1.
```
pekkaklarck: You seem to have `str` as the typing but it should be `str | bytes`.

src/robot/libraries/String.py
```diff
@@ -229,13 +256,19 @@ def format_string(self, template, /, *positional, **named):
...
+    def split_to_lines(
+        self,
+        string: str,
+        start: "int | str | None" = 0,
+        end: "int | str | None" = None,
```
pekkaklarck: This is one of the keywords that are affected by #5558. Please use `Literal['']` instead of `str` with `start` and `end`.

src/robot/libraries/String.py
```diff
@@ -262,7 +295,9 @@ def split_to_lines(self, string, start=0, end=None):
         logger.info(f"{len(lines)} line{s(lines)} returned.")
         return lines
 
-    def get_line(self, string, line_number):
+    # TODO(@pekkaklarck):
+    # does not work with bytes: ${{b'hello\nworld'}} -> Length of '["b'hello\\nworld'"]' should be 2 but is 1.
```
pekkaklarck: Use `str | bytes` and not `str` as the typing. Notice that also the return type must be `str | bytes`.

src/robot/libraries/String.py
```diff
@@ -229,13 +256,19 @@ def format_string(self, template, /, *positional, **named):
...
         return count
 
-    def split_to_lines(self, string, start=0, end=None):
+    # does not work with bytes: ${{b'hello\nworld'}} -> Length of '["b'hello\\nworld'"]' should be 2 but is 1.
+    def split_to_lines(
+        self,
+        string: str,
+        start: "int | str | None" = 0,
+        end: "int | str | None" = None,
+    ) -> list[str]:
```
pekkaklarck: This needs to be `"list[str | bytes]"`. Quotes are needed for Python 3.8 compatibility and `bytes` is needed because this keyword ought to work also with bytes and in that case the result is a list of bytes.

Check also other possible cases where you've used parameterized types like `list[str]` and make sure they are quoted.

src/robot/libraries/String.py
```diff
@@ -278,13 +313,14 @@ def get_line(self, string, line_number):
         line_number = self._convert_to_integer(line_number, "line_number")
         return string.splitlines()[line_number]
 
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
```
pekkaklarck: Another case where we need to check can the keyword be enhanced to work with bytes. As commented above, this needs to be separately from this PR. This is what I believe should be done with this keyword:
1. Use `str` as the typing.
2. Mention this in #5548.

src/robot/libraries/String.py
```diff
@@ -316,13 +352,14 @@ def get_lines_containing_string(
             contains = lambda line: pattern in line
         return self._get_matching_lines(string, contains)
 
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
```
pekkaklarck: Another keyword to be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -359,13 +396,14 @@ def get_lines_matching_pattern(
             matches = lambda line: fnmatchcase(line, pattern)
         return self._get_matching_lines(string, matches)
 
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
```
pekkaklarck: Another keyword to be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -403,13 +441,25 @@ def get_lines_matching_regexp(
         match = regexp.search if partial_match else regexp.fullmatch
         return self._get_matching_lines(string, match)
 
-    def _get_matching_lines(self, string, matches):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
```
pekkaklarck: This isn't a keyword so no need to mention this anywhere. The final typing depends on what we want to do with the above keywords.

src/robot/libraries/String.py
```diff
@@ -403,13 +441,25 @@ def get_lines_matching_regexp(
...
+        string: str,
+        matches: Callable[[str], "bool | re.Match[str] | None"],
+    ) -> str:
         lines = string.splitlines()
         matching = [line for line in lines if matches(line)]
         logger.info(f"{len(matching)} out of {len(lines)} lines matched.")
         return "\n".join(matching)
 
-    def get_regexp_matches(self, string, pattern, *groups, flags=None):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
```
pekkaklarck: To be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -403,13 +441,25 @@ def get_lines_matching_regexp(
...
+    def get_regexp_matches(
+        self,
+        string: str,
+        pattern: str,
+        *groups: object,
+        flags: Flags = None,
+    ) -> list["str | tuple"]:
```
pekkaklarck: I guess the correct typing for the return value would be `"list[str | tuple[str, ...]]"`. Notice that quotes need to be around the whole construct for Python 3.8 compatibility.

src/robot/libraries/String.py
```diff
@@ -448,13 +498,21 @@ def get_regexp_matches(self, string, pattern, *groups, flags=None):
...
-    def _parse_group(self, group):
+    def _parse_group(self, group: "str | int") -> "str | int":
         try:
             return int(group)
         except ValueError:
             return group
 
-    def replace_string(self, string, search_for, replace_with, count=-1):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
+    # e.g. b.replace(b"old", "new") -> TypeError
```
pekkaklarck, line 491: To be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -478,14 +536,15 @@ def replace_string(self, string, search_for, replace_with, count=-1):
         count = self._convert_to_integer(count, "count")
         return string.replace(search_for, replace_with, count)
 
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
```
pekkaklarck: To be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -519,7 +578,8 @@ def replace_string_using_regexp(
             flags=parse_re_flags(flags),
         )
 
-    def remove_string(self, string, *removables):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
```
pekkaklarck: To be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -543,7 +603,13 @@ def remove_string(self, string, *removables):
             string = self.replace_string(string, removable, "")
         return string
 
-    def remove_string_using_regexp(self, string, *patterns, flags=None):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
+    def remove_string_using_regexp(
```
pekkaklarck, line 587: To be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -564,8 +630,13 @@ def remove_string_using_regexp(self, string, *patterns, flags=None):
             string = self.replace_string_using_regexp(string, pattern, "", flags=flags)
         return string
 
-    @keyword(types=None)
-    def split_string(self, string, separator=None, max_split=-1):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
+    def split_string(
```
pekkaklarck, line 613: To be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -590,8 +661,13 @@ def split_string(self, string, separator=None, max_split=-1):
         max_split = self._convert_to_integer(max_split, "max_split")
         return string.split(separator, max_split)
 
-    @keyword(types=None)
-    def split_string_from_right(self, string, separator=None, max_split=-1):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
+    def split_string_from_right(
```
pekkaklarck, line 643: To be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -606,15 +682,25 @@ def split_string_from_right(self, string, separator=None, max_split=-1):
         max_split = self._convert_to_integer(max_split, "max_split")
         return string.rsplit(separator, max_split)
 
-    def split_string_to_characters(self, string):
+    # TODO(@pekkaklarck):
+    # def split_string_to_characters(self, string: "str | bytes") -> list["str | bytes"]:
+    # does not work:  Split String To Characters    ${{b'ab 12'}} -> [97, 98, 32, 49, 50]   int represtantion in robot hmmm
```
pekkaklarck: This is due to `bytes` being considered a sequence of integers in Python:
```python
>>> list(b'hello')
[104, 101, 108, 108, 111] 
```
I guess it would make sense in the context of this keyword to convert those integers to byte characters. Can be done like this:
```python
>>> [bytes([i]) for i in list(b'hello')]
[b'h', b'e', b'l', b'l', b'o']
```
Doing that as part of this issue doesn't make sense so better to just mention this in #5548. The initial typing could nevertheless be `str | bytes`.

src/robot/libraries/String.py
```diff
@@ -606,15 +682,25 @@ def split_string_from_right(self, string, separator=None, max_split=-1):
...
         """Splits the given ``string`` to characters.
 
         Example:
         | @{characters} = | Split String To Characters | ${string} |
         """
+
         return list(string)
 
-    def fetch_from_left(self, string, marker):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
```
pekkaklarck: To be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -624,7 +710,8 @@ def fetch_from_left(self, string, marker):
         """
         return string.split(marker)[0]
 
-    def fetch_from_right(self, string, marker):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
+    def fetch_from_right(self, string: str, marker: str) -> str:
```
pekkaklarck, line 682: To be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -697,8 +793,13 @@ def get_substring(self, string, start, end=None):
         end = self._convert_to_index(end, "end")
         return string[start:end]
 
-    @keyword(types=None)
-    def strip_string(self, string, mode="both", characters=None):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
```
pekkaklarck: To be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -792,8 +893,13 @@ def should_be_upper_case(self, string, msg=None):
         if not string.isupper():
             raise AssertionError(msg or f"{string!r} is not upper case.")
 
-    @keyword(types=None)
-    def should_be_title_case(self, string, msg=None, exclude=None):
+    # TODO(@pekkaklarck): some paramters could be string, some bytes -> risk of incompatibilies
```
pekkaklarck: To be mentioned in #5548.

src/robot/libraries/String.py
```diff
@@ -32,6 +32,13 @@
     "[ARABIC]": "".join(chr(c) for c in range(0x0600, 0x0700)),
     "[POLISH]": ascii_lowercase + ascii_uppercase + "ąćęłńóśźżĄĆĘŁŃÓŚŹŻ",
 }
+Error = Union[Literal["strict", "ignore", "replace"], str]
+Flags = Union[
+    Literal["IGNORECASE", "MULTILINE", "VERBOSE", "IGNORECASE | VERBOSE"], str, None
+]
```
pekkaklarck: I don't think listing these explicitly makes sense. We cannot anyway enumerate all possible combinations.

src/robot/libraries/String.py
```diff
@@ -32,6 +32,13 @@
...
+Markers = Union[
+    Literal["[LOWER]", "[UPPER]", "[LETTERS]", "[NUMBERS]", "[ARABIC]", "[POLISH]"], str
+]
```
pekkaklarck: Similarly as above, I don't think listing these explicitly is a good idea. These markers can also be used like `[NUMBERS]abcdef` which doesn't match any of the literal values. Using `str` allows that, but I believe all this still makes the signature unnecessarily complicated.

pekkaklarck:
At least some of the failures on CI are due to the use of parameterized types like `list[int]` without quotes. Quoting like `"list[int]"` is needed because we still support Python 3.8.

silentw0lf:
> I only looked at the code changes and they mostly looks good.
> 
> Keywords that don't currently work properly with strings and bytes should be listed in #5548 that I'll reopen. I can then go through them myself after this PR has been merged. Their initial typing as part of this PR can be either `str` or `str | bytes` depending on the context.

done.

src/robot/libraries/String.py
silentw0lf: Used just `str`

src/robot/libraries/String.py
silentw0lf: done.

src/robot/libraries/String.py
silentw0lf: done.

src/robot/libraries/String.py
silentw0lf: done. (listed all)

src/robot/libraries/String.py
silentw0lf: done. (listed all)

src/robot/libraries/String.py
silentw0lf: done. (listed all)

src/robot/libraries/String.py
silentw0lf: Is this the correct link? https://github.com/robotframework/robotframework/issues/5558 is just a link to "Add annotations to String", also there you reference 5558? 


Could you help me out? I can not see the explanation for `Literal['']`

src/robot/libraries/String.py
silentw0lf: Yes I left a comment there (which I now removed in new commits):

Testing showed:
```
does not work with bytes  (when using `str | bytes`): ${{b'hello\nworld'}} -> Length of '["b'hello\\nworld'"]' should be 2 but is 1.
```
This seems to be broken. 
Is also mentioned in  https://github.com/robotframework/robotframework/issues/5548.

src/robot/libraries/String.py
silentw0lf: * Used typing str | bytes as part of this PR. Preserve the current TypeError
* left a note referencing this comment in 5548

src/robot/libraries/String.py
silentw0lf: * use just str
* left a note referencing this comment in 5548

src/robot/libraries/String.py
silentw0lf: left a note referencing this comment in 5548

src/robot/libraries/String.py
silentw0lf: Update: I think I understand the problem and fixed it for now.

silentw0lf:
@pekkaklarck  I took time to answer your review notes and finished most of them.
 for the autotmatic tests: I have no access to the details of a test fail after 20 s
 
image

How is it possible to see the errors?



Update: 
In the artifacts I could see this only this error multiple times:

FAIL: String.Convert To.Convert To Title Case with excludes
Status of 'Convert To Title Case with excludes' should have been PASS but it was FAIL.

```
Error message:
Several failures occurred:
1) error: unterminated character set at position 0
2) error: unterminated character set at position 0
```

I tried to reproduce this error and figured out that  **`Iterable[str]`** is causing this problem for `python3.8`, while   i figures out that `list[str]` fixes this error



e.g.
```
    def convert_to_title_case(
        self,
        string: "str | bytes",
        exclude: "Iterable[str] | str | None" = None,
    ) -> str:
```
Tests are passed when using   `exclude: "list[str] | str | None" = None`


What should we do here? I remember , you hat the idea to use Iterable[str]
Any thaugts?

pekkaklarck:
I'd be surprised if `list[str]` would cause these failures. They look more like issues with tests themselves. Did you run tests locally? Do they pass?

silentw0lf:
> I'd be surprised if `list[str]` would cause these failures. They look more like issues with tests themselves. Did you run tests locally? Do they pass?

I made a mistake, writing it down.

Local Tests with python3.8  (focus on parameter `exclude: "Iterable[str] | str | None" = None,`

**fail** with   `exclude: "Iterable[str] ...."` (error: unterminated character set at position 0)
**pass** with `exclude: "list[str] ...."`

pekkaklarck:
Do all tests pass for you locally with Python 3.8? I can try that myself as well. Have you pushed all changes?

pekkaklarck:
I tested this locally and can confirm that four tests fail with Python 3.8 and 3.9 when using typing `"Iterable[str] | str | None"`, but not when using `"list[str] | str | None"`. I believe the reason is this:
1. Python 3.8 and 3.9 don't support the `x | y` typing syntax (that's why the type is in quotes to begin with).
2. Robot initially gets types with `typing.get_type_hints()`. It handles stringified types so that `"x | y"` syntax is evaluated, but because Python 3.8 and 3.9 don't support the `x | y` syntax, evaluation fails and `typing.get_type_hints()` raises an exception.
3. As a backup Robot parses the stringifies type itself using its own type hint parser. The parser apparently doesn't recognize `Iterable` and thus the typing is the same as `Unknown | str | None`. This only happens with Python 3.8 and 3.9, the newer see the typing correctly.
4. None of the types in `Unknown | str | None` match a list of strings and thus lists are converted. The first valid type is `str`, so the a list like `['a', 'an', 'the', 'to', 'is']` that's used in one of the failing tests is converted to a string.
5. The resulting string `['a', 'an', 'the', 'to', 'is']` is then split to excluded words from `,`. The list of excludes will become `["['a'", " 'an'", " 'the'", " 'to'", " 'is']"]`.
6. These excludes support regexps, so they are compiled like `exclude = [re.compile(e) for e in exclude]`.
7. The first item in the exclude list is string `['a'`. `[` is a special character in regexp and compiling this pattern fails:
   ```python
   >>> re.compile("['a'")
     ...
   re.error: unterminated character set at position 0
    ```

pekkaklarck:
The above explains why tests fail. There are three ways we can make them pass:
1. Enhance the stringified type parser so that it recognizes `Iterable`.
2. Change the typing to `Union[Iterable[str], str, None]` instead.
3. Use `"list[str] | str | None"`.

I believe the last option is the best one. I don't remember what I have earlier said about using `Iterable`, but at least now that I think about this, I don't see benefits with accepting iterables instead of lists. Automatic argument conversion anyway converts other sequences like tuples to lists and I see no reason to use any non-sequence iterables. Another benefit of `list[str]` over `Iterable[str]` is that users ought to be familiar with the term "list", but they may not understand "iterable".

atest/testdata/standard_libraries/string/string.robot
```diff
@@ -41,6 +41,12 @@ Split To Lines
     @{result} =    Split To Lines    ${EMPTY}
     Length Should Be    ${result}    0
 
+    # does not work
+    # @{result} =    Split To Lines    ${BYTES TEXT IN COLUMNS}
+    # Length Should Be    ${result}    2
+    # Should be equal    ${result}[0]    ${BYTES FIRST LINE}
+    # Should be equal    ${result}[1]    ${BYTES SECOND LINE}
```
pekkaklarck, line 48: Splitting bytes to lines should work. Is the typing of the keyword correct?

pekkaklarck (changes requested):
Looks very good. Only very minor changes needed. I can do them myself as well if you don't have time.

pekkaklarck:
When reviewing this PR, I noticed that our `str` conversion doesn't handle bytes too well at the moment. I submitted #5567 about this and have already implemented it. This may affect String library tests so merging master is a good idea.

It could now be a good idea to add tests for `Encode String To Bytes` with bytes values. That can be done as part of #5548 as well, though.

silentw0lf:
> Looks very good. Only very minor changes needed. I can do them myself as well if you don't have time.

Yes you can do the minor changes. I would at earliest have time tommorow evening.

pekkaklarck:
This is such a good shape that I'll fix the issues myself and merge this. I can then continue with other issues that are related to String.

pekkaklarck:
I changed `Iterable` to `list` and all tests passed. I merged the PR and will look at possible other changes as part of #5548.

Issue #5383, opened 2025-03-25
Request for CLI Argument --timeout in Robot Framework


Sam-Eldin:
I would like to request a new CLI argument, --timeout, to set an upper time limit for the entire Robot Framework test execution.

I am aware that timeouts can be defined at the suite and test levels, but I am building an automation system that runs Robot Framework tests for multiple teams. Many of these teams do not have timeouts set in their tests, and I need a way to enforce a global timeout for better control.

Currently, I have a timeout for the command handler itself (e.g., an SSH session), but this is not ideal because:

- If the timeout is reached, Robot Framework does not generate a complete report—tests are left in an incomplete state.
- Teams cannot determine which test or stage got stuck.
- Debugging test failures becomes difficult.

Proposed Solution:
Introduce a CLI argument like:

`robot --timeout 3h test.robot`

This argument would:

- Act as an upper limit for test execution.
- Override any timeouts already defined in test cases or suites.
- Ensure that a proper Robot Framework report is generated when the timeout is reached.

Would this be possible to implement? I believe it would greatly benefit users managing large-scale test automation.

Let me know if I can provide more details!

pekkaklarck:
I can see that this would be useful in some cases and believe implementing it wouldn't be too complicated.

Nowadays when Robot runs a keyword, it checks are there test or keyword timeouts active. If there are multiple, it uses the timeout that has least time remaining. Implementing this enhancement would basically require creating a new total timeout type that would be active during the whole execution. It would always be one of the timeout candidates and rest of the timeout functionality ought to work without bigger changes.

I won't have time to look at this myself in the near future, but would be happy to help if someone else is interested.

hassineabd:
@pekkaklarck 
it is similar to expand variable issue , and I have submitted a PR before , i Think i can take it and try to get it done together with the new design of expanding variable ticket

Sam-Eldin:
Good day mates. 
Is there any progress with the request?

Your efforts are really appreciated

hassineabd:
Hello @pekkaklarck ,  I have worked on this issue and  here are some details on the implementation:

- new CLI Option: --timeout 
- accepts standard Robot time formats (e.g., 30m, 1h) and supports runtime variables.
- total_timeout is exposed as a property in _ExecutionContext, delegating to EXECUTION_CONTEXTS.total_timeout for global accessibility.
[...]

PR #5594, opened 2026-01-10
Issue #5383 - Feature: Added global execution cli timeout

hassineabd:
Issue #5383

Key Changes:

- New CLI Option: --timeout 
- accepts standard Robot time formats (e.g., 30m, 1h) and supports runtime variables.
- total_timeout is exposed as a property in _ExecutionContext, delegating to EXECUTION_CONTEXTS.total_timeout for global accessibility.
- added _process_timeout in _BaseSettings to validate values immediately --> Invalid time strings or negative values trigger a failure before execution starts, ensuring fast feedback.
- a new class TotalTimeout to manage the global timer. It ensures that when the limit is hit, 
- execution stops with messages: "Execution stopped: Total execution timeout of {value} exceeded."

acceptance tests :  atest/robot/running/global_timeout.robot validates 16 distinct scenarios including nested suites, dry-run, variable expansion, and CLI validation.

### DIFF
--- a/atest/testresources/listeners/TimeoutIncreaseListener.py
+++ b/atest/testresources/listeners/TimeoutIncreaseListener.py
@@ -1,0 +1,30 @@
+
+from robot.running.context import EXECUTION_CONTEXTS
+from robot.running.timeouts import TotalTimeout
+
+class TimeoutIncreaseListener:
+    ROBOT_LISTENER_API_VERSION = 3
+
+    def __init__(self, timeout_string):
+        self.timeout_string = timeout_string
+
+    def start_suite(self, data, result):
+        try:
+             variables = EXECUTION_CONTEXTS.current.variables
+        except AttributeError:
+             variables = None 
+        
+        old_timeout = EXECUTION_CONTEXTS.total_timeout
+        print(f"DEBUG: Timeout before: {old_timeout}")
+        
+        # parse new timeout values using a temporary object
+        new_timeout_obj = TotalTimeout(self.timeout_string, variables)
+        
+        # time_left() relies on self.timeout and start_time
+        # update self.timeout (duration)
+        if new_timeout_obj.timeout is not None:
+            old_timeout.timeout = new_timeout_obj.timeout
+            old_timeout.string = new_timeout_obj.string
+            # this way do NOT reset start_time, so it counts from original start.
+                 
+        print(f"DEBUG: Timeout after: {EXECUTION_CONTEXTS.total_timeout}")
--- a/atest/robot/running/global_timeout.robot
+++ b/atest/robot/running/global_timeout.robot
@@ -90,3 +90,22 @@ Global Timeout with ExitOnFailure
     [Documentation]    Compatibility with --exitonfailure.
     Run Tests    --timeout 1s --exitonfailure --test "Long Sleep"    running/global_timeout/basic.robot
     Check Test Case    Long Sleep
+
+Global Timeout Via Listener
+    [Documentation]    Verify listener can set global timeout (0.1s) that overrides test timeout (default or none).
+    Run Tests    --listener TimeoutListener:0.1s --test "Long Sleep"    running/global_timeout/basic.robot
+    Check Test Case    Long Sleep    FAIL    Total Execution timeout 100 milliseconds exceeded.
+
+Increase Timeout Via Listener
+    [Documentation]    Verify listener can INCREASE global timeout (0.5s -> 3s) allowing a 2s test to pass.
+    Run Tests    --timeout 0.5s --listener TimeoutIncreaseListener:3s --test "Long Sleep"    running/global_timeout/basic.robot
+    Check Test Case    Long Sleep    PASS    ${EMPTY}
+    Stdout Should Contain    DEBUG: Timeout before: 500 milliseconds
+    Stdout Should Contain    DEBUG: Timeout after: 3 seconds
+
+Decrease Timeout Via Listener
+    [Documentation]    Verify listener can DECREASE global timeout (3s -> 0.1s) causing a 2s test to fail.
+    Run Tests    --timeout 3s --listener TimeoutIncreaseListener:0.1s --test "Long Sleep"    running/global_timeout/basic.robot
+    Check Test Case    Long Sleep    FAIL    Total Execution timeout 100 milliseconds exceeded.
+    Stdout Should Contain    DEBUG: Timeout before: 3 seconds
+    Stdout Should Contain    DEBUG: Timeout after: 100 milliseconds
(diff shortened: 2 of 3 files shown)


atest/testresources/listeners/TimeoutIncreaseListener.py
```diff
@@ -0,0 +1,30 @@
...
+        new_timeout_obj = TotalTimeout(self.timeout_string, variables)
+        
+        # time_left() relies on self.timeout and start_time
+        # update self.timeout (duration)
+        if new_timeout_obj.timeout is not None:
+            old_timeout.timeout = new_timeout_obj.timeout
+            old_timeout.string = new_timeout_obj.string
+            # this way do NOT reset start_time, so it counts from original start.
+                 
+        print(f"DEBUG: Timeout after: {EXECUTION_CONTEXTS.total_timeout}")
```
pekkaklarck, line 30: `EXECUTION_CONTEXT` is not part of the public API and we have a plan to replaces this global variable with execution related context object in the future. `EXECUTION_CONTEXT` can be used in tests if needed, but it shouldn't be used in examples demonstrating how users could utilize it.

In general I don't consider the possibility increase the global timeout important and it should be part of this overall issue. Setting a maximum timeout for the whole execution is fine, but there are already other ways to stop the whole execution conditionally if needed.

src/robot/conf/settings.py
```diff
@@ -379,6 +381,20 @@ def _validate_expandkeywords(self, values):
                     f"Expected 'TAG:<pattern>' or 'NAME:<pattern>', got '{opt}'.",
                 )
 
+    def _process_timeout(self, value):
+        from robot.variables.search import contains_variable
+        if not value or value.upper() == "NONE":
+            return None
+        if contains_variable(value):
+            return value
```
pekkaklarck, line 389: No other command line option supports variables and this option shouldn't support them either. Environment variables can be used like `--timeout $TIMEOUT` if something like this is needed.

src/robot/running/context.py
```diff
@@ -67,6 +67,7 @@ class ExecutionContexts:
     def __init__(self):
         self._contexts = []
         self._asynchronous = Asynchronous()
+        self.total_timeout = None
```
pekkaklarck, line 70: This should be `_total_timeout` to not encourage anyone to touch it. Such code will break at some point.

src/robot/running/context.py
```diff
@@ -100,10 +102,14 @@ def end_suite(self):
 
 class _ExecutionContext:
 
+    @property
+    def total_timeout(self):
+        return EXECUTION_CONTEXTS.total_timeout
```
pekkaklarck, line 107: This should be `_total_timeout` or not exist at all to not encourage users to touch it.

src/robot/running/context.py
```diff
@@ -100,10 +102,14 @@ def end_suite(self):
...
+
     def __init__(self, suite, namespace, output, dry_run=False, asynchronous=None):
         self.suite = suite
         self.test = None
-        self.timeouts = []
+        self.timeouts = [EXECUTION_CONTEXTS.total_timeout] if EXECUTION_CONTEXTS.total_timeout else []
```
pekkaklarck, line 112: The current execution context handling is a big mess that I hope to clean up at some point, but this code adds even more coupling between classes. A simple way to simplify this is adding `global_timeout=None` to `__init__` signature and then doing something like
```python
self.timeouts = [global_timeout] if global_timeout else []
```
here. If `self.timeouts` would need to be re-initialized later, `global_timeout` could be stored to `self.global_timeout`, but I hope we could avoid that.

src/robot/running/context.py
```diff
@@ -283,7 +290,7 @@ def _remove_timeout(self, timeout):
 
     def end_test(self, test):
         self.test = None
-        self.timeouts = []
+        self.timeouts = [EXECUTION_CONTEXTS.total_timeout] if EXECUTION_CONTEXTS.total_timeout else []
```
pekkaklarck, line 293: I think we should be able to pop the possible test timeout from `self.timeout` here instead of re-initializing whole `self.timeouts`. Earlier doing `self.timeouts = []` was so simple that it made sense, but popping ought to be simpler than re-initialization if global timeout exists.

src/robot/running/suiterunner.py
```diff
@@ -44,6 +44,14 @@ def __init__(self, output, settings):
...
         self.executed = [NormalizedDict(ignore="_")]
         self.skipped_tags = TagPatterns(settings.skip)
+        self.total_timeout = self._get_total_timeout()
+
+    def _get_total_timeout(self):
+        """Create total timeout if --timeout option was used."""
+        timeout = self.settings["Timeout"]
+        if timeout:
+            return TotalTimeout(timeout, self.variables)
+        return None
```
pekkaklarck, line 54: Does `SuiteRunner` need `total_timeout`? Wouldn't it be enough if contexts has it?

src/robot/running/suiterunner.py
```diff
@@ -128,6 +139,8 @@ def end_suite(self, suite: SuiteData):
         )
         with self.context.suite_teardown():
             failure = self._run_teardown(suite, self.suite_status, self.suite_result)
+            if not failure and self.context.total_timeout and self.context.total_timeout.timed_out():
+                failure = ExecutionFailed(self.context.total_timeout.get_message(), test_timeout=True)
```
pekkaklarck, line 143: `test_timeoout=True` isn't exactly correct in this case. Not sure how much it matters.

src/robot/running/timeouts/timeout.py
```diff
@@ -142,3 +142,25 @@ def any_timeout_occurred(self):
 
 class KeywordTimeout(Timeout):
     kind = "KEYWORD"
+
+
+class TotalTimeout(Timeout):
+    kind = "TOTAL EXECUTION"
```
pekkaklarck, line 148: As discussed on Slack, this should be `GlobalTimeout` instead of `TotalTimeout` and that terminology should be used consistently. The `kind` attribute should probably have value `GLOBAL`.

src/robot/running/timeouts/timeout.py
```diff
@@ -142,3 +142,25 @@ def any_timeout_occurred(self):
...
+            self._reported = True
+        return super().get_runner()
+
+    def timed_out(self):
+        timeout_occurred = super().timed_out()
+        if timeout_occurred and not self._reported:
+            from robot.output import LOGGER
+            msg = f"Execution stopped: Total execution timeout of {self.string} exceeded."
+            LOGGER.error(msg)
+            self._reported = True
```
pekkaklarck, line 165: Why is this situation reported separately as an error?

Issue #5610, opened 2026-02-04
TimeoutExceeded silenced by try-except Exception


JFoederer:
`TimeoutExceeded` is an exception defined by Robot Framework that is raised when a test or keyword times out. Its effect is very similar to that of Python's builtin `KeyboardInterrupt`, in that it is an exception that is raised independent from the code that is currently being executed.

Consider the following code:
```
try:
    float(value)
except (TypeError, ValueError, OverflowError, ZeroDivisionError):
    do_something()
```

The except clause catches, and implements the same behaviour, for all possible exceptions that can be raised by `float()`. Therefore you could expect that the exact same result can be achieved by writing:
```
try:
    float(value)
except Exception:
    do_something()
```

For `KeyboardInterrupt` this is true, because it is not derived from `Exception` but from `BaseException`. Exactly for this reason. However, `TimeoutExceeded` **is** caught in the second code snippet, even though it is not an exception raised by the sole enclosed call.

In practise the above example isn't the biggest issue. The real issue is that catch-all clauses are more common than we might like. Perhaps in our own libraries, because the underlying libraries we use are not 100% transparent about the exceptions they can raise, now or in the future. Or these libraries already have internal catch-all clauses of their own.

To prevent accidental silencing of Robot timeouts and make its behaviour more predictable and in line with similar constructs, like `KeyboardInterrupt`, the proposal is to derive `TimeoutExceeded` from `BaseException` instead of `RobotError`.

See also this [discussion on Slack](https://robotframework.slack.com/archives/C3C28F9DF/p1765558473110399).

pekkaklarck:
Using `except Exception:` is generally consider bad practice because it can catch exceptions that shouldn't be caught. I believe most linters warn about it. That said, I know it's used widely and it is certainly much better than using bare `except:` that catches all exceptions, including `SystemExit` and `KeyboardInterrupt`. Changing `TimeoutExceeded` so that it isn't caught by `except Exception:` could thus avoid problems with libraries and other tools that use the unsafe `except Exception:` pattern (but don't use the even worse `except:` pattern).

[...]

Issue #5377, opened 2025-03-21
Document how libraries can do cleanup activities if Robot's timeout occurs


pekkaklarck:
Issue #5376 proposes enhancing the Process library so that if Robot's timeout occurs when a process is waited to end, the process is killed instead of leaving it running on the background. That turned out to be surprisingly simple. Because other libraries may have similar needs, we should document how to do that in the User Guide. In addition to that, we should make sure that the related APIs are good.

When Robot's timeout occurs, `robot.errors.TimeoutError` is raised in the library code. It can basically occur at any point similarly as the standard `KeyboardInterrupt` or `SystemExit`. The simplest way to do cleanup activities if any exception occurs is using `try/finally`. It is really simple and straightforward, and it also has a benefit that the exception is automatically re-raised:

```python
def keyword():
    try:
        do_something()
    finally:
        do_cleanup()
```

In some cases there may be a need to catch Robot's `TimeoutError` explicitly. That's easy as well, but then the exception should be re-raised explicitly:

```python
from robot.errors import TimeoutError

def keyword():
    try:
        do_something()
    except TimeoutError:
        do_cleanup()
        raise
```

A problem with the above is that Python also has its own [TimeoutError](https://docs.python.org/3/library/exceptions.html#TimeoutError). This is problematic because `TimeoutError` shadows the standard exception and also because you don't know which exception `except TimeoutError:` refers to when you see just that line. Now that we are planning to make our `TimeoutError` part of the public API, I believe we should rename it to avoid confusion. I think `TimeoutExceeded` would be fine, it's different to `subprocess.TimeoutExpired`, but I'm open for other ideas as well. To avoid backwards compatibility problems, we should preserve the old name as an alias at least until RF 9.

If you wonder why we added `TimeoutError` in the first place when Python also has `TimeoutError`, the reason is that ours was there first. It existed already in RF 2.0, the first open source version, in 2008, but Python got its `TimeoutError` only in Python 3.3 in 2012.

PR #5611, opened 2026-02-05
Timeout exceeded not catchable as std Exception

JFoederer:
This pull request resolves the issue raised in #5610.


pekkaklarck:
Looks good. The new behavior should be documented in the User Guide in the new section added as part of #5377 in revision 29177a312d4078e9f670e99a7933198a12d475be, though.

themavik (commented):
Reviewed the changes — the implementation looks correct and follows the existing patterns.

themavik (commented):
nit: TimeoutExceeded docstring still says "cannot be caught be TRY/EXCEPT"—should be "by".

Worth a one-liner in the userguide that bare `except:` in library code still catches BaseException (including this), so the footgun is narrower than only `except Exception`.

Issue #5607, opened 2026-01-30
Add Keyword for easy send and read interact with PIPE for Process Library


IlfirinPL:
It would be use full to allow easy send and read for handling process

Methods like

Send Input To Process |  ${handle}  |  String
This keyword allow to send string to process

Read Output From Process | |  ${handle}  |  wait_for_line=${True} | ${minimum of line}=1
This keyword allow to read latest not read yet output, with option to wait for line of ouput. This last option will make code cleaner

Example Use
 ${handle}  |  Run Process | command | stdin=PIPE | stdout=PIPE
Send Input To Process |  ${handle}  |  String
${line} | Read Output From Process | |  ${handle}  |  wait_for_line=${True} | ${minimum of line}=1
Log | ${line} | Debug

pekkaklarck:
Something like this would be handy, but I'm afraid implementing it isn't too easy. If I've got it right, the only convenient API the underlying [subprocess](https://docs.python.org/3/library/subprocess.html) module provides for communication is [Popen.communicate](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate). It doesn't work like you propose, though, because it can only be used once and it (optionally) inputs text to the process and reads results in one go. Anything more complicated would probably require interacting with `stdin`, `stdout` and `stderr` directly. That may not sound too bad, but my understanding is that such code _very_ easily deadlocks and implementing safe, generic code is _really_ hard. Are you @IlfirinPL, or someone else, interested to study this and possibly create a prototype?

Kind of good news is that our `Wait For Process` already uses `communicate` if it needs to wait for the process to end. It thus shouldn't be too hard to enhance it so that it gets a text to write to `stdin` as an optional argument. I'm just not sure would that actually be any different than giving `stdin` already to `Start Process`. Anyone interested to test that? If there's a difference, this particular enhancement certainly can be done.

IlfirinPL:
I would try to make prototype, but I expect to make it crude.

Currently I set output to file (std and err), get whole file and "tail" last lines, based on that I make rest like Wait For Process To Output, but I expect problem with long running proces with bigger output.

IlfirinPL:
Added @pekkaklarck  Pull request with 3 methods (

- Send Input To Process, 
- Read Output From Process, 
- Process Output Should Contain 

Sending to input work without any issues, but reading part is not pretty require using files as output. I tried using steams but they didn`t provide any output during exaction. Method with files was only way that work reliable .

[...]

PR #5612, opened 2026-02-08
#5607 add methods Send Input To Process, Read Output From Process, Process Output Should Contain

IlfirinPL:
Add Following Methods
#5607 add methods Send Input To Process, Read Output From Process, Process Output Should Contain

### DIFF
--- a/atest/testdata/standard_libraries/process/interactive.robot
+++ b/atest/testdata/standard_libraries/process/interactive.robot
@@ -1,0 +1,49 @@
+*** Settings ***
+Resource           process_resource.robot
+
+Test Setup    Start Interactive Python
+Test Teardown    Terminate Process    ${PROCESS}
+
+*** Variables ***
+${PROCESS}    ${None}
+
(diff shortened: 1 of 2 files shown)


pekkaklarck:
I took a quick look at the code and I'm sorry to say that it doesn't look too good. Especially `Read Output From Process` is problematic. Requiring users to write outputs to files and then just reading content from them avoids deadlocks and other such problems, but it has various other problems:
1. Requiring outputs to be redirected to files is not convenient.
2. If outputs are redirected to files, users can as easily read content from them directly. 
3. If you read outputs multiple times, there's no way to get just the new content.
4. There is no reliable way to make sure the process has written all content. The keyword has a sleep in the beginning that apparently tries to accomplish that, but sleeping is not a good approach for synchronization. 

There are also various problems with the code style, test coverage, etc., but I don't want to spend time going through these problems in detail if the above issues aren't resolved. A problem is that I'm not sure is it even possible to fix the problems reliably.

Issue #2549, opened 2017-02-16
HTML support for set_test_documentation() and set_suite_documentation()


yahman72:
The title says it all: it would nice to have the \*HTML\* support also for the `set test documentation` and `set_suite_documentation` KW's. This would also be consistent with the `set test message` that supports the \*HTML\*.

My Use Case was:
I was trying to make links to execution logs (downloaded from SUT and stored in directories with timestamps) from the report.html. 
When I ran into problems with the `set test message` and HTML (see #2548) I tried to use `set test documentation` only to notice that it doesn't support HTML at all.

pekkaklarck:
Notice that you can use Robot's [custom documentation formatting](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#documentation-formatting) in docs already now. Being able to use HTML freely would be more powerful, though. A simple solution would be allowing `*HTML*` in the beginning of the documentation similarly as we allow it in exception messages. Are you interested to provide a PR?

mcoderr:
Hello,
may i can work on this ticket

pekkaklarck:
Go ahead!

geekypandey:
Hi! Is this issue stale, or can I take up this issue?

pekkaklarck:
Seems nobody is working with this so go ahead @geekypandey!

mcoderr:
Please go head

pekkaklarck:
This issue has some design decisions to be made that are related to #4068. The idea of that issue is to make test and suite messages available both as plain text and as HTML and there are s

We have #4068 about making test and suite messages available both as plain text and HTML. There are lot of similarities to this issue on implementation level. I just commented that issue related to [possible design](https://github.com/robotframework/robotframework/issues/4068#issuecomment-918354447).

pekkaklarck:
In addition to `set_xxx_documentation` methods supporting HTML, we should make it possible to use HTML when defining suite, test and keyword documentation in data:

```robotframework
*** Settings ***
Documentation    *HTML* Example <b>doc</b> in HTML.
```

Most likely supporting this wouldn't make the implementation any more complicated.

geekypandey:
okay! I will start looking in the issue.

dadwi91:
If this is still open, I am gonna pick this up

pekkaklarck:
Yes, this is still open. See also related #4068. Ping me on the #devel channel in our Slack if you need any help!

pekkaklarck:
[...]

Issue #5602, opened 2026-01-20
Support Markdown in suite and test/task documentation


pekkaklarck:
Currently suite and test documentation only supports Robot's own custom documentation format. We are planning to add Markdown support to Libdoc in RF 7.5 (#5304) and I believe we should support Markdown also in test and suite documentation. Because Markdown is so ubiquitous, we can also consider making it the default format in the future.

There are some design decisions to be made related to this. I'll cover them in separate comments.

pekkaklarck:
The first decision to be made is how to specify that documentation is Markdown. It would be easy to just switch from the Robot format to Markdown, but it would also be badly backwards incompatible. Somehow automatically detecting the format would be great, but I don't think that is possible reliably. That basically means that we need to allow configuring the format. This is what I propose we do:

1. Allow specifying the documentation format on a file level using `doc format: Markdown` before the first test data section. This is the same syntax we support when specifying localization language (e.g. `language: Finnish`).
   - This information should be available in the parsing model and added to the execution and result `TestSuite` objects.
   - `TestCase` objects can query the doc format from their parent suite.
   - This will also require storing the doc format in XML and JSON outputs.
2. Consider supporting the `doc format:` configuration in suite initialization files so that it is applied also to child suites. This would make it easier specify the format to larger suites. Child suites should naturally be able to override the configuration with their own `doc format:`. If this is done, we probably should make the `language:` configuration recursive as well, but that requires a separate issue.
3. Support `--doc-format` when running tests for specifying the default documentation format. Per-file configuration should have precedence over it.
4. Support the `doc format:` configuration also in resource file to allow specifying the documentation format for user keywords without needing `--doc-format` with Libdoc.

This is an example how this would look in data:
```robotframework
doc format: markdown

*** Settings ***
Documentation    Hello, [Robot](http://robotframework.org)! This is *Markdown*!
```

The exact spelling of the `doc format` marker can still be discussed. Alternatives include `doc_format`, `doc-format`, `documentation format`, etc. I kind of like the shorter version `doc format` over `documentation format` because we already use `--doc-format` and in the code we'd use `doc_format` to match existing `doc`. The marker will be case-insensitive and ignoring spaces and possibly also underscores and hyphens can be considered.

pekkaklarck:
[...]

PR #5624, opened 2026-03-03
Feature 2549 - support HTML in test and suite documentation

msamo-n:
Proposal for [#2549 - HTML support for set_test_documentation() and set_suite_documentation()](https://github.com/robotframework/robotframework/issues/2549)

### DIFF
--- a/atest/robot/output/js_model.robot
+++ b/atest/robot/output/js_model.robot
@@ -12,6 +12,11 @@ Strings like </script> logged as HTML are escaped in JS model
     Strings logged as HTML are escaped    log
     Strings logged as HTML are escaped    report
 
+HTML prefix in test documentation is rendered as HTML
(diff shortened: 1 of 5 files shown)


pekkaklarck:
Thanks for the PR. We are planning to add Markdown support to suite and test documentation in RF 7.5 (#5602) and adding HTML support at the same time would be nice. There's a problem that the design how to configure the documentation format discussed in #5602 is different to what was discussed in #2549. I believe the #5602 design is better and it should also be used with HTML.

The above basically means that this PR needs to be mostly rewritten. A good thing is that once #5602 is done, HTML support can easily utilize the same code and adding HTML support ought to be pretty simple. Anyway, it's best to wait a bit with this. I'll comment #2549 a bit more.

msamo-n:
Good, I'll close this one then.

PR #5631, opened 2026-03-19
fix: handle bool to int/float type conversion in user keywords

### LINKED ISSUE #5630: User keyword type conversion between int/float and bool not working correctly
## Bug Description
Type conversion for user keyword arguments does not handle bool to int/float correctly. For example, `Log As Int ${FALSE}` yields `False` (bool) instead of `0` (int).

## Expected Behavior
When a keyword parameter is typed as `int`, passing `${FALSE}` should convert to `0`, not keep the bool value.

Related to #5622.

themavik:
## Summary
Fixes #5630.
**Root cause:** `IntegerConverter.no_conversion_needed` returned `True` for booleans because `bool` is a subclass of `int`, so `isinstance(True, int)` passes. The boolean value was kept as-is instead of converting to 0/1.
**Fix:** Override `no_conversion_needed` in `IntegerConverter` to return `False` for booleans, and handle bool→int in `_non_string_convert`.
## Changes
- `src/robot/running/arguments/typeconverters.py`: IntegerConverter now converts bool to int (0/1); added bool to value_types
- `atest/testdata/keywords/type_conversion/annotations.robot`: added bool→int and bool→float test cases
- `atest/testdata/keywords/type_conversion/keyword_decorator.robot`: added bool→int test cases
- `doc/userguide/src/ExtendingRobotFramework/CreatingTestLibraries.rst`: updated supported types table — added bool_ as input type for integer, documented RF 7.5 behavior
## Testing
- Added `${TRUE}` → 1 and `${FALSE}` → 0 for Integer conversion
- Added `${TRUE}` → 1.0 and `${FALSE}` → 0.0 for Float conversion

### DIFF
--- a/atest/testdata/keywords/type_conversion/annotations.robot
+++ b/atest/testdata/keywords/type_conversion/annotations.robot
@@ -36,6 +36,8 @@ Integer
     Integer              100_e_-_2                 1
     Integer              ${41}                     41
     Integer              ${-4.0}                   -4
+    Integer              ${TRUE}                   1
+    Integer              ${FALSE}                  0
 
 Integer as hex
     Integer              0x0                       0
@@ -106,6 +108,8 @@ Float
     Float                ${4}                      4.0
     Float                ${-4.1}                   -4.1
     Float                ${FRACTION 1/2}           0.5
+    Float                ${TRUE}                   1.0
+    Float                ${FALSE}                  0.0
 
 Invalid float
     [Template]           Conversion Should Fail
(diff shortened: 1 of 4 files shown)


pekkaklarck:
The change looks good but needs explicit test or tests. Should also look at what the documentation says about integer conversion and possibly adjust them to explain that Boolean values are converted to actual integers staring from RF 7.5.

themavik:
Added acceptance tests for bool→int and bool→float in annotations.robot and keyword_decorator.robot. Updated the user guide table to list bool_ as a supported input type for integer conversion and added a note about the RF 7.5 behavior.

Issue #5603, opened 2026-01-21
Support embeddding tests/tasks to Markdown files


pekkaklarck:
This would in practice mean that if you have a Markdown file `example.md` with the below content, you could execute it with `robot example.md` and the test would be run.

```
This is just some text, *not* Robot data.

``‍`robotframework
*** Test Cases ***
Example
    Log    This is Robot data!
``‍`

More free text here. There could be additional code block with more Robot data as well.
```

We support this with reStructuredText, but I'm not sure how widely used the support is. With reST this works so that our custom reST parser finds all Robot Framework code blocks and then feeds them to normal Robot parser. The same approach would likely work here as well, but I believe the Markdown parser would be *really* simple. If we limit this support only to code blocks starting with the <code>```robotframework</code> header, the parser wouldn't even need to use any Markdown module. It could just go through lines one-by-one, search for the header, and store content between these headers and closing triple backticks.

When executing a directory like `robot example`, possible `*.md` files in it shouldn't be parsed by default, but it should be possible to use `--extension` or `--parse-include` options to change that. We also should automatically parse possible `*.robot.md` files. These are the same semantics we use with reStructuredText files.

Implementing this using the existing reStructuredText parser as an example would be pretty simple, but getting everything done, tested and documented is a relatively large task. Because this doesn't need deep knowledge about Robot's internals, this would be a very good issue for someone interested to contribute. I'll assign this tentatively to RF 7.5 scope that already has other Markdown related enhancements (#5304, #5602).

Notice that we already have #2610 about this topic. That issue has also proposed other Markdown related enhancement and its description is not that clear, though, so I decided to submit a new one and close the earlier as a duplicate.

seto:
Hello, I have implemented native support for Markdown (.md, .markdown, and .robot.md) for both test suites and resources. The implementation follows the same architectural pattern as the existing RST support, including a dependency-free reader to extract fenced code blocks.

You can find the implementation and the relative acceptance/unit tests in PR #5633.

Feedback and reviews are more than welcome!

pekkaklarck:
Thanks for the PR! I commented it briefly.

pekkaklarck:
I took a bit more detailed look at the code changes in the PR and they look pretty good. I especially like how little new code is needed.

[...]

PR #5633, opened 2026-03-23
Enable embedding tests/tasks to Markdown files #5603

seto:
This pull request introduces support for parsing test data and resource files written in Markdown format (`.md`, `.markdown`, `.robot.md`) in addition to the existing formats. It adds new parser logic, updates file extension handling throughout the codebase, and provides comprehensive acceptance tests to ensure correct behavior for Markdown files.

**Markdown file format support**

* Added a new `MarkdownParser` class in `parsers.py` to handle `.md`, `.markdown`, and `.robot.md` files, utilizing the ~~`read_markdown_data` utility~~ private method `_read_markdown_data` inside `MarkdownParser` for parsing.
* Updated builder logic to recognize and use `MarkdownParser` for Markdown file extensions in both test suite and resource file parsing.
* Extended the list of supported extensions in settings and suite structure builder to include `.robot.md`.

**Acceptance and Unit tests for Markdown support**

* Added new acceptance test suites and sample files under `atest/testdata/parsing/data_formats/markdown/` to verify correct parsing, error reporting, and suite structure for Markdown files, including edge cases and directory handling.
* Added Markdown resource files and updated resource extension tests to validate that `.md` and `.markdown` resources are parsed and imported correctly. 
* Updated test cases and data format references to account for the new Markdown extensions and increased test counts accordingly. 
* Added ~~`utest/parsing/test_markdownreader.py`~~ `utest/parsing/test_markdownparser.py` to validate the state-machine logic and edge cases (empty blocks, unclosed fences, etc.).

These changes ensure that Robot Framework can now seamlessly parse and execute tests and resources written in Markdown, with robust test coverage for the new functionality.


pekkaklarck:
Thanks for the PR! I don't have time for a proper review now, but based on a quick look at the implementation I have one request. Could you move the code related to reading Markdown data from `robot.utils` to `robot.running.builder`? The code is pretty simple, so it could possibly be made part of the new `MarkdownParser` class, but having that part in a separate module is fine as well.

I know there's `read_rest_data` under `robot.utils` and that I had instructed in the original issue to use the reStructuredText parser as an example. I had forgotten part of that implementation is under `robot.utils` and, more importantly, I don't see any good reasons why it's there. I guess part of the reason is to have conditional `docutils` imports in a separate module, but even then I see no reason to have that functionality under `robot.utils` when it's only used exactly once in parsing. Either there has been some good reason for this earlier or we just made a bad decision earlier. Anyway, let's not add the Markdown reading code under `robot.utils`. I'll separately consider deprecating `robot.utils.read_rest_data` and moving the related code elsewhere.

seto:
Hello Pekka, thanks for the quick look!
Ok, makes sense. I'll move the reading logic into `MarkdownParser` class and clean up `robot.utils` right away.
I'll ping you once the PR is updated.

seto:
Refactoring complete!

Moved the Markdown reading logic into `MarkdownParser` and cleaned up `robot.utils` as discussed. Unit tests have been updated and moved to reflect the new architecture.

All utest and atest are passing locally. Let me know if it looks good to you!

src/robot/running/builder/parsers.py
```diff
@@ -194,3 +194,35 @@ def _parse(self, method, source, defaults, init=False) -> TestSuite:
...
+            return self._read_markdown_data(reader)
+
+    def _read_markdown_data(self, mdfile: FileReader) -> str:
+        in_block = False
+        pending_separator = False  # Flag to add an empty line between multiple blocks.
+        indent = 0
+        lines = []
+
+        for line in mdfile.readlines():
+            line = line.rstrip("\n")
```
pekkaklarck: This seems to be unnecessary because all validations use `line.strip()` that anyway strips all whitespace. In the end lines should be joined together with `"".join(lines)`, though.

src/robot/running/builder/parsers.py
```diff
@@ -194,3 +194,35 @@ def _parse(self, method, source, defaults, init=False) -> TestSuite:
...
+            if not in_block:
+                if line.strip() in ("```robotframework", "```robot"):
```
pekkaklarck: This may be too strict. See, for example, what the CommonMark spec says about fenced code blocks:
https://spec.commonmark.org/0.31.2/#fenced-code-blocks

I'll comment the original issue about this separately to avoid having functionality discussion in PR review comments.

src/robot/running/builder/parsers.py
```diff
@@ -194,3 +194,35 @@ def _parse(self, method, source, defaults, init=False) -> TestSuite:
...
+                    in_block = True
+                    indent = line[: len(line) - len(line.lstrip(" \t"))]
+            else:
+                if line.strip() == "```":
+                    in_block = False
+                    pending_separator = True
```
pekkaklarck: Are you sure this `pending_separator` flag is needed? First of all, is an empty line needed between separate sections? If not, we don't need to add it and this flag is unnecessary. If an empty line is needed or convenient, it could be added here instead of raising the flag and doing it later. It would mean that there would be an empty line after the last section as well, but I doubt it matters and it could also be stripped at the end.

src/robot/running/builder/parsers.py
```diff
@@ -194,3 +194,35 @@ def _parse(self, method, source, defaults, init=False) -> TestSuite:
...
+        pending_separator = False  # Flag to add an empty line between multiple blocks.
+        indent = 0
+        lines = []
+
+        for line in mdfile.readlines():
+            line = line.rstrip("\n")
+            if not in_block:
+                if line.strip() in ("```robotframework", "```robot"):
+                    in_block = True
+                    indent = line[: len(line) - len(line.lstrip(" \t"))]
```
pekkaklarck: Use `line.lstrip()` to remove all whitespace, including non-breaking spaces and other Unicode spaces.

An alternative would be ignoring indent initially and passing lines of each block to [textwrap.dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent).

pekkaklarck (changes requested):
I only reviewed the code changes now. They look good in general and most of my comments were somewhat nitpickey. We do need to think about detecting starting code blocks, though, but I'll comment the original issue about that.

src/robot/running/builder/parsers.py
seto: I added `pending_separator` mainly for visual clarity in the extracted output, but I agree it's probably “overengineered”. I'll simplify.

pekkaklarck: The overall algorithm could actually be something like this:
```python
blocks = []
block = None
for line in mdfile.readlines():
    if block is None:
        if some-way-to-fgure-block-starts:
            block = []
    elif some-way-to-figure-block-ends:
        blocks.append(textwrap.dedent("".join(block)))
        block = None
    else:
        block.append(line)
return "\n".join(blocks)
```

seto:
I've updated the PR with the following changes:

- Regex-based detection: Refactored the logic to support backticks, tildes, variable fence lengths, and info strings.
- Unclosed fences: The parser will consume everything until EOF if the closing fence is missing, assuming we will strictly follow the CommonMark spec.
- Indentation: Switched to textwrap.dedent for cleaner and more robust block extraction.
- Refactoring: Removed the pending_separator and the unnecessary rstrip/strip calls during extraction to keep the logic leaner.
- Rebase: Performed a rebase to ensure the PR is clean and up-to-date with the latest master.

All local unit tests and acceptance tests are passing on my end.

pekkaklarck:
I took a quick look at the code and it looks great! There are some nitpickey things that could be changed, but I believe it's best I merge this PR and do such changes myself. I also need to take a more closer look at the test.

The main thing missing seems to be documentation. It can be added after this PR is merged, but certainly needs to be done before closing #5603. Do you @seto have interest to look at the docs as well?

utest/running/test_markdownparser.py
```diff
@@ -0,0 +1,205 @@
...
+        assert_equal(_read("# Title\n\nSome text.\n"), "")
+
+    def test_single_robotframework_block(self):
+        # ```robotframework
+        # *** Test Cases ***
+        # My Test    Log    Hello, RF!
+        # ```
+        md = (
+            "```robotframework\n*** Test Cases ***\nMy Test    Log    Hello, RF!\n```\n"
+        )
```
pekkaklarck, line 39: Formatting input strings using triple-quoted strings like
```
md = """
```robotframework
*** Test Cases ***
My Test    Log    Hello, RF!
```
"""
would make tests easier to read and avoid the need to have the comment with same data.

pekkaklarck (approved):

pekkaklarck:
Tests look great as well! It seems you used acceptance tests for other formats as a base which is great for consistency and maintainability. The only thing that I noticed that could/should be changed is how input data is formatted with unit tests. I can do that after I've merged this PR along with small code changes.

PR #5642, opened 2026-03-30
Fix type conversion between int/float and bool in user keywords

### LINKED ISSUE #5622: User keyword type conversion between int/float and bool not working
### Issue
When defining a user keyword with an argument that is typehinted as int, I expect that values are either converted to the type int or rejected. This works with e.g. strings or floats, but bool values appear to be accepted without conversion.
The same thing seems to happen in the reverse case, when passing int/float into a user keyword argument typehinted as bool.

### Example

```
*** Test Cases ***
Example Test
    Log As Int    ${0}
    Log As Int    0
    Log As Int    ${FALSE}
    Log As Int    ${TRUE}
    Log As Int    ${3.0}

    Log To Console    ---------

    Log As Bool    ${0}
    Log As Bool    0
    Log As Bool    ${FALSE}
    Log As Bool    ${TRUE}
    Log As Bool    ${3.0}


*** Keywords ***
Log As Int
    [Arguments]    ${value: int}
    Log To Console    ${value} ${{type(${value})}}

Log As Bool
    [Arguments]    ${value: bool}
    Log To Console    ${value} ${{type(${value})}}
```
Relevant console output:
```
0 <class 'int'>
0 <class 'int'>
False <class 'bool'>
True <class 'bool'>
3 <class 'int'>
---------
0 <class 'int'>
False <class 'bool'>
False <class 'bool'>
True <class 'bool'>
3.0 <class 'float'>
```

### Versions
Python 3.13.0
Robot Framework 7.4.1

chelslava:
## Summary

Fixes #5622

In Python, `bool` is a subclass of `int`, which caused incorrect type conversion behavior in user keyword arguments:

- `bool` values (`True`/`False`) passed to `int` or `float` arguments were not converted
- `int`/`float` values passed to `bool` arguments were not converted

### Root Cause

The `no_conversion_needed` method in `TypeConverter` checks `isinstance(value, type_info.type)`. Since `bool` is a subclass of `int`:
- `isinstance(True, int)` returns `True`
- So `no_conversion_needed(True)` incorrectly returned `True` for `int` arguments

### Changes

1. **BooleanConverter._non_string_convert**: Now converts `int`/`float` to `bool` using `bool(value)`

2. **IntegerConverter**:
   - Added `bool` to `value_types`
   - Override `no_conversion_needed` to return `False` for `bool` values
   - Added `bool` handling in `_non_string_convert`

3. **FloatConverter**:
   - Added `bool` to `value_types`
   - Override `no_conversion_needed` to return `False` for `bool` values
   - Added `bool` handling in `_convert`

### Before

```robotframework
Log As Int    ${TRUE}     # → True <class 'bool'>  (wrong!)
Log As Bool   ${3.0}      # → 3.0 <class 'float'>  (wrong!)
```

### After

```robotframework
Log As Int    ${TRUE}     # → 1 <class 'int'>      (correct!)
Log As Bool   ${3.0}      # → True <class 'bool'>  (correct!)
```

### DIFF
--- a/src/robot/running/arguments/typeconverters.py
+++ b/src/robot/running/arguments/typeconverters.py
@@ -27,7 +27,13 @@ from typing import Any, Literal, TYPE_CHECKING, Union
 from robot.conf import Languages
 from robot.libraries.DateTime import convert_date, convert_time
 from robot.utils import (
-    eq, get_error_message, plural_or_not as s, safe_str, Secret, seq2str, type_name
+    eq,
+    get_error_message,
+    plural_or_not as s,
+    safe_str,
+    Secret,
+    seq2str,
+    type_name,
 )
 
 if TYPE_CHECKING:
@@ -304,7 +310,7 @@ class BooleanConverter(TypeConverter):
     value_types = (str, int, float, NoneType)
 
     def _non_string_convert(self, value):
-        return value
+        return bool(value)
 
     def _string_convert(self, value):
         normalized = value.title()
@@ -322,9 +328,16 @@ class IntegerConverter(TypeConverter):
     type = int
     abc = Integral
     type_name = "integer"
-    value_types = (str, float)
+    value_types = (str, float, bool)
+
+    def no_conversion_needed(self, value: Any) -> bool:
+        if isinstance(value, bool):
+            return False
+        return super().no_conversion_needed(value)
 
     def _non_string_convert(self, value):
+        if isinstance(value, bool):
+            return int(value)
         if value.is_integer():
             return int(value)
         raise ValueError("Conversion would lose precision.")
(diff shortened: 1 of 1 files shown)


pekkaklarck:
The fix itself seems fine, but there are various problems otherwise:
- Issue #5622 reports two problems: 1) If you have `arg: int`, Boolean values are not converted to `int`. 2) If you have `arg: bool`, integer values are not converted to `bool`. This PR only changes the former issue. That is itself fine, as changing the latter is likely to cause some backwards incompatibility issues and could only be done in a major version. This PR doesn't anyway fix #5622 fully. A proper solution is agreeing what we want to change in #5622 and then possibly submitting a separate issue about the `arg: int` issue.
- As I commented the #5622, these conversion issues aren't related only to user keyword arguments. The title and description of this PR are thus somewhat misleading.
- Code doesn't follow our coding conventions. See contribution guidelines for details.
- Tests are missing. This is a rather big omission.
- The change should also be documented in the User Guide.

If you are interested to fix these issues and to get this PR merged, let's agree how to proceed in #5622.

PR #5651, opened 2026-04-07
TypeInfoParser crashes with UnboundLocalError or hangs on | inside type parameters

### LINKED ISSUE #5650: Invalid `|` usage in stringified type hints can crash or hang execution
## Summary

Two related crashes in `TypeInfoParser` when a `|` token appears inside `[...]` in an unexpected position. Both should produce a clean `ValueError` or `DataError` but instead crash or hang.

## Version

Robot Framework 7.4.2,
Python 3.12

## Case 1:  `UnboundLocalError` for `"list[|int]"`

**Python:**

```python
from robot.running.arguments.typeinfoparser import TypeInfoParser
TypeInfoParser("list[|int]").parse()
# UnboundLocalError: cannot access local variable 'type_' before assignment
```

**Robot:**

```robotframework
*** Settings ***
Library    typeinfoparser_bug_lib.py

*** Test Cases ***
TypeInfoParser raises UnboundLocalError for pipe before first type
    Run Keyword And Expect Error    ValueError: *    Parse Type String    list[|int]
```

## Case 2: Infinite loop / `MemoryError` for `"list[int,|]"`

```python
from robot.running.arguments.typeinfoparser import TypeInfoParser
TypeInfoParser("list[int,|]").parse()
# Hangs forever — eventually MemoryError
```

> **Warning:** Do not run this interactively — it hangs the process.

## Expected

A clean `ValueError` or `DataError`, e.g.:
`Invalid type string 'list[|int]': unexpected token '|'`

sudheerr937-ai:
Fixes #5650

Root cause: In `params()`, when a PIPE token appeared inside type 
parameters, none of the if/elif branches handled it causing:

- UnboundLocalError on first occurrence (param never assigned)
- Infinite loop/hang on second+ occurrence (never advances past PIPE)

Fix: Added else clause to raise a proper ValueError immediately.

Added 2 test cases to cover both failure scenarios:
- list[| str]
- list[int, | str]

### DIFF
--- a/src/robot/running/arguments/typeinfoparser.py
+++ b/src/robot/running/arguments/typeinfoparser.py
@@ -154,7 +154,8 @@ class TypeInfoParser:
                 param = TypeInfo()
                 param.nested = self.params()
             else:
-                self.error(f"Invalid type string '{self.source}': expected type name.")
+                self.error("Type name missing.")
+
             if literal:
                 param = self._literal_param(param)
             params.append(param)
@@ -220,3 +221,4 @@ class TypeInfoParser:
         raise ValueError(
             f"Parsing type {self.source!r} failed: Error at {position}: {message}"
         )
+        
--- a/src/robot/running/arguments/typeinfoparser.py
+++ b/src/robot/running/arguments/typeinfoparser.py
@@ -153,6 +153,8 @@ class TypeInfoParser:
                 self.advance()
                 param = TypeInfo()
                 param.nested = self.params()
+            else:
+                self.error("Type name missing.")
             if literal:
                 param = self._literal_param(param)
             params.append(param)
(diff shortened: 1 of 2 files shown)


src/robot/running/arguments/typeinfoparser.py
```diff
@@ -153,6 +153,8 @@ def params(self, literal: bool = False) -> "list[TypeInfo]":
                 self.advance()
                 param = TypeInfo()
                 param.nested = self.params()
+            else:
+                self.error("Type name missing.")
```
007bsd, line 157: I would suggest using a more generic error message:

```python
self.error(f"Invalid type string '{self.input}': expected type name.")
```

This keeps the message accurate for all unexpected tokens, not just `|`.

PR #5652, opened 2026-04-08
Update ControlStructures.rst

am9417:
Minor documentation change to avoid confusion between WHILE and "while"


pekkaklarck:
Yeah, starting a sentence discussing about WHILE loops with "While infinite loops..." is confusing and "Although infinite loops..." works much better. Thanks!

PR #5654, opened 2026-04-19
Fix get_index_from_list returning -1 for negative start when value is found

### LINKED ISSUE #5649: `get_index_from_list` incorrectly returns `-1` when value is found using negative `start`
## Summary

`get_index_from_list` uses `-1` to signal "not found". But when `start` is negative and the value is at the last position of the slice, the arithmetic `start + list_.index(value)` also produces `-1`, creating a false "not found" result for a value that IS in the list.

## Version

Robot Framework 7.4.2,
Python 3.12

## Reproduce

**Python:**

```python
from robot.libraries.Collections import Collections
c = Collections()
print(c.get_index_from_list([0], 0, start=-1))      # -1
print(c.get_index_from_list([1,2,3], 3, start=-1))  # -1
print(c.get_index_from_list([1,2,3], 3, start=-2))  # -1
```

**Robot:**

```robotframework
*** Settings ***
Library    Collections

*** Test Cases ***
Get Index From List with negative start returns wrong sentinel
    ${list}=    Create List    0
    ${result}=    Get Index From List    ${list}    0    start=-1
    Should Not Be Equal As Integers    ${result}    -1    msg=Value 0 IS in list but got -1 (not found)

Get Index From List negative start last element collision
    ${list}=    Create List    1    2    3
    ${result}=    Get Index From List    ${list}    3    start=-1
    Should Not Be Equal As Integers    ${result}    -1    msg=Value 3 IS in list but got -1 (not found)

Get Index From List negative start second-to-last collision
    ${list}=    Create List    1    2    3
    ${result}=    Get Index From List    ${list}    3    start=-2
    Should Not Be Equal As Integers    ${result}    -1    msg=Value 3 IS in list but got -1 (not found)
```

## Root Cause

`src/robot/libraries/Collections.py` line 317:

```python
return start + list_.index(value)
```

When `start=-1` and the value is the last element, `list_[start:].index(value) == 0`, so `-1 + 0 = -1`, which collides with the not-found sentinel.

## Expected

The actual index of the value in the original list (e.g. `0` for `[0]` with `start=-1`).

## Actual

```python
>>> c.get_index_from_list([0], 0, start=-1)
-1
>>> c.get_index_from_list([1, 2, 3], 3, start=-1)
-1
>>> c.get_index_from_list([1, 2, 3], 3, start=-2)
-1
>>> c.get_index_from_list([0, 1, 2, 3, 4], 4, start=-1)
-1
```

sudheerr937-ai:
get_index_from_list was returning -1 (not found) even when the value existed in the list, but only when a negative start like -1 or -2 was used.
Root cause: the math start + list_.index(value) accidentally produces -1 when start is negative and the value is at the end of the slice.
Fix: convert negative start to its real position in the list before doing the math
Added test cases in atest/testdata/standard_libraries/collections/list.robot covering negative start scenarios.
Fixes #5649

### DIFF
--- a/atest/testdata/standard_libraries/collections/list.robot
+++ b/atest/testdata/standard_libraries/collections/list.robot
@@ -120,11 +120,7 @@ Get Index From List
     Get Index From List    ${LONG}    43    4    7    expected=5     type=int
     Get Index From List    ${LONG}    43   end=8      expected=5     type=int
     Get Index From List    ${LONG}    nonex           expected=-1    type=int
-    Get Index From List    ${LONG}    ${2}     start=-1    expected=8     type=int
-    Get Index From List    ${LONG}    43       start=-4    expected=5     type=int
-    Get Index From List    ${LONG}    nonex    start=-3    expected=-1    type=int
-
-
+   
 Get Index From List with empty string as start index is deprecated
     [Template]    Verify Result
     Get Index From List    ${LONG}    43    ${EMPTY}    8    expected=5     type=int
@@ -134,6 +130,12 @@ Get Index From List with invalid index
     Get Index From List    ${LONG}    2    invalid    1     expected=${START ERROR2}
     Get Index From List    ${LONG}    2    1    invalid     expected=${END ERROR}
 
+Get Index From List with negative indices
+    [Template]    Verify Result
+    Get Index From List    ${LONG}    ${2}     start=-1    expected=8     type=int
+    Get Index From List    ${LONG}    43       start=-4    expected=5     type=int
+    Get Index From List    ${LONG}    nonex    start=-3    expected=-1    type=int
+    
 Copy List
     ${copy} =    Copy List    ${L2}
     Append To List    ${L2}      add to original
--- a/atest/testdata/standard_libraries/collections/list.robot
+++ b/atest/testdata/standard_libraries/collections/list.robot
@@ -120,6 +120,10 @@ Get Index From List
     Get Index From List    ${LONG}    43    4    7    expected=5     type=int
     Get Index From List    ${LONG}    43   end=8      expected=5     type=int
     Get Index From List    ${LONG}    nonex           expected=-1    type=int
+    Get Index From List    ${LONG}    ${2}     start=-1    expected=8     type=int
+    Get Index From List    ${LONG}    43       start=-4    expected=5     type=int
+    Get Index From List    ${LONG}    nonex    start=-3    expected=-1    type=int
+
 
 Get Index From List with empty string as start index is deprecated
     [Template]    Verify Result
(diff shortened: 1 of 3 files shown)


atest/testdata/standard_libraries/collections/list.robot
```diff
@@ -120,6 +120,10 @@ Get Index From List
     Get Index From List    ${LONG}    43    4    7    expected=5     type=int
     Get Index From List    ${LONG}    43   end=8      expected=5     type=int
     Get Index From List    ${LONG}    nonex           expected=-1    type=int
+    Get Index From List    ${LONG}    ${2}     start=-1    expected=8     type=int
+    Get Index From List    ${LONG}    43       start=-4    expected=5     type=int
+    Get Index From List    ${LONG}    nonex    start=-3    expected=-1    type=int
+
```
pekkaklarck, line 126: Adding new validation steps to this existing test is fine, but you could have also created a new test `Get Index From List with negative indices` or something like that. It would then need to be added to the matching file on the `atest/robot` side as well.

Please remove the unnecessary empty line.

src/robot/libraries/Collections.py
```diff
@@ -312,6 +312,8 @@ def get_index_from_list(
                 "keyword is deprecated. Use '0' instead."
             )
             start = 0
+        if isinstance(start, int) and start < 0:
+            start = max(len(list_) + start, 0)
```
pekkaklarck, line 316: 1. `start` is known to be `int` at this point so the `isinstance` check is unnecessary.
2. Using `max(len(list_) + start), 0)` looks a bit odd to me. Wouldn't `len(list_) + start)` be enough? I can see that if you use a negative index that's out-of-the-range (e.g. list has length 5 and index is -7) the latter can map the index to a value that is incorrect but valid (e.g. -2 with the earlier values), but using `max(..., 0)` just changes that to an other incorrect value.
3. I believe it would be better to map the `start` index to a positive value only *after* doing `self.get_slice_from_list(...)`. In that case possible invalid indices (e.g. -7 when list has length 5) cause an error already then and afterwards we can safely do `start = len(list_) + start` with negative indices.

src/robot/libraries/Collections.py
```diff
@@ -1328,3 +1330,4 @@ def report_error(default: str, message: "str | None", values: bool = False) -> N
     elif values:
         message += "\n" + default
     raise AssertionError(message)
+
```
pekkaklarck, line 1333: Please don't remove trailing newlines. As explained in CONTRIBUTING.rst, our coding style mandates that files should end with a newline.

pekkaklarck (changes requested):
The overall approach how to fix the issue looks good, but there are few things to be fixed.

pekkaklarck:
Should we add a note to keyword documentation about negative start indices yielding invalid results prior to RF 7.5?

sudheerr937-ai:
Hi @pekkaklarck ,

Thanks for the review! I've addressed all the feedback:

Removed isinstance check and max(), moved negative start conversion to after get_slice_from_list
Added a note to the docstring about RF 7.5
Moved the negative index test cases to a separate test Get Index From List with negative indices and added it to the atest/robot side as well
Removed the unnecessary empty line

sudheerr937-ai:
> Should we add a note to keyword documentation about negative start indices yielding invalid results prior to RF 7.5?

Yes, added a note to the docstring about this. It now mentions that prior to RF 7.5, using a negative start index could return -1 even when the value was found in the list.

Issue #5658, opened 2026-04-23
Type hints that use non-existing values or are invalid crash execution with Python 3.14


klaoude:
### Version information
- Robot Framework: 7.5.dev1
- Python: 3.14.4
- OS: Linux

`robot --version` output:
```text
Robot Framework 7.5.dev1 (Python 3.14.4 on linux)
```

### Steps to reproduce
Use this minimal example.
`issue.robot`:
```robot
*** Settings ***
Library  lib.py  AS  Lib

*** Test Cases ***
Test
    ${list}=  Create List  ${1}  ${2}
    ${result}=  Lib.my_sum    ${list}
    Should Be Equal  ${result}  ${3}
```
`lib.py`:
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing import Sequence


def my_sum(seq: Sequence[int]) -> int:
    return sum(seq)
```
Run:
```bash
PYTHONPATH="src" python -m src.robot issue.robot
```
### Expected result
Robot imports the library and executes the test normally.

### Actual result
Library import fails with:
```text
NameError: name 'Sequence' is not defined
```
```text
[ ERROR ] Unexpected error: NameError: name 'Sequence' is not defined
Traceback (most recent call last):
  File "./src/robot/utils/application.py", line 97, in _execute
    rc = self.main(arguments, **options)
  File "./src/robot/run.py", line 489, in main
    result = suite.run(settings)
  File "./src/robot/running/model.py", line 918, in run
    self.visit(runner)
    ~~~~~~~~~~^^^^^^^^
  File "./src/robot/model/testsuite.py", line 465, in visit
    visitor.visit_suite(self)
    ~~~~~~~~~~~~~~~~~~~^^^^^^
  File "./src/robot/model/visitor.py", line 129, in visit_suite
    if self.start_suite(suite) is not False:
       ~~~~~~~~~~~~~~~~^^^^^^^
  File "./src/robot/running/suiterunner.py", line 90, in start_suite
    ns.handle_imports()
    ~~~~~~~~~~~~~~~~~^^
  File "./src/robot/running/namespace.py", line 61, in handle_imports
    self._handle_imports(self._imports)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "./src/robot/running/namespace.py", line 72, in _handle_imports
    self._import(item)
    ~~~~~~~~~~~~^^^^^^
[...]

PR #5665, opened 2026-04-30
Python 3.14: avoid library import crash with deferred annotations

klaoude:
Issue: https://github.com/robotframework/robotframework/issues/5658

This PR fixes a Python 3.14 compatibility issue where library import can fail when Robot inspects keyword annotations.

On Python 3.14, `inspect.signature()` may evaluate annotations while creating keywords. That causes problems with deferred annotations such as names imported only under `TYPE_CHECKING`, and more generally with annotations that cannot be resolved at introspection time.

The fix uses `annotationlib.Format.FORWARDREF` with `inspect.signature()` on Python 3.14 to avoid the import-time crash.

For type hints, the fallback logic needs to preserve existing behavior with annotations like `Union["nonex", "references"]`. Because of that, `_get_types()` now first tries normal `get_type_hints()`, then falls back to raw `__annotations__`, and only if accessing `__annotations__` itself fails does it use `get_type_hints(..., format=Format.STRING)`.

The PR also adds acceptance tests for deferred annotations covering:
- `TYPE_CHECKING`-only generic annotations
- non-existing annotations

### DIFF
--- a/atest/robot/keywords/type_conversion/deferred_annotations.robot
+++ b/atest/robot/keywords/type_conversion/deferred_annotations.robot
@@ -1,0 +1,14 @@
+*** Settings ***
+Suite Setup      Run Tests    ${EMPTY}    keywords/type_conversion/deferred_annotations.robot
+Force Tags       require-py3.14
+Resource         atest_resource.robot
+
+*** Test Cases ***
+Type checking annotation
+    Check Test Case    ${TESTNAME}
+
+Type checking annotation with mixed types
+    Check Test Case    ${TESTNAME}
+
+Nonexisting annotation
+    Check Test Case    ${TESTNAME}
--- a/atest/robot/keywords/type_conversion/deferred_annotations.robot
+++ b/atest/robot/keywords/type_conversion/deferred_annotations.robot
@@ -1,14 +1,18 @@
 *** Settings ***
 Suite Setup      Run Tests    ${EMPTY}    keywords/type_conversion/deferred_annotations.robot
-Force Tags       require-py3.14
+Test Tags        require-py3.14
 Resource         atest_resource.robot
 
 *** Test Cases ***
+Deferred evaluation of annotations
+    [Documentation]    https://peps.python.org/pep-0649
+    Check Test Case    ${TESTNAME}
+
 Type checking annotation
     Check Test Case    ${TESTNAME}
 
-Type checking annotation with mixed types
+Nonexisting annotation
     Check Test Case    ${TESTNAME}
 
-Nonexisting annotation
+Type checking annotation with parameterized generic
     Check Test Case    ${TESTNAME}
--- a/atest/testdata/keywords/type_conversion/DeferredAnnotations.py
+++ b/atest/testdata/keywords/type_conversion/DeferredAnnotations.py
@@ -1,14 +1,23 @@
+from typing import TYPE_CHECKING
+
 from robot.api.deco import library
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
 
 
 class Library:
-
     def deferred_evaluation_of_annotations(self, arg: Argument) -> str:  # noqa: F821
         return arg.value
 
+    def type_checking_annotation(self, seq: Sequence[int]) -> int:  # noqa: F821
+        return sum(seq)
+
+    def nonexisting_annotation(self, arg: NonExisting):  # noqa: F821
+        return arg
+
 
 class Argument:
-
     def __init__(self, value: str):
         self.value = value
 
--- a/atest/testdata/keywords/type_conversion/DeferredAnnotations.py
+++ b/atest/testdata/keywords/type_conversion/DeferredAnnotations.py
@@ -3,21 +3,26 @@ from typing import TYPE_CHECKING
 from robot.api.deco import library
 
 if TYPE_CHECKING:
-    from collections.abc import Sequence
+    from typing import Sequence
+    class TypeCheckingOnly: ...
 
 
 class Library:
+
     def deferred_evaluation_of_annotations(self, arg: Argument) -> str:  # noqa: F821
         return arg.value
 
-    def type_checking_annotation(self, seq: Sequence[int]) -> int:  # noqa: F821
-        return sum(seq)
+    def type_checking_annotation(self, arg: TypeCheckingOnly) -> TypeCheckingOnly:  # noqa: F821
+        return arg
 
     def nonexisting_annotation(self, arg: NonExisting):  # noqa: F821
         return arg
 
+    def my_sum(self, seq: Sequence[int]) -> int:  # noqa: F821
+        return sum(seq)
 
 class Argument:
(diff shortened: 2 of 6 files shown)


atest/testdata/keywords/type_conversion/DeferredAnnotations.py
```diff
@@ -1,14 +1,23 @@
...
+
 from robot.api.deco import library
 
+if TYPE_CHECKING:
+    from collections.abc import Sequence
 
-class Library:
 
+class Library:
     def deferred_evaluation_of_annotations(self, arg: Argument) -> str:  # noqa: F821
```
pekkaklarck, line 12: Our coding style, see the Contribution guidelines, mandate having an empty row before methods. Please add the removed line back here and in the `Argument` class.

atest/testdata/keywords/type_conversion/DeferredAnnotations.py
```diff
@@ -1,14 +1,23 @@
...
         return arg.value
 
+    def type_checking_annotation(self, seq: Sequence[int]) -> int:  # noqa: F821
+        return sum(seq)
+
+    def nonexisting_annotation(self, arg: NonExisting):  # noqa: F821
+        return arg
```
pekkaklarck, line 19: For Robot, and any other code doing runtime type inspection, there's no difference with `Sequence` not being available due to a conditional `if TYPE_CHECKING` import and `NonExisting` simply not being defined at all. The difference between these two cases is that `Sequence` is recognized by Robot's stringly-typed-annotations parser and thus `Sequence[int]` is handled as expected and not simply ignored. It's thus good to separately test both of these cases, but the method names don't fully describe the actual thing that's tested. That probably doesn't matter and I don't immediately come up with better names, so leaving these as they are is fine.

atest/robot/keywords/type_conversion/deferred_annotations.robot
```diff
@@ -0,0 +1,14 @@
+*** Settings ***
+Suite Setup      Run Tests    ${EMPTY}    keywords/type_conversion/deferred_annotations.robot
+Force Tags       require-py3.14
```
pekkaklarck: `Force Tags` is deprecated. Use `Test Tags` instead.

atest/robot/keywords/type_conversion/deferred_annotations.robot
```diff
@@ -0,0 +1,14 @@
+*** Settings ***
```
pekkaklarck, line 1: Adding a separate suite for deferred annotations is a good idea. Please move the existing test for normal usage added in cca331cccbe644aabba538df2f5f45d9c2948d65 here as well (and make same changes to `robot/testdata/...` files).

atest/testdata/keywords/type_conversion/deferred_annotations.robot
```diff
@@ -0,0 +1,20 @@
+*** Settings ***
+Library    DeferredAnnotations.py
+
+*** Test Cases ***
+Type checking annotation
+    [Tags]    require-py3.14
```
pekkaklarck: Tags could be specified in settings using `Test Tags`, but tags aren't typically needed here on the `testdata` side so they probably can simply be removed.

src/robot/running/arguments/argumentparser.py
```diff
@@ -116,9 +122,18 @@ def _get_types(self, method):
         try:
             return get_type_hints(method)
         except Exception:  # Can raise pretty much anything
-            # Not all functions have `__annotations__`.
-            # https://github.com/robotframework/robotframework/issues/4059
-            return getattr(method, "__annotations__", {})
+            # Prefer raw annotations to preserve behavior with existing forward
+            # references like Union["nonex", ...] used in current tests.
```
pekkaklarck: Good catch that accessing `__annotations__` directly handles the `Union["nonex", ...]` case. I was wondering why those tests failed when I was prototyping the fix.

We don't generally use comments for explaining what the code does and instead try to write self-explaining code and cover all corner cases with tests. The reason I added the `# Not all functions ...` comment below was that I wasn't able to create a failing test and I wanted to avoid someone (e.g. me) refactoring `getattr(method. "__annotations_", {})` to simply `method.__annotations__`. I have since noticed that `partial` objects don't have `__annotations__`, so keyword implemented using them could be used in tests, so that comments isn't actually needed if such a test is added. Anyway, you can leave these comments for now and I may look at them again later.

src/robot/running/arguments/argumentparser.py
```diff
@@ -116,9 +122,18 @@ def _get_types(self, method):
...
+            try:
+                # Not all functions have `__annotations__`.
+                # https://github.com/robotframework/robotframework/issues/4059
+                return getattr(method, "__annotations__", {})
+            except Exception:
+                # With deferred annotations on Python 3.14, evaluating
+                # `__annotations__` itself can fail (e.g. TYPE_CHECKING-only imports).
+                if Format:
+                    return get_type_hints(method, format=Format.STRING)
+                return {}
```
pekkaklarck, line 134: Is there any possibility that `getatt(method, "__annotations__", {})` can fail with Python < 3.14? If not, the final exception handling block could be simplified to `return get_type_hints(method, format=Format.STRING)`. Linters might not like `Format.STRING` being used without `if Format`, though.

pekkaklarck (changes requested):
Looks very good in general. See my comments for things that could be enhanced still. Most of them are not somewhat nitpickey, but at least some of the test related things that should be fixed.

atest/testdata/keywords/type_conversion/DeferredAnnotations.py
klaoude: I changed `Sequence[int]` to a dummy `TypeCheckingOnly` class imported only under `TYPE_CHECKING`. This triggers the same runtime introspection failure, but avoids interaction with Robot’s string-annotation parser, so the tests stay coherent and focus on the intended behavior.

src/robot/running/arguments/argumentparser.py
klaoude: I've removed this comment and I've added a link to the issue for the second one.

I may take a look for `partial` objects test as well as for `a: 1/0` and `a: list[int` errors.

src/robot/running/arguments/argumentparser.py
klaoude: I think it can only raise for python >= 3.14.
Linter is indeed not happy without the `if Format`, I've keep it as it is.

atest/testdata/keywords/type_conversion/DeferredAnnotations.py
klaoude: And I've rename the sum test with mixed parameters `Type checking annotation with parameterized generic`

pekkaklarck: Looks good to me. 👍

Issue #5618, opened 2026-02-24
Option to suppress result file paths in console output when using programmatic API


oboehmer:
## Description

When using Robot Framework programmatically (via `robot.run()`, `robot.rebot()`, `TestSuite.run()`, or through tools like pabot), the final result file listing is always printed to stdout:

```
Output:  /path/to/results/output.xml
XUnit:   /path/to/results/xunit.xml
Log:     /path/to/results/log.html
Report:  /path/to/results/report.html
```

There doesn't seem to be a way to suppress this output while still showing test execution progress (dots, verbose output, etc.).

## Use Case

When building custom test runners or orchestration tools that wrap Robot Framework:

1. The wrapper tool provides its own summary/results output
2. The default file listing becomes redundant noise that clutters the output
3. Redirecting stdout entirely via `stdout=StringIO()` is not desirable because it also hides test execution progress

Example scenario: A test orchestrator runs Robot Framework executions (or uses pabot for parallel execution) alongside other test frameworks, and displays a consolidated summary at the end. The individual file listings from each execution add visual clutter without providing value, since the orchestrator already reports where results are located at the end.

## Current Workarounds

1. **Redirect stdout entirely** (`stdout=StringIO()`) - suppresses everything including useful test progress
2. **Post-process stdout** - capture and filter output, but this is fragile and adds complexity
3. **Accept the noise** - live with duplicate/redundant output

## Proposed Solution

Add an option to suppress result file path output, either:

1. **New parameter for programmatic API**: 
   ```python
   robot.run('tests/', suppress_result_paths=True)
   robot.rebot('output.xml', suppress_result_paths=True)
   ```

2. **New console mode or option**:
   ```
   --console dotted:quiet-results
   # or
   --quiet-results
   ```

3. **Extend existing console parameter** to accept additional flags

The key requirement is that test execution progress (controlled by `--console`) should remain visible while only the final file listing is suppressed.

## Note

I plan to raise a similar feature request to pabot to expose this option once it becomes available in Robot Framework. I'm happy to provide a PR for this feature if the approach is agreed upon.

## Environment

- Robot Framework version: 7.x
- This applies to both `robot` and `rebot` when invoked programmatically

pekkaklarck:
The feature itself is fine, but I'm not sure what's the best way to implement it. Configuring this kind of details with a dedicated argument on the` robot.run` level somehow doesn't feel that good to me. The main problem I see is that there could be other such configuring needs that would then require similar arguments.

[...]

PR #5671, opened 2026-05-11
ConsoleOutput: Replace LoggerApi by ListenerFacade inheritance

oboehmer:
Refactor console output classes (VerboseOutput, DottedOutput, QuietOutput, NoOutput) so they no longer inherit from LoggerApi. Instead, Logger wraps the selected console with ListenerFacade.create(), making them usable exactly like listener-based loggers.

This is step 1 of #5618 (support custom console loggers), enabling step 2 where custom loggers can be passed to --console and handled uniformly through the same ListenerFacade wrapping.

To avoid a circular import between logger.py and listeners.py, both sides use deferred imports: logger.py imports ListenerFacade inside register_console_logger(), and listeners.py imports LOGGER at the 3 call sites that need it (all error-handling paths).

Console classes now define explicit output_file/report_file/etc. methods (previously inherited from LoggerApi's default delegation to result_file).

### DIFF
--- a/src/robot/output/console/dotted.py
+++ b/src/robot/output/console/dotted.py
@@ -65,25 +65,25 @@ class DottedOutput:
         if msg.level in ("WARN", "ERROR") and msg.console:
             self.stderr.error(msg.message, msg.level)
 
-    def output_file(self, path: "Path"):
+    def output_file(self, path: Path):
         self.result_file("Output", path)
 
-    def report_file(self, path: "Path"):
+    def report_file(self, path: Path):
         self.result_file("Report", path)
 
-    def log_file(self, path: "Path"):
+    def log_file(self, path: Path):
         self.result_file("Log", path)
 
-    def xunit_file(self, path: "Path"):
+    def xunit_file(self, path: Path):
         self.result_file("XUnit", path)
 
-    def debug_file(self, path: "Path"):
+    def debug_file(self, path: Path):
         self.result_file("Debug", path)
 
     def result_file(
         self,
         kind: "Literal['Output', 'Report', 'Log', 'XUnit', 'Debug']",
-        path: "Path",
+        path: Path,
     ):
         self.stdout.result_file(kind, path)
 
--- a/src/robot/output/console/dotted.py
+++ b/src/robot/output/console/dotted.py
@@ -14,19 +14,19 @@
 #  limitations under the License.
 
 import sys
-from typing import TYPE_CHECKING
+from pathlib import Path
+from typing import Literal, TYPE_CHECKING
 
 from robot.model import SuiteVisitor
 from robot.utils import plural_or_not as s, secs_to_timestr
 
-from ..loggerapi import LoggerApi
 from .highlighting import HighlightingStream
 
 if TYPE_CHECKING:
     from robot.result import TestCase, TestSuite
 
 
-class DottedOutput(LoggerApi):
+class DottedOutput:
 
     def __init__(self, width=78, colors="AUTO", links="AUTO", stdout=None, stderr=None):
         self.width = width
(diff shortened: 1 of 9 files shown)


src/robot/output/console/dotted.py
```diff
@@ -65,7 +65,26 @@ def message(self, msg):
         if msg.level in ("WARN", "ERROR") and msg.console:
             self.stderr.error(msg.message, msg.level)
 
-    def result_file(self, kind, path):
+    def output_file(self, path: "Path"):
```
pekkaklarck: No need for quotes around `Path`.

src/robot/output/console/dotted.py
```diff
@@ -65,7 +65,26 @@ def message(self, msg):
...
+    def log_file(self, path: "Path"):
+        self.result_file("Log", path)
+
+    def xunit_file(self, path: "Path"):
+        self.result_file("XUnit", path)
+
+    def debug_file(self, path: "Path"):
+        self.result_file("Debug", path)
+
+    def result_file(
```
pekkaklarck, line 83: I didn't remember that `LoggerApi` has `log_file`, `report_file` etc. in addition to `result_file` that the former methods call by default. That's convenient and I understand you used that approach here as well. Would it, however, make more sense to add this functionality to the listener API? It would obviously be a separate enhancement to this one, it should get its own issue and it shouldn't be done as part of this PR. Even if we agree that's a good idea, we could merge this PR and then just remove the then unnecessary `log_file`, `report_file`, etc. methods here.

src/robot/output/console/verbose.py
```diff
@@ -73,7 +77,32 @@ def message(self, msg):
...
+    def log_file(self, path: "Path"):
+        self.result_file("Log", path)
+
+    def xunit_file(self, path: "Path"):
+        self.result_file("XUnit", path)
+
+    def debug_file(self, path: "Path"):
+        self.result_file("Debug", path)
+
+    def result_file(
```
pekkaklarck, line 101: Adding `result_file` to the listener API would avoid the need to have the above `log_file`, `report_file`, etc. methods here as well.

src/robot/output/listeners.py
```diff
@@ -57,6 +56,7 @@ def _import_listeners(self, listeners, library=None) -> "list[ListenerFacade]":
             except DataError as err:
                 if library:
                     raise
+                from .logger import LOGGER
```
pekkaklarck: These imports should be moved on top of the function where they are done and there should be an empty row after then. It seems we can keep these as normal, module level imports, though.

src/robot/output/logger.py
```diff
@@ -109,9 +109,10 @@ def register_console_logger(
         stdout=None,
         stderr=None,
     ):
-        self._console = ConsoleOutput(
-            type, width, colors, links, markers, stdout, stderr
-        )
+        from .listeners import ListenerFacade
+
+        console = ConsoleOutput(type, width, colors, links, markers, stdout, stderr)
+        self._console = ListenerFacade.create(console)
```
pekkaklarck: I think it would be cleaner if the `ConsoleOutput` factory method would wrap whatever it returns with `ListenerFacade`. This code shouldn't care how `ConsoleOutput` works.

`ConsoleOutput` being a factory method, not a normal class, is a bit ugly. We used that kind of pattern quite a lot earlier, but have since stopped that and many factory methods have been defactored to classes. In this case `ConsoleOutput` could itself inherit `LogerApi` and then delegate method calls to the underlying output objects, but doing it as part of this issue makes no sense. It could be that a factory method is even simpler in this particular case.

pekkaklarck (changes requested):
Looks good, but some changes needed. Next steps:
- Merge my changes from `master`. They should avoid cyclic import issues with `LOGGER`.
- See my review comments.
- Let's discuss adding `result_file` to the listener API separately.

src/robot/output/listeners.py
oboehmer: done, thanks for the upstream enhancement

src/robot/output/logger.py
oboehmer: good idea, done..

pekkaklarck (approved):

PR #5672, opened 2026-05-11
Add support for custom console loggers

oboehmer:
## Summary

- Add support for custom console loggers specified as `--console path/to/MyConsole.py` or programmatically via `run(..., console=MyConsole())`
- Custom consoles can be a class (by path, module name, or dotted name), a module with bare functions, or a pre-instantiated object — using the same import mechanism as listeners
- Error messages use `"console"` context (e.g., "Taking console 'X' into use failed") instead of `"listener"`

## Changes

**Production code:**
- `ConsoleOutput()` now delegates unrecognized types to `ListenerFacade.create(type, kind="console")` instead of raising `DataError`. Extracted `_create_builtin_output()` helper for readability.
- `ListenerFacade.create()` / `_import_listener()` accept a `kind` parameter for contextual error messages and importer naming.
- `run.py`: Moved `--pythonpath` setup before `register_console_logger()` so module-name console imports work.

**Test fixtures** (`atest/testresources/consoles/`):
- `CustomConsole.py` — class-based console with configurable marker and result file methods
- `module_console.py` — module-level functions, no result file methods (verifies V3 facade no-op behavior)
- `console_classes.py` — module with differently-named class for dotted import testing

**Tests:**
- 8 acceptance tests in `console_type.robot` (path, path+args, module name, module functions, dotted name, named args, wrong args, non-existing)
- 4 unit tests in `test_logger.py` (object, path, path+args, bad import)
- 2 API tests in `test_run_and_rebot.py` (programmatic `run(console=obj)` with lifecycle and result file assertions)

Implements step 2 of #5618.


src/robot/output/console/__init__.py
```diff
@@ -31,18 +29,23 @@ def ConsoleOutput(
 ):
     from ..listeners import ListenerFacade
 
+    output = _create_builtin_output(type, width, colors, links, markers, stdout, stderr)
+    if output is not None:
+        return ListenerFacade.create(output, kind="console")
+    # Custom console: type is a path/module name (str) or an object instance.
+    return ListenerFacade.create(type, kind="console")
```
pekkaklarck: Having a separate helper function that returns a built-in logger or `None` is unnecessary complicated in my opinion. As the code above shows, `ListenerFacade.create()` is called in both cases, so there's no real difference between a built-in or a custom logger anymore. I believe it would be simpler to have code like this:

```python
if isinstance(logger, str):
    upper = logger.upper()
    if upper == "VERBOSE":
        logger = VerboseOutput(...)
    elif upper == "DOTTED":
        logger = DottedOutput(...)
    ...
return ListenerFacade.create(logger, kind="console")
```

In the above I've renamed the `type` argument to `logger`. I'm not sure is it the best possible name, but it's certainly better than `type` now that the code accepts also custom loggers even as objects.

I don't particularly like nested `if` structures like above and a helper function could avoid that. Having the whole logic visible in one place has its benefits, though, especially when the logic isn't overly complicated. Once we have Python 3.10 as our minimum version, we can replace the inner `if` structure with `match` statement and make this even simpler.

Adding `kind` argument to `ListenerFacade.create()` is a good idea to enhance and also simplify error reporting. I'm not sure would `console logger` be better value than `console` to use here. If you prefer the latter, that's certainly fine.

atest/robot/cli/console/console_type.robot
```diff
@@ -37,9 +40,60 @@ None
     Stdout Should Be    empty.txt
     Stderr Should Be    empty.txt
 
-Invalid
-    Run Tests Without Processing Output    --Console Invalid    misc/pass_and_fail.robot
-    Stderr Should Be Equal To    [ ERROR ] Invalid console output type 'Invalid'. Available 'VERBOSE', 'DOTTED', 'QUIET' and 'NONE'.${USAGE TIP}\n
+Custom console by path
```
pekkaklarck: It would probably be better to move these new tests into a file like `custom_loggers.robot`. The current `console_type.robot` doesn't really match them and these tests are also in general different than tests for built-in console loggers. With the built-in loggers we want to test the actual output, but with custom loggers we are only interested that they are called properly.

Related to the above, `console_type.robot` should probably also be renamed to something more appropriate. Would `builtin_loggers.robot` be good? Whatever the name is, it should somehow match the file name used with custom loggers.

atest/robot/cli/console/console_type.robot
```diff
@@ -37,9 +40,60 @@ None
...
+Custom console with named argument
+    Run Tests    --console ${CONSOLES}${/}CustomConsole.py:marker=NAMED    misc/pass_and_fail.robot
+    Stdout Should Contain    NAMED: Suite 'Pass And Fail' started
+    Stdout Should Contain    NAMED: Test 'Pass' PASS
+    Stdout Should Contain    NAMED: Closing
+    Stderr Should Be Empty
+
+Custom console with wrong arguments
+    Run Tests Without Processing Output    --console ${CONSOLES}${/}CustomConsole.py:too:many:args    misc/pass_and_fail.robot
+    Stderr Should Start With    [ ERROR ] Taking console '
```
pekkaklarck: It would be good to test the full error message. If the problem here is that the path format depends on the system, you could use something like this:
```
${path} =    Normalize Path    ${CONSOLES}${/}CustomConsole.py
```
If the error message contains a portion that comes directly from Python, replacing it with a pattern is typically a good idea. In this case I think we control the whole error message, though.

atest/testresources/consoles/console_classes.py
```diff
@@ -0,0 +1,21 @@
+"""Module with a differently-named console class for dotted import testing."""
```
pekkaklarck: There's no need to have identical classes in different modules. The test for dotted import could just use `CustomConsole.CustomConsole` and this whole module could be removed.

atest/testresources/consoles/CustomConsole.py
```diff
@@ -0,0 +1,29 @@
+import sys
+
+
+class CustomConsole:
+
+    def __init__(self, marker="CUSTOM"):
+        self.marker = marker
```
pekkaklarck: Using `marker` variable and `DOTTED` value is a bit confusing, because the built-in loggers support `--console-markers` and there's a dedicated `DottedOutput`. In here the idea is to test passing arguments, so the variable could be something like `name="DEFAULT"` to avoid confusion.

atest/testresources/consoles/CustomConsole.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def start_suite(self, data, result):
+        sys.__stdout__.write(f"{self.marker}: Suite '{result.name}' started\n")
```
pekkaklarck: Adding a helper method like below would remove duplication. Not a big issue in test code, though.

```python
def write(self, message):
    sys.__stdout__.write(message + "\n")
```

src/robot/output/listeners.py
```diff
@@ -126,20 +126,23 @@ def create(
         listener: "str | object",
         log_level: "LogLevel | SettableLevel" = "INFO",
         library: object = None,
+        kind: str = "listener",
```
pekkaklarck, line 129: 👍

utest/output/test_logger.py
```diff
@@ -1,9 +1,21 @@
 import unittest
+from os.path import abspath, dirname, join
```
pekkaklarck: We prefer `pathlib.Path` with new code, including test code.

pekkaklarck (changes requested):
Looks very good! Some things could be enhanced to make code and tests simpler.

atest/testresources/consoles/CustomConsole.py
oboehmer: Addressed in 6fc37b1

atest/testresources/consoles/CustomConsole.py
oboehmer: Addressed in 6fc37b1

atest/testresources/consoles/console_classes.py
oboehmer: Addressed in fcf973a

src/robot/output/console/__init__.py
oboehmer: Addressed in 747532d and 9ede77c

utest/output/test_logger.py
oboehmer: Addressed in e4c6025

atest/robot/cli/console/console_type.robot
oboehmer: Addressed in 15ef6e8

atest/robot/cli/console/console_type.robot
oboehmer: Addressed in 99dc0893d

atest/robot/cli/console/custom_loggers.robot
```diff
@@ -0,0 +1,62 @@
...
+    Stderr Should Be Empty
+
+Custom console with wrong arguments
+    ${path} =    Normalize Path    ${CONSOLES}${/}CustomConsole.py
+    Run Tests Without Processing Output    --console ${CONSOLES}${/}CustomConsole.py:too:many:args    misc/pass_and_fail.robot
+    Stderr Should Be Equal To    [ ERROR ] Taking console logger '${CONSOLES}${/}CustomConsole.py:too:many:args' into use failed: Importing console logger '${path}' failed: Console logger 'CustomConsole' expected 0 to 1 arguments, got 3.${USAGE TIP}\n
+
+Non-existing custom console
+    Run Tests Without Processing Output    --console NonExistent    misc/pass_and_fail.robot
+    Stderr Should Match Regexp    \\[ ERROR \\] Taking console logger 'NonExistent' into use failed: Importing console logger 'NonExistent' failed: ModuleNotFoundError: No module named 'NonExistent'.*\\n\\nTry --help for usage information\\.
```
pekkaklarck: The error message with `ModuleNotFoundError` has been changing between Python versions and variants. It would be better to ignore everything after `ModuleNotFoundError:` by using `Stderr Should Start With`. If version passes with all Python versions and variants we currently support then its fine, but we'll learn that only when this PR is merged because PRs aren't tested against all variants.

atest/testresources/consoles/CustomConsole.py
```diff
@@ -0,0 +1,32 @@
+import sys
+
+
+class CustomConsole:
+
+    def __init__(self, name="DEFAULT"):
+        self.name = name
+
+    def write(self, message):
+        sys.__stdout__.write(message + "\n")
```
pekkaklarck: Just realized that `print(message, file=sys.__stdout__)` would be a bit better spelling for this. This version is fine as well, though.

pekkaklarck (approved):
Looks good. The main missing thing is updating the documentation of the `--console` option in the `--help` text. That would be done in `src/robot/run.py` file and I would slightly prefer it to be done as part of this PR that adds the new functionality. The User Guide documentation can be done separately, though, and I'm fine with `--help` text being updated then as well.

atest/robot/cli/console/custom_loggers.robot
oboehmer: addressed via f5c36b0

atest/testresources/consoles/CustomConsole.py
oboehmer: addressed via f5c36b0

oboehmer:
> Looks good. The main missing thing is updating the documentation of the `--console` option in the `--help` text. That would be done in `src/robot/run.py` file and I would slightly prefer it to be done as part of this PR that adds the new functionality. The User Guide documentation can be done separately, though, and I'm fine with `--help` text being updated then as well.

Updated the help text via 4cada46, please review

src/robot/conf/settings.py
```diff
@@ -639,7 +639,7 @@ def skip_teardown_on_exit(self):
     @property
     def console_output_config(self):
         return {
-            "type": self.console_type,
+            "logger": self.console_type,
```
pekkaklarck: I didn't realize earlier that console configuration comes here. Having `"logger": self.console_type` is somewhat inconsistent and something like `"logger": self.console_logger` doesn't look good either. I believe it would be best to change this to `"console": self.console`. That would require renaming `RobotSettings.console_type` property, but I consider that fine and this class isn't part of the public API. This would also require changing `ConsoleOutput` signature so that `logger` is renamed to `console`. Now that I think about this, that's probably better argument name in general.

src/robot/output/console/__init__.py
```diff
@@ -13,15 +13,13 @@
...
-from robot.errors import DataError
-
 from .dotted import DottedOutput
 from .quiet import NoOutput, QuietOutput
 from .verbose import VerboseOutput
 
 
 def ConsoleOutput(
-    type="verbose",
+    logger="verbose",
```
pekkaklarck: As discussed above, this could be better as `console`. Sorry for making you to redo these changes! If you are getting annoyed, I could do the changes myself (and I was very close to do that already now).

src/robot/output/logger.py
```diff
@@ -106,15 +106,15 @@ def __exit__(self, *exc_info):
 
     def register_console_logger(
         self,
-        type="verbose",
+        logger="verbose",
```
pekkaklarck: Also this should be renamed to `console`.

src/robot/run.py
```diff
@@ -329,12 +329,19 @@
                           arguments the same way as with --listener.
     --parser parser *     Custom parser class or module. Parser classes accept
                           arguments the same way as with --listener.
-    --console type        How to report execution on the console.
+    --console type or class_or_module       How to report execution on the console.
```
pekkaklarck: The option values we've used elsewhere have been either single words like `parser` above or list of values in format `x|y|z`. For consistency reasons the same approach should be used here instead of the current `type or class_or_module`. I don't think it makes sense to list all values and believe just `console` would be good enough.

src/robot/run.py
```diff
@@ -329,12 +329,19 @@
...
+                          This option takes pre-defined types or custom classes or modules
+                          for further customization. Pre-defined types are listed below:
```
pekkaklarck: These lines are two long and cause tests to fail. Lines must be max 79 characters to make sure the `--help` text is shown properly on 80 character wide consoles like the old Windows Command Prompt.

src/robot/run.py
```diff
@@ -329,12 +329,19 @@
...
                           verbose:  report every suite and test (default)
                           dotted:   only show `.` for passed test, `s` for
                                     skipped tests, and `F` for failed tests
                           quiet:    no output except for errors and warnings
                           none:     no output whatsoever
+                          Other values passed are interpreted as Python
+                          class or module to customize console output.
+                          Argument format is the same as with --listener.
+                          Examples: --console MyConsole
+                                    --console path/to/Console.py:arg1:arg2
```
pekkaklarck: One example with built-in consoles would be good.

pekkaklarck (changes requested):
The `--help` text has lines that are too long and that causes tests to fail. I also noticed few other things to change, but I can handle them myself as well.

src/robot/conf/settings.py
oboehmer: ack, will adjust..

oboehmer: resolved in c84df06

src/robot/run.py
oboehmer: resolved in ffe7e22

src/robot/run.py
oboehmer: resolved in ffe7e22

src/robot/output/logger.py
oboehmer: addressed in c84df06

src/robot/run.py
oboehmer: addressed in ffe7e22

src/robot/output/console/__init__.py
oboehmer: addressed in c84df06

pekkaklarck (approved):
Looks great! I like `console` as an argument/variable name.

Tests had failed due to my mistake with unrelated Windows tests. I rerun them now and unless there's something strange, will merge this when they are done.

pekkaklarck:
Tests still fail because this PR hasn't been rebased with `master` that has the fix. There are no real benefits in doing that so I'll simply merge this now.

PR #5677, opened 2026-05-13
Support configuring console logger with Rebot (#5674)

### LINKED ISSUE #5674: Support configuring console logger with Rebot
### Background

PR #5671 (issue #5618) added support for custom console loggers via `--console` for `robot`. It was agreed that `rebot` should get equivalent support in a separate issue/PR.

Currently `rebot` always uses `VerboseOutput` and has no way to control console output type.

### Proposal

Add the following options to `rebot`:

- **`--console verbose|quiet|none|path.to.CustomLogger`** — Select console output type, including custom loggers (same mechanism as `robot`).
- **`--quiet`** — Shortcut alias for `--console quiet`.
- **Programmatic API**: `rebot(..., console=obj)` — Allow passing a custom console logger object.

### Design considerations

- **No `--dotted` shortcut**: `dotted` vs `verbose` is meaningless for `rebot` since there's no test execution progress to show. It could still be accepted as a value (mapped to `verbose`) for consistency, or simply omitted.
- **No `--consolewidth` / `--consolemarkers`**: These configure execution progress output that doesn't apply to `rebot`.

oboehmer:
## Summary

Add `--console` and `--quiet` options to `rebot`, mirroring the mechanism added to `robot` in #5671/#5672. This allows controlling Rebot's console output type, including custom console loggers.

## Changes

- **`--console verbose|quiet|none|path.to.CustomLogger`** — select console output type for Rebot
- **`--quiet`** — shortcut for `--console quiet`
- **`--console dotted`** — rejected with a clear error (dotted is meaningless without test execution progress)
- **Programmatic API** — `rebot(..., console='quiet')` or `rebot(..., console=my_logger_obj)` both work
- Move `pythonpath` setup before console logger registration in `Rebot.main()` so custom console modules on `--pythonpath` can be imported
- Skipped `--dotted`, `--consolewidth`, `--consolemarkers` as they don't apply to Rebot's non-execution context

## Files changed

| File | Change |
|------|--------|
| `src/robot/conf/settings.py` | Add `ConsoleType`/`ConsoleTypeQuiet` to `RebotSettings`, add `console` property with dotted rejection, update `console_output_config`, prevent console propagation in `get_rebot_settings()` |
| `src/robot/rebot.py` | Add `--console`/`--quiet` to USAGE, reorder pythonpath before console registration, update `rebot()` docstring |
| `atest/resources/atest_resource.robot` | Add `Stdout Should Be Empty` keyword |
| `atest/robot/cli/rebot/console.robot` | New: 12 acceptance tests |

Fixes #5674

### DIFF
--- a/atest/robot/cli/rebot/console.robot
+++ b/atest/robot/cli/rebot/console.robot
@@ -67,3 +67,7 @@ Custom console as module with functions
 Non-existing custom console
     Run Rebot Without Processing Output    --console NonExistent    ${INPUT FILE}
     Stderr Should Start With    [ ERROR ] Taking console logger 'NonExistent' into use failed:
+
+Dotted is not supported
+    Run Rebot Without Processing Output    --console dotted    ${INPUT FILE}
+    Stderr Should Be Equal To    [ ERROR ] Console type 'dotted' is not supported with Rebot.${USAGE TIP}\n
--- a/atest/robot/cli/rebot/console.robot
+++ b/atest/robot/cli/rebot/console.robot
@@ -38,7 +38,3 @@ Custom console by module name
 Non-existing custom console
     Run Rebot Without Processing Output    --console NonExistent    ${INPUT FILE}
     Stderr Should Start With    [ ERROR ] Taking console logger 'NonExistent' into use failed:
-
-Dotted is not supported
-    Run Rebot Without Processing Output    --console dotted    ${INPUT FILE}
-    Stderr Should Be Equal To    [ ERROR ] Console type 'dotted' is not supported with Rebot.${USAGE TIP}\n
--- a/atest/robot/cli/rebot/console.robot
+++ b/atest/robot/cli/rebot/console.robot
@@ -1,0 +1,69 @@
+*** Settings ***
+Suite Setup       Run tests to create input file for Rebot
+Resource          rebot_cli_resource.robot
+
+*** Variables ***
+${CONSOLES}       ${CURDIR}${/}..${/}..${/}..${/}testresources${/}consoles
+
+*** Test Cases ***
+Default is verbose
+    Run Rebot    ${EMPTY}    ${INPUT FILE}
+    Stdout Should Contain    Output:
+
+Verbose
+    Run Rebot    --console verbose    ${INPUT FILE}
+    Stdout Should Contain    Output:
+
+Verbose with log and report
+    Run Rebot    --console verbose --log log.html --report report.html    ${INPUT FILE}
+    Stdout Should Contain    Output:
+    Stdout Should Contain    Log:
+    Stdout Should Contain    Report:
+
+Quiet
+    Run Rebot    --console quiet    ${INPUT FILE}
+    Stdout Should Be Empty
+    Stderr Should Be Empty
+
+--quiet shortcut
+    Run Rebot    --quiet    ${INPUT FILE}
+    Stdout Should Be Empty
+    Stderr Should Be Empty
+
+None
+    Run Rebot    --console none    ${INPUT FILE}
+    Stdout Should Be Empty
+    Stderr Should Be Empty
+
+Custom console by path
+    Run Rebot    --console ${CONSOLES}${/}CustomConsole.py --log log.html --report report.html
+    ...    ${INPUT FILE}
+    Stdout Should Contain    DEFAULT: Output:
+    Stdout Should Contain    DEFAULT: Log:
+    Stdout Should Contain    DEFAULT: Report:
+    Stdout Should Contain    DEFAULT: Closing
+
+Custom console by path with argument
+    Run Rebot    --console ${CONSOLES}${/}CustomConsole.py:ARGUMENT --log log.html --report report.html
+    ...    ${INPUT FILE}
+    Stdout Should Contain    ARGUMENT: Output:
+    Stdout Should Contain    ARGUMENT: Log:
+    Stdout Should Contain    ARGUMENT: Report:
+    Stdout Should Contain    ARGUMENT: Closing
+
+Custom console by module name
+    Run Rebot    --console CustomConsole --pythonpath ${CONSOLES} --log log.html --report report.html
+    ...    ${INPUT FILE}
+    Stdout Should Contain    DEFAULT: Output:
+    Stdout Should Contain    DEFAULT: Closing
+
+Custom console as module with functions
+    Run Rebot    --console ${CONSOLES}${/}module_console.py    ${INPUT FILE}
+    Stdout Should Not Contain    Output:
+    Stdout Should Not Contain    Report:
+    Stdout Should Not Contain    Log:
+    Stdout Should Contain    MODULE: Closing
+
+Non-existing custom console
+    Run Rebot Without Processing Output    --console NonExistent    ${INPUT FILE}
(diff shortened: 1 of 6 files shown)


src/robot/conf/settings.py
```diff
@@ -705,6 +707,8 @@ class RebotSettings(_BaseSettings):
         "StartTime"         : ("starttime", None),
         "EndTime"           : ("endtime", None),
         "Merge"             : ("merge", False),
+        "ConsoleType"       : ("console", "verbose"),
+        "ConsoleTypeQuiet"  : ("quiet", False),
```
pekkaklarck: These are common with `RobotSettings` and should thus be moved to `_BaseSettings`. It already has other console related settings are that common with Robot and Rebot.

src/robot/conf/settings.py
```diff
@@ -764,9 +768,19 @@ def _resolve_background_colors(self):
...
         return self["Merge"]
 
+    @property
+    def console(self):
+        if self["ConsoleTypeQuiet"]:
+            return "quiet"
+        console = self["ConsoleType"]
+        if isinstance(console, str) and console.upper() == "DOTTED":
+            raise DataError("Console type 'dotted' is not supported with Rebot.")
+        return console
```
pekkaklarck: 1. I don't think there's need to actively disallow using `--console dotted`. We probably shouldn't document it and absolutely shouldn't add `--dotted`.
2. The existing `console` property could probably be made generic and moved to `_BaseSettings`.

atest/robot/cli/rebot/console.robot
```diff
@@ -0,0 +1,73 @@
...
+    ...    ${INPUT FILE}
+    Stdout Should Contain    DEFAULT: Output:
+    Stdout Should Contain    DEFAULT: Closing
+
+Custom console as module with functions
+    Run Rebot    --console ${CONSOLES}${/}module_console.py    ${INPUT FILE}
+    Stdout Should Not Contain    Output:
+    Stdout Should Not Contain    Report:
+    Stdout Should Not Contain    Log:
+    Stdout Should Contain    MODULE: Closing
```
pekkaklarck: I'm not sure do we need test custom loggers this thoroughly here. We know the code is exactly the same and that it is already tested with Robot, so having just one test could be fine.

Typically having some duplication in tests isn't a problem, but in this case each test runs Robot separately and that is somewhat slow.

pekkaklarck (commented):
I took a quick look and the PR looks great. I'll take another look with a bit fresher eyes later, but it may take few days before I have time for that. See separate comments for things I noticed now.

src/robot/conf/settings.py
oboehmer: ack, done via 57d60fa6a

src/robot/conf/settings.py
oboehmer: ad 1: implemented via 5a79f4cfe
ad 2: The shared implementation would need to catch non-existent `ConsoleTypeDotted` in rebot, i.e.

```diff
     def console(self):
         if self["ConsoleTypeQuiet"]:
             return "quiet"
-        if self["ConsoleTypeDotted"]:
-            return "dotted"
+        try:
+            if self["ConsoleTypeDotted"]:
+                return "dotted"
+        except KeyError:
+            pass
         return self["ConsoleType"]
```

would you prefer this over the trivial and only slightly redundant current implementation? The behaviour is, after all, different for rebot and robot? 

I could also remove `ConsoleTypeDotted` out of the shared property and do something like below, but this looks less obvious, and would be an exception as all other properties in `_BaseSettings` are exhaustive for both robot and rebot.

```diff
--- a/src/robot/conf/settings.py
+++ b/src/robot/conf/settings.py
@@ -458,6 +458,12 @@ class _BaseSettings:
     def pre_rebot_modifiers(self):
         return self["PreRebotModifiers"]
 
+    @property
+    def console(self):
+        if self["ConsoleTypeQuiet"]:
+            return "quiet"
+        return self["ConsoleType"]
+
     @property
     def console_colors(self):
         return self["ConsoleColors"]
@@ -652,11 +658,9 @@ class RobotSettings(_BaseSettings):
 
     @property
     def console(self):
-        if self["ConsoleTypeQuiet"]:
-            return "quiet"
         if self["ConsoleTypeDotted"]:
             return "dotted"
-        return self["ConsoleType"]
+        return super().console
 
     @property
     def console_width(self):
@@ -766,12 +770,6 @@ class RebotSettings(_BaseSettings):
     def merge(self):
         return self["Merge"]
 
-    @property
-    def console(self):
-        if self["ConsoleTypeQuiet"]:
-            return "quiet"
-        return self["ConsoleType"]
-
     @property
     def console_output_config(self):
         return {
```

atest/robot/cli/rebot/console.robot
oboehmer: fair point, addressed via 0cfdfe36a

src/robot/conf/settings.py
pekkaklarck: The latest approach looks fine to me although the original with some duplication in two `console` implementations is fine as well.

Another alternative that I think would work is listing also `ConsoleTypeDotted` in the base. If `--dotted` isn't listed in Rebot's `--help`, the value should always be `False`. I'm not sure is the some validation that a specified setting is actually used, though, so this may not work.

Finally, there's a plan to rewrite the whole settings system at some point. It's old and not that great code that was written when the only command line option parsing module in Python standard library we could use was `getopt`. The code has worked, and being able so specify options directly in `--help` text is convenient, so we haven't touched it. I believe we should add configuration file support somewhat soon, and at that point this should be refactored. This means that whatever code we use is likely going to be rewritten soonish.

atest/robot/cli/rebot/console.robot
```diff
@@ -0,0 +1,40 @@
...
+    Run Rebot    --quiet    ${INPUT FILE}
+    Stdout Should Be Empty
+    Stderr Should Be Empty
+
+None
+    Run Rebot    --console none    ${INPUT FILE}
+    Stdout Should Be Empty
+    Stderr Should Be Empty
+
+Custom console by module name
```
pekkaklarck: There are no other custom console tests here, so this test should probably be named just "Custom console".

src/robot/rebot.py
```diff
@@ -256,6 +256,18 @@
...
+                          Built-in consoles:
+                          verbose: show output file paths (default)
+                          quiet:   no output except for errors/warnings
+                          none:    no output whatsoever
+                          Other values are interpreted as a custom
+                          console class or module. Argument format is
+                          the same as with --listener.
+                          Examples: --console verbose
+                                    --console quiet
+                                    --console path/to/Con.py:arg1:arg2
```
pekkaklarck: This could be a bit more explicit if it was changed to `path/to/Console.py:arg`, but this version is fine too. If you change it, change also the example on Robot side.

pekkaklarck (approved):
Looks very good! I added some nitpickey comments that you may want to look at. Write a comment here or ping me on Slack when you are done with possible changes and I merge this.

src/robot/conf/settings.py
oboehmer: ok, then I would leave it as is, it clearly expresses the intended behavior.

oboehmer:
@pekkaklarck , all comments addressed, ready to be merged.

PR #5680, opened 2026-05-25
Document custom console loggers for robot and rebot in User Guide

oboehmer:
Add documentation for the --console option accepting custom console logger classes and modules, introduced in PRs #5672 and #5677.

- Add "Custom console loggers" subsection under "Console output type" in ConfiguringExecution.rst with usage examples and a ProgressConsole class example.
- Add "Controlling Rebot console output" section in PostProcessing.rst documenting --console and --quiet support for Rebot.
- Update CommandLineOptions.rst appendix for both robot and rebot.
- Add cross-reference from ListenerInterface.rst noting that custom console loggers use the listener version 3 API.

related to #5618 and #5674

### DIFF
--- a/doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
+++ b/doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
@@ -817,6 +817,84 @@ Examples::
     robot --console quiet tests.robot
     robot --dotted tests.robot
 
+Custom console loggers
+~~~~~~~~~~~~~~~~~~~~~~
+
+In addition to the built-in values listed above, the :option:`--console` option
+accepts a path or name of a custom console logger class or module. The argument
+format is the same as with :option:`--listener`: a path to a Python file, a
+module name, or a dotted class name, with optional arguments separated by colons.
+
+Custom console loggers receive the same `listener version 3`__ method calls as
+normal listeners. Only methods that are implemented are called — missing methods
+are silently ignored. This means a minimal console only needs to implement the
+hooks it is interested in.
+
+__ `Listener interface`_
+
+Examples::
+
+    robot --console path/to/MyConsole.py tests.robot
+    robot --console MyConsole --pythonpath /path/to/consoles tests.robot
+    robot --console MyConsole.py:arg1:arg2 tests.robot
+
+The following example shows a custom console that provides a compact progress
+view with elapsed time and a running pass/fail counter — useful in CI pipelines
+or when the verbose output is too noisy but `dotted` does not provide enough
+context:
+
+.. sourcecode:: python
+
+    import sys
+    import time
+
+
+    class ProgressConsole:
+
+        def __init__(self):
+            self.passed = 0
+            self.failed = 0
+            self.skipped = 0
+            self.start_time = None
+
+        def start_suite(self, data, result):
+            if self.start_time is None:
+                self.start_time = time.time()
+
+        def end_test(self, data, result):
+            if result.passed:
+                self.passed += 1
+            elif result.failed:
+                self.failed += 1
+            else:
+                self.skipped += 1
+            elapsed = time.time() - self.start_time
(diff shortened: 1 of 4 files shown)


doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
```diff
@@ -817,6 +817,84 @@ Examples::
...
+format is the same as with :option:`--listener`: a path to a Python file, a
+module name, or a dotted class name, with optional arguments separated by colons.
+
+Examples::
+
+    robot --console path/to/myconsole.py tests.robot
+    robot --console MyConsole --pythonpath /path/to/consoles tests.robot
+    robot --console CustomConsole.py:arg1:arg2 tests.robot
+
+Custom console loggers receive the same `listener version 3`__ method calls as
```
pekkaklarck: We also support listener v2, that support was free, so there's no need to talk about the version here. Probably the linked text could be just `listener method calls`.

doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
```diff
@@ -817,6 +817,84 @@ Examples::
...
+
+        def result_file(self, kind, path):
+            print(f"{kind}:  {path}", file=sys.__stdout__)
+
+        def close(self):
+            elapsed = time.time() - self.start_time
+            total = self.passed + self.failed + self.skipped
+            print(f"\n{total} tests in {elapsed:.1f}s: "
+                  f"{self.passed} passed, {self.failed} failed, "
+                  f"{self.skipped} skipped", file=sys.__stdout__)
```
pekkaklarck: Cool example!

Is `file=sys.__stdout__` needed? If it is, I can take a look could that limitation be lifted. If it cannot, we probably should document it explicitly.

I'm starting to get used to Black style and would prefer this kind of formatting:
```python
print(
    f"\n{total} tests in {elapsed:.1f}s: "
    f"{self.passed} passed, {self.failed} failed, {self.skipped} skipped",
    file=sys.__stdout__
)
```

doc/userguide/src/ExtendingRobotFramework/ListenerInterface.rst
```diff
@@ -1105,3 +1105,7 @@ that covering them fully here in the User Guide is not possible. For more exampl
 you can see the `acceptance tests`__ using theses methods in various ways.
 
 __ https://github.com/robotframework/robotframework/tree/master/atest/testdata/output/listener_interface/body_items_v3
+
+The listener version 3 API is also used by `custom console loggers`_ that can
+replace the built-in console output (verbose, dotted, etc.) with a user-defined
+implementation.
```
pekkaklarck: Having a note that also console loggers use the listener interface is a good idea, but it should probably be somewhere in the beginning of the Listener interface section. As I commented earlier, we support also v2 listeners so no need to mention version information at all. Perhaps there could just be something like this:

```
.. note:: The listener interface is used also by `custom console loggers`_.
```

pekkaklarck (changes requested):
Very good! We need to check is `file=sys.__stdout__` needed and there's no need to talk about listener versions at all.

doc/userguide/src/ExtendingRobotFramework/ListenerInterface.rst
oboehmer: done

doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
oboehmer: dropped `sys.__stdout__`, the calls are happening outside the keyword execution context, so no need to redirect. Also changed formatting

doc/userguide/src/ExecutingTestCases/ConfiguringExecution.rst
oboehmer: done

PR #5495, opened 2025-09-10
[Enhancement] custom metadata support for testcases and keywords

febb0e:
# Custom Metadata Feature

Closes #4409 

## Overview

Introduces custom metadata feature for test cases and user keywords. The implementation allows users to define metadata for both test cases and keywords using the `[CustomName]` syntax.

## Implementation Details

- **Custom Metadata Syntax:**
  - Any setting in square brackets (e.g., `[Owner]`, `[Requirement]`, `[My Custom Metadata]`) is now treated as custom metadata.
  - Custom metadata is supported in both test cases and user keywords.
  - Metadata values can be single or multi-valued, and support variable resolution.
- **Parsing and Output:**
  - Custom metadata is parsed and stored in the test/keyword model.
  - Custom metadata is included in the output XML and HTML reports by default.
  - Output and parsing logic was adapted to treat all unknown square bracket settings as custom metadata.
- **Backwards Compatibility:**
  - Test cases that previously expected errors for undefined square bracket settings (e.g., `[UndefinedSetting]`) will now treat these as valid custom metadata.

## CLI Functionality: `--custommetadata` Flag

- **Purpose:**
  - The `--custommetadata` flag allows users to filter which custom metadata keys are considered "valid" for display and parsing in output files.
- **Behavior:**
  - **No flag set:** All custom metadata keys are valid and included in output and parsing.
  - **Flag set with one or more keys:** Only the specified custom metadata keys are considered valid and included in output and parsing. Others are ignored. Test cases and user keywords do not fail.
  - **Important:** The flag does **not** influence the execution of tests or keywords. It only affects which custom metadata is displayed in output or parsed through execution.

## Test Coverage

- **Acceptance Tests:**
  - Acceptance tests were created to cover major scenarios:
    - Basic custom metadata parsing
    - Edge cases (empty values, special characters, Unicode, etc.)
    - Performance (many metadata entries)
    - Variable resolution and built-in expressions
    - CLI filtering with `--custommetadata`
    - Output and report validation
- **Unit Tests:**
  - Unit tests ensure parsing, model storage, and output generation for custom metadata.

## Notable Changes

- **Error Handling:**
  - Square bracket settings that do not match known settings are now always treated as custom metadata. This means tests expecting errors for undefined settings will no longer fail; instead those settings are parsed as metadata.

## Summary

[...]

PR #5685, opened 2026-06-15
Test Metadata Feature

### LINKED ISSUE #4409: Test metadata

febb0e:
Closes https://github.com/robotframework/robotframework/issues/4409

# Overview

Add support for testcase metadata via the normal test setting syntax:

- `[Metadata]    Name    Value`

This metadata is now carried through execution, written to `output.xml`, parsed back from XML, into the log.

## What changed

- Test case settings now accept `Metadata`.
- Test model (`running` + `result`) now includes `metadata` on test cases.
- Suite runner copies test metadata from parsed data to result objects.
- XML logger writes test metadata as `<meta name="...">...</meta>` under each `<test>`.
- XML result parsing accepts and restores test-level `<meta>` entries.
- JS model/log rendering now includes test metadata (same as already used for suite metadata).

## Compatibility

- Existing suite metadata behavior is unchanged.
- XML parsing remains compatible.

## Tests

Added acceptance suite for the core flow:

- parse test metadata from test settings
- verify metadata in in-memory result model
- verify metadata in `output.xml`
- verify metadata survives `rebot` processing
- verify metadata is present in log output

## Notes

This change is only about explicit `[Metadata]` in test cases.
It does not introduce custom bracket settings like `[Owner]` or a new CLI filtering flag.

## Usage example

```robotframework
*** Test Cases ***
Login With Valid Credentials
	[Metadata]    Owner     QA Team
	[Metadata]    Ticket    RF-4409
	[Metadata]    Component     Authentication
	Open Application
	Login With Credentials
	Click Login Button
	Dashboard Should Be Visible
	[Teardown]    Close Application
```


Noordsestern:
Makes #5495 obsolete

src/robot/htmldata/rebot/log.html
```diff
@@ -280,6 +280,12 @@ <h2>Execution Errors</h2>
           <td class="doc">{{html doc()}}</td>
         </tr>
         {{/if}}
+        {{each metadata}}
+        <tr>
+          <th>{{html $value[0]}}:</th>
+          <td class="doc">{{html $value[1]}}</td>
+        </tr>
+        {{/each}}
```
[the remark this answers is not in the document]
pekkaklarck: This seems to be true and we also seem to have the same issue with existing suite metadata.

src/robot/running/suiterunner.py
```diff
@@ -152,6 +152,7 @@ def visit_test(self, data: TestData):
             self._resolve_setting(data.tags),
             self._get_timeout(data),
             data.lineno,
+            metadata=data.metadata.copy(),
             start_time=datetime.now(),
         )
```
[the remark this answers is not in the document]
pekkaklarck: `data.metadata` is a `Metadata` instance so this is safe.

pekkaklarck:
I took a somewhat quick look at the changes and they look great! It's pretty cool how little changes were actually needed. I also looked at the AI review comments and agreed with most of them. Things to do before this can be merged:

1. Fix conflicts. I didn't actually spot any conflicts myself, but GitHub UI complains about them and at the moment merge isn't even possible.
2. Fix issues reported by AI.
3. A bit more thorough review by me. At that point I plan to test the changes locally as well.
4. Fix possible issues found in the review. I can fix simple things myself.

After this has been merged, the functionality still needs to be documented. Could you @febb0e take a look at that as well? Probably better to open a separate PR for that.

Overall this looks so good that I move the related issue back to RF 7.5 scope. This most likely won't make it to the first alpha/beta that I plan to release this week, but I see no problems getting this in before the final version.

pekkaklarck:
One thing that's missing is that XML and JSON schemas need to be updated. I can take care of that if needed.

febb0e:
Hey @pekkaklarck
thanks for your response. I will take care of the review suggestions and update this PR accordingly.
Will also look into why there are any merge-conflicts showing up.
you will have an update on this soon from me. And yes I should document the functionality, will do this as you requested in a separate PR.

febb0e:
Hey @pekkaklarck

Thanks again for the review! I went through all of the AI comments and fixed the ones that turned out to be real issues:

- changed the metadata type annotations in `model/testcase.py` to match `Mapping[str, str] | None` and be consistent with the suite side.
- Fixed `to_dict()` to return a plain dictioanry instead of the Metadata object.
- Added the missing test-level metadata entries to XML and JSON schemas.

One comment (the log.html escaping one) is a false positive in my opinion, the key and value are already escaped via the same html_escape/html_format that the suite metadata uses, so I left that as is.

While working on this I found a handful of small bugs in the surrounding code that showed up when testing the feature end-to-end.

- Test metadata was missing from native JSON output (--output out.json) and from the Listener API v2 attribute dict — both places already had that for suites.
- Variables in test-level `[Metadata]    Key    ${VAR}` values weren't being resolved at all (suite metadata resolves them fine), so they were showing up as `${VAR}` in the output.
- The Testdoc tool was missing per-test metadata in its model.
- Also had to update one pre-existing test `test_case_settings.robot` that explicitly asserted `[Metadata]` is invalid in a test case, which is now false with this PR :)

I did consider adding a `Set Test Metadata keyword + ${TEST_METADATA}` variable for symmetry with the suite metadata side. But that's a new keyword/variable, so I didn't want to put it into this PR. Let me know what you think here.

Also let me know if you want anything split out or done differently!
I will now continue on the documentation of the feature, but this will happen in a different PR :)

pekkaklarck:
Thanks for the update! I'm back from holiday and will review this now. Hopefully we get this in soon and can then create RF 7.5 rc 1!

Great that you found and fixed those issues surrounding the main logic. It's hard to remember to get all of such cases covered, and I'd be surprised if we didn't need to make some adjustments still in the subsequent releases.

I agree `Set Test Metadata` keyword and `${TEST METADATA}` variable can be left out for now. Adding them before RF 7.5 final is still possible, but I don't consider it a big problem if they were added only in the next release. Let's anyway concentrate on getting the main logic implemented, tested and documented.

atest/robot/output/testcase_metadata.robot
```diff
@@ -0,0 +1,41 @@
+*** Settings ***
+Resource       atest_resource.robot
+Suite Setup    Run Tests    --log log.html --report report.html    output/testcase_metadata.robot
```
pekkaklarck, line 3: Could add `validate output=True` here to explicitly validate output.xml against the schema.

Schema validation can be enabled for all cases where output.xml is parsed with like `atest/run.py -S`. It slows execution a bit so it's not done by default, but explicitly enabling it where it makes sense is fine.

atest/robot/output/testcase_metadata.robot
```diff
@@ -0,0 +1,41 @@
...
+    Test Case Metadata Should Be Correct
+
+Test Case Metadata Is Written To XML Output
+    ${owner} =    Get Element    ${OUTFILE}    .//test[@name="Test With Metadata"]/meta[@name="Owner"]
+    Should Be Equal    ${owner.text}    Team Robot
+    ${ticket} =    Get Element    ${OUTFILE}    .//test[@name="Test With Metadata"]/meta[@name="Ticket"]
+    Should Be Equal    ${ticket.text}    RF-9999
+    ${html} =    Get Element    ${OUTFILE}    .//test[@name="Test With Metadata"]/meta[@name="Html"]
+    Should Be Equal    ${html.text}    <b>value</b> & data
+    Element Should Not Exist    ${OUTFILE}    xpath=.//test[@name="Test Without Metadata"]/meta
```
pekkaklarck, line 16: We haven't tested other constructs in XML this way. We have just validated that information is available in the model after parsing XML, like the previous test does, and also validated that outputs match the schema. That said, I don't see real harm in validating XML also like this so probably we can leave this test as it is.

atest/robot/output/testcase_metadata.robot
```diff
@@ -0,0 +1,41 @@
...
+Test Case Metadata Is Included In Log Model
+    Run Tests    --log log.html --report NONE    output/testcase_metadata.robot
+    File Should Contain    ${OUTDIR}/log.html    Team Robot
+    File Should Contain    ${OUTDIR}/log.html    RF-9999
+    File Should Contain    ${OUTDIR}/log.html    &lt;b&gt;value&lt;/b&gt; &amp; data
+
+Test Case Metadata Is Included In JSON Output
+    Copy Previous Outfile
+    Run Tests Without Processing Output    -o output.json    output/testcase_metadata.robot
+    Outputs Should Contain Same Data    ${OUTFILE COPY}    ${OUTDIR}/output.json    ignore_timestamps=True
```
pekkaklarck, line 32: It would be good to explicitly validate that tests contains correct metadata also in this case. Validating that results contain same data regardless the output file format is probably a good idea as well, but that functionality may be buggy and ignore test metadata.

atest/testdata/output/testcase_metadata.robot
```diff
@@ -0,0 +1,9 @@
+*** Test Cases ***
+Test With Metadata
+    [Metadata]    Owner    Team Robot
+    [Metadata]    Ticket    RF-9999
+    [Metadata]    Html    <b>value</b> & data
```
pekkaklarck, line 5: Perhaps "Escaped" would be a better name than "HTML". The point here is that the value is escaped.

Related to the above, suite metadata values support formatting using Robot's own syntax (and the plan is to support Markdown in the future). There should thus be an entry like this as well:

```
[Metadata]    Formatting    Robot *rocks*!
```
and in HTML the value should be `Robot <b>rocks</b>!`.

src/robot/htmldata/rebot/log.html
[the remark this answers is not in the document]
pekkaklarck: Was it @febb0e so that also keys are escaped and using `{{html ...}}` is safe (and actually required to avoid double escaping)?

pekkaklarck (approved):
Looks really good!

My comments seem to be only related to acceptance tests and they are mostly about style. Formatting metadata values definitely needs to be tested on acceptance test level, but it seems to be unit tested so the functionality ought to work fine.

I'm pretty sure there are some corner cases we have missed, but let's merge this now and fix possible problems afterwards. I can also fix the minor test issues myself at the same time as I experiment with this feature locally.

pekkaklarck:
As I wrote in the review, this looks good enough to be merged, but I'll wait that test run finishes before that. If you @febb0e are online, you can look and possibly fix the small issues I commented while tests run, but I can also fix them myself after merging this.

The next major task related to this is documentation. After that we can still think about `Set Test Metadata` and `${TEST METADATA}`. I also want to get a release candidate with this feature out soon so that your users have a possibility to test this before the final release.

febb0e:
Hey @pekkaklarck 
I have started to implement the issues you have mentioned. I am trying to create a PR for every issue, starting with the missing report entry for the testcase metadata and enhancements to the test cases as mentioned.
See: #5743 

Next thing was to add documentation, this was something I already prepared and just adjusted to your review.
See: #5744 

And I also added the &{TEST METADATA} and Set Test Metadata implementation.
See: #5745

PR #5700, opened 2026-07-07
Fix type_name crashing on objects without __class__ (#5699)

### LINKED ISSUE #5699: Automatic type conversion throws AttributeError on arguments that have no `__class__` attribute
Image

The objects causing this issue come from the SDK / Python library from a vendor and I'd consider these objects very non-Python; dunder-attributes like __class__, __str__, __dict__, etc.  are all simply not present. On a breakpoint, the repr seems to be `<frozen abc>`.

Note that `isinstance` does work with these objects,

While automatic conversion is not reasonably possible for objects like this, the automatic type conversion should not raise an exception.

kranthi9558:
Fixes #5699.

## Problem

`type_name` raised `AttributeError` on objects that don't expose
`__class__`, such as objects from some vendor SDKs (e.g. Squish GUI
testing). These objects are very non-Python — dunder attributes like
`__class__`, `__str__`, and `__dict__` are simply not present — but
`isinstance` still works with them.

The crash happened because `type_name` used `isinstance(item, IOBase)`
on the *instance*. Since `IOBase` is an abstract base class, this check
invokes ABC machinery that accesses the object's `__class__`, which
raises `AttributeError` for these objects.

This is especially problematic because `type_name` is used mainly to
build error messages (e.g. in the type-conversion failure path). So
instead of the real "cannot be converted" error, users got a confusing
`AttributeError` masking it.

## Fix

Compute the type once with `type(item)` (which works at the interpreter
level even when `__class__` is inaccessible) and check
`issubclass(type(item), IOBase)` against the *type* rather than the
instance. This avoids touching the object's attributes.

Behavior is unchanged for all normal inputs (files still resolve to
"file", etc.). For objects without `__class__`, `type_name` now returns
a usable name instead of raising, so the real conversion error surfaces.

## Tests

Added a unit test in `utest/utils/test_robottypes.py` that reproduces
the failure (an object whose instance-check raises `AttributeError`) and
verifies `type_name` returns a name instead of crashing. Full unit suite
passes (2395 tests).

### DIFF
--- a/src/robot/utils/robottypes.py
+++ b/src/robot/utils/robottypes.py
@@ -70,18 +70,19 @@ def type_name(item, capitalize=False):
         # Prior to Python 3.10, typing special forms (Any, Union, ...) didn't
         # have `__name__` but instead they had `_name`.
         name = item.__name__ if hasattr(item, "__name__") else item._name
-    elif isinstance(item, IOBase):
-        name = "file"
     else:
         typ = item if isinstance(item, type) else type(item)
-        named_types = {
-            str: "string",
-            bool: "boolean",
-            int: "integer",
-            type(None): "None",
-            dict: "dictionary",
-        }
-        name = named_types.get(typ, typ.__name__.strip("_"))
+        if issubclass(typ, IOBase):
+            name = "file"
+        else:
+            named_types = {
+                str: "string",
+                bool: "boolean",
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Thanks for the PR. The fix looks good to me.

Issue #5709, opened 2026-07-17
Convert standard library docs to Markdown and take argument documentation syntax to use


pekkaklarck:
Now that Libdoc supports Markdown (#5304) and arguments, return values and exceptions can be explicitly documented (#5604), we should take these great enhancements into use with standard libraries. Especially argument documentation will make resulting docs much better. Using Markdown instead of Robot's custom format doesn't matter that much for end users, but it certainly helps when libraries are shown as part of API docs.

It would be great to get all libraries converted in RF 7.5, but that may be too big a task. We should, however, convert at least some libraries so that they can be used as examples.

External help with conversion would be highly appreciated in general. This is an especially good issue for those who want to contribute to Robot Framework to get started. If you are interested, please comment here what library you'd like to look at first. Please also create PRs so that there are changes only to one library in one PR.

As part of this change, also the standard type documentation should be updated. That requires a small change on Robot side as well.

Shaurya2k06:
I want to pick this up! 

I'll start with Dialogs and Screenshot since they're small and self-contained, good first examples. Converting both to ROBOT_LIBRARY_DOC_FORMAT = "Markdown" with Args:/Returns: docs for every keyword, verified via libdoc. I'll open PRs, one library each. After that I'll move on to String, DateTime, and Collections,moving to the larger ones.

pekkaklarck:
Awesome! Doing libraries one-by-one is great, because we can then decide on conventions as needed.

One thing to decide already now is what to do with exceptions. For example, does it make sense do add something like

    Raises:
        AssertionError: If `first` and `second` are not equal.

to `Should Be Equal` or is that too redundant?

coyaSONG:
I would like to take the OperatingSystem library in one separate PR, without touching Dialogs, Screenshot, String, DateTime, or Collections. I will convert its library and keyword documentation to Markdown, add explicit argument and return documentation, keep exception documentation only where it adds useful information beyond the existing prose, and validate the generated Libdoc output plus the relevant tests. OpenAI Codex is assisting with the work, and I will disclose that in the PR.

Shaurya2k06:
Hey @pekkaklarck , gave it some thought, and I think we should only use the former when it adds information the prose does not already make clear. What do you think?

Shaurya2k06:
Hey @pekkaklarck 
[...]

PR #5710, opened 2026-07-17
Convert OperatingSystem documentation to Markdown

coyaSONG:
## What

- Convert the OperatingSystem library and keyword documentation to Markdown.
- Add structured Args and Returns documentation for all exposed keyword arguments and typed return values.
- Convert Robot Framework documentation tables, links, headings, and examples to Markdown equivalents.
- Add a Libdoc regression test that prevents undocumented arguments and return values.
- Update the OperatingSystem acceptance-test expectation for the converted deprecation link.

## Why

This is the OperatingSystem-only part of #5709. It does not touch Dialogs, Screenshot, String, DateTime, or Collections.

## How

The library now declares ROBOT_LIBRARY_DOC_FORMAT = "Markdown". Existing documentation content and behavior are preserved while syntax is converted, and exception behavior remains in the prose where separate Raises entries would be redundant.

## Testing

- invoke format on the three changed files
- invoke library-docs OperatingSystem
- Libdoc API unit tests: 6 passed
- Full Libdoc unit tests: 48 passed
- OperatingSystem acceptance tests: 279 passed
- Libdoc acceptance tests: 437 passed
- Generated JSON: 56 keywords, 0 undocumented arguments, 0 undocumented typed returns
- AST comparison: implementation unchanged apart from documentation and the documentation-format declaration

## AI assistance

OpenAI Codex assisted with the documentation conversion, tests, and validation. I reviewed the resulting diff and verification evidence.


src/robot/libraries/OperatingSystem.py
```diff
@@ -39,144 +39,156 @@
...
-    | ``[a-z]``    | matches one character from the range in the bracket     |
-    | ``[!a-z]``   | matches one character not from the range in the bracket |
+    | Pattern    | Matches                                                  |
+    | ---------- | -------------------------------------------------------- |
+    | `*`        | Any string, even an empty string                         |
+    | `?`        | Any single character                                     |
+    | `[chars]`  | One character in the bracket                             |
+    | `[!chars]` | One character not in the bracket                         |
+    | `[a-z]`    | One character from the range in the bracket              |
+    | `[!a-z]`   | One character not from the range in the bracket          |
```
pekkaklarck: Headers could be aligned and the extra whitespace from the end removed.

src/robot/libraries/OperatingSystem.py
```diff
@@ -39,144 +39,156 @@
...
-    Because the backslash character (``\``) is an escape character in
+    Because the backslash character (`\`) is an escape character in
     Robot Framework data, possible backslash characters in regular
-    expressions need to be escaped with another backslash like ``\\d\\w+``.
+    expressions need to be escaped with another backslash like `\\d\\w+`.
     Strings that may contain special characters but should be handled
-    as literal strings, can be escaped with the `Regexp Escape` keyword
-    from the BuiltIn library.
+    as literal strings, can be escaped with the
+    [Regexp Escape](BuiltIn.html#Regexp%20Escape) keyword from the BuiltIn library.
```
pekkaklarck: Making references to external libraries actual links is a good idea, but I'm not sure can we count on `BuiltIn.html` being on the same directory as the docs of this library. When the docs are hosted online that is
the case, but if someone downloads  the doc or views this doc via an editor, the link won't work.

Safer alternatives:
- Link to the online version with a full URL like `[Regexp Escape](https:///...#Regexp%20Escape)`. This has a problem that the URL is pretty long (I needed to cut it in this comment as well) making the source look a bit bad. It may not matter much in general, but if it does, creating a reference target and then using just `[Regexp Escape]` in the actual documentation would help.
- Just format the keyword name and make only `BuiltIn` a link that points to online documentation. Also this requires a longish URL, but this time a reference target for `[BuiltIn]` could be used also in other places. A drawback is that users would need to find the actual referenced keyword themselves. A direct link would probably be better, even if it would require multiple reference targets.
- Don't do any linking. We haven't had links earlier so not adding links now is obviously fine as well. I have been thinking that adding links like this would be a good idea, though.

src/robot/libraries/OperatingSystem.py
```diff
@@ -39,144 +39,156 @@
...
-    Paths beginning with ``~`` or ``~username`` are expanded to the current or
+    Paths beginning with `~` or `~username` are expanded to the current or
     specified user's home directory, respectively. The resulting path is
-    operating system dependent, but typically e.g. ``~/robot`` is expanded to
-    ``C:\Users\<user>\robot`` on Windows and ``/home/<user>/robot`` on Unixes.
+    operating system dependent, but typically e.g. `~/robot` is expanded to
+    `C:\Users\<user>\robot` on Windows and `/home/<user>/robot` on Unixes.
 
-    = pathlib.Path support =
+    # pathlib.Path support
```
pekkaklarck: The `pathlib.Path` part could probably be code formatted with backticks. I think the table of contents and everything else will work even if a header has formatting, but that would need to be tested.

src/robot/libraries/OperatingSystem.py
```diff
@@ -39,144 +39,156 @@
...
+        (`\\n`) after the output to make it easier to read in the
         console. To ease processing the returned output, this possible
         trailing newline is stripped by this keyword.
 
+        Args:
+            command: Command to execute.
+
+        Returns:
+            The command output with a possible trailing newline removed.
+
```
pekkaklarck: I'm not entirely sure what is the best place for documenting arguments and return values. I guess it to some extend depends on the keyword, but perhaps it would be best to always have them in the beginning after the "short doc" as in the example below. That's the approach we've used with other libraries.

```python
def example(arg: int) -> int:
    """This is the short doc.

    Args:
        arg: Argument doc.

    Returns:
        Retrun value doc.

    The main keyword doc.
    """
 ```

File: src/robot/libraries/OperatingSystem.py [hunk 5]
```diff
@@ -39,144 +39,156 @@
...
+        ```robotframework
+        ${output} =            Run       ls -lhF /tmp
+        Log                    ${output}
+        ${result} =            Run       ${CURDIR}${/}tester.py arg1 arg2
+        Should Not Contain     ${result}    FAIL
+        ${stdout} =            Run       /opt/script.sh 2>/tmp/stderr.txt
+        Should Be Equal        ${stdout}    TEST PASSED
+        File Should Be Empty   /tmp/stderr.txt
+        ```
```
pekkaklarck, line 192: These examples must have the `*** Test Cases ***` header and a test case name to be syntax highlighted correctly. In most cases using the name of the documented keyword as the test case name is fine. If there are needs for multiple slightly different tests, they should be based based on what they demonstrate.

src/robot/libraries/OperatingSystem.py
```diff
@@ -39,144 +39,156 @@
...
         """
         return self._run(command)[1]
 
     def run_and_return_rc(self, command: str) -> int:
         """_This keyword is considered deprecated. Use the
-        [http://robotframework.org/robotframework/latest/libraries/Process.html|
-        Process] library instead._
+        [Process](http://robotframework.org/robotframework/latest/libraries/Process.html)
+        library instead._
```
pekkaklarck: This is the second time the Process library is referenced and there likely is at least one more occurrence. We can avoid the need to repeat the URL by creating a reference target for `[Process]` in the library introduction like we did with [BuiltIn](https://github.com/robotframework/robotframework/blob/afe79a87a54b738ff85c8ba5c7a00d26fd4ca451/src/robot/libraries/BuiltIn.py#L5246). After that the doc here can be simplified to `use the [Process] library instead`.

src/robot/libraries/OperatingSystem.py
```diff
@@ -185,37 +197,57 @@ def run_and_return_rc(self, command: str) -> int:
         some operating systems (notable Windows) original return codes
         can be something else, but this keyword always maps them to
         the 0-255 range. Since the return code is an integer, it must be
-        checked e.g. with the keyword `Should Be Equal As Integers`
-        instead of `Should Be Equal` (both are built-in keywords).
+        checked e.g. with the keyword
+        [Should Be Equal As Integers](BuiltIn.html#Should%20Be%20Equal%20As%20Integers)
+        instead of [Should Be Equal](BuiltIn.html#Should%20Be%20Equal)
```
pekkaklarck: See my earlier comment about using `BuiltIn.html` as tarted path instead of an online version. If there are more occurrences like this, I'm not going to comment them separately anymore.

src/robot/libraries/OperatingSystem.py
```diff
@@ -185,37 +197,57 @@ def run_and_return_rc(self, command: str) -> int:
...
+        (both are built-in keywords).
+
+        Args:
+            command: Command to execute.
+
+        Returns:
+            The command return code as an integer in the range 0-255.
```
pekkaklarck: See my earlier comment about the Args/Returns placement. I'm not going to comment about this separately with latter occurrences.

src/robot/libraries/OperatingSystem.py
```diff
@@ -185,37 +197,57 @@ def run_and_return_rc(self, command: str) -> int:
...
-        | Should Be Equal As Integers | ${rc} | 0 |
-        | ${rc} = | Run and Return RC | /path/to/example.rb arg1 arg2 |
-        | Should Be True | 0 < ${rc} < 42 |
 
-        See `Run` and `Run And Return RC And Output` if you need to get the
+        ```robotframework
+        ${rc} =                       Run and Return RC    ${CURDIR}${/}script.py arg
+        Should Be Equal As Integers    ${rc}    0
+        ${rc} =                       Run and Return RC    /path/to/example.rb arg1 arg2
+        Should Be True                0 < ${rc} < 42
```
pekkaklarck: See my earlier comment about the required headers and names. I'm not going to comment latter occurrences of this problems anymore.

src/robot/libraries/OperatingSystem.py
```diff
@@ -236,26 +268,34 @@ def get_file(
...
-        - ``CONSOLE``: Use the console encoding. Outside Windows this is same
+        - `SYSTEM`: Use the default system encoding.
+        - `CONSOLE`: Use the console encoding. Outside Windows this is same
           as the system encoding.
 
-        ``encoding_errors`` argument controls what to do if decoding some bytes
-        fails. All values accepted by ``decode`` method in Python are valid, but
+        `encoding_errors` argument controls what to do if decoding some bytes
+        fails. All values accepted by `decode` method in Python are valid, but
         in practice the following values are most useful:
```
pekkaklarck: A link to docs about the accepted error handles could be added. We have just added it to a String library keyword that allow encoded values. See here:
https://github.com/robotframework/robotframework/blob/78dd2a947c0aad931b64715160e9e21e5181c19b/src/robot/libraries/String.py#L236

src/robot/libraries/OperatingSystem.py
```diff
@@ -1108,63 +1341,90 @@ def log_environment_variables(
...
+        ${path} =    Join Path    my           path
+        ${p2} =      Join Path    my/          path/
+        ${p3} =      Join Path    my           path     my       file.txt
+        ${p4} =      Join Path    my           /path
+        ${p5} =      Join Path    /my/path/     ..       path2
+        Should Be Equal    ${path}    my/path
+        Should Be Equal    ${p2}      my/path
+        Should Be Equal    ${p3}      my/path/my/file.txt
+        Should Be Equal    ${p4}      /path
+        Should Be Equal    ${p5}      /my/path2
```
pekkaklarck: 👍  for replacing the old way to show results with `Should Be Equal`. An alternative would be showing the result with a comment after the keyword call like we did with some [BuiltIn keywords](
https://github.com/robotframework/robotframework/blob/afe79a87a54b738ff85c8ba5c7a00d26fd4ca451/src/robot/libraries/BuiltIn.py#L264), but actually validating the return value makes it easier for us to test these examples.

A problem here is that on Windows you get `my\path` instead of `my/path` so the example won't pass. We could only add a note about that, but I think it's better to change the test so that it uses the built-in variable `${/}` that is either `/` or `\`, depending on the operating system. In practice that would mean using `my${/}path`instead of `my/path`.

src/robot/libraries/OperatingSystem.py
```diff
@@ -1108,63 +1341,90 @@ def log_environment_variables(
...
-        - @{p2} = ['/example', '/my/base/other']
-        - @{p3} = ['my/base/example/path', 'my/base/other', 'my/base/one/more']
+
+        ```robotframework
+        @{p1} =    Join Paths    base       example         other
+        @{p2} =    Join Paths    /my/base   /example        other
+        @{p3} =    Join Paths    my/base    example/path/   other    one/more
+        Should Be Equal    ${p1}    ${{['base/example', 'base/other']}}
+        Should Be Equal    ${p2}    ${{['/example', '/my/base/other']}}
+        Should Be Equal    ${p3}    ${{['my/base/example/path', 'my/base/other', 'my/base/one/more']}}
```
pekkaklarck: There's the same problem with `/` and `\` as with the previous example. In this case using `${/}` doesn't work, because it would be replaced with `\` on Windows before evaluating the inline Python expression and something like `'base\other'` isn't valid. That can be fixed easily by using raw strings like `r'base${/}other'`, but I'm slightly worried this gets rather technical. I don't see any other easy solution so perhaps that's fine.

src/robot/libraries/OperatingSystem.py
```diff
@@ -1108,63 +1341,90 @@ def log_environment_variables(
...
+        ${path3} =    Normalize Path    abc/./def//ghi
+        ${path4} =    Normalize Path    ~robot/stuff
+        Should Be Equal    ${path1}    abc
+        Should Be Equal    ${path2}    def
+        Should Be Equal    ${path3}    abc/def/ghi
+        Should Be Equal    ${path4}    /home/robot/stuff
+        ```
+
+        On Windows result would use `\\` instead of `/` and home directory
         would be different.
```
pekkaklarck, line 1464: Oh, this keyword already had a note about the `/` and `\` issue. I'm fine copying the note (minus the home directory part) to the above keywords instead of using `${/}`.

utest/libdoc/test_libdoc_api.py
```diff
@@ -47,6 +47,22 @@ def test_LibraryDocumentation(self):
...
+            for argument in keyword.args:
+                self.assertTrue(
+                    argument.doc,
+                    f"{keyword.name}: argument '{argument.name}' has no documentation",
+                )
+            data = keyword.to_dictionary()
+            if data["returnType"]:
+                self.assertTrue(
+                    data["returnDoc"],
+                    f"{keyword.name}: return value has no documentation",
```
pekkaklarck, line 63: Cool that you added a test for this! We probably should make this generic so that it tests all standard libraries. Leaving it like it is now is obviously fine in the context of this PR.

pekkaklarck (changes requested):
Looks really good in general! There are some recurring issues with examples, Args/Returns placement and linking to other libraries as well as some other smallish issues.

Are you @coyaSONG around so that you can take a look at my comments and fix the issues? If you are busy, I can also merge this PR in its current format and me or someone else can fix the problems.

pekkaklarck:
Did you @coyaSONG notice the review? It would be great if you could take a look at my comments. If you don't have time for that, you could just remove the draft status from this PR and we can make those adjustments ourselves.

coyaSONG:
@pekkaklarck The follow-up at 5deb3e4 addresses the review feedback: Markdown formatting and reference targets were corrected, `Args`/`Returns` were moved after the short descriptions, all keyword examples now include test-case headers and names, the encoding-error documentation link was added, and the cross-platform path examples now carry the Windows-result note. I also rechecked the current head and resolved the addressed threads. When convenient, could you please re-review? The PR remains draft until you confirm the documentation direction.

src/robot/libraries/OperatingSystem.py
```diff
@@ -40,183 +40,223 @@
...
-    | ``[a-z]``    | matches one character from the range in the bracket     |
-    | ``[!a-z]``   | matches one character not from the range in the bracket |
+    | Pattern | Matches |
+    | ------- | ------- |
+    | `*` | Any string, even an empty string |
+    | `?` | Any single character |
+    | `[chars]` | One character in the bracket |
+    | `[!chars]` | One character not in the bracket |
+    | `[a-z]` | One character from the range in the bracket |
+    | `[!a-z]` | One character not from the range in the bracket |
```
pekkaklarck, line 85: Columns should be aligned to make source look better.

src/robot/libraries/OperatingSystem.py
```diff
@@ -40,183 +40,223 @@
...
+        Create File          ${PATH}    Some text
+        File Should Exist    ${PATH}
+        Copy File            ${PATH}    ~/file.txt
+    ```
+
+    [Process]: https://robotframework.org/robotframework/latest/libraries/Process.html "Process library"
+    [Regexp Escape]: https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Regexp%20Escape "BuiltIn.Regexp Escape"
+    [Should Be Equal]: https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Should%20Be%20Equal "BuiltIn.Should Be Equal"
+    [Should Be Equal As Integers]: https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Should%20Be%20Equal%20As%20Integers "BuiltIn.Should Be Equal As Integers"
+    [Log]: https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Log "BuiltIn.Log"
```
pekkaklarck, line 142: With other libraries we have decided that when referring to external keywords, we only link to the library, not to individual keywords. As an example, see the Collections library.

src/robot/libraries/OperatingSystem.py
```diff
@@ -40,183 +40,223 @@
...
-        | ${rc} | ${stdout} =  | Run and Return RC and Output | /opt/script.sh 2>/tmp/stderr.txt |
-        | Should Be True       | ${rc} > 42      |
-        | Should Be Equal      | ${stdout}       | TEST PASSED |
-        | File Should Be Empty | /tmp/stderr.txt |
+
+        ```robotframework
+        *** Test Cases ***
+        Run And Return RC And Output
+            ${rc}    ${output} =             Run and Return RC and Output    ${CURDIR}${/}mytool
+            Should Be Equal As Integers      ${rc}    0
```
pekkaklarck, line 253: With other libraries we've decided to use
```
Should Be Equal    ${rc}    0    type=int
```
style instead of `Should Be Equal As Integers`. That works especially well in cases like this where arguments are aligned based on the longest keyword name.

src/robot/libraries/OperatingSystem.py
```diff
@@ -1114,63 +1367,100 @@ def log_environment_variables(
...
+
+        ```robotframework
+        *** Test Cases ***
+        Join Paths
+            @{p1} =    Join Paths    base       example         other
+            @{p2} =    Join Paths    /my/base   /example        other
+            @{p3} =    Join Paths    my/base    example/path/   other    one/more
+            Should Be Equal    ${p1}    ${{['base/example', 'base/other']}}
+            Should Be Equal    ${p2}    ${{['/example', '/my/base/other']}}
+            Should Be Equal    ${p3}    ${{['my/base/example/path', 'my/base/other', 'my/base/one/more']}}
```
pekkaklarck, line 1424: The inline Python evaluation syntax is somewhat complex. Using the newish type conversion functionality with `Should Be Equal` would allows this style that would probably be cleaner:

```
Should Be Equal    ${p1}    ['base/example', 'base/other']    type=list
```

pekkaklarck (approved):
Looks very good! I added some comments so please take a look at them.

pekkaklarck:
I took a pretty quick look at this and didn't see major problems. It would be great if you could fix the few issues I reported, but I think this could be merged already in this shape. I anyway need to take the final look locally so that I can also see how the generated docs look like.

Could you @Vinni82 take a look do you see something that could be fixed/enhanced based on the PRs you've worked with earlier?

pekkaklarck:
There has been no reaction to my latest review comments, but I consider this to be in good enough shape to be merged. @Vinni82 can do more enhancements after that, including fixing the issues I mentioned, and I'll do a final cleanup after that.

PR #5711, opened 2026-07-17
Libdoc: Convert standard type documentation to Markdown

Shaurya2k06:
## What

- Convert `STANDARD_TYPE_DOCS` in `standardtypes.py` from Robot documentation format to Markdown.
- Render standard type docs with `DocToHtml("MARKDOWN")` instead of `"ROBOT"` in `model.py`.

## Why

Part of #5709. Standard type docs shown for linked types (`int`, `list`, …) should use the same Markdown format as libraries.

## How

Links use `[text](url)`, inline code uses single backticks, and paragraphs avoid soft line breaks so Markdown HTML matches previous Robot rendering. Existing Libdoc datatype acceptance tests still pass without expectation changes.

## Testing

- `utest/run.py libdoc` — 47 passed
- `atest/run.py atest/robot/libdoc/datatypes_py-json.robot atest/robot/libdoc/datatypes_xml-json.robot` — 28 passed

### DIFF
--- a/atest/robot/libdoc/datatypes_py-xml.robot
+++ b/atest/robot/libdoc/datatypes_py-xml.robot
@@ -53,7 +53,7 @@ Standard
     ...    Any value is accepted. No conversion is done.
     DataType Standard Should Be    1
     ...    boolean
-    ...    Strings ``TRUE``, ``YES``, ``ON``, ``1`` and
+    ...    Strings `TRUE`, `YES`, `ON`, `1` and
     DataType Standard Should Be    6
     ...    Literal
     ...    Only specified values are accepted.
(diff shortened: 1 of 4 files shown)


pekkaklarck:
I noticed that longer paragraphs are not anymore split into multiple lines on the source. I also noticed the comment about Markdown preserving soft line breaks so there's clear motivation for it.

I tested that

```python
markdown.markdown("""
Hello,
world!
""")
```

yields `<p>Hello,\nworld!</p>` so the comment is correct. But does that really matter? Isn't the `\n` ignored when the content is rendered? Or do we use some CSS that preserves them and results don't look good?

If the `\n` doesn't affect the rendered output, I'd prefer longer paragraphs to be split. It makes diffs from possible future edits simpler (and also ought to make the diff of this PR simpler). If the `\n` actually affects the rendered output and the reason is styles, I'd prefer those styles to be adjusted and paragraphs to be split.

If the `\n` causes problems that cannot be fixed by adjusting styles, we should make sure we don't introduce them in the future either. There are two ways how to do that:

1. Add a unit test that validates that there are no `\n` inside paragraphs.
2. Chane Libdoc so that it removes the offending `\n`. It could either merge a paragraph that has been split to multiple lines into a single line before conversion, or it could remove `\n` from paragraphs after conversion.

Unfortunately neither of the above alternatives are easy to do reliably. Let's first verify does the `\n` actually cause problems and can possible problems be fixed by adjusting styles.

pekkaklarck:
Longer paragraphs still aren't split. This looks good otherwise so I'll merge this and look those paragraphs myself.

pekkaklarck:
I noticed that splitting paragraphs to multiple lines broke a unit test. The test compared earlier generated JSON spec with new spec and that failed, because the old spec didn't have newlines inside `<p>...</p>` in HTML converted docs but the new one had. That can be fixed by regenerating the JSON file.

PR #5712, opened 2026-07-17
Screenshot: Convert docs to Markdown and use Args/Returns syntax

Shaurya2k06:
## What

- Convert Screenshot library and keyword documentation to Markdown.
- Add structured `Args:` / `Returns:` documentation (and `Raises:` where useful, e.g. missing directory).

## Why

Screenshot-only part of #5709; small library suitable as an example.

## How

Declares `ROBOT_LIBRARY_DOC_FORMAT = "Markdown"`. Keyword examples use fenced `robotframework` blocks; internal links use `[Keyword]` / `[importing]` style.

## Testing

- Libdoc: `doc_format=MARKDOWN`, Args/Returns/Raises parsed for public keywords
- `ruff` / `black` clean on `Screenshot.py`
- Screenshot acceptance tests need a display (`require-screenshot`); behavior unchanged (docs only)

### DIFF
--- a/src/robot/libraries/Screenshot.py
+++ b/src/robot/libraries/Screenshot.py
@@ -44,41 +44,41 @@ class Screenshot:
     need to be installed separately. Taking screenshots also requires tests
     to be run with a physical or virtual display.
 
-    == Table of contents ==
+    ## Table of contents
 
     %TOC%
 
-    = Supported screenshot taking tools and modules =
+    ## Supported screenshot taking tools and modules
 
     How screenshots are taken depends on the operating system. On OSX
-    screenshots are taken using the built-in ``screencapture`` utility. On
+    screenshots are taken using the built-in `screencapture` utility. On
     other operating systems you need to have one of the following tools or
     Python modules installed. You can specify the tool/module to use when
-    `importing` the library. If no tool or module is specified, the first
+    [importing] the library. If no tool or module is specified, the first
     one found will be used.
 
-    - wxPython :: http://wxpython.org :: Generic Python GUI toolkit.
-    - PyGTK :: http://pygtk.org :: This module is available by default on most
+    - [wxPython](http://wxpython.org) — Generic Python GUI toolkit.
+    - [PyGTK](http://pygtk.org) — This module is available by default on most
       Linux distributions.
-    - Pillow :: http://python-pillow.github.io ::
+    - [Pillow](http://python-pillow.github.io) —
       Only works on Windows. Also the original PIL package is supported.
-    - Scrot :: http://en.wikipedia.org/wiki/Scrot :: Not used on Windows.
-      Install with ``apt-get install scrot`` or similar.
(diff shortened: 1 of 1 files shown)


src/robot/libraries/Screenshot.py
```diff
@@ -44,48 +44,49 @@ class Screenshot:
     need to be installed separately. Taking screenshots also requires tests
     to be run with a physical or virtual display.
 
-    == Table of contents ==
+    ## Table of contents
 
     %TOC%
 
-    = Supported screenshot taking tools and modules =
+    ## Supported screenshot taking tools and modules
```
pekkaklarck, line 51: I'm not sure should we use `#` or `##` level headers in introduction. I think we automatically convert `#` to `<h2>` (and `##` to `<h3>`) as part of Markdown conversion and thus `#` would be better. I'll test that locally myself and comment the issue about this.

src/robot/libraries/Screenshot.py
```diff
@@ -44,48 +44,49 @@ class Screenshot:
...
+    screenshots are taken using the built-in `screencapture` utility. On
     other operating systems you need to have one of the following tools or
     Python modules installed. You can specify the tool/module to use when
-    `importing` the library. If no tool or module is specified, the first
+    [importing] the library. If no tool or module is specified, the first
     one found will be used.
 
-    - wxPython :: http://wxpython.org :: Generic Python GUI toolkit.
-    - PyGTK :: http://pygtk.org :: This module is available by default on most
+    - [wxPython](http://wxpython.org) — Generic Python GUI toolkit.
```
pekkaklarck, line 60: Using links instead of URLs makes sense. I'm not sure is using a non-ASCII hyphen a good idea, though. Tools obviously should handle non-ASCII stuff just fine, and Libdoc itself certainly does, but there aren't much benefits compared to using just a normal `-` or `::` like the docs used to do. I'll test this locally and change the current hyphen to something else if needed.

pekkaklarck (approved):
Looks great!

As the inline comments explain, there are some design decisions to be made, but making those decisions requires seeing how things actually look like in HTML. I think it's best if I just merge this now, then test this locally and make possible adjustments afterwards. If there are decisions that affect also other libraries, I'll comment the original issues.

pekkaklarck:
I tested this locally and really liked the results. I did some fine-tuning that I'll commit shortly:
- Keep using `## Section` style with other sections. Using `# Section` would have meant that sections have same level as `Introduction`. We may want to consider changing this so that `# Section` would automatically be lower than automatically added `Introduction`, `Importing` and `Keywords`, though.
- Use `### Table of contents` to avoid adding the header to the actual table of contents.
- Change non-ASCII hyphen to a normal hyphen.
- Align comments in an example.
- Small fine tuning to some texts.

The first two points are something that are related to other libraries as well. I'll add a separate comment about them to #5709. As I wrote above, we may still consider adjusting header level handling in general.

PR #5713, opened 2026-07-17
Dialogs: Convert docs to Markdown and use Args/Returns syntax

Shaurya2k06:
## What

- Convert Dialogs library and keyword documentation to Markdown.
- Add structured `Args:` / `Returns:` / `Raises:` for public keywords.

## Why

Dialogs-only part of #5709.

## How

Module-level `ROBOT_LIBRARY_DOC_FORMAT = "Markdown"`. `Raises: RuntimeError` documented where Cancel fails the keyword; no redundant raises beyond that.

## Testing

- Libdoc: `MARKDOWN` with Args/Returns/Raises on keywords that need them
- `ruff` / `black` / `isort` clean

### DIFF
--- a/src/robot/libraries/Dialogs.py
+++ b/src/robot/libraries/Dialogs.py
@@ -15,11 +15,11 @@
 
 """A library providing dialogs for interacting with users.
 
-``Dialogs`` is Robot Framework's standard library that provides means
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Looks great! I noticed some other things that can be enhanced related to these docs, but I'll do those changes separately.

PR #5715, opened 2026-07-17
DateTime: Convert docs to Markdown and use Args/Returns syntax

Shaurya2k06:
## What

- Convert DateTime library and keyword documentation to Markdown.
- Add structured Args and Returns for all keywords.

## Why

DateTime-only part of #5709.

## Testing

- Libdoc: `MARKDOWN`, 8 keywords, all with Args and Returns docs

### DIFF
--- a/src/robot/libraries/DateTime.py
+++ b/src/robot/libraries/DateTime.py
@@ -15,139 +15,152 @@
 
 """A library for handling date and time values.
 
-``DateTime`` is a Robot Framework standard library that supports creating and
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Looks great again! Some examples seem to miss the `*** Test Cases ***` header that, somewhat unfortunately, is needed in order to get the example highlighted correctly. It might also make sense to use reference links instead of inline links to avoid long URLs messing up paragraphs. That makes sense especially if the same link target is needed multiple times.

I'll merge this and fix missing headers and other possible problems locally.

PR #5720, opened 2026-07-17
BuiltIn: Convert docs to Markdown and use Args/Returns syntax

Shaurya2k06:
## What

- Convert BuiltIn library and keyword documentation to Markdown.
- Add structured Args/Returns for keywords. Do not add redundant `Raises: AssertionError` on `Should *` keywords (failure condition is already stated in prose).

## Why

BuiltIn-only part of #5709 (largest library; last).

## Testing

- Libdoc: `MARKDOWN`, 107 keywords, 103 with Args, 39 with Returns

### DIFF
--- a/atest/robot/libdoc/cli.robot
+++ b/atest/robot/libdoc/cli.robot
@@ -47,7 +47,7 @@ Library argument matching resource extension when import fails
 
 Override name and version
     --name MyName --version 42 String ${OUTHTML}    HTML    MyName    42
-    -n MyName -v 42 -f xml BuiltIn ${OUTHTML}       XML     MyName    42
+    -n MyName -v 42 -f xml BuiltIn ${OUTHTML}       XML     MyName    42    docformat=MARKDOWN
 
 Missing destination subdirectory is created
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Previous PRs updating standard library docs had been pretty good, so I merged this one after only a cursory look at the changes. When going through the changes locally with an idea to make smallish enhancements like with earlier PRs, it turned out that there were _a lot_ of problems. Most of the argument docs were totally wrong, mostly likely due to AI hallucination, and needed to be rewritten. In addition to that, all tables had been converted to `robotframework` code blocks regardless did they contain Robot Framework examples or actual tabular data.

Fixing the docs took hours, but obviously it would have taken a lot of time for me to start from scratch as well. I also enhanced docs of various keywords at the same time which made the overall task bigger, but resulted with better docs in the end. I'll commit my changes shortly.

Shaurya2k06:
I'm so sorry about this! I only had a quick glance before pushing code, It's my mistake. I'll ensure better code quality from now on.

PR #5724, opened 2026-07-22
Fix invalid section header default-name error

### LINKED ISSUE #5723: `SectionHeader.from_params(Token.INVALID_HEADER)` raises `KeyError` when header name is not given
> bug-report format in `CONTRIBUTING.rst` ("Reporting bugs": version info, SSCCE, traceback).
> Correctness bug, not a security issue.
> `from_params KeyError`. No match; the `from_params` cousins #4359/#4357/#4670/#4729 are
> matches for `INVALID_HEADER`.

---

**Title:** `SectionHeader.from_params(Token.INVALID_HEADER)` raises `KeyError` because `zip` truncates 7 registered types against 6 default names

## Version information

- Robot Framework: reproduced on released **7.4.2** (pip) and source **7.5b2.dev1** @ `88058afbaadf36081544ab644bec99b246f404d5` (current `master` HEAD)
- Python: 3.14.6 (not version-sensitive)
- OS: platform-independent (pure model construction)

## What happens

`SectionHeader.handles_types` registers 7 token types. The parser routes any unrecognized
`*** ... ***` header to `INVALID_HEADER`, so a parsed invalid header has `.type == "INVALID HEADER"`.
When `SectionHeader.from_params` builds the default header name, it zips `handles_types`
against a local `names` tuple that lists only the 6 nameable sections. `zip` truncates to the
shorter sequence, `INVALID_HEADER` never enters the name map, and the subscript raises
`KeyError: 'INVALID HEADER'`. The 6 other registered header types construct with a default
name; `INVALID_HEADER` alone cannot.

This bites a model-rebuild workflow: the parser produces `INVALID_HEADER`, and feeding that
parsed `.type` back into `SectionHeader.from_params(header.type)` crashes.

## Steps to reproduce (SSCCE)

```python
from robot.parsing.lexer import Token
from robot.parsing.model.statements import SectionHeader

# all 6 nameable types build fine with a default name
for t in (Token.SETTING_HEADER, Token.VARIABLE_HEADER, Token.TESTCASE_HEADER,
          Token.TASK_HEADER, Token.KEYWORD_HEADER, Token.COMMENT_HEADER):
    SectionHeader.from_params(t)          # OK

# INVALID_HEADER is registered in handles_types, but:
SectionHeader.from_params(Token.INVALID_HEADER)     # -> KeyError: 'INVALID HEADER'
```

Traceback:

```
  File ".../robot/parsing/model/statements.py", line 334, in from_params
    name = dict(zip(cls.handles_types, names))[type]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
KeyError: 'INVALID HEADER'
```

Model-rebuild variant, where the parser itself produces the type that then crashes:

```python
import tempfile, os
from robot.api import get_model
p = tempfile.mktemp(suffix=".robot")
open(p, "w").write("*** Foobar ***\n*** Test Cases ***\nT\n    Log    hi\n")
m = get_model(p); os.unlink(p)
hdr = m.sections[0].header
print(hdr.type)                       # 'INVALID HEADER'
SectionHeader.from_params(hdr.type)   # -> KeyError: 'INVALID HEADER'
```

(`model.save()` round-trips an invalid header byte-identically, because `save()` serializes
existing tokens and never calls `from_params`. The crash is specific to the `from_params`
default-name path.) An explicit name works:
`SectionHeader.from_params(Token.INVALID_HEADER, name="Foobar")` → `"*** Foobar ***"`,
which isolates the fault to the default-name mapping.

## Expected vs actual

- **Expected:** `from_params` on any registered `handles_types` member either builds a header or raises a clear, typed error.
- **Actual:** an unhandled `KeyError: 'INVALID HEADER'` from an internal `dict(zip(...))` subscript on the 7th registered type.

## Root cause

`src/robot/parsing/model/statements.py`, `SectionHeader.from_params`, line 334:

```python
handles_types = (SETTING_HEADER, VARIABLE_HEADER, TESTCASE_HEADER,
                 TASK_HEADER, KEYWORD_HEADER, COMMENT_HEADER, INVALID_HEADER)  # 7
...
names = ("Settings", "Variables", "Test Cases", "Tasks", "Keywords", "Comments")  # 6
name = dict(zip(cls.handles_types, names))[type]    # zip truncates to 6 -> INVALID_HEADER missing
```

`zip` truncates to the 6-element `names` without a warning, so `INVALID_HEADER` has no default
name and the subscript raises `KeyError`.

## Suggested fix

Give `INVALID_HEADER` a default name (e.g. `"Invalid"` / `"Comments"`), or raise a clear
`ValueError`/`TypeError` stating that `INVALID_HEADER` requires an explicit `name=`. For example:

```python
names = ("Settings", "Variables", "Test Cases", "Tasks", "Keywords", "Comments", "Invalid")
```

so the two tuples stay the same length (a length assertion would guard against future drift).
Root cause at `src/robot/parsing/model/statements.py:334`.

---
*Found via property-based & differential bug-hunting, part of an effort to scale PBT (DepTyCheck-based) testing across the OSS ecosystem.*

*If this is intended / by-design: I'm really sorry, please just close it — no need to flag or ban me. I'm trying to scale property-based testing across the whole ecosystem and my publishing agents may have gotten this one wrong. I read every issue and follow up on each.*

Sanjays2402:
Closes #5723

`SectionHeader.from_params()` now raises a descriptive `ValueError` when an invalid header omits its required explicit name, instead of leaking an internal `KeyError`. The existing explicit-name path remains unchanged, and a regression covers the missing-name case.

### DIFF
--- a/src/robot/parsing/model/statements.py
+++ b/src/robot/parsing/model/statements.py
@@ -323,6 +323,8 @@ class SectionHeader(Statement):
         eol: str = EOL,
     ) -> "SectionHeader":
         if not name:
+            if type == Token.INVALID_HEADER:
+                raise ValueError("Invalid header requires an explicit name.")
             names = (
                 "Settings",
                 "Variables",
--- a/utest/parsing/test_statements.py
+++ b/utest/parsing/test_statements.py
@@ -111,6 +111,12 @@ class TestCreateStatementsFromParams(unittest.TestCase):
                 tokens, SectionHeader, type=token_type, name=f"*** {name} ***"
             )
 
+    def test_invalid_section_header_requires_name(self):
+        with self.assertRaisesRegex(
+            ValueError, "Invalid header requires an explicit name"
+        ):
+            SectionHeader.from_params(Token.INVALID_HEADER)
+
     def test_SuiteSetup(self):
         # Suite Setup    Setup Keyword    ${arg1}    ${arg2}
         tokens = [


pekkaklarck:
Thanks for the PR. I think just giving name `*** Invalid ***` to invalid headers without names is slightly better than raising a better exception. Most importantly that's easier to implement and test. I already have that fix done locally so I'll just close this one.

Issue #5726, opened 2026-07-28
Change Finnish translation of "Tags" from "Tagit" to "Tunnisteet"


asimell:
"Tagit" is not a proper Finnish word. It's a word in spoken Finnish or slang, but the proper native translation would be "Tunnisteet".

pekkaklarck:
I agree "Tunnisteet" is much better than "Tagit". Luckily we have just implemented support to deprecate settings and headers (#5668) so we can easily change the term without breaking existing tests/tasks using "Tagit". In RF 7.5 both terms will work without a warning, but in RF 7.6 (or RF 8.0) we'll start emitting deprecation warnings with the old term.

PR #5727, opened 2026-07-28
Change Finnish translation of "Tags"

asimell:
Implementation for #5726.

"Tagit" is not proper Finnish. This PR deprecates Finnish translations for Tags, Keyword Tags, Test Tags, and Task Tags to use the word "Tunnisteet", which is the proper Finnish word.

### DIFF
--- a/src/robot/conf/languages.py
+++ b/src/robot/conf/languages.py
@@ -598,10 +598,10 @@ class Fi(Language):
     task_template_setting = "Tehtävän Malli"
     test_timeout_setting = "Testin Aikaraja"
     task_timeout_setting = "Tehtävän Aikaraja"
-    test_tags_setting = "Testin Tagit"
-    task_tags_setting = "Tehtävän Tagit"
-    keyword_tags_setting = "Avainsanan Tagit"
-    tags_setting = "Tagit"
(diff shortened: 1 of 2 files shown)


src/robot/conf/languages.py
```diff
@@ -614,6 +614,11 @@ class Fi(Language):
     but_prefixes = ["Mutta"]
     true_strings = ["Tosi", "Kyllä", "Päällä"]
     false_strings = ["Epätosi", "Ei", "Pois"]
+    deprecations = {"Tagit": ("Tunnisteet", "7.5"),
+                    "Avainsanan Tagit": ("Avainsanan Tunnisteet", "7.5"),
+                    "Testin Tagit": ("Testin Tunnisteet", "7.5"),
+                    "Tehtävän Tagit": ("Tehtävän Tunnisteet", "7.5")
+                    }
```
pekkaklarck, line 622: This doesn't follow our code formatting guidelines. Easiest way to fix it is running `invoke format`. Can you @asimell still do that or should I do it after merging this? And yes, we should somehow automate formatting validation.

pekkaklarck: After a second thought, I'll fix this myself to get the PR and the related issue closed.

PR #5729, opened 2026-08-01
Add Add Months To Date and Subtract Months From Date keywords to the DateTime library

### LINKED ISSUE #4890: Feature Request :Add option to add months in Current date
Hi, 
In Robot currently, we don't have the option to add months to the current data in the DateTime library. 
We can now add Hours, Days, and, Years using Add Time to data.

 ### ${one_month_later}    Add Time To Date    ${current_date}    1 months

it we try to add the above it will not recognize as a keyword.

dilrukshax:
Closes #4890

## Summary

The `DateTime` standard library supports adding/subtracting fixed-duration
time intervals via `Add Time To Date` / `Subtract Time From Date`, but it
drops day-precision when one wants to add calendar months - which have a
variable number of days. Issue #4890 requests an `Add Months To Date`
keyword so users can write things like:

\`\`\`robot
\${one_month_later}=    Add Months To Date    \${current_date}    1
\`\`\`

This change adds:

- `Date.add_months(months)` helper, performing calendar-month arithmetic
  using Python's stdlib `calendar.monthrange`, no new dependencies.
  The day-of-month is clamped to the last valid day of the target month
  - e.g. Jan 31 + 1m = Feb 29 in a leap year (or Feb 28 otherwise);
  Mar 31 + 1m = Apr 30. Hours/minutes/seconds/milliseconds preserved.

- Public keyword `Add Months To Date` and symmetric
  `Subtract Months From Date` exported via `__all__` and present in the
  libdoc index. They follow the same argument conventions as
  `Add Time To Date` (date, amount, result_format, exclude_millis,
  date_format).

- Acceptance test data + runner suites:
  - `atest/testdata/standard_libraries/datetime/add_months_to_date.robot`
  - `atest/testdata/standard_libraries/datetime/subtract_months_from_date.robot`
  - `atest/robot/standard_libraries/datetime/{add,subtract}_months_to_date.robot`

## Why a calendar keyword and not just expand `Time`

Months are not a fixed-duration time interval. The existing `Time` class
is built on `datetime.timedelta`, which represents a fixed number of
seconds. Modeling "1 month" as 30.44 days would not give users the
calendar semantics they expect (e.g. Jan 31 + 1m should reach end of Feb,
not Mar 2). The new keyword intentionally stays calendar-aware and
end-of-month clamping keeps days monotonic.

## Verification

- All 17 new acceptance tests pass with `python -m robot --pythonpath src
  atest/testdata/standard_libraries/datetime/{add,subtract}_months_to_date.robot`
  on CPython 3.14.
- Existing `add_time_to_date`, `subtract_time_from_date`,
  `get_current_date`, `convert_date_result_format`, `convert_date_input_format`,
  `convert_time_*` suites are unchanged. The single unrelated
  `get_current_date.robot` "Invalid time zone" failure reproduces on
  master with the same Python version and predates this change.
- Followed CONTRIBUTING.rst: ruff and black run on the modified file;
  no new ruff findings are introduced (the remaining DTZ*** and N999
  warnings are pre-existing project-wide rules flagged on master too).
- No third-party dependencies added - the implementation uses
  `calendar.monthrange` from the standard library.

Follows style guidance from CONTRIBUTING.rst. Happy to adjust naming or
split end-of-month clamping into its own configurable argument if
maintainers prefer.

### DIFF
--- a/src/robot/libraries/DateTime.py
+++ b/src/robot/libraries/DateTime.py
@@ -304,24 +304,30 @@ Additionally, helper classes ``Date`` and ``Time`` can be used directly:
 |     # ...
 """
 
+import calendar
 import datetime
 import sys
 import time
 from typing import Literal, Union
 
 from robot.utils import (
-    elapsed_time_to_string, secs_to_timestr, timestr_to_secs, type_name
+    elapsed_time_to_string,
+    secs_to_timestr,
+    timestr_to_secs,
+    type_name,
 )
 from robot.version import get_version
 
 __version__ = get_version()
 __all__ = [
+    "add_months_to_date",
     "add_time_to_date",
     "add_time_to_time",
     "convert_date",
     "convert_time",
     "get_current_date",
     "subtract_date_from_date",
+    "subtract_months_from_date",
     "subtract_time_from_date",
     "subtract_time_from_time",
 ]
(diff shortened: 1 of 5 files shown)


src/robot/libraries/DateTime.py
```diff
@@ -519,6 +525,82 @@ def subtract_time_from_date(
...
+    | ${date} =       | Add Months To Date | 2024-01-31 12:05:03.111 | 1 |
+    | Should Be Equal | ${date}            | 2024-02-29 12:05:03.111 |
+    | ${date} =       | Add Months To Date | 2024-06-15 12:05:03.111 | -2 |
+    | Should Be Equal | ${date}            | 2024-04-15 12:05:03.111 |
+    """
+    return (
+        Date(date, date_format)
+        .add_months(int(months))
+        .convert(result_format, millis=not exclude_millis)
+    )
```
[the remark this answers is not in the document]
pekkaklarck: It's safe to simply replace `int(months)` with `months`. Our automatic argument conversion makes sure the value is an integer when used with Robot. If someone uses this programmatically, they should just pass an integer as the type hint suggest.

src/robot/libraries/DateTime.py
```diff
@@ -304,24 +304,30 @@
...
 import sys
 import time
 from typing import Literal, Union
 
 from robot.utils import (
-    elapsed_time_to_string, secs_to_timestr, timestr_to_secs, type_name
+    elapsed_time_to_string,
+    secs_to_timestr,
+    timestr_to_secs,
+    type_name,
```
pekkaklarck, line 317: To avoid wasting too much vertical space, we don't format normal imports so that each one is on its own line. That is handled automatically if you run `invoke format`.

src/robot/libraries/DateTime.py
pekkaklarck, line 564: The current method chaining is pretty complicated and I'm afraid automatic formatter (`invoke format`) will mess it up. If that happens, change the code to this is probably the best solution:
```python
date = Date(date, date_format).add_months(months)
return date.convert(result_format, millis=not exclude_millis)
```

src/robot/libraries/DateTime.py
```diff
@@ -519,6 +525,82 @@ def subtract_time_from_date(
...
+    | ${date} =       | Subtract Months From Date | 2024-02-29 12:05:03.111 | 1 |
+    | Should Be Equal | ${date}                  | 2024-01-29 12:05:03.111 |
+    | ${date} =       | Subtract Months From Date | 2024-04-15 12:05:03.111 | -2 |
+    | Should Be Equal | ${date}                  | 2024-06-15 12:05:03.111 |
+    """
+    return (
+        Date(date, date_format)
+        .add_months(-int(months))
+        .convert(result_format, millis=not exclude_millis)
+    )
```
pekkaklarck, line 601: As already noted above, the `int` conversion is unnecessary and splitting method chaining can be a good idea.

src/robot/libraries/DateTime.py
```diff
@@ -669,6 +751,19 @@ def __sub__(self, other: "Date | Time") -> "Date | Time":
             f"Can only subtract Date or Time from Date, got {type_name(other)}."
         )
 
+    def add_months(self, months: int) -> "Date":
+        if not isinstance(months, int):
+            raise TypeError(f"Months must be an integer, got {type_name(months)}.")
```
pekkaklarck, line 756: We can expect the value to be `int`. No need for extra type validation.

src/robot/libraries/DateTime.py
```diff
@@ -669,6 +751,19 @@ def __sub__(self, other: "Date | Time") -> "Date | Time":
...
+        dt = self.datetime
+        total = dt.year * 12 + (dt.month - 1) + months
+        year, zero_based_month = divmod(total, 12)
+        month = zero_based_month + 1
```
pekkaklarck, line 760: Calculating total months feels unnecessary and how month becoming zero is prevented is somewhat complicated. I believe this approach would be a bit cleaner:
```python
year_increment, month = divmod(dt.month + months, 12)
year = dt.year + year_increment
if month == 0:
    year -= 1
    month = 12
```
Notice that I haven't tested the above thoroughly, so I may have missed some corner cases.

src/robot/libraries/DateTime.py
```diff
@@ -669,6 +751,19 @@ def __sub__(self, other: "Date | Time") -> "Date | Time":
...
+        # Clamp the day to the last valid day of the target month so that
+        # adding a month to e.g. Jan 31 yields Feb 28/29 (end-of-month clamp).
+        last_day = calendar.monthrange(year, month)[1]
+        day = min(dt.day, last_day)
```
pekkaklarck, line 764: I agree that best way to handle the situation where the day is too big for the month is picking up the last valid day. I was thinking about catching ValueError caused by an invalid day and decrementing the day by one until it is valid, but getting the last day with the strangely named `calendar.monthrange` is much better. As a style issue, I'd probably use
```python
_, last_day = calendar.monthrane(year, month)
return Data(..., day=min(dt.day, last_day))
```
I'm also not sure is the comment necessary. Perhaps `add_months` should get a docstring where this behavior is explained instead.

pekkaklarck (changes requested):
Implementation looks good to me, but there are some things that can still be improved. Keyword docs should also get `New in Robot Framework 7.5.` notes at the end.

It would also be great if you could convert documentation to Markdown and take the new argument documentation syntax into use. We are doing that with all standard libraries (see #5709) and there's already a PR about converting existing DateTime docs that I try to get merged still today. If you don't have time for doc updates, we can do that separately after this PR has been merged.

I didn't look at the new test too thoroughly, but I'll do that once the above things are handled.

pekkaklarck:
Why did you close this @dilrukshax? The overall approach looked fine.

I've been thinking it could be better to add months, and possibly years, support to `Add Time To Date` and `Subtract Time From Date` keywords instead of adding new keywords. I'll comment the original issue about that.

Issue #5728, opened 2026-07-28
Libdoc: Long library names are not shown correctly on mobile


aaltat:
If library name is long, then in mobile view the language selection button is not correctly aligned the right side. Instead it moved more center. Happens with RF 7.4 and 7.5b. 
  
<img width="744" height="1026" alt="Image" src="https://github.com/user-attachments/assets/5d7c439e-dbc5-4db8-8685-980cf6351048" />

pekkaklarck:
Doesn't look too good. This is probably related to the keyword menu that's there only with small screens. Do you @aaltat have bandwidth to look at fixing this? One possibility would be integrating language selection with the other menu, but just moving the language selector somewhere else is probably simpler. I guess it could be in the lower right corner in mobile and possibly also in normal view.

aaltat:
Yes it does not look good. I guess mobile view is that popular, because no-one has complained and SL is popular library. In any case it is bug that should be fixed. 

Currently I am busy with Playwright update with Browser library, https://github.com/MarketSquare/robotframework-browser/pull/5074 , which is growing somewhat bigger thing, because we changing how BrowserBatteries is being build. Moving away from https://github.com/yao-pkg/pkg way build/package and actually putting node.exe inside of the wheel. This aligns how Playwright currently does it and moves us away from using some Playwright internal functionality which changes now and then. I assume that it will take some time. 

I will return work next week and I need to see what is going there.  Perhaps putting time with this one is possible at the same time when we do few of the improvements ideas for the Google doc string.

pekkaklarck:
We just need to decide is this important enough to be fixed in RF 7.5 or can it wait for the next release. This is a UI bug so there are no backwards compatibility problems or anything else that would prevent fixing this in a RF 7.5.1.

pekkaklarck:
You had this fixed already, right?

Issue #5604, opened 2026-01-21
Libdoc: Dedicated syntax for argument and return value documentation


pekkaklarck:
Libdoc shows keyword arguments separately and lists their names, "kind" (positional-only, varargs, ...), possible type (`str`, `int`, ...) and possible default value. There is, however, no way to include argument documentation in this listing. Instead, arguments need to be documented as part of the normal keyword documentation using whatever approach found usable. Libdoc also shows return value type, but it does not support documentation either.

We should change this and make it possible to specify argument and return value documentation explicitly so that it could be listed along with other argument and return value related information. This syntax should also play well with other tools such as IDEs and API doc generators. In practice we thus should use syntax that's used for arguments with Python normally. Unfortunately there is no official standard for that and different projects have used different approaches. The three most commonly used argument documentation styles are:

1. The [Sphinx style](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html). This is what we currently use with Robot's API docs. In that usage it has worked fine, but the syntax is pretty complicate and verbose. It also doesn't render nicely as Markdown that we plan to support with Libdoc (#5304).
2. The [NumPy style](https://numpydoc.readthedocs.io/en/latest/format.html##parameters). This style is used primarily in the scientific Python circles. but nobody would prevent us from using it. It's simpler than the Sphinx style, but it still requires quite a bit of markers and doesn't work with with Markdown.
3. The  [Google style](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods). This is the simplest format and works very well also in source code. En example using this style is below.

```python
def example(first: int, second: float) -> float:
    """Example keyword.

    Args:
        first: The doc for the first parameter.
        second: Also the second parameter has documentation.
          It is split to multiple lines.

    Return:
        The sum of the given arguments.
    """
    return a + b
```

The Google style is my clear favorite, but also it has some problems:

1. Also this syntax doesn't work too well with standard Markdown. For example, the block of tests starting with `Args:` is parsed as a single paragraph so you end up with a paragraph starting with  `Args: second: The doc for the first ...`. The situation gets better if you add an empty line after `Args:`. In that case rest of the block is considered a code block and rendered as preformatted text. Not great but massively better than alternatives.
[...]

PR #5734, opened 2026-08-07
Libdoc with new style for arguments

### LINKED ISSUE #5733: Libdoc: Add copy button to Markdown code blocks
If there code block in markdown which shows an example of test or keyword, it would be useful to add copy button to the code block so that users have an easy to copy the example. This is pretty common in many code examples in the web and is nice feature to the libdoc.

aaltat:
Fixes #5733 #5728
Part of #5604

Arguments now look like: 
image

And return and raises information:
image

Also in the previous implementation, the SchemathesisLibary looks better: 
image

I can even now consider keeping the `session` argument documentation as it is: 
image


emanlove:
I like a lot of what is here. Think it covers René's concern about unused space. I also think it puts this in a good spot for a release now and give people a chance to try it out and comment about it.

pekkaklarck:
I personally like the current styles a bit more that this. The current table has some wasted space, but I consider is easy to understand compared to this version.

I find it a bit strange that the data is kind of formatted as in a table but kind of as in a list. I would think using either a normal table, like we now have, or a list, like [this mkdocstring example does](https://mkdocstrings.github.io/reference/api/#mkdocstrings.MkdocstringsExtension), would be simpler. That said, I see problems with the table approach and I also see that complex type hints and long default values would probably cause problems with the list approach.

The above said, I'm not really against these styles, especially when the comments here have been positive. I'd like to hear opinions from a bit wider audience, though. Could you @aaltat and @Snooz82 create examples with the current styles and with these PR styles so that people could test them? It would be best if both versions would be somewhere available so that people could experiment with the real docs instead of just seeing images. Using a real library like SchemathesisLibary or/and BuiltIn would be best. If majority of the people like the new styles, and we got at least few votes, I'm fine with using these styles in RF 7.5. We can, and probably should, do a bigger Libdoc HTML update in the future anyway.

pekkaklarck:
If we go with this design, I'd somewhat strongly want to remove the equal sign from the default value. I didn't like it too much with the previous design either, but back then the default value was after the argument name and having `name = default <type>` was somewhat close to how this information is shown in the source. Now that the default is after the type, which I consider a good change in itself, the equal sign is just confusing.

aaltat:
Closing, repleaced by #5741 #5740 and #5739

PR #5735, opened 2026-08-07
markdown as living documentation

JFoederer:
* Extended the User Guide with information on how to keep Robot Framework syntax out of rendered md files for non-automation users.
* Test case extended to cover this use case.


pekkaklarck:
Thanks for the PR! This is a good use case for the new support to embed Robot data to Markdown files (#5603).

PR #5741, opened 2026-08-13
Libdoc: New table styles, theme toggle, doc clamping and tag filtering

aaltat:
Rework arguments, returns and raises into a single shared table style, and refresh the surrounding layout, spacing and colours.

Also included:
- theme toggle button; the choice is stored in localStorage and can be overridden with a ?theme= query parameter
- long argument docs are clamped with more.../less controls
- tags link to a keyword view filtered by that tag
- testdata.ts (development-only preview data) refreshed


pekkaklarck:
This has been replaced with newer PRs.

PR #5743, opened 2026-08-13
feat: added testcase metadata to report

febb0e:
This pull request enhances the HTML report by adding support for showing test case metadata in the details table. The changes help that metadata is shown only when present, and that sorting and column toogles work correctly with the new metadata column. Test automation and validation scripts are also updated.

**Rebot HTML report improvements:**

* Added a new "Metadata" column to the test details table, which is displayed only if at least one test case contains metadata.
* Updated JavaScript logic in `report.html` to detect the presence of metadata, render the column, and adjust table sorting.
* Added CSS rules for the new `.details-col-metadata` class to ensure proper column width and alignment.

**Test and validation updates:**

* Fixed a typo in test keywords and extended validation to check for metadata in all relevant HTML outputs, including those generated by Rebot.

### DIFF
--- a/atest/robot/parsing/test_metadata.robot
+++ b/atest/robot/parsing/test_metadata.robot
@@ -13,14 +13,16 @@ Metadata is included in log
     Validate metadata in HTML    ${OUTDIR}/log.html
 
 Metadata is included in report
-    Validate metadata in HTML    ${OUTDIR}/report.html
+    Validate metadata in HTML             ${OUTDIR}/report.html
+    Validate metadata column in report    ${OUTDIR}/report.html
 
 Metadata is preserved by Rebot
     Copy Previous Outfile
     Run Rebot    --log rebot-log.html --report rebot-report.html    ${OUTFILE COPY}
     Validate metadata in model
-    Validate metadata in HTML    ${OUTDIR}/rebot-log.html
-    Validate metadata in HTML    ${OUTDIR}/rebot-report.html
+    Validate metadata in HTML             ${OUTDIR}/rebot-log.html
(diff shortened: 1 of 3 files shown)


pekkaklarck:
Are you sure hiding the column if no test has metadata is worth the extra complexity? I believe it would be better to always show it because:
- We also show other colums regardless are they empty not. We could decide to change that, but then we should do that with all columns as part of a separate issue.
- We already have a mechanism for persistently collapsing columns.
- I'm slightly worried about the performance impact of going through all tests before rendering the table.
- We are planning to do larger log and report redesign in the future. We'll certainly revisit how all this info is shown then.

febb0e:
Hey @pekkaklarck, I totally agree, the conditional rendering is not worth it. Reverted that in this commit:

- The `Metadata` column is now always shown, exactly like `Documentation` and `Tags`. Empty metadata renders an empty cell.
- Dropped the `anyTestHasMetadata()` pre-render scan, so your performance concern is gone: `drawTestDetailsTable()`, `renderTestDetailsHeader()`, `renderTestDetails()` and `renderTestDetailsRows()` are now identical to master again.
- The column uses the existing persistent collapse mechanism so it can be hidden if it's in the way.

While checking this I noticed the existing HTML assertions don't actually verify anything about the column — the details table is rendered client-side, so `report.html` only contains embedded data, and `File Should Contain <Metadata>` passes without existing metadata, so I made the report tests check the templates.

pekkaklarck (approved):
Looks good to me! I need to test this locally to see that everything works, but I believe there are no issues and merge this and test later.

PR #5745, opened 2026-08-13
feat: added &{TEST METADATA} variable and Set Test Metadata

febb0e:
This pull request adds the built-in variable `${TEST_METADATA}` and the keywords `Set Test Metadata` and `Set Task Metadata` to Robot Framework. These changes allow users to set and access test-specific metadata dynamically during test execution.

The most important changes are:

### New Features

* Added the `${TEST_METADATA}` variable, which exposes the current test's metadata as a dictionary and is updated at test start and when metadata changes.
* Introduced the `Set Test Metadata` and `Set Task Metadata` keywords in `BuiltIn.py` to allow setting and appending metadata for the current test/task, with options for case/space/underscore-insensitive names and custom separators.

### Documentation

* Documented the new `${TEST_METADATA}` variable and its usage in the user guide, including its availability and the new keywords.

### DIFF
--- a/doc/userguide/src/CreatingTestData/Variables.rst
+++ b/doc/userguide/src/CreatingTestData/Variables.rst
@@ -1594,6 +1594,10 @@ can be changed dynamically using keywords from the `BuiltIn`_ library.
    |                        | dynamically using using :name:`Set Test Documentation`|            |
    |                        | keyword.                                              |            |
    +------------------------+-------------------------------------------------------+------------+
+   | &{TEST METADATA}       | The free metadata of the current test case. Can be    | Test case  |
+   |                        | set using :name:`Set Test Metadata` keyword.          |            |
+   |                        | New in Robot Framework 7.5.                           |            |
+   +------------------------+-------------------------------------------------------+------------+
    | ${TEST STATUS}         | The status of the current test case, either PASS or   | `Test      |
    |                        | FAIL.                                                 | teardown`_ |
    +------------------------+-------------------------------------------------------+------------+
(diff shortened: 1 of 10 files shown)


src/robot/libraries/BuiltIn.py
```diff
@@ -4734,6 +4734,75 @@ def set_test_documentation(
...
+        Args:
+            name: The name of the metadata to set.
+            value: The metadata value.
+            append: If true, the given `value` is added after the earlier
+              value instead of overwriting it.
+            separator: The separator to use between the old and the new
+              value when appending.
+
+        The metadata of the current test is available as a built-in variable
+        `${TEST METADATA}` in a Python dictionary. Notice that modifying this
```
pekkaklarck, line 4755: This could mention that the dict is dot-accessible like `${TEST METADATA.Example}`. Should also check what the Suite variant says about this.

doc/userguide/src/CreatingTestData/Variables.rst
```diff
@@ -1594,6 +1594,10 @@ can be changed dynamically using keywords from the `BuiltIn`_ library.
    |                        | dynamically using using :name:`Set Test Documentation`|            |
    |                        | keyword.                                              |            |
    +------------------------+-------------------------------------------------------+------------+
+   | &{TEST METADATA}       | The free metadata of the current test case. Can be    | Test case  |
+   |                        | set using :name:`Set Test Metadata` keyword.          |            |
```
pekkaklarck, line 1598: This implies that the keyword is the only way to set metadata. Changing "can be set" to "can be set dynamically" like with other similar variables would be better.

It would be best to also link to the section discussing test metadata, but for consistency reasons it should be done also with others. Not really related to this feature in general, but enhancing docs is obviously always good.

pekkaklarck (approved):
Lools great! I added two small comments about doc enhancements, but they don't prevent merging this. If you think those doc enhancements make sense, please create a new PR.

PR #5751, opened 2026-08-15
Process: Updated and splitted test cases names in new libdoc

Vinni82:


src/robot/libraries/Process.py
```diff
@@ -56,15 +56,21 @@ class ProcessResult:
...
+        Should Be Equal    ${result.rc}    0    type=int
+
+    Verify Output
+        Should Match    ${result.stdout}    Some t?xt*
+        Should Be Empty    ${result.stderr}
+
+    Verify Output Files
+        ${stdout}=    Get File    ${result.stdout_path}
+        Should Be Equal    ${stdout}    ${result.stdout}
+        File Should Be Empty    ${result.stderr_path}
```
pekkaklarck, line 72: Splitting the example to smaller pieces is a good idea, but having one test in the beginning to work as a setup is confusing and not a good practice in testing in general. Using a suite setup would be better, I'm afraid also that would add more complexity and thus problems than the smaller tests have benefits. I believe keeping the original format is better.

One thing to change in the example is alignment. Either all values should be aligned consistently to columns or, possibly better, the separator between keywords and arguments should be normal four spaces.

src/robot/libraries/Process.py
```diff
@@ -654,14 +664,22 @@ class Process:
...
+
```
pekkaklarck: This example has exactly the same problems as the one I commented earlier.

src/robot/libraries/Process.py
```diff
@@ -678,14 +696,21 @@ class Process:
...
+    Run Process
         ${result} =    Run Process    ${CURDIR}/script.py
         Should Not Contain    ${result.stdout}    FAIL
+
+    Terminate Process
         Terminate Process    ${handle}
+
+    Wait Process
         ${result} =    Wait For Process    First
         Should Be Equal As Integers    ${result.rc}    0
```
pekkaklarck, line 708: This test cannot be split like this because `${handle}` is needed in multiple tests. It could be made suite global variable, but I believe having just one test is better. That's probably better also in general, because the idea with the example is demonstrating how to run processes in parallel.

src/robot/libraries/Process.py
```diff
@@ -1334,19 +1366,23 @@ def get_process_result(
...
+            Should Be Empty    ${result.stderr}
+
+        Get one attribute
+            ${stdout}=    Get Process Result    ${PROCESS_ALIAS}    stdout=true
+            Should Be Equal    ${stdout}    Hello, world!
+
+        Multiple attributes
+            ${stdout}    ${stderr}=    Get Process Result    ${PROCESS_ALIAS}    stdout=yes    stderr=yes
+            Should Be Equal    ${stdout}    Hello, world!
+            Should Be Empty    ${stderr}
```
pekkaklarck, line 1380: This example is pretty long, so splitting it to smaller pieces has more benefits than there were with earlier examples. I still don't like the first test acting as a setup, though. I see two options:
- Use suite setup.
- Return to earlier style with comments instead of separate test cases. Empty lines could be added above comments in that case.

src/robot/libraries/Process.py
```diff
@@ -1457,7 +1493,7 @@ def join_command_line(self, *command: "Sequence[str] | str") -> str:
 
         ```robotframework
         *** Test Cases ***
-        Join command line
+        Build Command Line
```
pekkaklarck: I think the old name was better here and also with the following two examples. Using the name of the demonstrated keyword as the example name makes sense in general.

pekkaklarck (changes requested):
Looks pretty good. See the comments for things that could still be changed.

src/robot/libraries/Process.py
Vinni82: I removed the VAR section and seperated testcases. Made 1 testcase with comments above certain steps.
Reworked the whole document and kept normal four spaces seperator distance and also consistent with the variables a space e.g. ${stdout} =
There were some like this, ${stdout}=, I should have been consistent and reworked all the variables.

src/robot/libraries/Process.py
Vinni82: See previous comment

src/robot/libraries/Process.py
Vinni82: Created a single testcase with seperate sections with a comment above it e.g. # Wait Process followed by the two keywords.

src/robot/libraries/Process.py
Vinni82: Create one testcase with seperate comments and a empty line above the comment.

src/robot/libraries/Process.py
Vinni82: Changed the word Build -> Join for both testcases. 
Join command line
Join Command From List

PR #5753, opened 2026-08-17
feat: libdoc console output as Markdown

### LINKED ISSUE #5750: libdow show argument should return documentation in Markdown format
To enable fixing #5749, the libdoc `show` argument output to the console should be changed to also return library documentation in Markdown format. It should contain arguments, return value, raises information and tags. The return value and raises information is missing from the current output. Also other parts of the library documentation should be converted to Markdown.

aaltat:
Fixes: #5750


pekkaklarck:
Looks good to me. Code can be enhanced, but let's do that when we add Markdown output file support.

We also need to take a look should the User Guide be updated. I can do that.

PR #5758, opened 2026-08-19
Accept integer `maxerrorlines` via the programmatic API (#5113)

### LINKED ISSUE #5113: `maxerrorlines` does not appear to work via `robot.running.model.TestSuite.run`
I have some code that uses `robot.running.TestSuiteBuilder`:
```python
        builder = TestSuiteBuilder()
        suite = builder.build("some_suite.robot")
        # run
        suite.run(output="out.xml", loglevel="TRACE", maxerrorlines=None)
```
yet I am seeing the:
 `    [ Message content over the limit has been removed. ]`
in the test-case-level error message. 

And setting this to any `str` value (`int` throws `AttributeError: 'int' object has no attribute 'upper'`) seems to have no effect on the message length (I've tried `maxerrorlines="10"`, `="999"` as well). 

`loglevel` ***is*** working.

---

However, if I use `subprocess` to build a CLI command using `--maxerrorlines NONE"`, the error message is formatted properly.

---

Environment:
```
pip list | grep robot
robotframework                7.0

python --version
Python 3.10.14

cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

jaideeppyne:
## What

Fixes #5113. Passing `maxerrorlines` as an **int** through the programmatic API crashes the whole run:

```python
import robot
robot.run('tests.robot', maxerrorlines=10)
# [ERROR] Unexpected error: AttributeError: 'int' object has no attribute 'upper'
# → rc=255, the run is aborted
```

## Why

`RobotSettings._process_max_error_lines` checks the value like a string:

```python
def _process_max_error_lines(self, value):
    if not value or value.upper() == "NONE":   # assumes str
        return None
    value = self._convert_to_integer("MaxErrorLines", value)
    ...
```

That's fine for the CLI, where every option arrives as a string (`--maxerrorlines 30`). But the programmatic entry points (`robot.run(..., maxerrorlines=10)`, `TestSuite.run(maxerrorlines=10)`) pass **native Python types**, and an `int` — the natural type for a line count — hits `value.upper()` and raises `AttributeError`.

This is inconsistent with the sibling integer option `MaxAssignLength` (`_process_max_assign_length`), which accepts ints fine because it doesn't call `.upper()`.

## How

Guard the string-only `"NONE"` check with `isinstance(value, str)`, so an int flows straight to `_convert_to_integer`:

```diff
-        if not value or value.upper() == "NONE":
+        if not value or (isinstance(value, str) and value.upper() == "NONE"):
```

All existing string behaviour is preserved, and invalid values (int or string, e.g. `5`) still raise a clear `DataError` instead of `AttributeError`.

Verified on current `master`:

| input | before | after |
|---|---|---|
| `10` (int) | ❌ AttributeError, rc=255 | ✅ `10` |
| `'15'` | ✅ `15` | ✅ `15` |
| `None` | ✅ `None` | ✅ `None` |
| `'NONE'` | ✅ `None` | ✅ `None` |
| `5` (int) | ❌ AttributeError | ✅ `DataError` |

## Tests

Added `test_max_error_lines_accepts_integer` to `utest/conf/test_settings.py`. It fails on `master` (the `int` case raises `AttributeError`) and passes with the fix; the full `test_settings` suite (22 tests) stays green.

---

*Disclosure: this change was prepared with AI assistance and reviewed/verified by me before submission.*

### DIFF
--- a/utest/conf/test_settings.py
+++ b/utest/conf/test_settings.py
@@ -46,6 +46,18 @@ class TestRobotAndRebotSettings(unittest.TestCase):
         assert_equal(RobotSettings({"test": "one"})["TestNames"], ["one"])
         assert_equal(RebotSettings({"exclude": "two"})["Exclude"], ["two"])
 
+    def test_max_error_lines_accepts_integer(self):
+        # https://github.com/robotframework/robotframework/issues/5113
+        # The programmatic API (robot.run/TestSuite.run) passes native types, so
+        # an int must be accepted like the equivalent string from the CLI.
+        assert_equal(RobotSettings({"maxerrorlines": 10}).max_error_lines, 10)
+        assert_equal(RobotSettings({"maxerrorlines": "15"}).max_error_lines, 15)
+        assert_equal(RobotSettings({"maxerrorlines": None}).max_error_lines, None)
+        assert_equal(RobotSettings({"maxerrorlines": "NONE"}).max_error_lines, None)
+        # Invalid values (int or string) still raise a clear error, not AttributeError.
+        self.assertRaises(DataError, RobotSettings, {"maxerrorlines": 5})
+        self.assertRaises(DataError, RobotSettings, {"maxerrorlines": "5"})
+
     def test_output_files(self):
         for name in (
             "Output.xml",
--- a/src/robot/conf/settings.py
+++ b/src/robot/conf/settings.py
@@ -185,7 +185,7 @@ class _BaseSettings:
         return collect.level, show.level
 
     def _process_max_error_lines(self, value):
-        if not value or value.upper() == "NONE":
+        if not value or (isinstance(value, str) and value.upper() == "NONE"):
             return None
         value = self._convert_to_integer("MaxErrorLines", value)
         if value < 10:
--- a/src/robot/running/model.py
+++ b/src/robot/running/model.py
@@ -46,7 +46,7 @@ from robot.errors import (
 from robot.model import BodyItem, DataDict, TestSuites
 from robot.output import LOGGER, Output, pyloggingconf
 from robot.result import Result
-from robot.utils import format_assign_message, setter
+from robot.utils import format_assign_message, setter, text
 from robot.variables import VariableResolver
 
 from .bodyrunner import (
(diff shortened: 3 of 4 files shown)


utest/conf/test_settings.py
```diff
@@ -46,6 +46,18 @@ def test_multi_options_as_single_string(self):
         assert_equal(RobotSettings({"test": "one"})["TestNames"], ["one"])
         assert_equal(RebotSettings({"exclude": "two"})["Exclude"], ["two"])
 
+    def test_max_error_lines_accepts_integer(self):
```
yangfan-yf-yf, line 49: This verifies settings conversion only, so it passes while the issue's direct TestSuite.run() path still ignores the value. Please add a behavior-level regression here or under the running tests that executes a suite with a long failure message and asserts that maxerrorlines=None and a finite limit actually change the result.

yangfan-yf-yf (changes requested):
The settings conversion no longer raises AttributeError for integers, but I can still reproduce the original #5113 behavior at 20cda9e. A direct TestSuite.run() with maxerrorlines=None, 10, "NONE", or "999" leaves robot.utils.text.MAX_ERROR_LINES at 40 and returns the same 41-line truncated message in every case. The new settings test passes because it never exercises the suite execution path.

Please add a behavior-level TestSuite.run() regression and make the direct suite path apply the configured error-line limit for the duration of execution, restoring the previous global value in a finally path. The higher-level RobotFramework._execute() already performs that save/apply/restore step; TestSuite.run() currently does not.

Local validation on Windows with Python 3.12: utest/conf/test_settings.py passed (22 passed), the direct 60-line failure reproduction remained truncated for all four values above, and git diff --check passed.

yangfan-yf-yf (approved):
I rechecked `6fd34608`. The direct `TestSuite.run()` path now applies the configured error-line limit and restores both global limits in `finally`. The new behavior tests cover the default, integer and string limits, `None` / `NONE`, a high limit, and restoration after a normal run.

I also forced `TestSuite.visit()` to raise and verified that both globals were restored while the original exception propagated. The two affected test modules pass locally (`52 passed`). I do not see a remaining code blocker, so I am approving this head.

The Acceptance and Unit Actions are still waiting for repository approval and have not run; this review does not treat those hosted checks as passed.

jaideeppyne:
You're right, and thanks for digging into it — the settings fix only covered the conversion, not the execution path. I reproduced it: a direct `TestSuite.run()` leaves `MAX_ERROR_LINES` at 40 for every value, while `robot.run()` honours it because `_execute()` does the save/apply/restore and `TestSuite.run()` never did.

Pushed a commit that moves that step into `TestSuite.run()` with a `finally` restore (same treatment for `MAX_ASSIGN_LENGTH`, which had the identical gap), plus a behaviour-level test in `utest/running/test_running.py` that runs a suite and checks the limit is applied, can be disabled, and that the global is restored.

utest/conf/test_settings.py
jaideeppyne: Agreed, and that is in the PR now under `utest/running/test_running.py` as `TestMaxErrorLines`. It builds a suite whose test fails with a 60 line message, runs it through `TestSuite.run`, and checks the resulting message:

- default limit applies with no options
- `maxerrorlines=10` and `"10"` both give 11 lines
- `maxerrorlines=999` and `None` and `"NONE"` all give the full message back
- the global `MAX_ERROR_LINES` is restored afterwards

You were right that the settings test alone was not enough: `TestSuite.run` was still ignoring the value, so the fix needed a change in `running/model.py` as well, not just the conversion.

pekkaklarck:
Thanks for the PR! It's too late to get this into RF 7.5 and due to being busy with trying to get the release finished, I don't have time to review this now. I added the referenced issue to RF 7.6 scope to remind us to return to this when we start its development.

jaideeppyne:
No rush at all, and thanks for scoping the issue to 7.6 rather than letting it drop.

The branch will stay as it is, so there is nothing to redo when you get to it. If it has gone stale by then I will rebase it, just say the word.

Good luck with the 7.5 release.

PR #5764, opened 2026-08-25
libdoc: support for markdown file writing

### LINKED ISSUE #5749: Libdoc: Support Markdown as output format
# Markdown support
If library documentation is written as Markdown, as it is possible from 7.5 release, it would beneficial if libdoc could output library documentation also as Markdown. This is useful for many use cases. 

## AI/LLM 
Example LLM can easily understand documentation that is written in Markdown format. Because Markdown contains less extra characters, example compared to XML or JSON, it will consume less tokens from the LLM context, which helps LLM to digest bigger libraries documentation and do more meaningful work.

## Documentation publish systems
Many documentation publish systems do not accept standalone HTML pages, but they do accept Markdown documents. Then those documents can be easily published to different knowledge places, without needing to convert from one format to another.

aaltat:
Fixes #5749

### DIFF
--- a/atest/testdata/libdoc/MarkdownLibrary.py
+++ b/atest/testdata/libdoc/MarkdownLibrary.py
@@ -43,3 +43,15 @@ class MarkdownLibrary:
             *args (str): varargs doc.
             **kwargs (str): keyword args doc.
         """
+        pass
+
+    def only_arg_doc(self, arg: str):
+        """
+        Args:
+            arg: arg doc.
+        """
+        pass
+
+    def no_args_doc(self, arg: str):
+        """Only doc."""
+        pass
--- a/doc/userguide/src/SupportingTools/Libdoc.rst
+++ b/doc/userguide/src/SupportingTools/Libdoc.rst
@@ -8,10 +8,10 @@ Library documentation tool (Libdoc)
    :local:
 
 Libdoc is Robot Framework's built-in tool that can generate documentation for
-Robot Framework libraries and resource files. It can generate HTML documentation
-for humans as well as machine readable spec files in XML and JSON formats.
-Libdoc also has few special commands to show library or resource information
-on the console.
+Robot Framework libraries and resource files. It can generate HTML and Markdown
+documentation for humans as well as machine readable spec files in XML and JSON
+formats. Libdoc also has few special commands to show library or resource
+information on the console.
 
 Documentation can be created for:
 
@@ -44,24 +44,27 @@ Synopsis
 Options
 ~~~~~~~
 
-  -f, --format <html|xml|json|libspec>
-                           Specifies whether to generate an HTML output for humans or
-                           a machine readable spec file in XML or JSON format. The
-                           `libspec` format means XML spec with documentations converted
-                           to HTML. The default format is got from the output file
-                           extension.
+  -f, --format <html|xml|json|libspec|markdown>
+                           Specifies whether to generate an HTML output for humans,
+                           a machine readable spec file in XML or JSON format, or
+                           a Markdown file. The `libspec` format means XML spec with
+                           documentations converted to HTML. The default format is got
+                           from the output file extension. The `markdown` format is new
+                           in Robot Framework 7.5.
   -s, --specdocformat <raw|html>
                            Specifies the documentation format used with XML and JSON
                            spec files. `raw` means preserving the original documentation
                            format and `html` means converting documentation to HTML. The
                            default is `raw` with XML spec files and `html` with JSON
-                           specs and when using the special `libspec` format.
-  -F, --docformat <robot|html|text|rest>
+                           specs and when using the special `libspec` format. Not
+                           applicable with `html` or `markdown` outputs.
+  -F, --docformat <robot|markdown|html|text|rest>
                            Specifies the source documentation format. Possible
                            values are Robot Framework's documentation format,
-                           HTML, plain text, and reStructuredText. Default value
-                           can be specified in test library source code and
-                           the initial default value is `robot`.
+                           Markdown, HTML, plain text, and reStructuredText. Default
+                           value can be specified in test library source code and
+                           the initial default value is `robot`. Markdown support is
+                           new in Robot Framework 7.5.
   --theme <dark|light|none>
                            Use dark or light HTML theme. If this option is not used,
                            or the value is `none`, the theme is selected based on
@@ -70,7 +73,8 @@ Options
   --language <lang>
                           Set the default language in documentation. `lang`
                           must be a code of a built-in language, which are
-                          `en` and `fi`. New in Robot Framework 7.2.
+                          `en` and `fi`. Only applicable with HTML outputs.
+                          New in Robot Framework 7.2.
   -N, --name <newname>     Sets the name of the documented library or resource.
   -V, --version <newversion>  Sets the version of the documented library or
                            resource. The default value for test libraries is
@@ -158,10 +162,10 @@ This works if spec files use either :file:`*.xml`, :file:`*.libspec` or
 Generating documentation
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-Libdoc can generate documentation in HTML (for humans) and XML or JSON (for tools)
-formats. The file where to write the documentation is specified as the second
-argument after the library/resource name or path, and the output format is
-got from the output file extension by default.
+Libdoc can generate documentation in HTML, Markdown, XML or JSON formats.
+The file where to write the documentation is specified as the second argument
+after the library/resource name or path, and the output format is got from the
+output file extension by default.
 
 Libdoc HTML documentation
 '''''''''''''''''''''''''
@@ -268,6 +272,32 @@ The spec file format may change between Robot Framework major releases.
 
 __ https://json-schema.org/
 
+Libdoc Markdown documentation
+'''''''''''''''''''''''''''''
+
+Starting from Robot Framework 7.5, Libdoc can also write documentation into
+a Markdown_ file. Markdown files are human readable plain text files, but they
+can also be converted to HTML or otherwise processed by external tools.
+
+Libdoc automatically uses the Markdown format if the output file extension is
+:file:`*.md`. The format can also be set explicitly with the :option:`--format`
+option::
+
+   libdoc OperatingSystem OperatingSystem.md
+   libdoc test/resource.robot doc/resource.md
+   libdoc --format markdown MyLibrary MyLibrary.markdown
+
+Documentation is taken from the library or resource as-is without converting it
+to any other format. The result is thus proper Markdown only if the library uses
+the `Markdown documentation syntax`_ itself.
+
+The :option:`--specdocformat`, :option:`--theme` and :option:`--language` options
+are not applicable with Markdown outputs.
+
+.. note:: The Markdown output format is not guaranteed to stay stable between
+          Robot Framework versions. If you need a stable, machine readable format,
+          use `Libdoc spec files`_ instead.
+
 Viewing information on console
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -300,10 +330,11 @@ Examples::
   libdoc SeleniumLibrary version
 
 When showing documentation of a whole library or some keywords, the overall
-structure is formatted using Markdown_. This is especially convenient with
-libraries that use `Markdown documentation syntax`_ themselves, because then
-the whole output is in Markdown format. This is demonstrated by the following
-example from the beginning of the Dialogs library documentation::
+structure is formatted using Markdown_. This is the same formatting that is
+used with `Libdoc Markdown documentation`_ outputs. It is especially convenient
+with libraries that use `Markdown documentation syntax`_ themselves, because
+then the whole output is in Markdown format. This is demonstrated by the
+following example from the beginning of the Dialogs library documentation::
 
     # Dialogs
 
@@ -324,12 +355,14 @@ example from the beginning of the Dialogs library documentation::
 
     ### Execute Manual Step
 
-    **Arguments:**
+    #### Arguments
 
     * `message` (type: `str`) -
       The instruction shown in the initial dialog.
     * `default_error` (type: `str`, default: ``) -
       The default value shown in the possible error message dialog.
+
+    #### Documentation
 
     Pauses execution until user sets the keyword status.
 
@@ -340,7 +373,8 @@ example from the beginning of the Dialogs library documentation::
 
 .. note:: The console output format is not guaranteed to stay stable between
           Robot Framework versions. If you need a stable, machine readable format,
-          use `Libdoc spec files`_ instead.
+          use `Libdoc spec files`_ instead. If you want to save the documentation
+          in this format, use `Libdoc Markdown documentation`_ outputs.
 
 Writing documentation
 ---------------------
--- a/src/robot/libdoc.py
+++ b/src/robot/libdoc.py
@@ -73,19 +73,21 @@ is running the module as a script like `python path/to/robot/libdoc.py`.
 Options
 =======
 
- -f --format HTML|XML|JSON|LIBSPEC
+ -f --format HTML|XML|JSON|LIBSPEC|MARKDOWN
                           Specifies whether to generate an HTML output for
-                          humans or a machine readable spec file in XML or JSON
-                          format. The LIBSPEC format means XML spec with
-                          documentations converted to HTML. The default format
-                          is got from the output file extension.
+                          humans, a machine readable spec file in XML or JSON
+                          format, or a Markdown file. The LIBSPEC format means
+                          XML spec with documentations converted to HTML. The
+                          default format is got from the output file
+                          extension. MARKDOWN is new in RF 7.5.
  -s --specdocformat RAW|HTML
                           Specifies the documentation format used with XML and
                           JSON spec files. RAW means preserving the original
                           documentation format and HTML means converting
                           documentation to HTML. The default is RAW with XML
                           spec files and HTML with JSON specs and when using
-                          the special LIBSPEC format.
+                          the special LIBSPEC format. Not applicable with HTML
+                          or MARKDOWN outputs.
  -F --docformat ROBOT|MARKDOWN|HTML|TEXT|REST
                           Specifies the source documentation format. Possible
                           values are Robot Framework's documentation format,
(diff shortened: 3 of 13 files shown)


src/robot/libdocpkg/consoleviewer.py
```diff
@@ -65,19 +61,6 @@ def version(self):
     def _console(self, msg, end="\n"):
         print(console_encode(msg), end=end)
```
pekkaklarck, line 64: There might be room for cleanup here now that formatting has been moved elsewhere.

src/robot/libdocpkg/markdownformatter.py
```diff
@@ -0,0 +1,150 @@
...
+        md += self._format_tags(self.keyword.tags)
+        md += self._format_doc(self.keyword.doc)
+        return md
+
+    def _format_args(self, args: ArgumentSpec) -> str:
+        if not args:
+            return ""
+        md = "#### Arguments\n\n"
+        for arg in args:
+            if arg.kind in (ArgInfo.POSITIONAL_ONLY_MARKER, ArgInfo.NAMED_ONLY_MARKER):
```
pekkaklarck: Can actually use `arg.is_marker` here. I had forgotten about it myself.

Possibly could also change to `if not arg.is_marker` to avoid `continue`.

src/robot/libdocpkg/markdownformatter.py
```diff
@@ -0,0 +1,150 @@
...
+        for tag in tags:
+            md += f"* `{tag}`\n"
+        return md + "\n"
+
+    def _format_doc(self, doc: str) -> str:
+        if not doc:
+            return ""
+        md = "#### Documentation\n\n"
+        return md + doc + "\n\n"
+        # return f"{doc}\n\n" if doc else ""
```
pekkaklarck: This needs to be removed.

src/robot/libdocpkg/markdownformatter.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class LibdocMarkdownWriter:
+    """Writes the full library/resource documentation as a single Markdown file.
+
+    Reuses the same rendering `ConsoleViewer` uses for `libdoc ... show`: doc
+    strings are inserted as-is, converted to Markdown only when the source is
+    already documented in the Markdown doc format.
```
pekkaklarck: The end here is pretty confusing, because nothing is ever converted to Markdown. Perhaps the paragraph could we written like this:

> Formatting is the same that is used when viewing documentation on console. Actual documentation is got directly from the library/resource, so the result is proper Markdown only if the library uses Markdown as its documentation format.

src/robot/libdocpkg/markdownformatter.py
```diff
@@ -0,0 +1,150 @@
...
+        md += f"* Scope: {lib.scope}\n"
+    if lib.version or lib.type == "LIBRARY":
+        md += "\n"
+    if lib.doc:
+        md += "## Introduction\n\n"
+        md += lib.doc + "\n\n"
+    return md
+
+
+class KeywordFormatter:
```
pekkaklarck: For consistency reasons it would be good to add a separate function for keyword formatting. The function could use `KeywordFormatter` intenally:
```python
def format_keyword(keyword: "KeywordDoc", init: bool = False) -> str:
    return KeywordFormatter(keyword, init).format()
```
Users could then use `format_intro` and `format_keyword` instead of `format_intro` and `KeywordFormatter().format`.

Adding separate `format_init` could be considered as well, then `format_keyword` wouldn't need the `init` argument.

src/robot/libdocpkg/markdownformatter.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def __init__(self, keyword: "KeywordDoc", init=False):
+        self.keyword = keyword
+        self.init = init
+
+    def format(self) -> str:
+        md = f"### {self.keyword.name}\n\n" if not self.init else ""
+        md += self._format_args(self.keyword.args)
```
pekkaklarck: Looking at this code now, I believe this would be a slightly better style:
```python
def __init__(self, init: bool = False):
    self.init = init

def format(self, keyword: "KeywordDoc") -> str:
    md = f"### {keyword.name}\n\n" if not self.init else ""
    md += self._format_args(keyword.args)
```
The idea is that you could then use this like
```python
init_formatter = KeywordFormatter(init=True)
init_formatter.format(init)
kw_formatter = KeywordFormatter()
for kw in keywords:
    kw.formatter.format(kw)
```
It may be that the code is never used like that, but I still believe this would be better OO design.

src/robot/libdocpkg/consoleviewer.py
pekkaklarck: New `LibdocMarkdownWriter.write` has the same formatting logic as here. Duplication isn't good in general and in this case it means we needed to make enhancements and fixes to two places. We may, for example, add type docs to Markdown outputs in the future.

I believe we need a separate class like below and use it in these two places:
```python
class MarkdownFormatter:

    def format(self, libdoc) -> str:
        self._format_introduction(...)
        self._format_importing(...)
        self._format_keywords(...)
```
With this design it would be easy to add `self._format_type_docs(...)` and to make other such changes so that both usages would benefit.

This `MarkdownFormatter` could then use the existing `KeywordFormatter` internally. These two classes could possibly also be merged, but it's probably better to keep them separate to avoid `MarkdownFormatter` growing too big. That's a risk especially if we add type docs or make other such changes later.

src/robot/libdocpkg/markdownformatter.py
```diff
@@ -0,0 +1,150 @@
...
+    def write(self, libdoc, output):
+        with file_writer(output) as outfile:
+            outfile.write(format_intro(libdoc))
+            if libdoc.inits:
+                outfile.write("## Importing\n\n")
+                for init in libdoc.inits:
+                    outfile.write(KeywordFormatter(init, init=True).format())
+            outfile.write("## Keywords\n\n")
+            for kw in libdoc.keywords:
+                outfile.write(KeywordFormatter(kw).format())
```
pekkaklarck: As already noted in an earlier comment, this duplicated code should be moved to a common class.

src/robot/libdoc.py
```diff
@@ -73,19 +73,21 @@
 Options
 =======
 
- -f --format HTML|XML|JSON|LIBSPEC
+ -f --format HTML|XML|JSON|LIBSPEC|MD
```
pekkaklarck: This should be `MARKDOWN` instead of `MD`. I hope this isn't tied to the output file extension, though.

src/robot/libdoc.py
```diff
@@ -73,19 +73,21 @@
...
                           Specifies whether to generate an HTML output for
-                          humans or a machine readable spec file in XML or JSON
-                          format. The LIBSPEC format means XML spec with
-                          documentations converted to HTML. The default format
-                          is got from the output file extension.
+                          humans, a machine readable spec file in XML or JSON
+                          format, or a Markdown file. The LIBSPEC format means
+                          XML spec with documentations converted to HTML. The
+                          default format is got from the output file
+                          extension. MD is new in RF 7.5.
```
pekkaklarck: `MD is new in RF 7.5.` -> `Markdown support is new in RF 7.5.`

src/robot/libdoc.py
```diff
@@ -73,19 +73,21 @@
...
  -s --specdocformat RAW|HTML
                           Specifies the documentation format used with XML and
                           JSON spec files. RAW means preserving the original
                           documentation format and HTML means converting
                           documentation to HTML. The default is RAW with XML
                           spec files and HTML with JSON specs and when using
-                          the special LIBSPEC format.
+                          the special LIBSPEC format. Not applicable with HTML
+                          or MD outputs.
```
pekkaklarck: `or MD outputs` -> `or Markdown outputs`

src/robot/libdoc.py
```diff
@@ -100,7 +102,7 @@
     --language lang       Set the default language used in HTML outputs.
                           `lang` must be one of the built-in language codes:
 {format_languages()}
-                          New in RF 7.2.
+                          New in RF 7.2. Not applicable with MD outputs.
```
pekkaklarck: This isn't really applicable with specs either. I believe the added note can be removed. Changing it to something like `Only applicable with HTML outputs.` would be fine as well, but I'm 98% sure we'd forget to update it if we actually enable translations with Markdown. 😄

src/robot/libdoc.py
```diff
@@ -113,21 +115,30 @@
 Creating documentation
 ======================
 
-When creating documentation in HTML, XML or JSON format, the output file must
-be specified as the second argument after the library or resource name or path.
+When creating documentation in HTML, XML, JSON or MD format, the output file
```
pekkaklarck: `MD` -> `Markdown`

src/robot/libdoc.py
```diff
@@ -113,21 +115,30 @@
...
+special `*.libspec` extensions which means XML spec with actual library and
+keyword documentation converted to HTML. The format can also be set explicitly
+using the `--format` option.
+
+Markdown output (`*.md`) writes the library or resource documentation as a
+single Markdown file, including the introduction, importing and all keywords.
+It is written as-is, with no keyword name filtering. Documentation is only
+converted to Markdown when the library itself uses the Markdown documentation
+format (see `--docformat`); otherwise the original documentation is included
+unconverted. MD output is new in RF 7.5.
```
pekkaklarck: This paragraph is confusing similarly as the earlier API doc. There's no conversion to Markdown at all. This could be enhanced, but I'm not sure is this relevant information in `--help` text in general. If this is preserved, use `Markdown` instead of `MD` and `Robot Framework 7.5` instead of `RF 7.5.`.

Notice that `RF 7.5` is fine when documenting options, because there's not much room for docs. Here we don't have such a problem so better use proper names instead of abbreviations. `MD` isn't a generally used abbreviation so I'd avoid it altogether (except in `*.md`).

src/robot/libdoc.py
```diff
@@ -207,6 +218,11 @@ def main(
         format, specdocformat = self._get_format_and_specdocformat(
             format, specdocformat, output
         )
+        lang = self._validate_lang(language)
+        # TODO: Markdown output doesn't support --language yet. Remove this guard,
+        # and thread `lang` through to `LibdocMarkdownWriter`, once it does.
+        if format == "MD" and lang:
+            raise DataError("The --language option is not applicable with MD outputs.")
```
pekkaklarck: Do we need to add this at all?

pekkaklarck (changes requested):
Looks very good, but I believe the formatting logic should be restructured and few other smaller things fixed/enhanced.

src/robot/libdoc.py
pekkaklarck: Related to the above, this feature needs to be documented in the User Guide.

src/robot/libdocpkg/markdownformatter.py
aaltat: Done

src/robot/libdocpkg/markdownformatter.py
aaltat: My bad, done

src/robot/libdocpkg/markdownformatter.py
aaltat: I agree. Made the change.

src/robot/libdoc.py
aaltat: Well, there are similar type of errors raised example in here: https://github.com/robotframework/robotframework/blob/master/src/robot/libdoc.py#L266 and few other places. We can also silently ignore the invalid command line switch too. or do you mean the comment?

atest/testdata/libdoc/MarkdownLibrary.py
```diff
@@ -43,3 +43,18 @@ def varargs_and_kwargs(self, *args: str, **kwargs: str):
...
             **kwargs (str): keyword args doc.
         """
+        pass
+
+    def no_doc(self, arg: str):
+        """
+        Args:
+            arg: arg doc.
+        """
+        pass
```
pekkaklarck, line 53: This is a bit oddly named keyword because the method has a doc. Having a keyword without any doc (and any args) would probably be a good idea in addition or instead of this.

atest/testdata/libdoc/MarkdownLibrary.py
```diff
@@ -43,3 +43,18 @@ def varargs_and_kwargs(self, *args: str, **kwargs: str):
...
+
+    def no_args_doc(self, arg: str):
+        """Args is here but no arg defined.
+
+        Args:
+        """
```
pekkaklarck: This is probably testing the DocStringParser more than anything else. Could probably be removed.

doc/userguide/src/SupportingTools/Libdoc.rst
```diff
@@ -44,24 +44,27 @@ Synopsis
...
+  -F, --docformat <robot|markdown|html|text|rest>
                            Specifies the source documentation format. Possible
                            values are Robot Framework's documentation format,
-                           HTML, plain text, and reStructuredText. Default value
-                           can be specified in test library source code and
-                           the initial default value is `robot`.
+                           Markdown, HTML, plain text, and reStructuredText. Default
+                           value can be specified in test library source code and
+                           the initial default value is `robot`. Markdown support is
+                           new in Robot Framework 7.5.
```
pekkaklarck, line 67: Good catch that Markdown was missing here.

doc/userguide/src/SupportingTools/Libdoc.rst
```diff
@@ -158,10 +162,10 @@ This works if spec files use either :file:`*.xml`, :file:`*.libspec` or
 Generating documentation
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-Libdoc can generate documentation in HTML (for humans) and XML or JSON (for tools)
-formats. The file where to write the documentation is specified as the second
-argument after the library/resource name or path, and the output format is
-got from the output file extension by default.
+Libdoc can generate documentation in HTML and Markdown (for humans) and in XML
+or JSON (for tools) formats. The file where to write the documentation is
```
pekkaklarck: This "for humans" and "for tools" stuff gets a bit murky with Markdown that is kind of for both. Perhaps we should just remove that stuff and only list supported formats? Could possibly do that elsewhere too.

doc/userguide/src/SupportingTools/Libdoc.rst
```diff
@@ -268,6 +272,34 @@ The spec file format may change between Robot Framework major releases.
...
 __ https://json-schema.org/
 
+Libdoc Markdown documentation
+'''''''''''''''''''''''''''''
+
+Starting from Robot Framework 7.5, Libdoc can also write documentation into
+a single Markdown_ file. Markdown is similar to the HTML documentation format
+but it missing some features, like langue support or documenting the custom
+types used by keyword. The main advantage of that Markdown is that it is a both
+human and AI readable format that can be used as-is by both consumers.
```
pekkaklarck: I don't think it is a good idea to mention missing language support here. We'll certainly forgot to update this if/when such support is added. Markdown can also be used also by other tools than AI. AI could be mentioned, but I'm not sure is it necessary. Perhaps the paragraph could be rephrased like this:

> Starting from Robot Framework 7.5, Libdoc can also write documentation into
> a Markdown_ file. Markdown files are human readable plain text files, but they can
> also be converted to HTML or otherwise processed by external tools.

src/robot/libdocpkg/markdownformatter.py
```diff
@@ -0,0 +1,170 @@
...
+            md += kw_formatter.format(kw)
+        return md
+
+    def format(self) -> str:
+        md = self.format_introduction()
+        md += self.format_importing()
+        md += self.format_keywords()
+        return md
+
+class LibdocMarkdownWriter:
```
pekkaklarck: Should have two empty lines.

src/robot/libdocpkg/markdownformatter.py
```diff
@@ -0,0 +1,170 @@
...
+        md = "#### Documentation\n\n"
+        return md + doc + "\n\n"
+
+
+class MarkdownFormatter:
+
+    def __init__(self, libdoc):
+        self.libdoc = libdoc
+
+
```
pekkaklarck: Should only have one empty line.

src/robot/libdocpkg/markdownformatter.py
```diff
@@ -0,0 +1,170 @@
...
+        kw_formatter = KeywordFormatter()
+        for kw in self.libdoc.keywords:
+            md += kw_formatter.format(kw)
+        return md
+
+    def format(self) -> str:
+        md = self.format_introduction()
+        md += self.format_importing()
+        md += self.format_keywords()
+        return md
```
pekkaklarck: This is the "main" method of this class. I think it should be first one (after `__init__`).

src/robot/libdocpkg/markdownformatter.py
```diff
@@ -0,0 +1,170 @@
...
+
+from robot.model import Tags
+from robot.running import ArgInfo, ArgumentSpec
+from robot.utils import file_writer
+
+if TYPE_CHECKING:
+    from .model import KeywordDoc
+
+
+class KeywordFormatter:
```
pekkaklarck, line 84: This is a utility class only used by other classes in this module and it would be more logical to have it after them. I think a logical order would be:

- LibdocMarkdownWriter
- MarkdownFormatter
- KeywordFormatter

pekkaklarck (changes requested):
Looks very good but some fine tuning could still be done!

src/robot/libdocpkg/markdownformatter.py
aaltat: Done

src/robot/libdocpkg/markdownformatter.py
aaltat: Done

src/robot/libdocpkg/markdownformatter.py
aaltat: Done

src/robot/libdocpkg/markdownformatter.py
aaltat: Done

doc/userguide/src/SupportingTools/Libdoc.rst
aaltat: Done

doc/userguide/src/SupportingTools/Libdoc.rst
aaltat: Done in here.

atest/testdata/libdoc/MarkdownLibrary.py
aaltat: Changed.

atest/testdata/libdoc/MarkdownLibrary.py
aaltat: it is. changed keyword name.

PR #5766, opened 2026-08-27
Libdoc: Collections addressed review comments 

Vinni82:
- Reworked copy items towards *** Variables *** section. Tested every testcase
- Reworked the linking towards other libraries and put the library links in the docstring reference.
- Reworked line length max 88 chars, two are in Python code, I ignored those

### DIFF
--- a/src/robot/libraries/Collections.py
+++ b/src/robot/libraries/Collections.py
@@ -46,9 +46,9 @@ class _List:
             The converted list.
 
         Mainly useful for converting tuples and other iterable to lists.
-        Use [Create List] from the BuiltIn library for constructing new lists.
+        Use [Create List] from the [BuiltIn] library for constructing new lists.
 
-        Use [Split String To Characters] from the String library for splitting
+        Use [Split String To Characters] from the [String] library for splitting
         strings to a list of characters.
         """
         return list(item)  # type: ignore
@@ -72,14 +72,16 @@ class _List:
         Examples:
 
         ```robotframework
+        *** Variables ***
+        @{list_app_1}    a    b    c
+        @{list_app_2}    f    g    d
+
         *** Test Cases ***
         Append to list
-           ${list_1} =    Copy List    ${LIST_ABC}
-           ${list_2} =    Copy List    ${LIST_ABCDE}
-           ${appended_1} =    Append To List    ${list_1}    xxx
-           ${appended_2} =    Append To List    ${list_2}    x    y    z
+           ${appended_1} =    Append To List    ${list_app_1}    xxx
+           ${appended_2} =    Append To List    ${list_app_2}    x    y    z
            Should Be Equal    ${appended_1}    ["a", "b", "c", "xxx"]    type=list
-           Should Be Equal    ${appended_2}    ["a", "b", "c", "d", "e", "x", "y", "z"]    type=list
+           Should Be Equal    ${appended_2}    ["f", "g", "d", "x", "y", "z"]    type=list
         ```
 
         !!! note
@@ -121,14 +123,16 @@ class _List:
         Examples:
 
         ```robotframework
+        *** Variables ***
+        @{list_ins_1}    a    b    c
+        @{list_ins_2}    f    g    d
+
         *** Test Cases ***
         Insert in list
-            ${list_1} =    Copy List    ${LIST_ABC}
-            ${list_2} =    Copy List    ${LIST_ABC}
-            ${inserted_1} =    Insert Into List    ${list_1}    0    xxx
-            ${inserted_2} =    Insert Into List    ${list_2}    -1    xxx
-            Should Be Equal    ${inserted_1}    ["xxx", "a", "b", "c"]    type=list
-            Should Be Equal    ${inserted_2}    ["a", "b", "xxx", "c"]    type=list
+            ${insert_begin} =    Insert Into List    ${list_ins_1}    0    xxx
+            ${insert_end} =    Insert Into List    ${list_ins_2}    -1    xxx
+            Should Be Equal    ${insert_begin}    ["xxx", "a", "b", "c"]    type=list
+            Should Be Equal    ${insert_end}    ["f", "g", "xxx", "d"]    type=list
         ```
 
         !!! note
@@ -154,9 +158,13 @@ class _List:
         Examples:
 
         ```robotframework
+        *** Variables ***
+        @{LIST_COMB_1}      a    b    c
+        @{LIST_COMB_2}    a    b    c    d    e
+
         *** Test Cases ***
         Combine lists
-            ${combined} =    Combine Lists    ${LIST_ABC}    ${LIST_ABCDE}
+            ${combined} =    Combine Lists    ${LIST_COMB_1}    ${LIST_COMB_2}
             Should Be Equal    ${combined}    ["a", "b", "c", "a", "b", "c", "d", "e"]    type=list
         ```
         """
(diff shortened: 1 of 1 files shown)


src/robot/libraries/Collections.py
```diff
@@ -46,9 +46,9 @@ def convert_to_list(self, item: object) -> list:
             The converted list.
 
         Mainly useful for converting tuples and other iterable to lists.
-        Use [Create List] from the BuiltIn library for constructing new lists.
+        Use [Create List] for constructing new lists.
 
-        Use [Split String To Characters] from the String library for splitting
+        Use [Split String To Characters] for splitting
```
pekkaklarck: I initially wrote a comment here that using the `[Create List]` link style doesn't work with keywords that don't exist in this library, and that the reference to BuiltIn/String shouldn't be removed. Only at the end I noticed that you had actually crated link targets for `[Crete List]` and other external keywords. That obviously works, but that approach has some problems:

- It would anyway be good to mention where these keywords come from. Most keyword links are internal, so it's strange if some of them jump to other libraries without any visual indication.
- Adding and maintaining links to all external keywords requires work.
- We should either link all external keyword or none of them. Not there's a big table of related BuiltIn keywords without linking.

To avoid the above problems, I believe it's better to link only to libraries and just use code formatting with the keyword name. In practice that means using this:
```
Use `Create list` from the [BuiltIn] library for ...
```

I have already used that approach with the table of related BuiltIn keywords, so the `[BuiltIn]` target already exists, but a target should be added to `[String]` as well.

src/robot/libraries/Collections.py
```diff
@@ -72,14 +72,17 @@ def append_to_list(
         Examples:
 
         ```robotframework
+        *** Variables ***
+        @{list_app_1}    a    b    c
+        @{list_app_2}    f    g    d
```
pekkaklarck: Few things to fix:
- Global variables should use capital letters.
- The unnecessary `app` part should be removed.
- Other examples have used format `@{LIST1}` without a `_`. I think that looks a bit better than `@{LIST_1}`, but I'm fine with both. We should just use same convention at least within a single library.

I understand the need for the `app` if it is there to separate these lists from lists in other examples and without it running all examples doesn't work. In that case it should be something more explicit like `APPEND` and in that case the `LIST` prefix gets unnecessary. Something like `@{APPEND1}` would probably work well.

An other alternative would be using the `VAR` syntax to create local variables. Then there's no risk for variable conflicts.

src/robot/libraries/Collections.py
```diff
@@ -72,14 +72,17 @@ def append_to_list(
...
-           ${list_1} =    Copy List    ${LIST_ABC}
-           ${list_2} =    Copy List    ${LIST_ABCDE}
-           ${appended_1} =    Append To List    ${list_1}    xxx
-           ${appended_2} =    Append To List    ${list_2}    x    y    z
+           ${appended_1} =    Append To List    ${list_app_1}    xxx
+           ${appended_2} =    Append To List    ${list_app_2}    x    y    z
            Should Be Equal    ${appended_1}    ["a", "b", "c", "xxx"]    type=list
-           Should Be Equal    ${appended_2}    ["a", "b", "c", "d", "e", "x", "y", "z"]    type=list
+           Should Be Equal    ${appended_2}    ["f", "g", "d", "x", "y", "z"]
+           ...    type=list
```
pekkaklarck: I think it would be better to keep this last example on one line even if it would mean that the source line would be over our 88 character max line length. That limit is not enforced in docs and breaking it in cases like this isn't a big problem. Line could also be made shorter by keeping lists a bit shorter.

src/robot/libraries/Collections.py
```diff
@@ -121,14 +124,16 @@ def insert_into_list(
         Examples:
 
         ```robotframework
+        *** Variables ***
+        @{list_ins_1}    a    b    c
+        @{list_ins_2}    f    g    d
```
pekkaklarck: See the earlier comment. Probably using `@{INSERT1}` and `@{INSERT2}` here and `@{APPEND1}` and `@{APPEND2}` earlier would work well.

src/robot/libraries/Collections.py
```diff
@@ -154,10 +159,15 @@ def combine_lists(self, *lists: ListLike) -> list:
         Examples:
 
         ```robotframework
+        *** Variables ***
+        @{LIST_COMB_1}      a    b    c
+        @{LIST_COMB_2}    a    b    c    d    e
```
pekkaklarck: See earlier comments. I'd use `@{COMBINE1}` style here. Also align the values.

I noticed these same issues also with other examples. I'm not going to comment the remaining ones, but they should obviously be updated as well.

src/robot/libraries/Collections.py
```diff
@@ -761,8 +790,8 @@ def lists_should_be_equal(
 
         ```robotframework
         *** Variables ***
-        @{USER1}         Jane    Doe    jane@example.com
-        @{USER2}         John    Doe    john@example.com
+        @{USER1}    Jane    Doe    jane@example.com
+        @{USER2}    John    Doe    john@example.com
```
pekkaklarck: The old separator between the variable name and value was better. Although the separator between items in Robot data typically is four spaces, settings in the Settings section and variables in the Variables section are an exception and with them values are typically aligned. Due to the alignment level depending on the length of the setting/variable name, values are generally aligned so that there's room for longer names.

With variables a good rule of thumb is starting values from the level where the `*** Variables ***` header ends like this example did earlier. That level should also be used with all other examples to make them look consistent.

In the long run we should use Robocop for automatically formatting examples.

src/robot/libraries/Collections.py
```diff
@@ -1708,8 +1762,8 @@ def should_contain_match(
         [regular expression syntax](http://docs.python.org/library/re.html).
         Notice that the backslash character often used with regular
         expressions is an escape character in Robot Framework data and needs
-        to be escaped with another backslash like `regexp=\\d{6}`. See the
-        [BuiltIn] keyword `Should Match Regexp` for more details.
+        to be escaped with another backslash like `regexp=\\d{6}`.
+        See [Should Match Regexp] for more details.
```
pekkaklarck: As I wrote in the first comment, the reference to BuiltIn should not removed. I also believe it's enough to link to BuiltIn, not directly to `Should Match Regexp`. In the latter case also other external keywords should be linked and I don't think that's worth the initial or the maintenance effort.

src/robot/libraries/Collections.py
```diff
@@ -1848,7 +1902,7 @@ def get_matches(
 
         ```robotframework
         *** Variables ***
-        @{LIST}          foo    bar    abc 123
+        @{LIST}    foo    bar    abc 123
```
pekkaklarck: Preserve the original separator. See an earlier comment about this for details.

Notice that if you end up using example specific variables like `@{APPEND1}` with other examples, this one should probably also be something like `@{MATCHED}` or `@{MATCHED_ITEMS}`.

pekkaklarck (changes requested):
I like the new style of introducing variables with each example a lot more than the earlier. There are, however, various easy-to-fix issues with variable naming and alignment explained in comments.

Another thing to change is how external keywords are linked to. I had earlier linked only to external libraries, not to individual keywords, but this PR adds direct links and removes notes where the keywords originate from. I believe linking only to libraries would be adequate and requires less work now (we should link all keywords) and in the future (maintenance), but I'm not totally against linking each keyword (as long as it is done consistently). Notes where keywords originate should be preserved regardless are individual keywords linked or not, though.

src/robot/libraries/Collections.py
```diff
@@ -46,10 +46,10 @@ def convert_to_list(self, item: object) -> list:
             The converted list.
 
         Mainly useful for converting tuples and other iterable to lists.
-        Use [Create List] from the BuiltIn library for constructing new lists.
+        Use Create List from the [BuiltIn] library for constructing new lists.
```
pekkaklarck, line 49: Should use code formatting with external keyword names like:
> Use `Create List` from ...

pekkaklarck (approved):
Looks good! I noticed one minor issue that can be fixed (see comment), but I'll do that myself after merging this.

PR #5771, opened 2026-08-29
Add Listener API v3 pre-import callbacks

### LINKED ISSUE #5667: Extend listener with pre-import events
## Context

Born from this discussion on Slack: https://robotframework.slack.com/archives/C015KB1QSDN/p1775116818448659

Robot includes the [Listener API](https://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#listener-interface):

> Robot Framework's listener interface provides a powerful mechanism for getting notifications and for inspecting and modifying data and results during execution. 

Recently, I tried to use this API to centrally change library arguments for every library import. In doing so, I found out that the listener API only provides events _after_ the importing is already done. For my use case, I must have an event _before_ the import starts. This applies to both listener v2 and v3.

## Goal

Extend the listener API with the following events:
- Before Library import starts
- Before Resource import starts
- Before Variables import starts

## Possible design

The current listeners already support the events `library_import`, `resource_import`, and `variables_import`. These events don't need to change.

Add the following events to listener v3:

- `start_library_import`
- `start_resource_import`
- `start_variables_import`

Adding these events may look somethings like this:

### Option 1: Reusing importer

```robotframework
class Example:
    ROBOT_LISTENER_API_VERSION = 3

    def start_library_import(self, importer: robot.running.resourcemodel.Import):
        # Does not currently exist

    def library_import(self, library: robot.running.testlibraries.TestLibrary, importer: robot.running.resourcemodel.Import):
        # Already exists, is unchanged

    def start_resource_import(self, importer: robot.running.resourcemodel.Import):
        # Does not currently exist

    def resource_import(self, resource: robot.running.resourcemodel.ResourceFile, importer: robot.running.resourcemodel.Import):
        # Already exists, is unchanged

    def start_variables_import(self, importer: robot.running.resourcemodel.Import):
        # Does not currently exist

    def variables_import(self, attrs, importer: robot.running.resourcemodel.Import):
        # Already exists, is unchanged
```

### Option 2: Reusing parser nodes

```robotframework
class Example:
    ROBOT_LISTENER_API_VERSION = 3

    def start_library_import(self, node: robot.parsing.model.statements.LibraryImport):
        # Does not currently exist

    def library_import(self, library: robot.running.testlibraries.TestLibrary, importer: robot.running.resourcemodel.Import):
        # Already exists, is unchanged

    def start_resource_import(self, node: robot.parsing.model.statements.ResourceImport):
        # Does not currently exist

    def resource_import(self, resource: robot.running.resourcemodel.ResourceFile, importer: robot.running.resourcemodel.Import):
        # Already exists, is unchanged

    def start_variables_import(self, node: robot.parsing.model.statements.VariablesImport):
        # Does not currently exist

    def variables_import(self, attrs, importer: robot.running.resourcemodel.Import):
        # Already exists, is unchanged
```

## Optional additions

Make an alias for the following, already existing, functions:

`library_import` -> `start_library_import`
`resource_import` -> `start_resource_import`
`variables_import` -> `start_variables_import`

> This could be implemented in backwards compatible manner so that we'd add new start/end_library/resource/variables_import hooks. Old library/resource/variables_import hooks could be preserved and called if the matching end_xxx_import methods weren't defined. Thus there would be no need changes to the listener API version.
> ~Pekka

## Acceptance criteria

- The new `start_*_import` events are called before the relevant file is imported.
- The new `start_*_import` events receive data on what will be imported.
- The listener can control imports using the new `start_*_import` events
  - The listener can change the file that will be imported
  - The listener can change the arguments with which the file is imported
  - The listener can change the alias with which the library is imported (only applies to library imports)

sunilkumarvalmiki:
## Summary

- Add Listener API v3 `start_*_import` callbacks for libraries, resources, and variable files.
- Pass the mutable import model before name/argument resolution so listeners can change the import target, arguments, and library alias.
- Fix Windows path-plus-arguments parsing when an existing path contains parent segments; this restores the custom-console logger regression.

## Tests

- `python -m pytest utest\\output utest\\running\\test_namespace.py utest\\utils\\test_text.py -q` — 181 passed, 3 existing collection warnings.
- `python -m unittest utest.output.test_listeners_issue5667 utest.running.test_namespace_issue5667` — 3 passed.
- `atest/run.py --exclude no-ci atest/robot/output/listener_interface/listening_imports.robot` — 3 passed.
- `python -m compileall` on changed sources and tests — passed.
- Ruff on new tests and parser regression — passed.

The implementation follows Option 1 described in #5667 and leaves existing
after-import events unchanged.

Closes #5667

### DIFF
--- a/atest/robot/output/listener_interface/listening_imports.robot
+++ b/atest/robot/output/listener_interface/listening_imports.robot
@@ -12,6 +12,7 @@ All imports are usable
 
 Listen Imports
     Init Expect
+    Start Expect    Library    BuiltIn    []
     Expect
     ...    Library
     ...    BuiltIn
@@ -19,11 +20,14 @@ Listen Imports
     ...    importer: None
     ...    originalname: BuiltIn
     ...    source: //BuiltIn.py
+    Start Expect    Resource    resource_with_imports.robot    []
+    Start Expect    Resource    another_resource.robot    []
     Expect
     ...    Resource
     ...    another_resource
     ...    importer: //resource_with_imports.robot
     ...    source: //another_resource.robot
+    Start Expect    Library    Process    []
     Expect
     ...    Library
     ...    Process
@@ -36,6 +40,7 @@ Listen Imports
     ...    resource_with_imports
     ...    importer: //imports.robot
     ...    source: //resource_with_imports.robot
+    Start Expect    Library    String    []
     Expect
     ...    Library
     ...    String
@@ -43,6 +48,7 @@ Listen Imports
     ...    importer: //imports.robot
     ...    originalname: String
     ...    source: //String.py
+    Start Expect    Library    local_lib.py    [\${2}]
     Expect
     ...    Library
     ...    Aliased
@@ -50,6 +56,7 @@ Listen Imports
     ...    importer: //imports.robot
     ...    originalname: local_lib
     ...    source: //local_lib.py
+    Start Expect    Library    pythonmodule    []
     Expect
     ...    Library
     ...    pythonmodule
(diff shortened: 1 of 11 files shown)


pekkaklarck:
Thanks for the PR! Unfortunately it's too late to get big enhancements like this to RF 7.5 anymore. I'll take a look at this when we start RF 7.6 (or RF 8.0) development.

PR #5772, opened 2026-08-30
Libdoc: XML library

Vinni82:
- Added indentation for XML codeblocks
- Moved Args, Returns, Raises under the first line
- Looked at max 88 chars
- Added a linking towards Collections / BuiltIn lib

### DIFF
--- a/src/robot/libraries/XML.py
+++ b/src/robot/libraries/XML.py
@@ -83,7 +83,7 @@ else:
         Examples:
         ```robotframework
         *** Test Cases ***
-        Parse xml
+        Parsed Element
             # Get element as a return value:
             ${elem} =    Parse XML    ${CURDIR}/example.xml
             # Use element as an argument with other keywords:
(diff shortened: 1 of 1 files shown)


src/robot/libraries/XML.py
```diff
@@ -93,7 +93,9 @@ class Element:
 
         ```
 
-        [Element]: https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element "Element"
+        [Element]:
+        https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element
+        "Element"
```
pekkaklarck, line 98: I'm not sure does slitting reference targets like this work nor is it a good idea in general. I know the original line is over our 88 max line length, but that's something we can ignore in docs in cases like this.

pekkaklarck (approved):
Looks good! I'll test the reference link target syntax locally after merging this.

The failure on CI is due to me breaking one Markdown related test.

PR #5779, opened 2026-09-03
Report: Rework test details table layout 

Snooz82:
- Render test metadata as a table instead of a flat list
- Show tags as wrapping pills in `report.html` and `log.html`

### DIFF
--- a/src/robot/htmldata/rebot/log.html
+++ b/src/robot/htmldata/rebot/log.html
@@ -102,10 +102,6 @@ function popFromIterator(iterator, upTo) {
     return result;
 }
 
-function tagList(tags) {
-    return typeof tags === 'string' ? tags.split(', ') : tags;
-}
-
 function makeElementVisible(id) {
     window.testdata.ensureLoaded(id, function (ids) {
         util.map(ids, expandElementWithId);
@@ -346,10 +342,10 @@ function addExecutionLog(main) {
           <td class="doc">{{html doc()}}</td>
         </tr>
         {{/if}}
-        {{if tags}}
+        {{if tags.length}}
         <tr>
           <th>Tags:</th>
-          <td><div class="tags">{{each tagList(tags)}}<span class="label">{{html $value}}</span>{{/each}}</div></td>
+          <td><div class="tags">{{each tags}}<span class="label">{{html $value}}</span>{{/each}}</div></td>
         </tr>
         {{/if}}
         {{if timeout}}
--- a/src/robot/htmldata/rebot/log.html
+++ b/src/robot/htmldata/rebot/log.html
@@ -289,7 +289,7 @@ function addExecutionLog(main) {
         {{if tags.length}}
         <tr>
           <th>Tags:</th>
-          <td>{{html tags.join(', ')}}</td>
+          <td><div class="tags">{{each tags}}<span class="label">{{html $value}}</span>{{/each}}</div></td>
         </tr>
         {{/if}}
         {{if timeout}}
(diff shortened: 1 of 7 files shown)


src/robot/htmldata/rebot/log.html
```diff
@@ -345,7 +345,7 @@ <h2>Execution Errors</h2>
         {{if tags}}
         <tr>
           <th>Tags:</th>
-          <td>{{html tags}}</td>
+          <td><div class="tags">{{each tags}}<span class="label">{{html $value}}</span>{{/each}}</div></td>
```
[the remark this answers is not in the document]
Snooz82: fixed. i forgot the keyword tags.

src/robot/htmldata/rebot/log.html
```diff
@@ -289,7 +293,7 @@ <h2>Execution Errors</h2>
         {{if tags.length}}
         <tr>
           <th>Tags:</th>
-          <td>{{html tags.join(', ')}}</td>
+          <td><div class="tags">{{each tags}}<span class="label">{{html $value}}</span>{{/each}}</div></td>
```
pekkaklarck, line 292: Having each tag in a separate `<span>` is a good idea. Is wrapping them with `<div>` necessary? Isn't `<td>` enough? It seems we've used `<div>`s inside `<td>`s also elsewhere so probably that doesn't matter much.

src/robot/htmldata/rebot/log.html
```diff
@@ -345,7 +349,7 @@ <h2>Execution Errors</h2>
         {{if tags}}
         <tr>
           <th>Tags:</th>
-          <td>{{html tags}}</td>
+          <td><div class="tags">{{each tagList(tags)}}<span class="label">{{html $value}}</span>{{/each}}</div></td>
```
pekkaklarck: Is it so that with keywords `tags` is a string and with tests its a list? That's _really_ confusing and should be changed if that's easy.

src/robot/htmldata/rebot/report.html
```diff
@@ -840,8 +840,8 @@ <h3>Patterns</h3>
     </td>
   {{/if}}
     <td class="details-col-doc"><div class="doc details-limited">{{html doc()}}</div></td>
-    <td class="details-col-metadata"><div class="doc details-limited">{{each metadata}}<div>{{html $value[0]}}: {{html $value[1]}}</div>{{/each}}</div></td>
-    <td class="details-col-tags"><div>{{html tags.join(', ')}}</div></td>
+    <td class="details-col-metadata"><div class="details-limited">{{if metadata.length}}<table class="metadata">{{each metadata}}<tr><th>{{html $value[0]}}:</th><td class="doc">{{html $value[1]}}</td></tr>{{/each}}</table>{{/if}}</div></td>
+    <td class="details-col-tags"><div class="tags">{{each tags}}<span class="label">{{html $value}}</span>{{/each}}</div></td>
```
pekkaklarck, line 844: There's a conflict caused by the earlier PR that formatted test metadata as a table.

pekkaklarck (changes requested):
Conflict needs to be resolved.

src/robot/htmldata/rebot/log.html
Snooz82: that div defines the `tags` class, so that the `<span>` can be leaner and don't have to repeat the tags context multiple times.

pekkaklarck: I would assume that `<td class="tags">` works as well, but we have used `<divs>` in cases like this also elsewhere so let's use them here as well.

src/robot/htmldata/rebot/log.html
Snooz82: should be fixed now

